PR #475, opened 2014-12-15
Hg 124 (Extract critical, "above the fold" CSS and place it inline)

bkoval:
Summary of the changes:
* Extracting "above the fold" SCSS to preload/ directory, which is included in the style tag in the head of the html document through baseline.scss.

* Changing rem-calcs to rem where needed, since we wanted to get rid of dependencies in SCSS to keep inline css as small as possible.

* Loading seperate stylesheet (app.css) using defer rather than rendering it blockingly (as we did before) or lazy-loading - defer makes it render in non-blocking way just before DOMContentLoaded, which helps in getting rid of weird transition effects and flickering inline SVGs while removing .preload class.

* Rendering article title only for existing articles (if we'd keep rendering in a way we did, we would visually change bad article title from the request (e.g. "Rachel Berry2") into "Error XYZ" on the front-end side. Right the title comes empty and let's front-end code generate the error title.

* There is a significant boost when we take a look at the first time user can see the article on the screen. I will run the tests again and report them on JIRA.

ping @hakubo @kenkouot @rogatty @mklucsarits


public/scripts/main/app.ts
```diff
@@ -29,7 +29,6 @@ App.initializer({
 			LOG_TRANSITIONS: debug,
 			LOG_TRANSITIONS_INTERNAL: debug
 		});
-
```
rogatty: Unnecessary.

public/styles/baseline.scss
```diff
@@ -1,3 +1 @@
-// Styles to be inlined into base layout
-body {
-}
+@import 'preload/index';
```
rogatty: Maybe we should stick to a single name - `preload` or `baseline`. IMO it's confusing right now. I think `preload` is the more obvious one.

public/styles/component/_site-head.scss
```diff
@@ -1,16 +1,7 @@
...
 	direction: ltr;
-	height: rem-calc(50);
-	left: 0;
-	margin: 0 auto;
-	position: fixed;
-	right: 0;
-	text-align: center;
-	top: 0;
 	z-index: $z-site-head;
 }
```
rogatty: I understand these rules are left to minimize HTML size. But won't it cause any jumping after they're loaded?

public/styles/preload/_article.scss
```diff
@@ -0,0 +1,22 @@
...
+	background-color: rgb(223,223,223);
+	height: auto;
+	width: 100%;
+}
+
+.article-title {
+	font-size: 2rem;
+	line-height: 2rem;
+
+	margin:{
```
rogatty: Add `\s` and I think there is no need for `\n`.

public/styles/preload/_layout.scss
```diff
@@ -0,0 +1,58 @@
...
+body {
+	font-family: "Helvetica Neue", Helvetica, Roboto, Arial, sans-serif;
+	font-size: 16px;
+	margin: 0;
+	word-wrap: break-word;
+}
+
+.preload {
+	.page-wrapper {
+		margin-top: 99px;
```
rogatty: What is it for?

public/styles/preload/_layout.scss
```diff
@@ -0,0 +1,58 @@
...
+	text-align: center;
+	top: 0;
+}
+
+.wiki-title {
+	font-size: 0.8125rem;
+	margin: 0 0 1.25rem 0;
+
+	a {
+		color: #006cb0;
```
rogatty: It's already defined for `a` at the beginning of this file.

server/routes.ts
```diff
@@ -87,6 +88,7 @@ function beforeArticleRender (request: Hapi.Request, result: any): void {
 	}
 
 	result.displayTitle = title;
```
rogatty: This line is now duplicated.

views/application.hbs
```diff
@@ -6,5 +6,8 @@
 	</div>
 </nav>
 <div class="page-wrapper" lang="{{wiki.language.content}}" dir="{{wiki.language.contentDir}}">
+	<h2 class="wiki-title">
+		<a href="#">{{wiki.siteName}}</a>
```
rogatty: What is this `href="#"` for? Can you just remove it?

views/layout.hbs
```diff
@@ -15,11 +15,11 @@
 
 		{{#if isRtl}}
 			<!-- build:css /public/styles/app.rtl.css -->
-			<link rel="stylesheet" href="/public/styles/app.rtl.css">
+			<link rel="stylesheet" href="/public/styles/app.rtl.css" defer>
 			<!-- endbuild -->
 		{{else}}
 			<!-- build:css /public/styles/app.css -->
-			<link rel="stylesheet" href="/public/styles/app.css">
+			<link rel="stylesheet" href="/public/styles/app.css" defer>
```
rogatty: Did you try method [described here](https://developers.google.com/speed/docs/insights/OptimizeCSSDelivery?hl=en)?

public/styles/baseline.scss
bkoval: OK - just so you know, renaming baseline.scss will require changing gulp watcher :)

public/styles/component/_site-head.scss
bkoval: They should not as all the layout styling for site-head is just moved inline.

public/styles/preload/_layout.scss
bkoval: To match the padding and margin on the article container that is injected by ember, but it's not there in the html response. Reduces jumpiness.

public/styles/component/_site-head.scss
rogatty: What about these?
```
@extend .row;
@extend .small-12;
```

What do they add and are they even needed anymore?

public/styles/preload/_layout.scss
rogatty: How is this number calculated? What would cause it to change? Maybe you could add some comments (in other .scss files?) to make the connection clear.

public/styles/component/_site-head.scss
bkoval: This is foundation boiler-plate, takes care of styling ::before and ::after elements etc.

public/styles/preload/_layout.scss
bkoval: I'll do that, good idea.

gulp/tasks/watch.js
```diff
@@ -23,7 +23,7 @@ gulp.task('watch', ['build', 'build-views'], function () {
 		 * Baseline is a scss file that gets inlined, so the views must be recompiled
```
rogatty: `preload.scss is a scss file`

public/README.md
```diff
@@ -47,7 +47,7 @@ that is compiled to
 ```
   styles
     app.css
-    baseline.css
+    preload.css
```
rogatty: Is it still compiled to a file or just inlined in HTML?

public/styles/component/_site-head.scss
rogatty: Yeah, I know but I think my question is still valid :) How do they change layout? Did you check if they add margins or something else? Will site jump after they're loaded? Should they be inlined too?

public/styles/preload/_layout.scss
```diff
@@ -0,0 +1,58 @@
...
+body {
+	font-family: "Helvetica Neue", Helvetica, Roboto, Arial, sans-serif;
+	font-size: 16px;
+	margin: 0;
+	word-wrap: break-word;
+}
+
+.preload {
+	.page-wrapper {
+		//To match fully rendered article's distance from top (50px for site-head + 49 of margin)
```
rogatty: Where are the values `50px` and `49px` defined? Can you add comments there too so it won't break after someone modifies them?

views/layout.hbs
bkoval: I did. Google's method causes the external css to load in a first free frame after the whole inline layout is rendered and browsers is not executing any other job. I think it's too late for us since .preload class is removed more quickly than the new css is applied, which causes jumping and fouc just as "classic" lazy-loading would.

public/README.md
bkoval: Both.

public/styles/component/_site-head.scss
bkoval: Nah, they don't cause jumping, everything that would potentially cause jumping in site-head was moved inline. I'll check if we can safely remove them, but they introduce a couple of things that might break "something somewhere" if deleted ;)

public/styles/preload/_layout.scss
```diff
@@ -0,0 +1,58 @@
...
+	word-wrap: break-word;
+}
+
+.preload {
+	.page-wrapper {
+		//60px to mirror top leaderboard + 2rem ~~40px for margin top on wiki container
+		margin-top: 100px;
+	}
+
+	svg {
```
hakubo: svg:not(.wikia-logo) ?

public/styles/preload/_layout.scss
```diff
@@ -0,0 +1,58 @@
...
+		display: none;
+
+		&.wikia-logo {
+			display: inline;
+		}
+	}
+}
+
+.site-head {
+	background-color: rgb(255, 255, 255);
```
hakubo: white?

views/application.hbs
```diff
@@ -6,5 +6,8 @@
 	</div>
 </nav>
 <div class="page-wrapper" lang="{{wiki.language.content}}" dir="{{wiki.language.contentDir}}">
```
hakubo: single quotes please

rogatty:
One last thing and it LGTM

rogatty:
See https://github.com/Wikia/mercury/pull/498

PR #481, opened 2014-12-23
Upgrade spinner

hakubo:
https://wikia-inc.atlassian.net/browse/HG-497

use vertical color for its color
create global class for vertical
refactor SB code to use this class


rogatty:
LGTM

PR #484, opened 2014-12-29
Move to Jenkins

hakubo:


README.md
```diff
@@ -1,4 +1,3 @@
-[![Build Status](https://travis-ci.org/Wikia/mercury.svg)](https://travis-ci.org/Wikia/mercury)
```
rogatty: Should another link be here instead? Jenkins is not accessible without VPN so maybe link to the Wikia One's article with proper link inside?

rogatty:
At first glance it was confusing for me what's `app` and what's `service`. Maybe because in gulp we have tasks `docs-back`, `docs-front`, `scripts-back` and `scripts-front`. How about unifying it all?

But then... Should the directory `/public` be `/app` instead? It would make obvious what is what.

hakubo:
Good point, we should unify naming convention a bit,
let talk about it tomorrow.

PR #487, opened 2014-12-30
Remove hammer and use native touch events

hakubo:


public/scripts/main/views/MediaLightboxView.ts
```diff
@@ -223,60 +216,152 @@ App.MediaLightboxView = App.LightboxView.extend({
...
 	},
 
-	gestures: {
-		swipeLeft: function (): void {
-			if (this.get('isGallery') && !this.get('isZoomed')) {
-				this.nextMedia();
-			}
-		},
+	gesture: {
+		minSwipe: 50,
```
rogatty: This is being set in two places (for now). Should we abstract it now or wait until there are more cases?

public/scripts/main/views/MediaLightboxView.ts
```diff
@@ -223,60 +216,152 @@ App.MediaLightboxView = App.LightboxView.extend({
...
+	 * resets data after pinch and pan
+	 * @param event
+	 * @returns {boolean}
+	 */
+	touchEnd: function (event: {originalEvent: {touches: any[]}}): void {
+		var x = Em.get(event, 'originalEvent.changedTouches.0.clientX');
+
+		//swipe
+		if (this.get('isGallery') && !this.get('isZoomed')) {
+			if (Math.abs(this.get('gesture.startX') - x) > this.gesture.minSwipe) {
```
rogatty: How about abstracting this condition to utils? Then all the settings could be there too.

public/scripts/main/views/MediaLightboxView.ts
```diff
@@ -223,60 +216,152 @@ App.MediaLightboxView = App.LightboxView.extend({
...
+		}
+
+		//doubleTap
+		if (event.originalEvent.touches.length === 0) {
+			if (this.get('gesture.tapped')) {
+				return this.onDoubleTap(x);
+			}
+
+			this.set('gesture.tapped', true);
+			setTimeout(() => this.set('gesture.tapped', false), 200);
```
rogatty: Maybe abstract this number to `gesture`.

public/scripts/main/views/MediaLightboxView.ts
hakubo: I am not sure about overloading utils module
maybe a mixin could help us somehow :)

PR #489, opened 2014-12-31
Add Crowdin config

hakubo:
This is configuration to crowdin update tools
This is a tool that will enable us to update and fetch translations.


kenkouot:
Excited about this, can you add a description for this in the PR? That way if someone comes back and looks at this, they have some context.

PR #490, opened 2014-12-31
Close lightbox on back button

hakubo:
Initial idea - this puts every image into history stack


rogatty:
See https://github.com/Wikia/mercury/pull/501

PR #500, opened 2015-01-12
Use the same meta tag description as MW

rogatty:
https://wikia-inc.atlassian.net/browse/HG-488

Use article.article.description instead of article.details.abstract as the meta tag description value.

Don't merge before Wikia/app#5847 is deployed on production (January 14th).


hakubo:
LGTM

PR #501, opened 2015-01-12
Close lightbox on back button

rogatty:
https://wikia-inc.atlassian.net/browse/HG-197

Initial idea - this puts every image into history stack.

Same as https://github.com/Wikia/mercury/pull/490 but with `dev` branch as target.


hakubo:
@rogatty @RodriGomez 
This would work like this:
open image -> go back - closes lightbox
open image -> go to next/prev image -> go back - prev image -> go back closes lightbox

Is this something that we want?

rogatty:
What about your comment in JIRA? :D
>I feel like back button should just close gallery.

But yeah, both solutions make sense for me. I'd merge it and we can change it if the product owner says it should be the other way.

public/scripts/main/controllers/MediaLightboxController.ts
```diff
@@ -144,6 +144,17 @@ App.MediaLightboxController = App.LightboxController.extend({
 	}.observes('currentMedia').on('init'),
 
 	/**
+	 * observes file property
```
rogatty: No way, Captain Obvious! ;) I'm against comments like this one, it's redundant with `.observes('file')`.

RodriGomez:
Some time ago me and Stanley agreed that tapping back button should close modal although we didn't agree if it should be in all cases. Let's implement it and change if needed. My only concern is that behaviour used should be consistent in all of Mercury skin when a lightbox is opened.

hakubo:
I tried to merge dev to it, but there are too many changes. I will create new branch for it.

PR #502, opened 2015-01-12
Remove hammer and use native touch events

rogatty:
https://wikia-inc.atlassian.net/browse/HG-502

Same as https://github.com/Wikia/mercury/pull/487 but with `dev` branch as target.


public/scripts/main/views/MediaLightboxView.ts
```diff
@@ -223,60 +216,153 @@ App.MediaLightboxView = App.LightboxView.extend({
...
 	},
 
-	gestures: {
-		swipeLeft: function (): void {
-			if (this.get('isGallery') && !this.get('isZoomed')) {
-				this.nextMedia();
-			}
-		},
+	gesture: {
+		minSwipe: 50,
```
rogatty: This is being set in two places (for now). Should we abstract it now or wait until there are more cases?

public/scripts/main/views/MediaLightboxView.ts
```diff
@@ -223,60 +216,153 @@ App.MediaLightboxView = App.LightboxView.extend({
...
+	 * resets data after pinch and pan
+	 * @param event
+	 * @returns {boolean}
+	 */
+	touchEnd: function (event: {originalEvent: {touches: any[]}}): void {
+		var x = Em.get(event, 'originalEvent.changedTouches.0.clientX');
+
+		//swipe
+		if (this.get('isGallery') && !this.get('isZoomed')) {
+			if (Math.abs(this.get('gesture.startX') - x) > this.gesture.minSwipe) {
```
rogatty: How about abstracting this condition to utils? Then all the settings could be there too.

> @hakubo:
> I am not sure about overloading utils module
> maybe a mixin could help us somehow :)

front/scripts/main/views/ArticleView.ts
```diff
@@ -172,37 +172,43 @@ App.ArticleView = Em.View.extend(App.AdsMixin, {
 		}
 	},
 
-	hammerOptions: {
-		touchAction: 'auto'
+	minSwipe: 50,
```
kenkouot: Are these intentionally just for the article view? Seems like they should be application wide?

hakubo: Yes, or even mixin, but I would test it this way.
And if it works and if we use this in more places, then I would abstract it away.

public/scripts/main/views/MediaLightboxView.ts
hakubo: I would wait, I would like to see if this implementation does not degrade UX now.

public/scripts/main/views/MediaLightboxView.ts
hakubo: When we will have more places with such interaction, We will create a mixin for that

hakubo:
@tomasznapieralski Could you let me know how testing went? Better then previous interaction? Worse?

hakubo:
It was fun to try it,
But I don't think we should maintain another implementation of gestures
hammer is good enough for us, if wee need something else we can always add it to it

PR #504, opened 2015-01-13
Set API timeout to match MediaWiki app's timeout

kenkouot:
NewRelic was throwing errors because of timeouts caused by long parses. Mercury's timeouts were capped at 30s while MW's are capped at 5m. This pull request raises our timeout to match that.


mklucsarits:
LGTM

hakubo:
:( I get the point,
we could serve empty article with a spinner and lazy load article? - just an idea

PR #512, opened 2015-01-13
HG-523 Fix multi-line captions in linked galleries

mklucsarits:
https://wikia-inc.atlassian.net/browse/HG-523

This change fixes vertical spacing in linked image galleries with
multi-line captions by only showing the first line of a caption.


rogatty:
Wouldn't it be cleaner and safer (and slower, I know) to replace `<br>` elements with spaces or removing the rest with `caption.split('<br>')[0]`? I'm not a fan of introducing another magic number.

mklucsarits:
@rogatty That was my original approach before @kenkouot suggested simply using CSS to hide it. I'm not too opinionated toward one approach or the other. However, on desktop, we'll most likely want to show full, multi-line captions, so it would make sense to always return the full caption for the media component's caption property, rather than shorten it for a special case (mobile). For that reason I'd be more in favor of a presentation-only solution, but I'm open to counter-arguments. :)

hakubo:
Yup, presentation only fix is sound to me as well.

I would be for not doing any 'post production' of what have user created.

PR #516, opened 2015-01-16
Catchpoint support

rogatty:
Disable location API for Catchpoint tests.


hakubo:
too bad we need this, LGTM

PR #518, opened 2015-01-16
Reminder about whitespace

kenkouot:
Remember to check for style in your code reviews.

Two main whitespace errors here are the:
* space after function name
* no whitespace surrounding arguments/conditions in parens

Remember, our TS guidelines fall back to JS guidelines where applicable. Developers should familiarize themselves with the TS and JS guidelines [here](https://github.com/Wikia/guidelines).

### DIFF
--- a/server/lib/Utils.ts
+++ b/server/lib/Utils.ts
@@ -121,12 +121,12 @@ module Utils {
 	}
 
 	/**
-	* @desc Get vertical color from localSettings
-	*
-	* @param {LocalSettings} localSettings
-	* @param {string} vertical
-	* @return {string}
-	*/
+	 * @desc Get vertical color from localSettings
+	 *
+	 * @param {LocalSettings} localSettings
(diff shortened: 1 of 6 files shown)


server/lib/Utils.ts
```diff
@@ -121,12 +121,12 @@ module Utils {
 	}
 
 	/**
-	 * @desc Get vertical color from localSettings
-	 *
-	 * @param {LocalSettings} localSettings
-	 * @param {string} vertical
-	 * @return {string}
-	 */
+	* @desc Get vertical color from localSettings
```
mklucsarits: This formatting was correct before it was changed. For Javadoc-style doc comments, every line after the first should be indented one space, so the asterisks line up with the top line.

rogatty: Yeah, it's how PhpStorm does autoformatting too.

hakubo:
LGTM

PR #522, opened 2015-01-20
Add logging when there is no response object or response.header function

rogatty:
https://wikia-inc.atlassian.net/browse/HG-532


hakubo:
Mkay, lets have a butchers! LGTM

PR #529, opened 2015-01-22
Search placeholder and cancel button translations

Warkot:
https://wikia-inc.atlassian.net/browse/HG-506


front/locales/vi/translation.json
```diff
@@ -55,6 +55,8 @@
     "smartbanner-install-ios":"Get the app",
     "smartbanner-install-android":"Install",
     "smartbanner-price":"FREE",
-    "view-more":"View more"
+    "view-more":"View more",
+    "search-placeholder":"Search",
```
kenkouot: I am being nitpicky, but I think this would be more explicit as `search-placeholder-string` or text, or label, etc..,

front/templates/main/components/side-nav.hbs
```diff
@@ -6,7 +6,7 @@
 }}
 <div {{bind-attr class=':drawer controller.isInSearchMode:search'}}>
 	<div class='search-container'>
-		{{input placeholder='Search' focus-in='searchFocus' value=searchQuery action='enter'}}
+		{{input placeholder=searchPlaceholder focus-in='searchFocus' value=searchQuery action='enter'}}
```
kenkouot: same as comment above re: explicitness

front/locales/vi/translation.json
Warkot: I was even thinking of making this key's usage even broader as it's pretty universal—thus naming it just like `search-label`.

rogatty: I agree that `search-label` is more future proof. It's possible that we will have to add some kind of label or button with this string after redesign.

hakubo: As a general rule we should not reuse translations
String in english in 2 different contexts can be the same but this does not have to be true for other languages.

Warkot: I was also thinking of that, but when we decide to change it to narrow the context to something like 'Type here to search', then we can add a new key.

front/scripts/main/components/SideNavComponent.ts
```diff
@@ -9,6 +9,9 @@ App.SideNavComponent = Em.Component.extend({
 	isCollapsed: true,
 	isInSearchMode: false,
 	searchQuery: '',
+	searchPlaceholderLabel: function () {
```
rogatty: `function (): string {`

rogatty:
One little thing and it LGTM.

PR #531, opened 2015-01-23
Close lightbox on back button

hakubo:
This would work like this:
open image -> go back - closes lightbox
open image -> go to next/prev image -> go back - prev image -> go back closes lightbox

Is this something that we want?

### DIFF
--- a/front/scripts/main/controllers/MediaLightboxController.ts
+++ b/front/scripts/main/controllers/MediaLightboxController.ts
@@ -144,6 +144,18 @@ App.MediaLightboxController = App.LightboxController.extend({
 	}.observes('currentMedia').on('init'),
 
 	/**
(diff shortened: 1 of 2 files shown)


front/scripts/main/controllers/MediaLightboxController.ts
```diff
@@ -144,6 +144,18 @@ App.MediaLightboxController = App.LightboxController.extend({
 	}.observes('currentMedia').on('init'),
 
 	/**
+	 * closes lightbox when file queryParam is not set
+	 * otherwise tries to open image lightbox with appropriate image
+	 */
+	fileObserver: function (): void {
+		if (this.get('file') == null) {
```
kenkouot: strict operator here please

PR #534, opened 2015-01-23
I18nMixin && uselang={lang} param support

dianafa:
https://wikia-inc.atlassian.net/browse/HG-442

Generally, the idea of this solution is to make Mixin handling translations and dynamically update them when their dependencies change e.g. number of comments (computedProperties). 
Translation keys are kept in view and are called from view (from it’s I18nMixin).

Using uselang={lang} param you can change the language of static article elements. 

To test how it works you can paste it for example in ArticleView#didInsertElement:

```
setTimeout(() => {
    this.set('controller.uselang', 'pl');
},5000);

setTimeout(() => {
    this.set('commentsCount', 7);
},10000);
```

What also would be nice to change:
- this.get('controller.uselang’) - by geting and setting it we use in fact article.controller not, the application.controller, but having this also in article controller can be useful too.
- variable readyToTranslate - is this name siginificant enough?

Also ‘app:’ namespace is removed from translations in order to use it more in the future, dividing translations into more useful namespaces.


rogatty:
Please merge the latest dev to your branch.

public/scripts/main/controllers/ApplicationController.ts
```diff
@@ -2,9 +2,10 @@
 'use strict';
 
 App.ApplicationController = Em.Controller.extend({
-	queryParams: [{noAds: 'noads'}],
+	queryParams: [{noAds: 'noads'}, {uselang: 'uselang'}],
```
rogatty: How about `useLang: 'uselang'`? Would be more consistent with `noAds`.

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,55 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../../../typings/jquery/jquery.d.ts" />
+'use strict';
+
+App.I18nMixin = Em.Mixin.create({
+	//this param is used to trigger translate() function when language changed
+	readyToTranslate: false,
+
+	changeLng: function(): void {
```
rogatty: I'd use the full word everywhere except of query param name (`useLang`) so for example `changeLanguage` here.

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,55 @@
...
+		var lng = this.get('controller.uselang');
+		i18n.setLng(lng, () => {
+			this.notifyPropertyChange('readyToTranslate');
+		});
+	}.observes('controller.uselang'),
+
+	translate: function(): void {
+		Object.keys(this.translations).forEach((translationKey: string) => {
+			var translationParams = this.translations[translationKey] || {};
+			var computedProperty = [
```
rogatty: Single `var` per function.

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,55 @@
...
+				'readyToTranslate',
+				() => {
+					var paramsHash : { [index: string]: any } = {};
```
rogatty: There shouldn't be a space after `paramsHash`.

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,55 @@
...
+					Object.keys(translationParams).forEach((paramKey: string) => {
+						paramsHash[paramKey]= this.get(translationParams[paramKey]);
```
rogatty: There should be a space after `[paramKey]`

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,55 @@
...
+	readyToTranslate: false,
+
+	changeLng: function(): void {
+		var lng = this.get('controller.uselang');
+		i18n.setLng(lng, () => {
+			this.notifyPropertyChange('readyToTranslate');
+		});
+	}.observes('controller.uselang'),
+
+	translate: function(): void {
```
rogatty: Please, add JSDoc explaining what kind of magic is going on in this function.

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,55 @@
...
+	}.observes('readyToTranslate'),
+
+	init: function (): void {
+		this._super();
+
+		i18n.init({
+			resGetPath: '/public/locales/__lng__/translation.json',
+			detectLngQS: 'uselang',
+			fallbackLng: App.get('language'),
+			debug: true,
```
rogatty: Do we really want debug to be always turned on?

hakubo:
While doing review, please also focus on overall functionality.
Is this something that you would like to use?

Or you rather use i18n helper in templates?

public/scripts/main/models/ArticleModel.ts
```diff
@@ -49,6 +49,7 @@ App.ArticleModel = Em.Object.extend({
 });
 
 App.ArticleModel.reopenClass({
+
```
rogatty: Nope :no_good:

public/scripts/main/views/ArticleView.ts
```diff
@@ -14,9 +15,22 @@ interface HTMLElement {
 	scrollIntoViewIfNeeded: () => void
 }
 
-App.ArticleView = Em.View.extend(App.AdsMixin, {
+App.ArticleView = Em.View.extend(App.AdsMixin, App.I18nMixin, {
 	classNames: ['article-wrapper'],
 
+	translations: { //mixin looks at this object and updates it
```
rogatty: I can't find it in guidelines but I think there was discussion that we shouldn't put comments on the right side of code. Move it one line up, please.

public/scripts/main/mixins/I18nMixin.ts
dianafa: good point, I forgot about this

public/scripts/main/controllers/ArticleController.ts
```diff
@@ -11,6 +11,7 @@ App.ArticleController = Em.ObjectController.extend({
 	commentsPage: null,
 	map: null,
 	noAds: Em.computed.alias('controllers.application.noAds'),
+	uselang: Em.computed.alias('controllers.application.useLang'),
```
rogatty: Maybe `useLang: `?

public/templates/main/article.hbs
```diff
@@ -7,7 +7,7 @@
 			{{#collapsible-menu
 				showMenuIcon=true
 				additionalClasses='table-of-contents'
-				tLabel='article-toc-label'
+				tLabel= view.T-article-toc-label
```
rogatty: There shouldn't be a space.

public/templates/main/components/collapsible-menu.hbs
```diff
@@ -3,7 +3,7 @@
 	{{svg 'toc' viewBox='0 0 18 13' class='icon toc'}}
 	{{/if}}
 	{{svg 'chevron' viewBox='0 0 12 7' class='icon chevron'}}
-	{{i18n tLabel count=tCount}}
+	{{tLabel}}
```
rogatty: It bothers me a little that this one is different from the rest. Is it possible to use `view.T-article-categories-list-label` here?

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,65 @@
...
+	 * function which should launch when any of dependencies change. 
+	 * After, if translation contains any params, adds the rest of the dependencies
+	 * so if any of params change, function also should be launched.
+	 * Finally, the property (with 'T-' prefix) is created in object which Mixin is added to.
+	 */
+	translate: function(): void {
+		Object.keys(this.translations).forEach((translationKey: string) => {
+			var translationParams = this.translations[translationKey] || {},
+				computedProperty = [
+				'readyToTranslate',
```
bognix: please fix indentation.

public/scripts/main/mixins/I18nMixin.ts
```diff
@@ -0,0 +1,65 @@
...
+					var paramsHash: { [index: string]: any } = {};
+					Object.keys(translationParams).forEach((paramKey: string) => {
+						paramsHash[paramKey] = this.get(translationParams[paramKey]);
+					});
+					return i18n.t(translationKey, paramsHash);
+				}
+			];
+
+			Object.keys(translationParams).forEach((paramKey: string) => {
+				computedProperty.unshift(translationParams[paramKey]);
```
bognix: does it have to be unshift istead of push? Unshift performance if very poor compared to push, http://jsperf.com/push-pop-vs-unshift-shift/3

rogatty: I'm guessing that following code needs a function to be the last element of an array, and have dependencies before it:
```
Ember.defineProperty(this, 'T-' + translationKey, 
	Ember.computed.apply(this, computedProperty)
);
```

@dianafa can you push all dependencies to `computedProperty` and only after that add the anonymous function there?

rogatty:
`/front/scripts/main/helpers/timeAgo.ts` should be updated too

rogatty:
When I open site using this branch, in console I get:
- `i18next.js:790 currentLng set to: pl` - 7 times
- `i18next.js:790 i18next not finished initialization. you might have called t function before loading resources finished.` - 12 times

and then:
```
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
i18next.js:790 loaded: /public/locales/pl/translation.json
i18next.js:790 loaded: /public/locales/en/translation.json
```

It doesn't look like something we want ;)

kenkouot:
First off, I'd like to say great job on the quality of code in this PR @dianafa. The solutions are well implemented.

That being said, I'm against merging this. I think there are a few things that are at play here:
* This PR assumes that for some reason, we need to be able to update language dynamically during the same application instance. This is evidenced by Diana's `setTimeout` example in the description. This requirement is completely new to me and actually opposite of some decisions @hakubo and I made earlier in the application where we decided that changing languages would for simplicity's sake, only be done at application bootstrap and as such, would necessitate a full page refresh each time.
** I don't like the idea of managing and updating this state within an application and rerendering all the messages within a view, etc.., For the same reason that going to a different wiki simply is a full page refresh–it's more elegant and reduces the complexity and cognitive load on the developer and application to not have to manage this.
** The user does not gain much benefit from this. I bet the numbers of users who use ?uselang are fairly low in the overall traffic and this is really a tool for devs. Even so, we've confusing being overload language using a queryParam, with the ability to do so within the same single page application instance.
* The second part of this PR actually changes the mechanism by which developers can implement messages into Mercury views and actually **adds complexity and coupling** to the workflow, without increasing value (except to be able to refresh in same SPA instance, which is of dubious usefulness as I pointed out above).
** The new method, just adds code. You still have to reference the message variable in the view AND you have to pass in a translations object to the View. Compared to the previous implementation, we've just increased the bloat for our view. Consider views which may have many messages, the file will not grow eloquently. In terms of file size, this new approach approximately **doubles** the space that messages take up in compiled JS (the `{{view.messageKey}}` is still there **and** the `translations` hash in the view is redundant.
** This approach creates a coupling between messages and the view model. Now, this technically breaks the encapsulation of MVC and also introduces an extra layer of abstraction that wasn't needed before.
** It's just plain cumbersome compared to the previous method.
** It has no precedence in other frameworks. Keys should be declarative, and should only need to be declared once. Proxying through a mixin/view so that it can respond to Ember handling of the queryParam seems like unintended use.

Anyways, TLDR: good work on the implementation, but we should not switch to this method.

kenkouot:
Also, instantiating each key to null seems like the same kind of application pattern that you look at 3 years down the line and think, "Why did we go that route?"

lizlux:
I agree with Ken that we probably don't want to list out all the messages used in each template in the view JS if we can avoid it. Also, doing a page refresh to change the language seems reasonable to me.

kenkouot:
By comparison, a simpler solution would just be to read the queryParams on the server and overload the language variable serialized into (layout.hbs)[https://github.com/Wikia/mercury/blob/dev/server/views/layout.hbs#L34]

dianafa:
Thanks Ken (and Liz) for long explanation. It was interesting task and I've learned a lot playing with Ember's issues. Thanks @hakubo for help!

mklucsarits:
I agree with Ken's comments. I think this adds unwanted complexity for little gain, as it is. Translation keys should not be duplicated (or be there at all) in the views/components. And this type of on-the-fly translation is not necessary, in my opinion, when a page refresh should instead take care of language change.

hakubo:
Closed pull request for now, so we don't merge it accidentally now.

Mea culpa for not chiming in on why we were playing with this approach. We wanted to understand better the problem and possible solutions.
It gives us some advantages like: way to update translations if language changes (uselang param is just a side effect to that not the core usage), one place to look for what messages are used in a view, better caching, cleaning up the namespaces.

Having that said, I do agree with all the comments. We were looking for alternative options around.

@kenkouot we have to stop inlining translations with content as this will make caching hard, this PR do solve that, but of course this can be implemented using other techniques, and to my knowledge using separate resource will be no worse for FastBoot.

But the biggest improvement that this code gives us is that translations become `ComputedProperties`
that not only can respond to language change but also to dependant keys. So number of comments, or time ago message can be easily updated when needed.


I see 2 main issues that we have to still fix/improve on:
- caching of article currently is bound to a language needlessly
- messages are not updated when keys that a translation depends on changes


What do you guys think about those?

PR #535, opened 2015-01-26
Increase the tap area for lightbox close button by 20px

rogatty:
https://wikia-inc.atlassian.net/browse/HG-327


bkoval:
LGTM

bkoval:
Still LGTM if the size change is safe :) :jack_o_lantern:

PR #536, opened 2015-01-26
Revert method for setting body unscrollable to the one that was working

rogatty:
https://wikia-inc.atlassian.net/browse/HG-457

Nothing new, this code was here before. This causes top bar to hide/show after toggling side nav but it's a lower priority task than menu not being scrollable.

### DIFF
--- a/front/scripts/main/views/ApplicationView.ts
+++ b/front/scripts/main/views/ApplicationView.ts
@@ -33,7 +33,7 @@ App.ApplicationView = Em.View.extend({
 
 	smartBannerVisible: Em.computed.alias('controller.smartBannerVisible'),
 	sideNavCollapsed: Em.computed.alias('controller.sideNavCollapsed'),
-	scrollable: Em.computed.alias('controller.scrollable'),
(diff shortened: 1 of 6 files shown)


hakubo:
I remember there was some problem with lightbox, does it work with this implementation

front/scripts/main/views/ApplicationView.ts
```diff
@@ -128,13 +129,22 @@ App.ApplicationView = Em.View.extend({
 
 	actions: {
 		setScrollable: function (): void {
-			var $element = $(this.get('element'));
-			$element.off('scroll touchmove mousewheel', this.preventDefault);
+			Em.$('body')
+				.removeClass('no-scroll')
```
hakubo: `no-scroll` could be property, so it never diverge by accident

rogatty:
@hakubo thanks for tips. We weren't actually making page unscrollable when opening lightbox. It should be fine now.

hakubo:
Can we test it first?
@tomasznapieralski please do not merge yet.

PR #537, opened 2015-01-27
Add play icon to video thumbnail in comment

rogatty:
https://wikia-inc.atlassian.net/browse/HG-489

### DIFF
--- a/front/scripts/main/components/ArticleCommentComponent.ts
+++ b/front/scripts/main/components/ArticleCommentComponent.ts
@@ -65,7 +65,7 @@ App.ArticleCommentComponent = Em.Component.extend({
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


front/scripts/main/components/ArticleCommentComponent.ts
```diff
@@ -83,6 +83,10 @@ App.ArticleCommentComponent = Em.Component.extend({
 				$figcaption: JQuery,
 				$figure = $('<figure/>');
 
+			if (image.type === 'video') {
```
bkoval: Could the current object in a loop be called thumbnail, media or sth more universal? :dancer:

rogatty: @bkoval you're right, fixed.

PR #538, opened 2015-01-27
Fixed problem with handling links when comments are opened

rogatty:
https://wikia-inc.atlassian.net/browse/HG-539


hakubo:
LGTM

PR #539, opened 2015-01-27
Fix loading images in horizontal galleries after window resize

rogatty:
https://wikia-inc.atlassian.net/browse/HG-536

### DIFF
--- a/front/scripts/main/components/GalleryMediaComponent.ts
+++ b/front/scripts/main/components/GalleryMediaComponent.ts
@@ -31,8 +31,7 @@ App.GalleryMediaComponent = App.MediaComponent.extend(App.ArticleContentMixin, {
 		this.setProperties({
 			media: mediaArray,
 			limit: this.incrementLimitValue,
-			galleryLength: mediaArray.length,
-			galleryWidth: this.$().width()
+			galleryLength: mediaArray.length
 		});
 	},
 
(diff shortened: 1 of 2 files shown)


front/scripts/main/components/GalleryMediaComponent.ts
```diff
@@ -82,11 +84,13 @@ App.GalleryMediaComponent = App.MediaComponent.extend({
...
 
 		thisGallery.on('scroll', () => {
-			this.onScroll(thisGallery, galleryWidth, maxImages);
+			this.onScroll(thisGallery, maxImages);
 		});
 	},
 
-	onScroll: function (thisGallery: JQuery, galleryWidth: number, maxImages: number): void {
+	onScroll: function (thisGallery: JQuery, maxImages: number): void {
+		var galleryWidth = this.get('galleryWidth');
```
hakubo: shouldn't we just use this here `articleContent.width`?

front/scripts/main/components/GalleryMediaComponent.ts
```diff
@@ -105,5 +109,11 @@ App.GalleryMediaComponent = App.MediaComponent.extend({
 				}
 			}
 		}, 100);
-	}
+	},
+
+	articleContentWidthObserver: function (): void {
+		if (this.get('_state') === 'hasElement') {
```
hakubo: not a fan of playing with internals, do we have to check it?

rogatty: This observer is being called before the component has element and it was throwing error about `this.$()` being undefined. I could check for `typeof this.$()` but is it better?

front/scripts/main/components/GalleryMediaComponent.ts
```diff
@@ -105,5 +109,11 @@ App.GalleryMediaComponent = App.MediaComponent.extend({
...
+			this.set('galleryWidth', this.$().width());
```
hakubo: how about `galleryWidth` becoming a computed property that is dependant on `articleContent.width`

rogatty:
@hakubo I've tried making it computed property:
```
galleryWidth: function () {
	return this.$().width();
}.property('articleContent.width'),
```
but it doesn't run when `articleContent.width` is updated by `ArticleContentMixin`.

I modified observer to be:
```
articleContentWidthObserver: function (): void {
	console.log('### articleContentWidthObserver');
	this.notifyPropertyChange('galleryWidth');
}.observes('articleContent.width')
```
and it runs when `articleContent.width` is updated by `ArticleContentMixin` but `galleryWidth` property is still not being recalculated.

I believe this is the same issue that @bognix was struggling with.

If you don't have a better idea than us why the Ember is behaving like that, I see two options:
1. Leave it like that and be prepared that it might break after Ember update (tests).
2. Stop caching `galleryWidth` and get it from DOM after each `this.onScroll` call. It's debounced anyway and shouldn't be very heavy. We're making `thisGallery.find('img:not(.loaded)')` call in there already.

hakubo:
LGTM, but something does not feel right about it.
I will review this along with refactor of code for gallery component

Thanks!

PR #540, opened 2015-01-27
Fix label on comment replies

rogatty:
https://wikia-inc.atlassian.net/browse/HG-540

This code was trying to access `this['comment.comments.length']` instead of `this.comment.comments.length`


hakubo:
LGTM, but the fact that scope is passed here, does not make me comfortable

PR #546, opened 2015-01-28
I18n update

Warkot:
**TL;DR:** New translations for Mercury.

---
There are few keys left untranslated in few languages—mostly latest additions, which are `map-lightbox-error`, `search-label` and `cancel-label`.

Also I've changed lately some key values in the source file (English). It was just a style fix—like `"months-ago-label":"one month ago"` swapped to `"months-ago-label":"a month ago"`. It doesn't change the meaning at all, but it still lead to invalidation of those keys' values translations in CrowdIn. Fortunately, I've downloaded a copy of the locales before uploading the feral(-ish) revision of the source file, which enabled me to merge all of the changes. While we're waiting for the re-styled translations, the ones here can surely be used.

The Polish file looks a little different than other ones—that's because Polish language handles different plurals in different manner (1 piwo, <2; 4> piwa, <5; 21> piw, <22; 24> piwa, <25; 31> piw… and so on). This is supported by i18next but not by CrowdIn (hopefully just yet). We could try to hack our code around it somehow, but there are probably more languages which have similar *features* but for different plurals.


front/locales/pl/translation.json
```diff
@@ -62,10 +63,11 @@
     "username-anonymous":"A Wikia Contributor",
     "smartbanner-store-ios":"W App Store",
     "smartbanner-store-android":"W Google Play",
-    "smartbanner-install-ios":"Pobierz aplikację",
+    "smartbanner-install-ios":"Pobierz",
     "smartbanner-install-android":"Zainstaluj",
-    "smartbanner-price":"FREE",
+    "smartbanner-price":"Za darmo",
     "view-more":"Zobacz więcej",
+    "wikia-map-view-map":"View",
```
rogatty: Come on ;)

rogatty:
All the files have permissions changed: `100755 → 100644`
Is it expected? Won't it break anything?

Warkot:
The downloaded files from CrowdIn have this permissions set, but it shouldn't break anything as those files are not executables for the operating system. They're a kind of a database for the plugin. This database then is just read—not executed. I would even set 0644 to the translation source file (the English one) to keep it consistent.

front/locales/pl/translation.json
Warkot: Come on what? :stuck_out_tongue_winking_eye: What should it be? I was considering few: "Zobacz", "Przeglądaj", "Otwórz" or it could be even something else. :confused: :thought_balloon:

hakubo:
Ok, thanks!

PR #551, opened 2015-01-30
Fix for chevron position on single linked image with caption

rogatty:
Single linked image with caption generates following element:
`<figure class="ember-view media-component article-image has-caption visible" data-ref="4">`.

I've created `MercuryLinkedImagesNew` article on mediawiki119@devbox to test it (http://mediawiki119.wikia.com/wiki/MercuryLinkedImages on production). During QA, please check if all possible cases are there.

/cc @vforge

### DIFF
--- a/front/styles/module/article/_media.scss
+++ b/front/styles/module/article/_media.scss
@@ -33,6 +33,8 @@
 		z-index: $z-1;
 	}
 
+	//First line is for single images
+	//Second is for galleries that have multiple figure tags
   	&.has-caption .chevron,
 	.has-caption .chevron {
 		margin-bottom: 50px;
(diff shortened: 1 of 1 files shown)


hakubo:
Probably somebody asked it but can we have
`has-caption` class always on the same element?

It does not feel right to do this via 2 selectors

rogatty:
@hakubo we always put this class on the `<figure>` element but:
- there is only one `<figure>` element for `ImageMediaComponent`
- there are multiple `<figure>` elements for `GalleryMediaComponent` / `LinkedGalleryMediaComponent`

I don't see a way to do this with single CSS selector without introducing bad HTML.

hakubo:
mkay 
LGTM

tomasznapieralski:
Chevron vertical position is correct in all cases.
Horizontal position is correct for linked-gallery (2 small img in row),
but is incorrect for article-image, because chevrons are aligned to article-content and if image has smaller width, chevron is outside figure.
You should decide is that acceptable or not. The main problem was fixed, so we can merge it.

PR #553, opened 2015-02-05
add consul support

owend:
* split mediawikiHost into devboxDomain and consulDomain
* if consulDomain is set, API requests go through consul service discovery (DNS)
* add unit test for consul url rewrite 
* updated README 
* changed a few function names to make them more clear/explicit/unique.  
  * I didn't like ArticleRequest.fetch() and MediaWiki.fetch() being so similar in names because one fetch() calls the other one.  It was kind of hard to know what was going on there when looking at it with fresh eyes.  Because ArticleRequest.comments() will get comments, I changed it to ArticleRequest.article() to match that.  
  * changed getDomainName to getFallbackDomain since it's only used for fallback case
  * changed getFallbackSubdomain to getFallbackWiki


kenkouot:
Once merge, should developers just expect this to work? Or is there a migration path? If so, can you specify the steps for migration in the PR? We'll have to consider this a "breaking change" too!

owend:
I intended for consulDomain to be optional so it should be backwards compatible.  That's the goal anyway, if it isn't for some reason I'd like to fix that.  I can add instructions for testing locally with consul.

hakubo:
Nice!

I just don't see any definition of consul domain, or any getter from ENV
will that have to be defined in localSetting always?

Thanks for making the code more readable! :bow:

config/localSettings.d.ts
```diff
@@ -21,7 +21,8 @@ interface LocalSettings {
 	environment: any;
 	host: any;
 	loggers: LoggerInterface;
-	mediawikiHost: string;
+	devboxDomain?: string;
+	consulDomain?: string;
```
mklucsarits: Small request here: can you please move these config entries to keep them in alphabetical order? It will help keep the file maintainable as it grows.

owend:
There is 1 example of the consul domain in the README (mediawiki.service.consul) but I'm not sure if that's going to be the permanent name yet because we haven't added it to chef yet.  It looks like the app generally uses localSettings for everything so I expect it can be defined in localSettings for now.  We should have a wider discussion of how we want to expose/define those in the future, as it does seem like it won't scale to have hardcoded service names in every service configuration.   

@hakubo  So... you want me to change something or was that just a comment?  :)

hakubo:
Just a comment.

I would rename consulDomain to mediawikiDomain and maybe this should be defined like this:
```
mediawikiDomain: process.env.MEDIAWIKI_CONSUL_DOMAIN;
```

As soon we might use Pandora service somewhere here and we will have to have something like
pandoraDomain, or something like that.

owend:
Should it only come from environment?  Or should there be a config->environment hierarchy?
Our env vars generally look like WIKIA_SOMETHING so maybe WIKIA_MEDIAWIKI_CONSUL?

owend:
PR updated with name change.  The env var thing we can do at any point, but we don't actually have a good way to define those env vars now so I'm going to skip that and just leave the config driven setup.  Is that ok?

README.md
```diff
@@ -6,18 +6,23 @@
 * `npm install -g bower jshint gulp forever tsd typescript-formatter bower-installer` to install global dependencies
 * `bower install` will install client dependencies
 * `tsd update` will update typings folder with ambient files
-* Copy `config/localSettings.example.ts` to your own copy of `localSettings.ts` and set the `mediawikiHost`, `port` and `wikiFallback`.
-    The `mediawikiHost` should be set to your devbox name without 'dev-' prefix.
+* Copy `config/localSettings.example.ts` to your own copy of `localSettings.ts` and set the `port` and `devboxDomain`.
+    The devboxDomain must have your devbox name (without the dev- prefix) in it.
     The `wikiFallback` is useful but is not obligatory.
+    If you want to test with consul, add mediawikiDomain: 'mediawiki.service.consul' or env var:
```
kenkouot: wrap code in backticks `mediawikiDomain: 'mediawiki.service.consul`

Also, this line is really confusing to me–especially the "or env var:" part.

README.md
```diff
@@ -27,6 +32,7 @@
 
 ## Access Mercury
 Open http://muppet.127.0.0.1.xip.io:8000/wiki/Gonzo in your browser
+curl -H "Host:muppet.wikia-dev.com" "http://dev-joe:8000/wiki/Gonzo"
```
kenkouot: wrap code snippets in backticks, prefix CLI commands with $

kenkouot:
@owend I don't mind leaving it config driven for now, but we do use ENV vars in our runit script: https://github.com/Wikia/chef-repo/blob/master/cookbooks/_mercury/templates/default/sv-mercury-run.erb

owend:
I fixed the formatting of the readme and removed the reference to environment variables since it is not implemented in this PR anyway.

hakubo:
LGTM

owend:
Ok, just doing some more background work (dev-consul setup) before this can be merged.

hakubo:
@owend so you will merge this whenever it's ready? will you deploy it as well to test it on prod?

Thanks

owend:
Yeah, I'm trying to figure out how to get this QA'd and approved for merging.  I can definitely do the deployment and testing if that's the preferred approach.

owend:
Status:  the backwards compatible mode works on sandbox-mercury.  The consul mode does not, because sandbox-mercury is not configured with the proper role.  I will get that working and test it as well but that's all chef stuff.  For now we can merge this just to get the config changes propagated.   Ken said it was ok. :)

hakubo:
@kenkouot ok, can you merge it and see if all is good?

owend:
sandbox-mercury testing was successful yesterday.

kenkouot:
Magical merge moment!

PR #554, opened 2015-02-06
Call preventdefault only when we do something with a click

hakubo:
I know it's duplicated, but logic to deduplicate that seems like an overhead to me know


rogatty:
LGTM but please test all possible click cases before releasing.

Warkot:
LGTM, too. Mentioned before possible click cases should contain all possible buttons on different parts of the link (and out of it, too). :shipit:

tomasznapieralski:
Comparing to prod, no changes.

PR #557, opened 2015-02-12
Call event.preventDefault directly

mklucsarits:
This change removes some outdated and unnecessary code, which was
introduced with
https://github.com/Wikia/mercury/commit/53d14ee7c0fa9f37527734230244a02c
7ea471b4

The ```setScrollable``` and ```setUnscrollable``` actions no longer
exist, so this work-around for ```event.preventDefault()``` is not
needed.


rogatty:
LGTM

PR #558, opened 2015-02-12
Replace current implementation of sending queryParams to client

kenkouot:


mklucsarits:
Looks ok to me at a quick glance.

server/lib/Utils.ts
```diff
@@ -134,6 +134,21 @@ module Utils {
 		}
 		return null;
 	}
+
+	export function parseQueryParams (obj: any): any {
```
rogatty: Looks like a good candidate for unit tests.

server/lib/Utils.ts
```diff
@@ -134,6 +134,21 @@ module Utils {
...
+		var parsed = {},
+			key,
+			prop;
```
rogatty: `server/lib/Utils.ts(140,4): 7005 Variable 'key' implicitly has an 'any' type.`
Both of them do.

server/lib/Utils.ts
hakubo: any (:)) chance for anything better then any?

hakubo: and actually doesn't hapi give you query params parsed? - I am sure they do
https://github.com/hapijs/qs

kenkouot: @hakubo Parsed to strings, but this casts to int when possible (for obvious, non-silly cases)

PR #564, opened 2015-02-13
Stop escaping already escaped captions

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-111

We don't escape captions in `image-media.hbs` and shouldn't do it in `video-media.hbs`. They're escaped in MW already.


hakubo:
LGTM

PR #570, opened 2015-02-17
Mercury login

dianafa:
tl;dr:
I used Ember Simple Auth.
Currenty user can login, see his avatar, click on link to his profile, logout.
I wrote also some basic tests.

---------

How does it works:
1. To handle communication with Helios it uses the ember-simple-auth with simple-auth-oauth2 (OAuth) authenticator, since it seems to be the best and still supproted/enhanced library for Ember.js for authentication/authorization.

2. ember-simple-auth automatically maintains the 'session' variable which is available in all routes, controllers and components.

3. To keep all login connected functions I've created AuthMixin which I add to SiteHeadComponent. AuthMixin collects data from input and calls simple-auth 'authenticate' function.

4. Plugin sends POST for '/token' url and then in routes.ts we proxy it to https://id.wikia-services.com/token?CREDENTIALS

CREDENTIALS look like:
* grant_type=password <- this part is mantained by ember-simple auth by setting:
window.ENV['simple-auth'] = {
   authorizer: 'simple-auth-authorizer:oauth2-bearer'
};

* client_id=MEDIAWIKI_CLIENT_ID
* client_secret=MEDIAWIKI_CLIENT_SECRET
above two are set as a localSettings in localSettings.ts

* username=USERNAME
* password=PASS

6. With received user_id it asks wikia.com/api/v1 for user's username & avatar
http://www.wikia.com/api/v1/User/Details?ids=USER_ID


rogatty:
@dianafa bad news - this branch is based from master too.

front/scripts/main/controllers/ApplicationController.ts
```diff
@@ -8,6 +8,7 @@ App.ApplicationController = Em.Controller.extend({
 	noScroll: false,
 	noAds: '',
 	isLoading: false,
+	LoginFormCollapsed: true,
```
rogatty: Is there a good reason for this to not be camelCased?

front/scripts/main/mixins/AuthMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+	init: function () {
+		this._super();
+		if (this.get('session.isAuthenticated')) {
+			this.getUserData();
+		}
+	},
+
+	getUserData: function () {
+		Ember.$.ajax({
+			url: '/user/' + this.get('session').content.user_id,
```
rogatty: This will throw exception if `this.session.content` is undefined.

front/scripts/main/mixins/AuthMixin.ts
```diff
@@ -0,0 +1,57 @@
...
+	init: function (): void {
+		this._super();
+		if (this.get('session.isAuthenticated')) {
+			this.getUserData();
+		}
+	},
+
+	getUserData: function (): void {
+		Ember.$.ajax({
+			url: '/user/' + this.get('session').content.user_id,
```
rogatty: This will throw exception if this.session.content is undefined. The same for line 37.

front/templates/main/login-form.hbs
```diff
@@ -0,0 +1,21 @@
+{{#if session.isAuthenticated}}
+	<div class="login-form-toggle">
+		<a {{bind-attr href=currentUser.url}}> Profile {{currentUser.name}} </a>
```
rogatty: All labels should be translatable. Also, these whitespaces shouldn't be here.

front/templates/main/login-form.hbs
```diff
@@ -0,0 +1,21 @@
...
+		<br>
+		<br>
```
rogatty: Try to use CSS instead of `<br>`

server/routes.ts
```diff
@@ -155,6 +155,47 @@ function routes (server: Hapi.Server) {
 			}
 		};
 
+	var loginConfig = {
```
rogatty: We already have `config`, move it there (`config.login`).

server/routes.ts
```diff
@@ -155,6 +155,47 @@ function routes (server: Hapi.Server) {
...
+		handler: (request: any, reply: any) => {
+			request.payload += `&client_id=${localSettings.auth.client_id}&client_secret=${localSettings.auth.client_secret}`;
+			reply.proxy({
+				host: 'id.wikia-services.com',
+				protocol: 'https',
+				passThrough: true,
+				xforward: true,
+				onResponse : (err: any, res: any, request: any, reply: any, settings: any, ttl: any) => {
+					reply(err, res);
+				}
```
rogatty: Do we need to override this property? It doesn't really do anything, right?

front/scripts/main/controllers/ApplicationController.ts
dianafa: hmmm, no

server/routes.ts
```diff
@@ -155,6 +155,47 @@ function routes (server: Hapi.Server) {
...
+		}
+	});
+
+	server.route({
+		method: 'GET',
+		path: '/user/{id}',
+		config: loginConfig,
+		handler: (request: any, reply: any) => {
+			reply.proxy({
+				host: 'www.wikia.com/api/v1/User/Details?ids=' + request.params.id + '&size=50',
```
rogatty: I don't like that we're setting image size in `server/routes.ts` but am not sure what would be the perfect solution. I see two options:
1. Create new component with all login-related stuff, remove this parameter and use `Mercury.Modules.Thumbnailer` in the new component.
2. Move this parameter to localSettings.

I like the first one much more.

server/routes.ts
dianafa: yes, it was useful but is not anymore, thanks

server/views/layout.hbs
```diff
@@ -61,6 +61,22 @@
...
+				crossOriginWhitelist: ['https://id.wikia-services.com/token'],
+				authenticator: 'simple-auth-authenticator:oauth2-password-grant',
+				serverTokenEndpoint: '/token'
+			};
+
+			window.ENV['simple-auth'] = {
+				authorizer: 'simple-auth-authorizer:oauth2-bearer'
+			};
+
+			</script>
```
rogatty: Can this script be put in a component or somewhere else in `/front`? If not (SimpleAuth requirements, I guess) can you move it to a partial?

server/routes.ts
dianafa: This size is not super-important, if it's not set, by default we get avatar 100x100px. However, creating a new component can be nice.

front/locales/en/translation.json
```diff
@@ -59,6 +59,12 @@
 		"view-more": "View more",
 		"wikia-map-view-map": "View",
 		"search-label": "Search",
-		"cancel-label": "Cancel"
+		"cancel-label": "Cancel",
+		"enter-login-label": "Enter login",
+		"enter-password-label": "Enter password",
+		"logout": "Logout",
```
mklucsarits: Logout and Login should not be used as verbs. It should be "Log out" and "Log in".

front/scripts/main/components/GalleryMediaComponent.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference path="../app.ts" />
 /// <reference path="./MediaComponent.ts" />
+/// <reference path="../mixins/ArticleContentMixin.ts" />
```
mklucsarits: Please make sure you're not including files not related to authentication. It looks like changes to this file should be in another branch.

front/scripts/main/mixins/AuthMixin.ts
```diff
@@ -0,0 +1,65 @@
...
+			type: 'GET',
+			dataType: 'json',
+			contentType: 'application/x-www-form-urlencoded'
+		}).then((response: any): void => {
```
mklucsarits: I think the response data should be validated somehow, before assuming it's in the expected format.

hakubo:
@kenkouot @mklucsarits Are you going to take this work?
Should we finish it?

I know that you had a plan to work on it. Is it still on your backlog?

kenkouot:
We are taking it :) I will share our backlog with you offline for a bit more transparency

rogatty:
@dianafa @hakubo should this still be open if we have https://github.com/Wikia/mercury/pull/592?

dianafa:
no :P

PR #571, opened 2015-02-17
Search query visible in search input

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-121

Now the search query should be displayed in search input after displaying the results.


vforge:
lgtm

hakubo:
I think you created a branch of master and want to merge it to dev,
can you create your branch of dev, so we don't have all those commits here? Thanks!

PR #572, opened 2015-02-17
Fixed height on top leaderboard

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-84

Doesn't show ad-slot when noAds is set.


rogatty:
LGTM

hakubo:
Why do we have those merge pull here?

front/templates/main/application.hbs
```diff
@@ -1,7 +1,9 @@
 {{smart-banner isVisible=view.smartBannerVisible}}
 {{site-head smartBannerVisible=view.smartBannerVisible sideNavCollapsed=view.sideNavCollapsed trackClick='trackClick'}}
 {{side-nav isCollapsed=view.sideNavCollapsed}}
-{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
+{{#unless noAds}}
```
hakubo: not displaying ads is handled inside the component so we don't have to do it here.

rogatty: We're always rendering HTML for it. With `noads=1` it's:
```
<div id="MOBILE_TOP_LEADERBOARD" class="ad-slot" style="display: none !important;"></div>
```

Do you think it should be rendered and hidden?

hakubo: yes, that was requested by ads team, this way they can always fully debug what is going on with ads

rogatty: Actually, good point. We check `noAds` in `AdSlotComponent.ts` and condition logic is more complicated than `if (noAds)`. This shouldn't be done in template.

dianafa: aaaaa, ok!

kenkouot:
Yeah, can we just redo this branch without all the merges

PR #573, opened 2015-02-18
Fixed top leaderboard ad height

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-84

`noAds` as a computed property and class binding.

### DIFF
--- a/front/scripts/main/components/AdSlotComponent.ts
+++ b/front/scripts/main/components/AdSlotComponent.ts
@@ -6,30 +6,28 @@
 
 App.AdSlotComponent = Em.Component.extend({
 	classNames: ['ad-slot-wrapper'],
-	classNameBindings: ['nameLowerCase', 'noAdsClass:no-ads'],
+	classNameBindings: ['nameLowerCase', 'noAds'],
 	//This component is created dynamically, and this won't work without it
 	layoutName: 'components/ad-slot',
 
 	name: null,
-	// noAds is being passed from ApplicationController where it's also casted to a string
-	noAds: '',
-	noAdsClass: false,
 
 	nameLowerCase: function () {
 		return this.get('name').toLowerCase().dasherize();
 	}.property('name'),
 
-	didInsertElement: function () {
-		var noAds = this.get('noAds');
-
-		if (noAds === '' || noAds === '0') {
+	// noAds is being passed from ApplicationController
+	noAds: function (key: string, value: any) {
+		if (value === '' || value === '0') {
 			Em.Logger.info('Injected ad:', this.get('name'));
 			Mercury.Modules.Ads.getInstance().addSlot(this.get('name'));
+			return false;
 		} else {
 			Em.Logger.info('Ad disabled for:', this.get('name'));
-			this.set('noAdsClass', true)
+			return true;
 		}
-	},
+		return false;
+	}.property(),
 
 	willDestroyElement: function() {
 		var name = this.get('name');
--- a/front/scripts/main/components/AdSlotComponent.ts
+++ b/front/scripts/main/components/AdSlotComponent.ts
@@ -18,14 +18,17 @@ App.AdSlotComponent = Em.Component.extend({
 
 	// noAds is being passed from ApplicationController
 	noAds: function (key: string, value: any) {
-		if (value === '' || value === '0') {
-			Em.Logger.info('Injected ad:', this.get('name'));
-			Mercury.Modules.Ads.getInstance().addSlot(this.get('name'));
-			return false;
-		} else {
-			Em.Logger.info('Ad disabled for:', this.get('name'));
-			return true;
+		if (arguments.length > 1) {
+			if (value === '' || value === '0') {
+				Em.Logger.info('Injected ad:', this.get('name'));
+				Mercury.Modules.Ads.getInstance().addSlot(this.get('name'));
+				return false;
+			} else {
+				Em.Logger.info('Ad disabled for:', this.get('name'));
+				return true;
+			}
 		}
+		return this.get('noAds')
 	}.property(),
 
 	willDestroyElement: function() {
--- a/front/scripts/main/components/AdSlotComponent.ts
+++ b/front/scripts/main/components/AdSlotComponent.ts
@@ -17,7 +17,7 @@ App.AdSlotComponent = Em.Component.extend({
 	}.property('name'),
 
 	// noAds is being passed from ApplicationController
-	noAds: function (key: string, value: any) {
+	noAds: function (key?: string, value?: any) {
 		if (arguments.length > 1) {
 			if (value === '' || value === '0') {
 				Em.Logger.info('Injected ad:', this.get('name'));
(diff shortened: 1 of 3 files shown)


rogatty:
LGTM

hakubo:
LGTM

front/scripts/main/components/AdSlotComponent.ts
```diff
@@ -26,6 +26,7 @@ App.AdSlotComponent = Em.Component.extend({
 			Mercury.Modules.Ads.getInstance().addSlot(this.get('name'));
 		} else {
 			Em.Logger.info('Ad disabled for:', this.get('name'));
+			this.set('no-ads', true)
```
rogatty: We have `noAds` and `no-ads` properties. No-no :no_good:

front/scripts/main/components/AdSlotComponent.ts
```diff
@@ -6,7 +6,7 @@
 
 App.AdSlotComponent = Em.Component.extend({
 	classNames: ['ad-slot-wrapper'],
-	classNameBindings: ['nameLowerCase'],
+	classNameBindings: ['nameLowerCase', 'no-ads'],
```
rogatty: Look at http://emberjs.com/api/classes/Ember.View.html#property_classNameBindings
>If the value of the property is a Boolean, the name of that property is added as a dasherized class name.

front/scripts/main/components/AdSlotComponent.ts
```diff
@@ -6,28 +6,27 @@
...
 	name: null,
 	// noAds is being passed from ApplicationController where it's also casted to a string
-	noAds: '',
-
-	nameLowerCase: function () {
-		return this.get('name').toLowerCase().dasherize();
-	}.property('name'),
-
-	didInsertElement: function () {
+	noAdsBoolean: function () {
```
hakubo: @dianafa I know what are you trying to do here.

Please read about setters in Ember, and let's quickly chat on Monday! :trumpet:

front/scripts/main/components/AdSlotComponent.ts
```diff
@@ -6,28 +6,28 @@
...
+		if (value === '' || value === '0') {
 			Em.Logger.info('Injected ad:', this.get('name'));
 			Mercury.Modules.Ads.getInstance().addSlot(this.get('name'));
+			return false;
 		} else {
 			Em.Logger.info('Ad disabled for:', this.get('name'));
+			return true;
 		}
-	},
+		return false;
```
rogatty: This will never be executed.

front/scripts/main/components/AdSlotComponent.ts
```diff
@@ -6,28 +6,28 @@
...
 	nameLowerCase: function () {
 		return this.get('name').toLowerCase().dasherize();
 	}.property('name'),
 
-	didInsertElement: function () {
-		var noAds = this.get('noAds');
-
-		if (noAds === '' || noAds === '0') {
+	// noAds is being passed from ApplicationController
+	noAds: function (key: string, value: any) {
```
rogatty: It returns boolean.

rogatty: Please, read http://emberashes.com/blog/2012/12/23/getters-and-setters-in-ember-dot-js/ and use:
```
if (arguments.length === 1) { // get
	// ...
} else {                      // set
	// ...
}
```
because now you're logging `Ad disabled...` every time the getter is being used.

test/specs/front/scripts/main/components/AdSlotComponent.js
```diff
@@ -40,51 +40,45 @@ test('Name lower case', function () {
 test('Component behaves correctly depending on noAds value', function () {
 	var testCases = [{
 			'properties': {
-				'name': 'Test ad 1'
-			},
-			'expectedLength': 1,
-			'message': 'Element added to slot because no noAds property was passed'
-		}, {
-			'properties': {
-				'name': 'Test ad 2',
```
dianafa: I removed above test case because it didn't set noAds variable at all. After changes in AdSlotComponent the behaviour is different, noAds is now a computed property, and not existing noAds is not possible. Removed case (when there is no noAds variable in URL) works correctly when checked manually.

front/scripts/main/components/AdSlotComponent.ts
```diff
@@ -6,28 +6,30 @@
...
+				Em.Logger.info('Injected ad:', this.get('name'));
+				Mercury.Modules.Ads.getInstance().addSlot(this.get('name'));
+				return false;
+			} else {
+				Em.Logger.info('Ad disabled for:', this.get('name'));
+				return true;
+			}
 		}
-	},
+		return value;
```
rogatty: This will always be undefined, right? I've tried to think of some better way to do this but came up with nothing :disappointed:
Setter is easy but where should getter get the value from?
@hakubo do you have any ideas?

hakubo: I actually proposed this, but yeah you're right :)

be default this should be false
so this should be `false` here actually.

test/specs/front/scripts/main/components/AdSlotComponent.js
rogatty: @dianafa this case was written by @nandy-andy to make sure that we show ads even if `noAds` isn't set in template:
```
{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
```

Let me take over, I have an idea :)

hakubo:
LGTM, thanks

nandy-andy:
:+1: @PiotrGackowski was able to reproduce the issue @rogatty discussed with me on Slack. Piotr is now trying to reproduce it on real device but those changes most likely didn't introduce the issue, so you're safe to move on with them :santa:

tomasznapieralski:
LGTM

front/styles/module/_ads.scss
```diff
@@ -7,6 +7,7 @@
 }
 
 .mobile-top-leaderboard {
+	height: 50px;
```
nandy-andy: Can you set `min-height` instead?

rogatty: Is there a good reason for doing that? Can you create a ticket with details? Thanks.

nandy-andy: We can have [higher leaderboards than 50px](http://qa-s2.wikia-prod:8080/view/Ads%20Synthetic/job/ads-mobile-synthetic/groups=TestSynthetic,label=qa-vm10-webdriver/lastSuccessfulBuild/artifact/logs/log.html ). What project should I create the ticket in?

dianafa: project CONCF.

nandy-andy: Yes, but AFAIK it's supposed to work in Mercury like in WikiaMobile (after all the container class is also named `mobile-top-leaderboard` :smiling_imp:). Therefore, there is possibility an ad with 100px height will be displayed in the slot.

Unfortunately, I've learnt about it too late - after the changes were merged. I also learnt it's a good practice not to set `height` in containers with ads slots. Better way is to use `min-height` and the best is to use created by AdEng team `default-height` class on the container (but it might not exist in Mercury).

PR #575, opened 2015-02-18
CONCF-7 | 'Full site' footer link fix (switches skin to the wiki's default skin)

Warkot:
Skin choosing process is controlled by cookie on Varnish. → https://github.com/Wikia/wikia-vcl/blob/master/control-skin.vcl

### DIFF
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -21,11 +21,14 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			var controller = this.controllerFor('article'),
 				model = controller.get('model'),
 				trackingCategory = target.dataset.trackingCategory,
-				info = M.getLinkInfo(model.get('basePath'),
+				info = M.getLinkInfo(
+					model.get('basePath'),
 					model.get('title'),
 					target.hash,
 					target.href
-				);
+				),
+				domainNameRegExpMatchArray: any = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname),
+				domainName: string = domainNameRegExpMatchArray ? domainNameRegExpMatchArray[0] : '.wikia.com';
 
 			/**
 			 * Handle tracking
@@ -43,8 +46,7 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			 */
 			if (target.className.indexOf('external') > -1) {
 				if (target.href.indexOf('useskin=oasis') > -1) {
-					// If using Oasis skin, set Mercury cookie
-					document.cookie = 'useskin=oasis; path=/';
+					document.cookie = 'useskin=oasis; domain=' + domainName + '; path=/';
 				}
 				return window.location.assign(target.href);
 			}
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -1,6 +1,5 @@
 /// <reference path="../app.ts" />
 /// <reference path="../../mercury/utils/articleLink.ts" />
-/// <reference path="../../../../typings/jquery.cookie/jquery.cookie.d.ts" />
 'use strict';
 
 App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -27,8 +27,26 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 					target.hash,
 					target.href
 				),
-				domainNameRegExpMatchArray: any = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname),
-				domainName: string = domainNameRegExpMatchArray ? domainNameRegExpMatchArray[0] : '.wikia.com';
+				domainNameRegExpMatchArray: any,
+				cookieDomain: string,
+				cookiePath: string;
+
+			if (window.Mercury.wiki.cookieDomain) {
+				cookieDomain = window.Mercury.wiki.cookieDomain;
+			} else {
+				domainNameRegExpMatchArray = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname);
+				if (domainNameRegExpMatchArray) {
+					cookieDomain = domainNameRegExpMatchArray[0];
+				} else {
+					cookieDomain = '.wikia.com';
+				}
+			}
+
+			if (window.Mercury.wiki.cookieDomain) {
+				cookiePath = window.Mercury.wiki.cookiePath;
+			} else {
+				cookiePath = '/';
+			}
 
 			/**
 			 * Handle tracking
@@ -46,7 +64,7 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			 */
 			if (target.className.indexOf('external') > -1) {
 				if (target.href.indexOf('useskin=oasis') > -1) {
-					document.cookie = 'useskin=oasis; domain=' + domainName + '; path=/';
+					document.cookie = 'useskin=oasis; domain=' + cookieDomain + '; path=' + cookiePath;
 				}
 				return window.location.assign(target.href);
 			}
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -49,7 +49,7 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			 * handle links that are external to the application like ?useskin=oasis
 			 */
 			if (target.className.indexOf('external') > -1) {
-				if (target.href.indexOf('useskin=oasis') > -1) {
+				if (target.href.indexOf('useskin=' + defaultSkin) > -1) {
 					document.cookie = 'useskin=' + defaultSkin + '; domain=' + cookieDomain + '; path=' + cookiePath;
 				}
 				return window.location.assign(target.href);
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -1,5 +1,6 @@
 /// <reference path="../app.ts" />
 /// <reference path="../../mercury/utils/articleLink.ts" />
+/// <reference path="../../../../typings/jquery.cookie/jquery.cookie.d.ts" />
 'use strict';
 
 App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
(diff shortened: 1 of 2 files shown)


front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -43,8 +46,8 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			 */
 			if (target.className.indexOf('external') > -1) {
 				if (target.href.indexOf('useskin=oasis') > -1) {
-					// If using Oasis skin, remove Mercury cookie
-					document.cookie = 'useskin=;expires=Thu, 01 Jan 1970 00:00:00 GMT;path=/';
+					// If using Oasis skin, set Mercury cookie
```
rogatty: It's not what the code says :)

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +21,14 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
 				trackingCategory = target.dataset.trackingCategory,
-				info = M.getLinkInfo(model.get('basePath'),
+				info = M.getLinkInfo(
+					model.get('basePath'),
 					model.get('title'),
 					target.hash,
 					target.href
-				);
+				),
+				domainNameRegExpMatchArray: any = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname),
```
rogatty: What will happen when using xip.io?

rogatty:
We already use https://github.com/carhartl/jquery-cookie in `SmartBannerComponent`. There is no need to set domain manually.

front/scripts/main/routes/ApplicationRoute.ts
bognix: consider using wgCookieDomain variable if it's accessible in Mercury?

rogatty: Currently it's not (http://mediawiki119.wikia.com/api/v1/Mercury/WikiVariables) but it would be easy to add. @bognix is it somehow different than current domain?

Warkot: Actually that RegExp match would work nicely with any domain—including `xip.io`.
Using `window.Mercury.wiki.cookieDomain` (which is now implemented) leads to failed cookie setting as its value is usually `.wikia.com` (but can also be e.g. `.wowwiki.com`). There could be some kind of an `if` statement checking the environment.
However, this cookie is not checked at all in Mercury itself (so on dev environment, too)—it's checked on Varnish.

Warkot:
We want to set cookie domain to `.wikia.com` not e.g. `.elderscrolls.wikia.com` which is a default behavior when setting a cookie without setting its domain.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +21,32 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
-				);
+				),
+				domainNameRegExpMatchArray: any,
+				cookieDomain: string,
+				cookiePath: string;
+
+			if (window.Mercury.wiki.cookieDomain) {
+				cookieDomain = window.Mercury.wiki.cookieDomain;
+			} else {
+				domainNameRegExpMatchArray = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname);
```
mklucsarits: I think exec() is overkill here. If you only need to test for a regex, use test(), as it's faster. Then set cookieDomain to window.location.hostname if true.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +21,32 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
+				if (domainNameRegExpMatchArray) {
+					cookieDomain = domainNameRegExpMatchArray[0];
+				} else {
+					cookieDomain = '.wikia.com';
```
mklucsarits: Not a big deal, but I don't like that this is hard-coded. Do we have a config value for the domain?

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +21,32 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
 					model.get('title'),
 					target.hash,
 					target.href
-				);
+				),
+				domainNameRegExpMatchArray: any,
+				cookieDomain: string,
+				cookiePath: string;
+
+			if (window.Mercury.wiki.cookieDomain) {
```
mklucsarits: Please use Em.get(Mercury, 'wiki.cookieDomain') to reference these wiki variables for now. We should implement a better getter method for those values in the future, though.

Warkot: Thanks, I will. :+1: I just didn't know it about it until @rogatty told me about it on Slack today.

front/scripts/main/routes/ApplicationRoute.ts
Warkot: I'm not testing here, I'm picking the cookie domain name `.wikia.com`, `.wowwiki.com` or whatever it is. But that's just a fallback here if no value is returned from API.

Warkot: But I'm more concerned with this https://github.com/Wikia/mercury/pull/575#discussion_r24921843.

mklucsarits: Oops, I didn't realize what you're actually doing there.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -43,8 +64,7 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			 */
 			if (target.className.indexOf('external') > -1) {
 				if (target.href.indexOf('useskin=oasis') > -1) {
-					// If using Oasis skin, remove Mercury cookie
-					document.cookie = 'useskin=;expires=Thu, 01 Jan 1970 00:00:00 GMT;path=/';
+					document.cookie = 'useskin=oasis; domain=' + cookieDomain + '; path=' + cookiePath;
```
hakubo: we would have to send default skin with wikivariables
as this might be... 'monobook' ...

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +21,32 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
+			} else {
+				domainNameRegExpMatchArray = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname);
+				if (domainNameRegExpMatchArray) {
+					cookieDomain = domainNameRegExpMatchArray[0];
+				} else {
+					cookieDomain = '.wikia.com';
+				}
+			}
+
+			if (window.Mercury.wiki.cookieDomain) {
```
hakubo: `cookiePath = Em.get(Mercury, 'wiki.cookieDomain') || '/'`

mklucsarits: Better yet, Em.getWithDefault()

front/scripts/main/routes/ApplicationRoute.ts
Warkot: True that. :+1: Or `uncyclopedia`. I'll add that to https://github.com/Wikia/app/pull/6344.

front/scripts/main/routes/ApplicationRoute.ts
Warkot: `Em.getWithDefault(Mercury, 'wiki.cookieDomain', '/')`?

mklucsarits: Yes

front/scripts/main/routes/ApplicationRoute.ts
Warkot: Fixed it here: https://github.com/Wikia/mercury/commit/d513c3803ad3c2f9e291857354d91cf428cd33d7

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +22,17 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
+				info = M.getLinkInfo(
+					model.get('basePath'),
 					model.get('title'),
 					target.hash,
 					target.href
-				);
+				),
+				// exec() returns an array of matches or null if no match is found.
+				domainNameRegExpMatchArray: string[] = /\.[a-z0-9\-]+\.[a-z0-9]{2,}$/i.exec(window.location.hostname),
+				cookieDomain: string = domainNameRegExpMatchArray ? domainNameRegExpMatchArray[0] : '.wikia.com',
```
rogatty: Try with `wowwiki.com` - it will be set to `.wikia.com` because it doesn't match your regex.

What's the point of setting cookie domain to `.wikia.com` if it's not in the current URL (otherwise it would be matched)? How about setting it to `'.' + window.location.hostname`?

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -21,11 +22,17 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
+				cookiePath: string = '/',
```
rogatty: Looks like it doesn't need a variable :)

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -42,9 +49,8 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			 * handle links that are external to the application like ?useskin=oasis
 			 */
 			if (target.className.indexOf('external') > -1) {
-				if (target.href.indexOf('useskin=oasis') > -1) {
-					// If using Oasis skin, remove Mercury cookie
-					document.cookie = 'useskin=;expires=Thu, 01 Jan 1970 00:00:00 GMT;path=/';
+				if (target.href.indexOf('useskin=' + defaultSkin) > -1) {
```
rogatty: What if there is a link with `useskin=wikiamobile` and it's not the default one? Don't we want to redirect it properly?

front/scripts/main/routes/ApplicationRoute.ts
Warkot: I tried it at `wowwiki.com`—user is redirected to `www.wowwiki.com` which is also the value of `window.location.hostname`, so RegExp works as expected and returns `[".wowwiki.com"]`.
But I like the idea to fallback to the full `window.location.hostname` (or `window.document.domain`). However, I'd go further and let the browser set this automatically, that is by not setting the `domain=…` at all in this case. What do you think, @rogatty?

front/scripts/main/routes/ApplicationRoute.ts
Warkot: Indeed. :+1:

front/scripts/main/routes/ApplicationRoute.ts
Warkot: Yup, we could. That was just done for the footer link. I could add a regex here.

Warkot:
I have a question, because PhpStorm is underlining this: `openLightbox` and `closeLightbox` functions have set return type to void, but actually have return statements. It doesn't seem right. :confused:  Ping: @hakubo @rogatty

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference path="../app.ts" />
 /// <reference path="../../mercury/utils/articleLink.ts" />
+/// <reference path="../../../../typings/jquery.cookie/jquery.cookie.d.ts" />
```
rogatty: This is obsolete.

front/scripts/main/routes/ApplicationRoute.ts
hakubo: I woould reload page if useskin= is present that's it

Warkot: But we have to set the right `useskin` cookie, don't we?

rogatty: Setting `useskin` cookie for every link with `useskin` seems like a bad idea (somebody will exploit it). It would be better to set it only for the 'Full site' link.

hakubo:
@Warkot if you can update the return types, I would appreciate that.

Warkot:
@hakubo: @rogatty created a PR for the types issue fix  https://github.com/Wikia/mercury/pull/582/files.

front/scripts/main/routes/ApplicationRoute.ts
Warkot: @rogatty, that convinces me. I'd leave it the current proposed shape. @hakubo? :shipit:

hakubo: kk, go for it! :)

rogatty:
LGTM

hakubo:
lGTM

PR #577, opened 2015-02-18
HG-566 Extended Optimizely integration

mklucsarits:
https://wikia-inc.atlassian.net/browse/HG-566

This change improves the Optimizely implementation by somewhat
decoupling the A/B testing code and allowing tracking of custom events.

### DIFF
--- a/front/scripts/mercury/utils/variantTesting.ts
+++ b/front/scripts/mercury/utils/variantTesting.ts
@@ -1,0 +1,40 @@
+/// <reference path="../../baseline/mercury.d.ts" />
+
+/**
+ * @define variantTesting
+ *
+ * Helper for variant testing using Optimizely
+ */
+'use strict';
+
+interface Window {
+	optimizely?: any;
+}
+
+module Mercury.Utils.VariantTesting {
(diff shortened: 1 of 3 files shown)


front/scripts/mercury/utils/variantTesting.ts
```diff
@@ -0,0 +1,40 @@
...
+
+module Mercury.Utils.VariantTesting {
+	var optimizely = window.optimizely;
+
+	/**
+	 * Activates all variant tests for the current page
+	 *
+	 * @returns {void}
+	 */
+	export function activate () {
```
rogatty: We should add something about `(): void {` to [guidelines](https://github.com/Wikia/guidelines/tree/master/TypeScript). It differs in our codebase, sometimes it's there and sometimes not. In my opinion it should be present.

mklucsarits: I agree with that. In this case I just forgot to add it.

kenkouot:
Add tests for these functions, mocking window.optimizely using sinon

PR #578, opened 2015-02-19
Hammer improvements

hakubo:
After some rework how we deal with gestures and actually implementing it on my own
I have noticed that actually supporting all the edge cases would make me 'reimplement' hammer.

We might take ownership of hammer as it currently lack a maintainer, but does everything we need currently.

I fixed most annoying issues that we have, but I think we should leave it for now.

### DIFF
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -19,6 +19,15 @@ var App: any = Em.Application.create({
 		apiBase: Mercury.apiBase || '/api/v1'
 	});
 
+window.emberHammerOptions = {
+	hammerOptions: {
+		//we are using fastclick so this is adviced by ember-hammer lib
+		ignoreEvents: [],
+		swipe_velocity: 0.1,
+		pan_threshold: 1
+	}
+};
+
 App.initializer({
 	name: 'preload',
 	initialize: (container: any, application: any) => {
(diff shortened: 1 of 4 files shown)


front/scripts/main/app.ts
```diff
@@ -41,5 +47,14 @@ App.initializer({
...
 		});
+
+		window.emberHammerOptions = {
+			hammerOptions: {
+				//we are using fastclick so this is adviced by ember-hammer lib
+				ignoreEvents: [],
+				swipe_velocity: 0.2,
+				pan_threshold: 0
+			}
+		};
```
rogatty: How about extracting it to a module or utils? Exactly what @mklucsarits did in https://github.com/Wikia/mercury/pull/577

hakubo: Sounds like a bit of overengeering, as this is just 3 vars now
and the whole file is about 60LOC

but We should keep that in mind, if this file grows to a ~100LOC

rogatty:
LGTM

hakubo:
@tomasznapieralski please try now, it should work much better!

tomasznapieralski:
Now we talking! :)
LGTM & automated tests!

front/scripts/main/app.ts
```diff
@@ -13,6 +19,15 @@ var App: any = Em.Application.create({
 		apiBase: Mercury.apiBase || '/api/v1'
 	});
 
+window.emberHammerOptions = {
```
kenkouot: I didn't see this before. Can we put this somewhere else? Seems just kind of randomly thrown in here.

hakubo: I'll try to think of something,
dedicated file makes more sense probably.

PR #579, opened 2015-02-19
Redirect to / when URI is malformed

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-70

`onPreResponse` event is the earliest place where we can detect that the request URI was malformed (`decodeURIComponent` failed in hapijs/call lib and `badrequest` method was set as a special route handler).
When MediaWiki gets request like that it redirects to the main page with code 301.
For now we don't want to send additional request to get title of the main page and are redirecting to / which causes user to get the main page eventually.

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -62,7 +62,23 @@ class App {
 
 		server.ext('onPreResponse', this.getOnPreResponseHandler(this.isDevbox));
 
-		/*
+		/**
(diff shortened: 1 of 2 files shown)


rogatty:
I still need to fix type definition for `Hapi.Request.route`

hakubo:
LGTM

PR #581, opened 2015-02-20
Track articleType with GA

hakubo:
I do not have better idea on how to pass stuff around if not leave the global:
`Mercury.article` and update it on article change

### DIFF
--- a/front/scripts/main/models/ArticleModel.ts
+++ b/front/scripts/main/models/ArticleModel.ts
@@ -1,6 +1,7 @@
+/// <reference path="../../baseline/mercury" />
 /// <reference path="../app.ts" />
 /// <reference path="../../mercury/utils/string.ts" />
-/// <reference path="../../mercury/modules/Ads.ts" />
+/// <reference path="../../mercury/modules/Ad$s.ts" />
 /// <reference path="../../../../typings/i18next/i18next.d.ts" />
 
 interface Response {
@@ -105,6 +106,7 @@ App.ArticleModel.reopenClass({
 			});
 		}
 
+		delete Mercury.article;
 		return article;
 	},
 
@@ -167,7 +169,7 @@ App.ArticleModel.reopenClass({
 
 		//We need to update global article object
 		//to allow eg. for analytics to take article context from it
-		Mercury.article = data;
+		M.provide('article.type', data.type);
 		model.setProperties(data);
 	}
 });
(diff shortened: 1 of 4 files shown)


front/scripts/main/models/ArticleModel.ts
```diff
@@ -1,6 +1,7 @@
+/// <reference path="../../baseline/mercury" />
 /// <reference path="../app.ts" />
 /// <reference path="../../mercury/utils/string.ts" />
-/// <reference path="../../mercury/modules/Ads.ts" />
+/// <reference path="../../mercury/modules/Ad$s.ts" />
```
rogatty: :no_good:

hakubo: haha
:)

front/scripts/main/models/ArticleModel.ts
```diff
@@ -165,6 +167,9 @@ App.ArticleModel.reopenClass({
 			}
 		}
 
+		//We need to update global article object
+		//to allow eg. for analytics to take article context from it
```
rogatty: This comment isn't very precise. There will be only `Mercury.article.type` and nothing else in `Mercury.article`, right?

hakubo: You're right

rogatty:
LGTM

front/scripts/main/models/ArticleModel.ts
```diff
@@ -165,6 +167,11 @@ App.ArticleModel.reopenClass({
 			}
 		}
 
+		// We could keep whole article in global but we want to discourage that but
+		// We need to update global article.type
+		// to allow eg. for analytics to use it
+		// TODO: Should analytics be part of ember? That should simplify how to pass stuff around.
+		M.provide('article.type', data.type);
```
kenkouot: Wait, what is article type? Is this different from namespace value? I'd prefer we don't use the article namespace (since I explicitly delete it after Ember bootstraps). If you have to store state, do it in `M.provide('_state.article.type')`. I will work on a getter/setter for Mercury state.

SebastianMarzjan:
Article type is a quality assigned to an article using machine learning. It states what the article is about, i.e. tv series, tv episode, character, video game
We plan to use that tracking to measure how the meaning of content influences user behavior.

front/scripts/main/models/ArticleModel.ts
SebastianMarzjan: Article type is a quality assigned to an article using machine learning. It states what the article is about, i.e. tv series, tv episode, character, video game
We plan to use that tracking to measure how the meaning of content influences user behavior.

kenkouot: Great, thanks for this reply. I have ideas on where to store this information in a more formalized way in the application, expect a PR @hakubo

PR #582, opened 2015-02-20
openLightbox and closeLightbox don't return anything

rogatty:
https://github.com/emberjs/ember.js/blob/v1.9.1/packages/ember-routing/lib/system/route.js#L1748 & https://github.com/emberjs/ember.js/blob/v1.9.1/packages/ember-routing/lib/system/route.js#L1840

/cc @Warkot


Warkot:
Thx for the change! :bow: That was my thought too.
LGTM :heavy_check_mark:

PR #583, opened 2015-02-24
HG-521 Integrate performance monitoring into Ember & persist with Weppy

kenkouot:


front/scripts/mercury/modules/Trackers/Perf.ts
```diff
@@ -0,0 +1,65 @@
...
+				sample: Mercury._state.weppyConfig.samplingRate,
+				aggregationInterval: Mercury._state.weppyConfig.aggregationInterval
+			});
+			super();
+		}
+
+		track (params: PerfTrackerParams): void {
+			var trackFn = this.tracker;
+
+			if (params.module && typeof params.module === 'string') {
```
mklucsarits: First condition here is not needed if checking typeof.

front/scripts/mercury/modules/Trackers/Perf.ts
```diff
@@ -0,0 +1,65 @@
...
+				trackFn = this.tracker.into(params.module);
+			}
+
+			if (params.context) {
```
mklucsarits: Why would the context change during a call to ```track```? Will context always be sent with track? If so, there doesn't seem to be a need to set it initially in the constructor.

front/scripts/main/app.ts
```diff
@@ -14,6 +16,30 @@ var App: any = Em.Application.create({
 	});
 
 App.initializer({
+	name: 'performanceMonitoring',
+	after: 'preload',
```
rogatty: If it's called after `preload` maybe the code should be below it too?

front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
```diff
@@ -3,19 +3,6 @@ interface Window {
...
-	prefix?: string;
-	// ie. 'UA-32129070-1'
-	id: string;
-	// sampling percentage, from 1 to 100
-	sampleRate: number;
-}
-
-interface GAAccountMap {
-	[name: string]: GAAccount;
-}
```
rogatty: Why did you remove this? What about "atomic PRs"?

front/scripts/mercury/modules/Trackers/Perf.ts
```diff
@@ -0,0 +1,70 @@
...
+		constructor () {
+			this.tracker = Weppy.namespace('mercury');
+			this.defaultContext = {
+				skin: 'mercury',
+				'user-agent': window.navigator.userAgent,
+			}
+			this.tracker.setOptions({
+				host: Mercury._state.weppyConfig.host,
+				transport: 'url',
+				context: this.defaultContext,
```
rogatty: Should we add current URL to the context?

front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
kenkouot: I removed this because after I added the WeppyConfig interface to a common location, and sourced it from other files (localSettings.d.ts), it created redundancies with these interfaces. So I removed these interfaces since they are already defined in localSettings.d.ts.

front/scripts/mercury/modules/Trackers/Perf.ts
kenkouot: Good idea! Will add that.

front/scripts/main/app.ts
kenkouot: Yeah I like that

front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
kenkouot: Added it on different line because it needs to parse URL at time of track, not at instantiation of tracker:
https://github.com/Wikia/mercury/pull/583/files#diff-758befc75e5df0779aa612c8722b5e72R48

front/scripts/mercury/modules/Trackers/Perf.ts
```diff
@@ -0,0 +1,73 @@
...
+
+		track (params: PerfTrackerParams): void {
+			var trackFn = this.tracker;
+
+			if (typeof params.module === 'string') {
+				trackFn = this.tracker.into(params.module);
+			}
+
+			// always set the current URL as part of the context
+			this.defaultContext.url = window.location.href.split('#')[0];
```
rogatty: Will this work if we track something, URL changes and then we track something else? It probably needs flushing after (before?) each transition.

front/scripts/mercury/utils/track.ts
```diff
@@ -160,11 +160,13 @@ module Mercury.Utils {
...
-			var trackerInstance = new trackers[tracker]();
+			var Tracker = trackers[tracker],
+				instance: TrackerInstance;
 
-			if (trackerInstance && trackerInstance.trackPageView) {
+			if (typeof Tracker.prototype.trackPageView === 'function') {
+				instance = new Tracker();
 				Em.Logger.info('Track pageView:', tracker);
-				trackerInstance.trackPageView(trackerInstance.usesAdsContext ? adsContext : context);
+				instance.trackPageView(instance ? adsContext : context);
```
rogatty: Seems like a typo. Why check `instance` while calling it?

front/scripts/mercury/modules/Trackers/Perf.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+module Mercury.Modules.Trackers {
+	export class Perf extends BaseTracker {
+		public static depsLoaded: boolean = typeof Weppy === 'function';
+		tracker: any;
+		defaultContext: {
+			skin: string;
+			url?: string;
+			'user-agent': string;
```
wladekb: Please add "env" here as we discussed: "prod" for production, something else for other envs incl. preview, verify, devs etc.

front/scripts/mercury/modules/Trackers/Perf.ts
```diff
@@ -0,0 +1,73 @@
...
+
+			if (typeof params.module === 'string') {
+				trackFn = this.tracker.into(params.module);
+			}
+
+			// always set the current URL as part of the context
+			this.defaultContext.url = window.location.href.split('#')[0];
+
+			// update context in Weppy with new URL and any explicitly passed overrides for context
+			trackFn.setOptions({
```
wladekb: use trackFn.flush() to trigger immediate send of the queue, but keep in mind all metrics you register afterwards will be bound to the new context

server/views/layout.hbs
```diff
@@ -83,6 +84,11 @@
 			<script src="/front/vendor/i18next/i18next.js"></script>
 			<script src="/front/vendor/vignette/dist/vignette.js"></script>
 			<script src="/front/vendor/scroll-depth/jquery.scrolldepth.js"></script>
+
+			{{#if weppyConfig.enabled}}
```
hakubo: does it work with
`gulp build --type prod` ?

This run during build:


but this is evaluated during runtime:
{{#if weppyConfig.enabled}}

hakubo:
@kenkouot what are the next steps with this?

front/scripts/mercury/modules/Trackers/Perf.ts
kenkouot: @wladekb So should I:
* setOptions -> flush -> (count || store || timer).send ?
* setOptions -> (count || store || timer).send -> flush

kenkouot: Feels like the second option

front/scripts/mercury/utils/trackPerf.ts
```diff
@@ -0,0 +1,16 @@
...
+/// <reference path="../modules/Trackers/Perf.ts" />
+'use strict';
+
+/**
+* @description Instantiates performance tracker
+*/
+module Mercury.Utils {
+	var instance: Mercury.Modules.Trackers.Perf;
+	export function trackPerf (obj: PerfTrackerParams) {
+		if (Mercury.Modules.Trackers.Perf.depsLoaded) {
```
mklucsarits: I think it's bad practice to reference a "class" property from another "class" directly, without using a getter. "depsLoaded" should instead be a method (in other words, a getter).

mklucsarits:
Looks mostly OK to me. I just have the one comment. It's up to you if you want to change that or not.

PR #584, opened 2015-02-25
Fix for crashing after empty response

rogatty:
Server was throwing exception when there was no payload in response (for example 503 from Varnish).


hakubo:
LGTM

PR #586, opened 2015-02-26
Replace all instances of _state access with getters/setters.

kenkouot:
WIP, need to add tests, feel free to start reviewing the code. Need to fix tests as well.
This PR aims to normalize how we define and access common state in our application. Previously, `M.provide` offered a way to declare some property arbitrarily, but no way to access it commonly and as such, we saw a few different ways of doing it: `Em.get` in Ember app code and accessing objects directly in Mercury core code.

In order to formalize it a bit and to enclose the state in a closure (to avoid abuse), I've created `M.state` as both a getter and setter for state within the application. The function can be used in the following ways:
```js
// M.prop (key: string, value: any, mutable: boolean = false): value
M.prop('foo.bar.baz', 666);
// M.prop (key: string): any
M.prop('foo.bar.baz') // => 666
``` 
@hakubo @rogatty


front/scripts/baseline/mercury.ts
```diff
@@ -41,14 +41,47 @@ module Mercury {
...
 	export module Utils {
-		export function provide(str: string, obj: any) {
+
+		var __state__: any = {};
+
+		function _getState (key: string): any {
+			var parts = key.split('.');
+			if (parts.length > 1) {
+				var i = 0;
+				var value = __state__;
```
rogatty: One `var` per function.

front/scripts/baseline/mercury.ts
```diff
@@ -41,14 +41,47 @@ module Mercury {
...
+					value = value[parts[i]];
+					i++;
+				}
+				return value;
+			}
+			return __state__[key];
+		};
+
+		function _setState (key: string, value: any): any {
+			if (value === undefined || value === null) {
```
rogatty: Not sure about this one http://stackoverflow.com/a/3550319/1050577

front/scripts/baseline/mercury.ts
```diff
@@ -41,14 +41,47 @@ module Mercury {
...
+				throw 'Cannot set property ' + key + ' to ' + value;
```
rogatty: Why can't we set property to `null`? I don't have an example but it seems possible that we'll have to reset some property.

mklucsarits: Yes I agree.

front/scripts/baseline/mercury.ts
```diff
@@ -41,14 +41,47 @@ module Mercury {
...
+			}
+			return namespacer(key, __state__, value);
+		}
+
+		export function state (key: string, value?: any): any {
+			if (value) {
```
mklucsarits: If we do want to support setting properties to ```null```, this condition will have to change also.

front/scripts/baseline/mercury.ts
```diff
@@ -41,14 +41,47 @@ module Mercury {
 	}
 
 	export module Utils {
-		export function provide(str: string, obj: any) {
+
+		var __state__: any = {};
+
+		function _getState (key: string): any {
```
hakubo: We could actually use Ember object for that right?
```
var State = Em.Object.create();

State.get('some.state');
State.set('some.state');
```

A nice outcome of this is that then state becomes observable, and I can see a way that we could use that.
What do you think?

mklucsarits:
I know we discussed this together I support this change, but I have a couple qualms with the implementation.
1) Although it may have been meaningful in the initial stages of Mercury, I think the word "state" is not appropriate in the way we're using it here now. It's just semantics, but many of these values describe properties not related to the internal state of the application. I'd rather describe them as "properties" (in the OOP sense of the word and not the JS sense) than "state". However, this is just semantics and maybe a weak argument on my part.
2) I think using the same method for a getter and setter is bad practice. I know some popular libraries do this for brevity (like ```$.css()```), but that doesn't doesn't necessarily make it a good idea. I think we should be explicit in naming functions to keep the code self-documenting and easier to read. I would suggest ```getState``` and ```setState```, or ```getProp``` and ```setProp```. In addition, using a separate getter could allow us to implement a "getWithDefault" method for things like ```apiBase```.
3) Related to (2), it would maybe be nice to have the setter accept an object of multiple key-value pairs, so as not to call the method over and over again as is done in layout.hbs.

These are minor gripes though, and this is a great step in the right direction.

front/scripts/baseline/mercury.ts
mklucsarits: I wrote my long-ish comment before I saw this, but I like this idea, @hakubo.

kenkouot: Talked to Matt about this and while it is nice, I don't support this for a couple of reasons:
* I don't want to create a dependency on Ember for state that needs to get set at the beginning of the application
* I believe, and this is becoming clearer now, that this should be for immutable state by default. Properties declared that need to be mutable should have to be explicitly set as such.
* The values stored here are intended to be shared everywhere, with no dependencies: in ember code, in mercury core js code, think of this as a replacement for the global state in the Wikia/app instance.
* The existence of this does not mean we should be any less critical of when we use a "global-like" namespace.

kenkouot:
Overall, I think the `getWithDefault` approach is pretty unnecessary. The alternative is usually less verbose and understood by every programmer. For example: `M.state('lastName') || 'Doe` works perfectly fine.

kenkouot:
I am considering the explicit setter and getter functions versus a combined approach. I think idiomatically it's fine and clean, but I understand that is a subjective opinion and would like some other viewpoints on this.

I agree that being able to set multiple properties at once would be a helpful method.

Current audit of methods:
```
function prop(key: string): value;
function prop(key: string, value: any, mutable: boolean = false): value;

// Allows you to set multiple properties, by default initialized to immutable  
function setProperties(value: any, mutable: boolean = false): value;
```

hakubo:
Hey Ken thanks for working on this, as I believe too that we need to do this right!
Therefore I have some questions:
What are the reasons behind not introducing Ember dependency to it?
AFAIK we can use Ember Objects outside Ember views and controllers.

Do we have to load all of this in head? I don't remember now but maybe this should be moved to bottom 

What are the advantages of having immutable state? 
For this we could actually use Ember.Freezable maybe?

I think we chose Ember to increase our productivity, if we work around it now should we reconsider our framework?

Thanks again!

kenkouot:
Thanks for the input Jakub!

Having immutable state and properties (and the control over their mutability) affords us a greater degree of safety and convention in our code and protects us from bugs, or abuse. Object types in JS are natively mutable while primitive types are immutable. Note, that better options for immutability are part of the ES6 proposal in the form of [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol).

Lets say we want to set an array, `protectedKeys`, that we want locked down. Once it's set, we do not want to be able to push anymore values into this array. Currently we have no way of **conveniently** doing this. Ember also doesn't offer this functionality even with Ember.Freezable. Take the following code:

```javascript
var FrozenObject = Ember.Object.extend(Ember.Freezable, {
  protectedKeys: ['onlyHasOneMember']
});

var test = new FrozenObject();
// Let's push a intruding member, developers can do this, so they will
test.get('arr').push('intrudingMember');
```

Compare that with my approach:
```javascript
// sets property, defaults to immutable
M.prop('protectedArr', ['onlyHasOneMember']);
// get the property and try to push intruder into it
M.prop('protectedArr').push('intruder'); // throws an error, cannot assign read only property
```

Immutability is helpful. It aligns with our vision leaning towards "convention over configuration". I do believe it is useful in practical cases. It makes developers think before they try to dump over overwrite variables in the global state. One developer may alter a value, without realizing it has side effects somewhere else in the application. We can protect certain variables this way. As to why I don't want to use Ember to manage this state, I'd like these options to be available right away to variables injected by the server. These properties don't need to be observable (as for the most part, we don't intend them to change so it's just overhead) and I don't want to create a dependency between Mercury JS and Ember Application code. I think it's important we keep that encapsulation clear as our application grows. As our needs change, or technology becomes available, or solutions improve–this encapsulation lends to loose coupling, which means we can be nimble in implementing positive upgrades in our code base. Think of Vignette and how easily we were able to abstract that code into an external library because of our dedication to separation of concerns.

kenkouot:
Going to merge this tomorrow if nothing huge comes up!

front/scripts/main/app.ts
```diff
@@ -38,7 +41,7 @@ App.initializer({
 		if (window.location.search.match(/debug=1/)) {
 			debug = true;
 		}
-	
+
 		App.setProperties({
 			apiBase: Mercury.apiBase || '/api/v1',
 			language: loadedLanguage || 'en',
```
dianafa: Why App language and apiBase propierties are defined in lines 18-19 if we override them here? I mean, why the `var App: any = Em.Application.create();` is not correct?

kenkouot: Good eye, fixed in dev: https://github.com/Wikia/mercury/commit/94c43e77a0d26f31e0b05b0ed582e8df5984f7d0

PR #587, opened 2015-02-26
Labels are missing for some languages

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-144

The issue was that when in layout.hbs condition #if wiki.language.user was fulfilled it expected user language even if in translations.js loading language failed (i.e. user language was not supported language). However there was still reference to Mercury._state.translations.USER_LANGUAGE which caused untranslated strings. Now, reference to Mercury._state.translations calls language returned by function in translations.js


hakubo:
I think the same fallback we have to make in app.ts
as if on server side you'll fallback from pt-pt to pt
on front end you will still try to load pt-pt and it'll fail right?

server/views/_helpers/translations.js
```diff
@@ -17,5 +23,7 @@ module.exports = function (language) {
 			}, 'Translation not found');
 		}
 	});
-	return JSON.stringify(translations);
+
+	response[foundLanguage] = translations;
```
bognix: can't you just use lang here instead of foundLang? FoundLanguage is used only here

rogatty: @bognix it's in a different scope `.some(function (lang) {`

server/views/_helpers/translations.js
```diff
@@ -2,10 +2,16 @@ var Logger = require('../../lib/Logger');
...
+		response = {};
 
-	[language, defaultLanguage].some(function (lang) {
-		var translationPath = '../../../front/locales/' + lang + '/translation.json';
+	[language, fallbackLanguage, defaultLanguage].some(function (lang) {
+		foundLanguage = lang;
+		translationPath = '../../../front/locales/' + lang + '/translation.json';
+		
 		try {
 			translations = require(translationPath);
```
rogatty: Shouldn't `foundLanguage = lang;` be under this line? It's not found before this succeed, right?

server/views/_helpers/translations.js
rogatty: @dianafa what will happen if no language was found and `foundLanguage === ''`?

server/views/layout.hbs
```diff
@@ -31,7 +31,7 @@
 
 		<script>
 			<!-- @include ../../www/front/scripts/baseline.js -->
-			M.provide('_state.translations.{{#if wiki.language.user}}{{wiki.language.user}}{{else}}en{{/if}}', {{{translations wiki.language.user}}});
+			M.provide('_state.translations', {{{json (translations wiki.language.user)}}});
```
rogatty: I don't think we need these brackets here, do we?

server/views/_helpers/translations.js
dianafa: Then it will load 'en' (defaultLanguage). So result will be {en : { translations : ... }}.

server/views/_helpers/translations.js
dianafa: yes but the next iteration will overwrite it (if this lang was not found). If the language loaded successfully next iteration'll never come.

front/scripts/main/app.ts
```diff
@@ -9,7 +9,7 @@
 declare var i18n: I18nextStatic;
 
 var App: any = Em.Application.create({
-		language: Em.getWithDefault(Mercury, 'wiki.language.user', 'en'),
+		language: Object.keys(Mercury._state.translations)[0] || 'en',
```
dianafa: What do you think about this solution?
Other options could be:
2) Involve the same fallback logic as in translation.js AGAIN here.
3) Create new Mercury var i.e. Mercury.language ?

rogatty: If this solution stays we have to fix tests because `Mercury._state` is still undefined when running Karma.

front/scripts/main/app.ts
```diff
@@ -8,22 +8,25 @@
 
 declare var i18n: I18nextStatic;
 
-var App: any = Em.Application.create({
-		language: Em.getWithDefault(Mercury, 'wiki.language.user', 'en'),
-		apiBase: Mercury.apiBase || '/api/v1'
-	});
+var App: any = Em.Application.create();
```
dianafa: Application.create options are moved to initializer to keep everything in one place. It allows for keeping code clean while computing the language (App property).

hakubo:
LGTM

tomasznapieralski:
Add support for pt-br lang

tomasznapieralski:
LGTM

server/views/_helpers/translations.js
```diff
@@ -2,10 +2,16 @@ var Logger = require('../../lib/Logger');
 
 module.exports = function (language) {
 	var translations = {},
-		defaultLanguage = 'en';
+		translationPath,
+		fallbackLanguage = language.split('-')[0],
+		defaultLanguage = 'en',
+		foundLanguage = '',
+		response = {};
```
kenkouot: I'd like to reserve the use of response for only server responses, can you change this to `wrapper` or something? @dianafa

PR #589, opened 2015-02-28
Hapi Refactor

kenkouot:
While working on a separate feature, I noticed that our routes file has really grown to be untenable. This is my fault as the patterns that I set when first working on the file didn't scale too well in readability or separation of concerns. This PR will refactor `routes.ts` to be declarative and will move all the business logic to `facets/` and `lib/` for common functions. This also moves the route bounding to `app.ts` and not individually in `routes.ts`. Overall, I think this is a great improvement.

Note, this PR changes have been intended for awhile but after seeing the [npm website](https://github.com/npm/newww) refactor to Hapi, I decided to adopt their patterns to clean up our code.


kenkouot:
This is ready for you guys to start CR'ing @mklucsarits @hakubo

server/facets/showArticle.ts
```diff
@@ -0,0 +1,122 @@
+/// <reference path="../../typings/hapi/hapi.d.ts" />
+
+import Article = require('../lib/Article');
+import Utils = require('../lib/Utils');
+import Tracking = require('../lib/Tracking');
+import Caching = require('../lib/Caching');
+import localSettings = require('../../config/localSettings');
+
+export = showArticle;
```
mklucsarits: I think the export should be at the end of the file, to keep with convention.

server/facets/api/presenters/resultWrapper.ts
```diff
@@ -0,0 +1,7 @@
+function resultWrapper (err: any, result: any) {
```
mklucsarits: Why not just name this function wrapResult, if that's how it's being called?

server/facets/showArticle.ts
```diff
@@ -0,0 +1,122 @@
...
+
+var cachingTimes = {
+	enabled: false,
+	cachingPolicy: Caching.Policy.Private,
+	varnishTTL: Caching.Interval.standard,
+	browserTTL: Caching.Interval.default
+};
+
+function showArticle (request: Hapi.Request, reply: Hapi.Response) {
```
mklucsarits: Please add :void to the end of this declaration, since it seems like that's what we're doing now.

mklucsarits:
Please add :void as the return type for the functions that need it in these files, since you're already updating all of them.

server/routes.ts
```diff
@@ -1,340 +1,82 @@
...
+		handler: require('./facets/api/search').get
+	}
+];
 
-			reply('Server status is: OK')
-				.header('X-Memory', String(memoryUsage.rss))
-				.header('X-Uptime', String(~~process.uptime()))
-				.code(200);
-		}
+indexRoutePaths = [
```
mklucsarits: I don't think "indexRoutes" is an appropriate name anymore for the article paths. We should start to make the app more non-article-centric where possible. Can it be renamed to articleRoutes maybe?

server/lib/Utils.ts
```diff
@@ -1,169 +1,183 @@
...
+export function getCachedWikiDomainName (localSettings: LocalSettings, host: string): string {
+	var wikiDomain: string;
 
-	/**
-	 * @desc Removes the port from hostname
-	 *
-	 * @param {string} host
-	 * @returns {string}
-	 */
-	export function clearHost (host: string): string {
```
mklucsarits: This is a weird function name for what it does.

kenkouot: I'll fix this in another PR

mklucsarits:
LGTM

PR #598, opened 2015-03-05
Check for error before trying to access response which can be undefined

rogatty:
This fixes `TypeError: Cannot read property 'statusCode' of undefined` reported by New Relic.


hakubo:
lgtm

PR #603, opened 2015-03-07
Update gulp-sass and foundation

rogatty:
To support io.js we have to update gulp-sass to v1.3.3 which uses node-sass v2.0.x.
This change allows Mercury to run on io.js v1.2.0.
Currently the latest io.js version is v1.5.0 but node-sass doesn't support it yet, see https://github.com/sass/node-sass/issues/694

Foundation 5.4.x didn't build properly and had to be updated too. This caused issues with margins, fixed in this PR. We should check if anything else has broken.


kenkouot:
I support switching over to io.js, at least until Joyent and the io.js team remerge their work back into core node and restructure, but we haven't talked much about outside of a bit in chat. Can you send out an e-mail with any migration suggestions? Also need a general knowledge share.

hakubo:
LGTM,

@kenkouot sure, what do you actually think about meeting this week and syncing a bit?

kenkouot:
@hakubo That sounds great! Want to schedule it?

mklucsarits:
I understand the desire to support io.js, but what are the reasons for switching from Node? I didn't hear anything about this except in this PR.

kenkouot:
Well, io.js is node! It's just node that is actively being worked on. The most striking and obvious change is that io.js uses the latest version of V8, so their are significant performance increases amongst the regular crop of patches.

For more on io.js, here is a podcast with isaacs himself: http://devchat.tv/js-jabber/147-jsj-io-js-with-isaac-schleuter-and-mikeal-rogers

@hakubo @rogatty I'm good with updating whenever you guys feel ready, just send an update with some of the observations!

rogatty:
I'm merging this to dev because https://github.com/Wikia/mercury/pull/625 broke it.

PR #605, opened 2015-03-09
Use MW API to get a random article name

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-124
https://wikia-inc.atlassian.net/browse/CONCF-181

1. Ember handles click on `Random Page`
2. Ember sends request to `Hapi @ /api/v1/article?random&titleOnly`
3. Hapi sends request to `MW @ /api.php?action=query&generator=random&grnnamespace=0&format=json`
4. Hapi replies with random article title
5. Ember makes a regular transition, exactly the same as after clicking a link

This can be optimized to return article content from `/api/v1/article?random` instead of only a title.
My solution was simpler to deliver.


typings/ember/ember.d.ts
```diff
@@ -648,7 +648,7 @@ declare module Ember {
         static metaForProperty(key: string): {};
         static isClass: boolean;
         static isMethod: boolean;
-        sendAction(action: string, context: any): void;
+        sendAction(action: string, ...contexts: any[]): void;
```
rogatty: https://github.com/emberjs/ember.js/blob/5fd2d035b30aa9ebfe73de824b3b283ec8e589cc/packages/ember-views/lib/views/component.js#L282

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -97,12 +97,32 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 			});
 		},
 
+		// This is used only in not-found.hbs template
 		expandSideNav: function (): void {
 			this.get('controller').set('sideNavCollapsed', false);
 		},
 
-		collapseSideNav: function (): void {
```
hakubo: is there anything using collapseSideNav?

hakubo:
it'd be nice to also have a route that gives you a random article... 
but this will come after we standarise our urls

server/routes.ts
```diff
@@ -57,6 +57,11 @@ unauthenticatedRoutes = [
 		method: 'GET',
 		path: localSettings.apiBase + '/search/{query}',
 		handler: require('./facets/api/search').get
+	},
+	{
+		method: 'GET',
+		path: localSettings.apiBase + '/randomArticle',
```
kenkouot: I'd like the route to follow either the same convention as comments, eg `article/comments/` or simply `article?random=1` (so built into the article route). In RESTful terms, a random article is not a new resource.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -97,12 +97,32 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
+		randomArticle: function (): void {
```
kenkouot: Generally all your interaction with the server should be done in the model. Please move this method to the ArticleMode, as a sibling to find() aka a static method. You can then easily use that method from here. Also, we should aim to keep the Application controller clean from specific resource responsibilities.

front/styles/component/_local-nav-menu.scss
```diff
@@ -40,6 +40,10 @@
 				transform: rotate(90deg);
 			}
 		}
+
+		.random-article {
```
kenkouot: Let's call it .random-article-link

front/templates/main/components/side-nav.hbs
```diff
@@ -19,5 +19,5 @@
 		<div {{action 'searchCancel'}} class='cancel'>{{i18n 'cancel-label'}}</div>
 	</div>
 	{{local-wikia-search sideNavCollapsed=isCollapsed query=searchQuery isInSearchMode=isInSearchMode}}
-	{{local-nav-menu sideNavCollapsed=isCollapsed}}
+	{{local-nav-menu sideNavCollapsed=isCollapsed randomArticle='randomArticle'}}
```
kenkouot: I'm not sure why you are passing in the string randomArticle to here and above. Can you explain?

front/scripts/main/components/LocalNavMenuComponent.ts
```diff
@@ -76,6 +76,10 @@ App.LocalNavMenuComponent = Em.Component.extend({
 
 		goBack: function (): void {
 			this.set('currentMenuItem', this.get('parentItem'));
+		},
+
+		randomArticle: function (): void {
+			this.sendAction('randomArticle');
```
Warkot: Shouldn't it be a verb instead of a noun? It's a label for an action, so maybe something like `getRandomArticle`, `goTo…` or `navigateTo…`?

kenkouot: great observation @Warkot . Yes, actions should be verbs.

front/scripts/main/routes/ApplicationRoute.ts
rogatty: There are two other actions called `collapseSideNav` (`LocalNavMenuComponent` and `LocalWikiaSearchComponent`) and they are being used. This one is not.

server/routes.ts
rogatty: Resource that we want to access here is an article title, not article. Does it make any difference?

I see one problem with your proposal - if we want to use `/article/randomTitle` then we won't be able to access an article with title `randomTitle`. We can use `/article/randomTitle/` but IMO it looks bad.

In the line 52 there is `// TODO: if you call to api/v1/article/comments/ without supplying an id, this actually calls /api/v1/article` so the comments route is not a perfect example. Shouldn't we use routes like:
- `/article/{articleTitle}`
- `/articleComments/{articleId}`
- `/articleRandomTitle`
?

front/templates/main/components/side-nav.hbs
rogatty: It's needed for `this.sendAction('randomArticle')`, see http://emberjs.com/guides/components/sending-actions-from-components-to-your-application/

We already do the same thing in https://github.com/Wikia/mercury/blob/dev/front/templates/main/application.hbs#L2

rogatty:
I've addressed all the issues, please take another look and see my comment (now hidden) about route names https://github.com/Wikia/mercury/pull/605#discussion_r26142464

server/routes.ts
Warkot: Shouldn't it be for the comments like this: `api/v1/article/{articleId}/comments/`? But I agree on the `/article/random`, that it looks just like an article with a title 'random'. Moreover, `/article/` returns a different resource – an article, while random gives just a title.

rogatty: @Warkot what if there is an article with title `123/comments/whatever`? (yes, it's possible http://gta.wikia.com/123/comments/whatever)

It's hard to have a universal route if our article titles can be so complex.

Warkot: It is, indeed. But what I had in mind is that we're using sometimes a title, sometimes an ID. In my example if we just used IDs, it would be very conventional.

kenkouot:
I'm still not quite satisfied about the API routing. It seems too close to ~RPC-ish and definitely seems to stray from a RESTful resource. The bug with article/{id}/comments/ should actually be easily fixed by first matching that route.

Beyond that, I feel that getting a random article should be an operation on the /article resource and that even though random is a collection of one article, it should be grouped together with the article resource.

So for instance, the route `/article?random=1` would be suitable to me. We're currently in ideation for our recommendation for URI reform, but we intend to recommend a few options that stick heavily to RESTful conventions. Checkout this link for a good link to RESTful nested resources: http://restfulrouting.com/mappings/resources

kenkouot:
Also, `random` seems like it is an augmentation or parameter to a resource and codifying that into it's own route, seems very RPCish to me. @drsnyder can you comment on this?

kenkouot:
Also look at the way reddit API handles random subreddits, and random listings from a specific subreddit:

```
// redirect
www.reddit.com/r/random
// json
www.reddit.com/r/random.json

// random listing from specific subreddit
// SOURCE: https://www.reddit.com/dev/api#GET_random
GET [/r/subreddit]/random
```
Reddit's API in general is a good collection of RESTful **resources** and nested resource operations. Explore more here: https://www.reddit.com/dev/api

kenkouot:
Also we should decide what casing to use for our APIs globally at Wikia.

kenkouot:
Talked with @rogatty just now and I suggest we go with:
`/article?random&titleonly` or something close to that.
In the future, our exposed APIs if they follow REST, `/article` without an ID would fetch a generalized collection, so we'd need to support other query params on GET /article anyways, ie: sort=ascdate&limit=10&filter=recent, etc..,

kenkouot:
Just a small comment, but looks good to me after! :+1:

drsnyder:
> Talked with @rogatty just now and I suggest we go with:
/article?random&titleonly or something close to that.
In the future, our exposed APIs if they follow REST, /article without an ID would fetch a generalized collection, so we'd need to support other query params on GET /article anyways, ie: sort=ascdate&limit=10&filter=recent, etc..,

Resource names should be [plural](https://github.com/Wikia/guidelines/tree/master/APIDesign#resources-and-naming). Ideally we would support `/articles/random` but `random` might conflict with a valid resource unless we specifically reserve `random` which we probably don't want to do.

Alternatively, I think `/articles/?random` is a reasonable second option. The problem with this is that it requires more documentation around the API but I think that's ok in this case.

hakubo:
Thanks for feedback!
I would proceed with /article/ as this will be consistent with our other resources

will 'fix' that with another PR

tomasznapieralski:
LGTM

PR #608, opened 2015-03-12
Folderify our gulp sass task

kenkouot:
This PR applies the same assets/folder strategy that we use for scripts, svg and templates to our scss/css files. As we laid the foundation for multi-tenancy in Mercury, @lizlux and I realized that this would make it easier to identify which assets were shared, and which were not.

For more information on gulp-folders, check out [hakubo/gulp-folders](https://github.com/hakubo/gulp-folders).


lizlux:
LGTM

front/styles/auth/app.scss
```diff
@@ -0,0 +1,3 @@
+body {
+	background: salmon;
```
kenkouot: This is just a seed file for @lizlux's login work. Salmon is also my favorite color code.

bkoval: :dancers:

hakubo:
could we just name it `main`?

kenkouot:
Sure!

gulp/tasks/sass.js
```diff
@@ -1,21 +1,25 @@
...
+gulp.task('sass', folders(paths.src, function (folder) {
 	return piper(
-		gulp.src(paths.src),
+		gulp.src([
+			'!' + path.join(paths.src, folder, paths.partials),
+			path.join(paths.src, folder, paths.compile)
+		]),
 		sass(options),
-		//currently support for map is broken
 		prefixer(['last 2 version', '> 1%', 'ie 8', 'ie 7'], {
```
lizlux: Why do we have IE 7 and 8 in here? If anything, shouldn't we have IE 9?

gulp/tasks/sass.js
```diff
@@ -1,21 +1,25 @@
...
 	options = require('../options').sass,
+	path = require('path'),
 	paths = require('../paths').styles;
 
-gulp.task('sass', function () {
+gulp.task('sass', folders(paths.src, function (folder) {
 	return piper(
-		gulp.src(paths.src),
+		gulp.src([
+			'!' + path.join(paths.src, folder, paths.partials),
```
lizlux: Just read [here](https://github.com/wearefractal/vinyl-fs#srcglobs-opt) that negations should follow positive globs. Looking at the output though, it didn't seem to make a difference for `ember-main.css` when I switched the order. Thoughts @kenkouot?

PR #610, opened 2015-03-12
CONCF-178 Load InstantGlobals from MediaWiki

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-178

I've made the resources loaded from MW tightly coupled with the ads related code. This probably can be refactored and converted to be more universal but I don't know if there is any need for loading anything else from MediaWiki.

Diana:
I've added a tests and $wgSitewideDisableAdsOnMercury variable in InstantGlobals which when set to true disables ads on Mercury. Second part of this ticket can be found here: https://github.com/Wikia/app/pull/6741


rogatty:
Tests have to be fixed too.

server/views/layout.hbs
```diff
@@ -29,10 +29,12 @@
 
 		<style><!-- @include ../../www/front/styles/baseline.css --></style>
 
+		{{> media-wiki-globals }}
```
kenkouot: Any benefit to doing this as a partial, rather than adding as a thing in `front/scripts/baseline`

server/views/_partials/media-wiki-globals.hbs
```diff
@@ -0,0 +1,7 @@
+<script>
```
kenkouot: Can we add a TODO comment that identifies this as technical debt.

front/scripts/mercury/modules/Ads.ts
```diff
@@ -53,7 +52,7 @@ module Mercury.Modules {
 						callback.call(this);
 					});
 				} else {
-					Em.Logger.error('Looks like ads asset has not been loaded');
+					Em.Logger.error('Looks like modules have not been loaded');
```
bkoval: I guess this error could tell something more :) ("AdEngine modules" instead of "modules" would certainly give more context during debugging).

bkoval:
Since this branch is here for a week, it diverged from dev too much, so there are merge conflicts. I guess you could backport dev to this branch so it's merge'able.

hakubo:
LGTM

server/views/layout.hbs
```diff
@@ -94,6 +94,8 @@
 			<script src="/front/vendor/scroll-depth/jquery.scrolldepth.js"></script>
 		<!-- endbuild -->
 
+		<script src="//{{server.mediawikiDomain}}/__load/-/cb={{wiki.cacheBuster}}&only=scripts/wikia.ext.instantGlobals,instantGlobalsOverride"></script>
```
rogatty: @gabrys: is this the correct value for the cachebuster? It's equal to `$wgCacheBuster` and is coming from `/api/v1/Mercury/WikiVariables`.

gabrys: The cb value is irrelevant (the requests are cached 5 mins on each 
tier). Cb can be even omitted.

tomasznapieralski:
LGTM

PR #611, opened 2015-03-12
updating server.view typing

lizlux:
I think this is how the typing should be defined for `server.views`. The old way was causing errors in my IDE. 

@kenkouot @hakubo


lizlux:
My IDE doesn't support all TS 1.4 features, so it was throwing errors. Code is correct though.

PR #592, opened 2015-03-05
Global Authentication (built in Hapi)

kenkouot:
Alrighty, so this is foundational work for the new global login page that the social team is doing. This PR provides a global login page that will validate your credentials with Helios and redirect to a route `/auth` implemented in your application. The `/auth` route allows your respective app to handle the credentials however you want (set your session handling, etc..,)

## Sample Flow
For instance, take this (hypothetical) flow in Mercury:

1. Start with an unauthenticated user
2. User attempts to access `//muppet.wikia.com/wiki/Elmo/edit` on Mercury *(VCL logic puts this user in Mercury test group, so requests get routed to Mercury backends)*
2. Application redirects to `//www.wikia.com/login?redirect=muppet.wikia.com/wiki/Elmo/edit` (New feature) and is served `/login` by Mercury
3. User provides credentials through log in form, hits submit
4. Mercury gets the POST, successfully grabs credentials from Helios
5. Mercury uses these credentials to set cookies
6. Mercury then redirects the user to the URI specified in the `redirect` param. MediaWiki and Mercury Ember apps are expected to understand these cookies.

## CSRF Protection
This PR also adds CSRF protection by default in our hapi routes. AJAX requests to `/login' will also need to provide a CSRF token as part of their request payload.

## Special Notes
* Helios config is stored in localSettings, which is .gitignore'd, please do not store any sensitive data in app code.

## Migration Strategy
* This PR also introduces changes to i18n usage. Please check #615 for details.

### Remaining Tasks
* [x] Add CSRF protection using [Crumb](https://github.com/hapijs/crumb)
* [x] Document new routes and expectations for referring clients
* [x] Normalize error states sent from server to client for login page
* [x] Implement "Remember Me"

* [x] Front-End for the Landing Page
* [x] Front-End for the Login Page
* [x] Validation for the Login Page
* [.] Front-End for the Signup Page
* [.] Validation for the Signup Page
* [.] Translations for the Login / Signup

PR #612, opened 2015-03-13
Soc 472 New Signup Landing Page

bkoval:
@lizlux Please take a look


hakubo:
We will need AJAX login soon anyway, why don't we try to create this login page also as a part of ember application and save our users a page refresh?

There are already written libraries that would handle that eg. https://github.com/simplabs/ember-simple-auth

I talked about that with Ken and he mentioned that you will be working on this as a static page, but the more I think about that, the more I lean towards using ajax login always. 

We could do fallback to normal form also of course, but percentage of users with disabled JS used to be so low that I don't think we have to optimise for them.

front/svg/symbols/main/wikia-logo-320.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="145px" height="41px" viewBox="0 0 145 41" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sketch="http://www.bohemiancoding.com/sketch/ns">
+    <!-- Generator: Sketch 3.2.2 (9983) - http://www.bohemiancoding.com/sketch -->
```
Warkot: What do you think about running an optimizer on this file which would remove the comments and simplify the rest?

server/routes.ts
```diff
@@ -119,7 +119,14 @@ unauthenticatedRoutes = [
 		method: 'GET',
 		path: '/breadcrumb',
 		handler: require('./facets/operations/generateCSRFView')
-	}
+	},
+
+
+	{
+		method: 'GET',
+		path: '/signupstart',
```
bkoval: @lizlux @kenkouot and everyone interested - I was struggling to find a good name of a path and view for the Signup / Login landing page. Do you have any ideas?

server/views/_layouts/wikia-static.hbs
```diff
@@ -3,7 +3,9 @@
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width">
-        <title>Login</title>
+        <title>{{{title}}}</title>
+        <link rel="stylesheet" href="/front/vendor/wikia-style-guide/dist/css/main.css">
```
bkoval: If I get this correctly, wikia-static will be used solely for login / signup, that's why I added styleguide css here. Am I correct in assuming this? @kenkouot @lizlux

server/views/signupstart.hbs
```diff
@@ -0,0 +1,28 @@
...
+        <div class="register-message">
+            <p>Register with</p>
+            <img src="/front/svg/wikia-logo-320.svg">
+        </div>
+    </div>
+    <div class="auth-options">
+        <ul class="signup-providers">
+            <li>
+                <button class="signup-provider-facebook">
+                    CONNECT WITH FACEBOOK
```
bkoval: Just a reminder that i18n will be handled in a different ticket.

server/views/signupstart.hbs
```diff
@@ -0,0 +1,28 @@
+<section class="signup-start">
+    <div class="register-message-wrapper">
+        <div class="register-message">
+            <p>Register with</p>
+            <img src="/front/svg/wikia-logo-320.svg">
```
Warkot: Including this via inline CSS would save one request. What do you think about such way? @kvas-damian

gulp/tasks/symbols.js
```diff
@@ -10,6 +10,9 @@ var gulp = require('gulp'),
 gulp.task('symbols', folders(paths.src, function (folder) {
 	return piper(
 		gulp.src(path.join(paths.src, folder, paths.files)),
+		gulpif('**/wikia-logo-320.svg', piper(
```
hakubo: Just place it in svg/images and it'll get copied
I have just noticed that this is missing from doc, will add it in a minute

but as a general rule, we should avoid such hardcoded file names

Warkot:
In https://github.com/Wikia/mercury/blob/SOC-472/front/images/faniverse-bg-544.png all the lines look very distorted, how was it resized?

lizlux:
@hakubo, the reason we're building this login flow outside of ember is b/c it needs to work within oasis too. These pages will be loaded into an iframe in oasis and we don't want to load the big ember library just for a simple login page. See https://github.com/Wikia/mercury/pull/592 for notes on the workflow.

bkoval:
@Warkot - right, there's something wrong with it, good catch. I didn't create the file myself, I asked a designer for it, so I didn't verify it along the way, will definitely get back to this on Monday!

front/styles/auth/signupstartpage.scss
```diff
@@ -0,0 +1,93 @@
+body {
+	background-color: #062d50;
+}
+
+.signup-start {
+	background-image: url('/front/images/faniverse-bg-544.png');
+	background-position: center center;
+	background-repeat: no-repeat;
+	background-size: cover;
```
lizlux: I played around with changing the screen size and I think I like it better without this property. The media query does a good job of handling smaller and larger mobile screens, so the background image doesn't need to stretch and distort.

hakubo:
@lizlux Ken mentioned to me that discourse is using ember as an embedable library and they are fine with it, performance wise.

Since our grand goal is to use Mercury everywhere does it make sense to optimise for oasis now?

kenkouot:
We're not optimizing for oasis, but really building this outside of Ember is part of our efforts to **not rebuild a monolith in Ember**. This login site is part of a bigger picture that may include micro sites and other corporate efforts. Think of Ember largely as our consumption app right now.

kenkouot:
In addition, my hapi work already supports AJAX login. So you could indeed implement this for the Ember app as well. However, I do NOT think that we need to build our global login page into Ember. Also, the global login page is going to be served from www.wikia.com/login. You can't change subdomains without a page refresh.

kenkouot:
Also, my words re: discourse is it's fine for a lazy loaded component but not necessarily fine to incur the load penalty of Ember (+ our Ember app) for a simple login page that requires a page refresh in and out of it anyways (due to subdomain restrictions).

front/styles/auth/signupstartpage.scss
```diff
@@ -0,0 +1,93 @@
+body {
```
kenkouot: there is something gravely wrong with this filename

server/views/_layouts/wikia-static.hbs
kenkouot: Assumption is correct!

bkoval:
I think the security and reusability of the iframe solution are biggest pros of this approach for now. From what I hear we should stick to it in this implementation so let's leave the static version for now :)

lizlux:
@bkoval, Crtistina has suggested we make the path to this page /join. As for what we call it in the code, we could do some variation of "join" or we could call it something like "auth-landing".

front/styles/auth/signupstartpage.scss
bkoval: True that, will change it as "signupstart" was just a placeholder name for the whole view.

front/styles/auth/signupstartpage.scss
```diff
@@ -0,0 +1,93 @@
...
+	p {
+		margin-bottom: 15px;
+	}
+}
+
+.signup-provider-facebook {
+	background-color: #38528a;
+	border: none;
+	margin-bottom: 15px;
+}
```
lizlux: Let's add `text-transform: uppercase` here and for the other buttons.

server/views/_layouts/wikia-static.hbs
lizlux: I was thinking it would be better to include the style guide's `main.css` inside `app.scss` to avoid the additional request. The other option I see is to use grunt to build one css file that includes all the css files needed. Let's discuss which way makes the most sense.

server/facets/auth/signupstart.ts
```diff
@@ -0,0 +1,19 @@
...
+
+function get (request: Hapi.Request, reply: any): void {
+	if (request.auth.isAuthenticated) {
+		return reply.redirect(request.query.redirect || '/');
+	}
+
+	return reply.view(
+		'signupstart',
+		{
+			title: "Signup"
```
lizlux: single quotes

front/styles/auth/signupstartpage.scss
lizlux: Can we call it `auth-landing`? This is the landing page for login and signup so "auth" would be more accurate than "signup".

server/views/signupstart.hbs
```diff
@@ -0,0 +1,28 @@
+<section class="signup-start">
+    <div class="register-message-wrapper">
```
lizlux: Is there a reason we need this wrapping element and `.sign-in-message-wrapper`? Can the HTML in this file be reduced?

bkoval:
@lizlux I decided to leave background-size: cover for now, as it might happen that the phone / tablet will have a resolution bigger than 768px (eg in landscape mode). Without this property bg would just stick to it's natural size, leaving empty space around. What do you think?

lizlux:
@bkoval, after 768px (including ipad in landscape) let's consider that desktop. We're going to have a different mockup for that screen size. I don't think we should support over 768px for this design.

bkoval:
I left wikia-logo-320.svg in /front/svg/images for reference but inlined it as data-uri into html so we don't make an additional request.

kvas-damian:
@bkoval manually inlined file sounds like a small hack - when I modify source file I expect to update it automatically, which does not take place here. Similar scenario is valid when you start from template, I'd like to get original file - how do I know which one is correct.

IMO we should create gulp task which will handle that. If you think it's out of scope for this ticket please create ticket and add ```TODO```.

bkoval:
@kvas-damian Yes, I was meant to create another ticket for that, thanks!

front/styles/auth/_landing-page.scss
```diff
@@ -0,0 +1,106 @@
...
+}
+
+ul {
+	list-style-type: none;
+}
+
+.sign-in-message {
+	display: inline-block;
+	margin-bottom: 0;
+	padding-right: 1.5rem;
```
lizlux: Looking at the mockup and the UI, I think this value is closer to `.75rem` than `1.5`.

front/styles/auth/_landing-page.scss
```diff
@@ -0,0 +1,106 @@
+body {
+	background-color: #062d50;
+}
+
+.signup-start {
```
lizlux: What do you think about renaming "signup-start" to "auth-landing-page" so they are consistent?

front/styles/auth/_landing-page.scss
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+.signup-provider-email {
+	background-color: transparent;
+	border-color: #fff;
+
+	&:focus, &:hover {
+		background-color: transparent;
+		border-color: #1A5EB8;
```
lizlux: Were these colors part of the design spec? They seem funny when viewed in chrome on desktop. If they are intentionally there, can we import the `$color-blue-light` var from the style guide?

server/views/auth-landing-page.hbs
```diff
@@ -0,0 +1,26 @@
...
+        </div>
+    </div>
+    <div class="auth-options">
+        <ul class="signup-providers">
+            <li>
+                <button class="signup-provider-facebook">
```
lizlux: I know these look like buttons but they act more like links. Let's use `a.button` instead.

front/styles/auth/_landing-page.scss
lizlux: Also, this component is really similar to the footer styles I'm working on. We probably want to unify these styles between the landing page and the footer.

lizlux:
I'm going to merge this PR so I have the code available to my branch. We can continue to improve upon this code after it's merged.

bkoval:
@kvas-damian https://wikia-inc.atlassian.net/browse/SOC-545

PR #614, opened 2015-03-13
Soc 479

lizlux:
This PR sets up header and footer styles. It's missing the "fanverse" background for now, but I'd like to use the same image that @bkoval is using in his PR https://github.com/Wikia/mercury/pull/612/files, so this can be updated later. 

Note: One difference between this PR and Bart's is that I'm including the style guide css inside app.scss as opposed to with a separate request. We'll have to decide what's best and sync up our two implementations. 

To do:
- [x] Use the "fanverse" background images that Bart will commit with his PR 

@bkoval @kenkouot


bkoval:
@lizlux I'm in for the single SCSS file, so I updated my PR to use the same way :)

server/facets/auth/login.ts
```diff
@@ -63,11 +63,19 @@ function authenticate (username: string, password: string, callback: AuthCallbac
...
+
 	if (request.auth.isAuthenticated) {
 		return reply.redirect(request.query.redirect || '/');
 	}
 
-	return reply.view('login', null, {
+	// TODO: use i18n once it's set up (SOC-526)
+	context = {
+		header: 'Welcome Back!',
+		footer: 'Don\'t have an account? <a href="#">Register now</a>'
```
bkoval: How about extracting all the HTML into Handlebars and passing two strings ('Don\'t have an account?' and 'Register now'). This way we won't need to use html directly in the i18n and all the view-rendering part will be done exclusively in .hbs file :)

front/styles/auth/app.scss
```diff
@@ -1,3 +1,37 @@
...
+}
+
+.auth-header, .auth-footer {
+	background: #092344;
+	color: #FFF;
+	text-align: center;
+}
+
+.auth-header {
+	font-size: 18px;
```
bkoval: What do you think about using rem values for fonts? Style-guide uses base font of 16px so this would be 1.125 rem (not sure if CSS will notice third number after the comma :)

front/styles/auth/app.scss
```diff
@@ -1,3 +1,37 @@
...
+}
+
+.auth-footer {
+	color: #FFFFFF;
+	font-size: 12px;
+	letter-spacing: 1.25px;
+	line-height: 18px;
+	padding: 15px;
+
+	a {
```
bkoval: It looks like it should have font-weight: bold as well :)

server/facets/auth/login.ts
lizlux: This has been an age-old issue with i18n when there's text followed by a call to action / link. The problem is that in some languages the link would make more sense if it were positioned before or within the text. Separating them into two different messages makes it hard to do this. The solution in the past has been to use wiktext for link so it can be placed anywhere in the message and the user can still avoid using HTML markup.

bkoval:
@lizlux Let's merge it, I don't see any major issues with the code and I like the new CSS structure, will apply it to landing page styles as well.

front/styles/auth/base/_body.scss
```diff
@@ -0,0 +1,9 @@
+body {
+	background-color: $color-blue-dark;
```
kenkouot: Actually don't use these absolute color variables directly, but rather, make your own settings file and declare semantic variables like:

```
$body-background-color: $color-blue-dark;
```

This is just 1 layer of abstraction that allows us to cascade changes very easily and from a central file (settings).

front/styles/auth/base/_buttons.scss
```diff
@@ -0,0 +1,6 @@
+button, a.button {
```
kenkouot: So this should really be in modules, base should just be for bare HTML elements styles and their pseudos

front/styles/auth/layout/_footer.scss
```diff
@@ -0,0 +1,14 @@
+.auth-footer {
+	color: $color-white;
+	text-align: center;
+	color: #FFFFFF;
```
kenkouot: repeat

front/styles/auth/layout/_footer.scss
```diff
@@ -0,0 +1,14 @@
+.auth-footer {
+	color: $color-white;
```
kenkouot: auth-footer color information should be in `modules/` (same for anchor tag below)

front/styles/auth/layout/_header.scss
```diff
@@ -0,0 +1,7 @@
+.auth-header {
+	color: $color-white;
```
kenkouot: non-layout information should be in theme or modules

front/styles/auth/layout/_splash.scss
```diff
@@ -0,0 +1,19 @@
+.splash {
+	background-image: url('/front/images/faniverse-bg-544.png');
```
kenkouot: this bg-image should be in themes/

front/styles/auth/module/_landing-page.scss
```diff
@@ -1,44 +1,19 @@
-body {
```
kenkouot: I think a lot of this page should be in `layout` or `theme`. To help, you can give the body tag a class using hapi.

server/facets/auth/join.ts
```diff
@@ -1,15 +1,34 @@
 /// <reference path='../../../typings/hapi/hapi.d.ts' />
 
+interface getContext {
```
kenkouot: interface names should be PascalCased. The name should also be a bit more descriptive, "JoinViewContext" or something

server/views/auth-landing-page.hbs
```diff
@@ -1,26 +1,24 @@
...
...0409.7%20C302.6%20405.7%20304.7%20404%20307.3%20404%20C309.6%20404%20311.4%20404.9%20312.5%20406.4%20L312.5%20413%20L312.5%20413%20Z%22%20sketch%3Atype%3D%22MSShapeGroup%22%2F%3E%3Cpath%20d%3D%22M230.1%20393.9%20C232.9%20393.9%20235.2%20391.6%20235.2%20388.9%20C235.2%20386.1%20233%20383.8%20230.2%20383.8%20C227.4%20383.8%20225.2%20386.1%20225.2%20388.9%20C225.1%20391.6%20227.4%20393.9%20230.2%20393.9%20%22%20sketch%3Atype%3D%22MSShapeGroup%22%2F%3E%3Cpath%20d%3D%22M225.1%20396.7%20L225.1%20422.8%20L237.7%20422.8%20L237.7%20417.3%20L233.8%20417.3%20L233.8%20402.2%20L237.7%20402.2%20L237.7%20396.7%20L225.1%20396.7%20%22%20sketch%3Atype%3D%22MSShapeGroup%22%2F%3E%3Cpath%20d%3D%22M284.2%20383.8%20C281.4%20383.8%20279.2%20386.1%20279.2%20388.9%20C279.2%20391.6%20281.5%20393.9%20284.2%20393.9%20C287%20393.9%20289.3%20391.6%20289.3%20388.9%20C289.2%20386.1%20287%20383.8%20284.2%20383.8%20%22%20sketch%3Atype%3D%22MSShapeGroup%22%2F%3E%3Cpath%20d%3D%22M276.6%20396.7%20L276.6%20402.2%20L280.6%20402.2%20L280.6%20417.3%20L276.6%20417.3%20L276.6%20422.8%20L289.2%20422.8%20L289.2%20396.7%20L276.6%20396.7%20%22%20sketch%3Atype%3D%22MSShapeGroup%22%2F%3E%3C%2Fg%3E%3C%2Fg%3E%3C%2Fg%3E%3C%2Fsvg%3E">
```
kenkouot: Have you seen how we do SVGs in the main mercury app? We compile them into a single file, then inline that file and reference them in our HTML using `<symbols>`. It's an SVG sprite approach that allows for great reusability and maintainability.

kenkouot:
Looking good so far. Remember that in SMACCS, declaration merging of CSS classes is okay. That is to say, you may have a class `.foo-module` is expressed like this:
```scss
// in modules/
.foo-module {
  width: 200px;
  height: 200px;
  border: 1px solid;
}

// in themes/
.foo-module {
  border-color: pink
}
```

Getting to know what type of styles belong in each SMACSS folder will help with this in the long term. Let me know if you need the e-book :dancer:

server/views/auth-landing-page.hbs
lizlux: @bkoval are you already handling this in a separate ticket?

PR #615, opened 2015-03-14
i18next revamp

kenkouot:
Alright, so this PR is now ready for code review. This PR accomplishes a few things:
* It implements i18next support on the server side through a plugin I wrote: [kenkouot/hapi-i18next](https://github.com/kenkouot/hapi-i18next/)
* It fixes a misuse of i18next namespaces and syntax. More on that below.

You'll notice that in the changeset, all translations have been changed. This is because we were misusing i18next's concept of namespaces. In i18next, namespaces are analogous to the file names, and the objects inside the files are accessed using `module.key` syntax whereas namespaces/filenames are expressed using `mynamespace:module.key`. i18next supports multiple namespaces and setting a default namespace. By fixing this, we gain the ability to break out our translations file-wise AND use the objects within the json document. For instance:

*In main.json*
```json
{
  "app":{
    "my-key":"foo"
  },
  "lightbox":{
    "my-key":"foo"
  }
}
```

*in other.json*
```json
{
  "onboarding":{
    "my-key":"foo"
  }
}
```

This, of course, opens doors to being able to request granular and smaller translations asynchronously, which may be very useful for asynchronously loaded modules in the future. As of now, it just gives us greater flexibility in translations since CrowdIn manages the reusability of translations in it's platform.

## Migration Strategy
For the most part, the migration strategy is simply to switch from using `:` to `.` in your translation keys. The `:` is meant to declare a namespace, which in i18next is intended to reference a different file which may already be loaded or loaded on demand. See the change set for example.

How to declare translations:
```handlebars
// from default ns
{{i18n 'app.hub-games'}}
// from another ns other than default
{{i18n 'otherns:my.key'}}
```

@lizlux @bkoval @hakubo


front/scripts/main/helpers/i18nHelper.ts
```diff
@@ -1,10 +1,19 @@
 /// <reference path="../app.ts" />
 
-Em.Handlebars.registerBoundHelper('i18n', function (value: string, options: any) {
-	var params: {
-			[key: string]: string;
-		} = {},
-		namespace = 'app';
+Em.Handlebars.registerBoundHelper('i18n', function () {
+var options = arguments[arguments.length - 1],
+	args = Array.prototype.slice.apply(arguments, [0, -1]),
```
mklucsarits: I think this slice will result in an empty array if options are not passed in, ie. ```{{i18n 'app.article-suggestions-label'}}```

Also, indentation needs fixing.

server/i18n-hapi.ts
```diff
@@ -0,0 +1,67 @@
...
+interface HapiPluginRegister {
+	(server: any, options: any, next: Function): void;
+	attributes?: {
+		name: string;
+		version: string;
+	};
+}
+
+var defaults = {
+	supportedLngs: ['en'],
```
mklucsarits: Can we please add one more letter and use "lang" to shorten "language"? Lng makes me think of long or longitude.

hakubo:
If we'll use i18next on the server side, that would mean we want be able to cache content of articles for logged in users. Can we think of a solution that would enable us, to leverage Varnishes for our content for everyone?

kenkouot:
@hakubo You're jumping the gun here. This is for pages that aren't Ember pages, eg, auth and onboarding flows and in the future, micro and marketing sites (think exec pages, careers, etc...,).

This has very little effect on varnish as the content of those sites will likely either not be cached in the same way. For instance, www.wikia.com/login will support i18n by parsing the accept-language header and set the language for the response that way. Default en will be cached, but the other languages will be served dynamically (only way to do this for that domain, if we hosted a login page at de.wikia.com/login, that would be cached separately as well). From a feature perspective, wikia.com/login will likely have a dropdown to set language cookie, but that likely won't be cached.

front/scripts/main/helpers/i18nHelper.ts
lizlux: This might be written more clearly/succinctly using `Array.prototype.pop` to remove the last argument and assign it to `options` at the same time.

server/app.ts
```diff
@@ -50,11 +50,24 @@ var plugins = [
...
+				cookieName: 'lang',
+				detectLngFromHeaders: true
+			}
+		}
 	}
 ];
 
 server.register(plugins, (err: any) => {
 	if (err) {
+		console.log(err);
```
bkoval: Development leftover? :)

lizlux:
LGTM (just one minor comment)

PR #618, opened 2015-03-17
CE-1431 Added /editorPreview/ route to preview given HTML in Mercury skin

gabrys:
This pull requests adds capability to preview an HTML in a Mercury skin.

The API is: POST /editor_preview/ the POST params are:

* parserOutput: JSON as returned by ArticleApi (article as JSON)
* title: article title
* mwHash: sha1 hash with salt to prove the parserOutput was indeed generated by MediaWiki

This commit reveals an issue of not properly encoding article title passed by MediaWiki API in https://github.com/Wikia/mercury/pull/618/files#diff-5a151be511cd8e3c3d6a4e795aad8d46R120 . We chose to fix this issue on the MediaWiki side https://wikia-inc.atlassian.net/browse/HG-564, but I would like us to revisit this issue and move the fix to Mercury, so everything works as expected.

The salt for recomputing mwHash is expected to appear as an environmental variable MW_PREVIEW_SALT.

At one point, we might want to simplify the whole flow by making /editor_preview/ accepting wiki text to parse instead of parsed HTML, but now it would mean crossing multiple layers between Mercury and MW Parser (the mwHash becomes useless then).

### DIFF
--- a/front/templates/main/application.hbs
+++ b/front/templates/main/application.hbs
@@ -4,6 +4,8 @@
 			{{outlet}}
 		</section>
 	</div>
+	{{loading-spinner active=isLoading}}
+	{{outlet 'lightbox'}}
 {{else}}
 	{{smart-banner isVisible=view.smartBannerVisible}}
 	{{site-head smartBannerVisible=view.smartBannerVisible sideNavCollapsed=view.sideNavCollapsed trackClick='trackClick'}}
--- a/front/templates/main/application.hbs
+++ b/front/templates/main/application.hbs
@@ -11,9 +11,7 @@
 	{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
 	<div class='page-wrapper' {{bind-attr lang=language.content}} {{bind-attr dir=language.contentDir}}>
 		<section class='wiki-container'>
-			{{#unless editorPreview}}
-				<h2 class='wiki-title' {{action 'trackClick' 'wikiname'}}>{{#link-to 'article' mainPageTitle alt=siteName}}{{siteName}}{{/link-to}}</h2>
-			{{/unless}}
+			<h2 class='wiki-title' {{action 'trackClick' 'wikiname'}}>{{#link-to 'article' mainPageTitle alt=siteName}}{{siteName}}{{/link-to}}</h2>
 			{{outlet}}
 		</section>
 	</div>
--- a/front/templates/main/application.hbs
+++ b/front/templates/main/application.hbs
@@ -1,12 +1,22 @@
-{{smart-banner isVisible=view.smartBannerVisible}}
-{{site-head smartBannerVisible=view.smartBannerVisible sideNavCollapsed=view.sideNavCollapsed trackClick='trackClick'}}
-{{side-nav isCollapsed=view.sideNavCollapsed loadRandomArticle='loadRandomArticle'}}
-{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
-<div class='page-wrapper' {{bind-attr lang=language.content}} {{bind-attr dir=language.contentDir}}>
-	<section class='wiki-container'>
-		<h2 class='wiki-title' {{action 'trackClick' 'wikiname'}}>{{#link-to 'article' mainPageTitle alt=siteName}}{{siteName}}{{/link-to}}</h2>
-		{{outlet}}
-	</section>
-</div>
-{{loading-spinner active=isLoading}}
-{{outlet 'lightbox'}}
+{{#if editorPreview}}
+	<div class="page-wrapper editor-preview" {{bind-attr lang=language.content}} {{bind-attr dir=language.contentDir}}>
+		<section class='wiki-container'>
+			{{outlet}}
+		</section>
+	</div>
+{{else}}
+	{{smart-banner isVisible=view.smartBannerVisible}}
+	{{site-head smartBannerVisible=view.smartBannerVisible sideNavCollapsed=view.sideNavCollapsed trackClick='trackClick'}}
+	{{side-nav isCollapsed=view.sideNavCollapsed loadRandomArticle='loadRandomArticle'}}
+	{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
+	<div class='page-wrapper' {{bind-attr lang=language.content}} {{bind-attr dir=language.contentDir}}>
+		<section class='wiki-container'>
+			{{#unless editorPreview}}
+				<h2 class='wiki-title' {{action 'trackClick' 'wikiname'}}>{{#link-to 'article' mainPageTitle alt=siteName}}{{siteName}}{{/link-to}}</h2>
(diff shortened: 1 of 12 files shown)


server/facets/showArticle.ts
```diff
@@ -14,11 +14,30 @@ var cachingTimes = {
...
+
+	return (computedHash === mwHash);
+}
+
 function showArticle (request: Hapi.Request, reply: Hapi.Response): void {
 	var path: string = request.path,
 		wikiDomain: string = Utils.getCachedWikiDomainName(localSettings, request.headers.host);
 
-	if (path === '/' || path === '/wiki/') {
+	if (path === '/editor_preview/') {
```
kenkouot: This route deserves it's own facet, right now it just seems kind of thrown in here.
I suggest `server/facets/editorPreview`

server/routes.ts
```diff
@@ -57,6 +57,11 @@ unauthenticatedRoutes = [
 		method: 'GET',
 		path: localSettings.apiBase + '/search/{query}',
 		handler: require('./facets/api/search').get
+	},
+	{
+		method: 'POST',
+		path: '/editor_preview/',
```
kenkouot: I'm unsure on the snake_case here. @hakubo on the front end, we have camelCase for query params. We need to decide now on a casing for our exposed URIs (api and application). I prefer snake_case personally as it is case insensitive but still have delimiters for spaces. If we change that, I'd prefer to change our the `camelCase` query params in the Ember routing too.

So, snake_case everywhere?

server/facets/showArticle.ts
```diff
@@ -14,11 +14,30 @@ var cachingTimes = {
 	browserTTL: Caching.Interval.default
 };
 
+function verifyMwHash (parserOutput: string, mwHash: string) {
```
kenkouot: this could likely be it's own facet, toss it in `operations/verifyMWHash`.

kenkouot:
Can I get some context on what this feature is? Why it requires a POST? Eventually we're moving most of the article resource responsibilities to the service being developed by @Wikia/platform-team. This service is in active development so we really shouldn't be building in new APIs, but rather making feature requests to platform.

server/routes.ts
```diff
@@ -57,6 +57,11 @@ unauthenticatedRoutes = [
 		method: 'GET',
 		path: localSettings.apiBase + '/search/{query}',
 		handler: require('./facets/api/search').get
+	},
+	{
+		method: 'POST',
```
kenkouot: This resource should be a GET. POST is to be used for create. Use url params to send params to the GET. Here is a resource for learning about REST conventions http://www.restapitutorial.com/lessons/httpmethods.html

server/facets/showArticle.ts
kenkouot: this editor preview code path likely deserves its own facet

kenkouot:
Good implementation with minor details, however, I'm not convinced that this fits in with our API patterns, now or in the future. I'd like to revisit the API parameters (why does it need parserOutput AND title) and the output.

server/routes.ts
kenkouot: Actually, I'll just add this casing to our URI structure initiative which is just kicking off today. Will report back (expect an email soon!)

server/routes.ts
gabrys: You can't transfer too much data over GET.

kenkouot: Yeah, this sort of adds to my point about this API design though. I'm not sure it's great design to GET or POST literally an entire parser output to Mercury, just to get back a view wrapped page. It's these types of patterns that come from product requests that end up growing our codebase in unmaintainable ways.

There are many ways to do this correctly. I assume this is for changes that aren't totally published yet. Ideas:

* MW app can generate a temporary URL with perishable article URL. MW app then can:
  * Simply curl Mercury with temporary URL params and get back the payload that way, without building any additional APIs
  * or MW app passes a Mercury API the perishable URL, Mercury returns a payload.

There are many other questions to be answered too. If this is done internally, behind consul from MW App to Mercury, then you don't need the hash either.

Overall, I think this feature could benefit from more thoughtful consideration on the absolute best way to implement this now and thinking towards the future. As it stands now, I am not wholly satisfied with the mechanics of it.

gabrys: Actually I agree with you 100%. We discussed this with @hakubo already 
and it looks like the perfect solution is Mercury /editor_preview/ to 
accept the wiki text instead of the full parser output (as explained in 
the pull request description).

I feel like this part could be done better by you guys, while I can 
enjoy having this little shortcut before all parts missing are taken 
care of (for starters: we need POST-ing to MediaWiki API from Mercury, 
which we never needed before, then lot's of design decisions on the way 
to parser: Mercury API, Article API, ArtictlesAsJson and some other in 
between).

I hope what I hacked around here is not "too hacky" and having this 
(this being the preview of articles in Mercury when editing from Oasis) 
sooner rather than done better gives us more value at this moment (with 
open discussion on how to implement this properly).

server/routes.ts
gabrys: Thanks! I'll be happy to apply whatever convention we settle for this.

server/routes.ts
gabrys: BTW, the idea of passing the whole parser output is because for RTE editor (the pre-VE one) needs a call to MW backend to generate the wiki text.

The same code then converts the wiki text to HTML for all the skins, so it was just easier to use the same function to generate HTML for Mercury as well (rather than exiting the API method in half and executing the second half from Mercury API context).

front/templates/main/application.hbs
```diff
@@ -1,12 +1,22 @@
...
+		</section>
+	</div>
+{{else}}
+	{{smart-banner isVisible=view.smartBannerVisible}}
+	{{site-head smartBannerVisible=view.smartBannerVisible sideNavCollapsed=view.sideNavCollapsed trackClick='trackClick'}}
+	{{side-nav isCollapsed=view.sideNavCollapsed}}
+	{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
+	<div class="page-wrapper" {{bind-attr lang=language.content}} {{bind-attr dir=language.contentDir}}>
+		<section class='wiki-container'>
+			{{#unless editorPreview}}
```
hakubo: is it needed here?
you already check for it on line 1

on the other hand it would be nice to have separate template for preview and have shared partial/view
This can be done though as a separate ticket

hakubo:
LGTM

It's a good start, we can improve upon it when this becomes more needed.

@kenkouot - I like snake_case as well, can you create ticket for Mercury team to standardise it, if you don't have one already?

@gabrys thanks for the work, as we talked please create ticket to play a TODO role, with the ideas that we had on how we can improve this solution.

server/routes.ts
kenkouot: Just sent out an e-mail to source the truth on this topic!

kenkouot:
@hakubo ticket created and assigned to myself!

tomasznapieralski:
LGTM

PR #619, opened 2015-03-17
Enable Optimizely

mklucsarits:
We don't yet have a hot-swappable configuration tool in place, and we need to enable Optimizely this week for an experiment that was expected to run two weeks ago, so this seems like the only thing to do. Alternative suggestions to enable this on production are welcome.

This change will result in the Optimizely JS snippet being loaded for every Mercury user.


hakubo:
this could come from process.env.optimizely or something like that and
or we could use consul to store such configuration

kenkouot:
merge dev into branch to fix build error

mklucsarits:
Changed to use env flag.

mklucsarits:
I need to get this on production on Friday, so please take a look at it before I merge.

kenkouot:
LGTM

PR #620, opened 2015-03-17
Add a rootElement for Ember to render into

kenkouot:
Without a specified root element, Ember defaults it's root element to render into to the `body` tag. This is confusing to the [Google PageSpeed](https://developers.google.com/speed/pagespeed/insights/) which sees that we have page content after scripts in the body even though we in fact loaded the scripts after the content.

This fixes that issue by explicitly setting rootElement to a node that is inserted before those scripts and should increase our PageSpeed score.

New mobile algorithms from Google take into effect 4/21.

@tedgill  @hobbs

### DIFF
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -14,7 +14,11 @@ interface Window {
 
 declare var i18n: I18nextStatic;
 
-var App: any = Em.Application.create();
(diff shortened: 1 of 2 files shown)


mklucsarits:
Looks OK to me.

PR #621, opened 2015-03-18
SOC-479 updating scss categorization

lizlux:
This is in response to code review feedback in https://github.com/Wikia/mercury/pull/614.


bkoval:
LGTM :dancers:

front/styles/auth/layout/_body.scss
```diff
@@ -1,9 +1,8 @@
 body {
-	background-color: $color-blue-dark;
 	display: flex;
 	flex-direction: column;
 	height: 100vh;
-	width: 100vw;
 	justify-content: flex-end;
 	margin: auto;
+	width: 100vw;
```
bkoval: I'm still wondering if we should use these or stick to %. I'll see how does it look like on Android default browsers under KitKat, but caniuse says they're not supported at all

lizlux: going to remove these for now since they are unnecessary on the body tag, but it would be good to know if we can use `vw` and `vh` going forward.

kenkouot:
There is a SCSS error, check jenkins console output

PR #622, opened 2015-03-18
Soc 484 New Login Page

bkoval:
@lizlux This is an implementation of the Front-End for New Login page, with floating labels and non-empty fields validation.

https://wikia-inc.atlassian.net/browse/SOC-484


bkoval:
I decided to try out the pure-css floating labels, the only question is about situation where user types something into the field that passes :valid pseudocheck in css and blurs the field afterwards. In that case, the floating label will be gone until the field is focused again.

front/styles/auth/app.scss
```diff
@@ -10,6 +10,7 @@
 
 // components
 @import 'component/close';
```
kenkouot: Components aren't really a SMACSS thing, we just have them in our Ember scss folder because we have Ember components, which are more encapsulated than modules (but should they be?). Just use the modules folder here!

front/styles/auth/base/_anchors.scss
```diff
@@ -1,3 +1,12 @@
 a {
 	color: inherit;
 }
+
+.forgot-password {
+	color: $color-blue-light;
```
kenkouot: don't use `color-*` variables here. Instead, declare a semantic variable in `auth/_settings` and use that variable.

```scss
/* in _anchors */
a {
  color: $anchor-color;
}

/* in settings *?
$anchor-color: $color-blue-light
```

front/styles/auth/base/_buttons.scss
```diff
@@ -4,3 +4,8 @@ button, a.button {
 	width: 100%;
 }
 
+.submit-button {
```
kenkouot: same thing, shouldn't be in base, should have a button module

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,46 @@
+.input-container {
+	position: relative;
+}
+
+input[type="text"],
```
kenkouot: haha, now these should be in base :D

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,46 @@
...
+	border: none;
+	border-bottom: 1px solid $color-blue-gray;
+	border-radius: 0;
+	box-shadow: none;
+	height: auto;
+	margin-top: 0;
+	padding: 1.25rem 0 0.5rem;
+}
+
+input.sign-in-password {
```
kenkouot: you don't need to prefix this with a element selector, just use the class

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,46 @@
...
+	color: $color-blue-gray;
+	padding: 0;
+	pointer-events: none;
+	position: absolute;
+	top: 1.25rem;
+	transition: font-size ease-in 0.15s, top ease-in 0.15s;
+}
+
+input[type="text"]:focus  + .floating-label,
+input[type="password"]:focus + .floating-label {
```
kenkouot: space

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,46 @@
...
+	font-size: 0.75rem;
+	top: 0;
+}
+
+input[type="text"]:valid  + .floating-label,
+input[type="password"]:valid + .floating-label {
+	display: none;
+}
+
+input[type="text"]:valid:focus  + .floating-label,
```
kenkouot: single quotes in SCSS

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,46 @@
...
+input.sign-in-password {
+	margin-bottom: rem-calc(5px);
+}
+
+.floating-label {
+	color: $color-blue-gray;
+	padding: 0;
+	pointer-events: none;
+	position: absolute;
+	top: 1.25rem;
```
kenkouot: don't use REM for anything but text (we may need to clarify this in documentation)

lizlux:
@bkoval @kenkouot once https://github.com/Wikia/mercury/pull/621/files is merged it will also help with some of the issues in this PR, like removing the components dir.

front/styles/auth/component/_input.scss
bkoval: Yup, work still in progress, will do sir!

front/styles/auth/component/_input.scss
bkoval: @kenkouot Could we treat this as an exception? This padding needs to match the height of the text + a bit of extra space. If I use px here it could be not enough for the desktops where rem base goes from 16px to 18px, right?

bkoval:
@lizlux @kenkouot I'll make the components -> modules switch for scss after we merge pull request from @lizlux with scss reorganization.

front/scripts/auth/auth.ts
```diff
@@ -0,0 +1,48 @@
...
+		} else {
+			this.deactivateSubmit();
+		}
+	};
+
+	/**
+	 * Checks if both username and password fields are not empty
+	 * @returns {boolean}
+	 */
+	validateIfNotEmpty = ():boolean => {
```
lizlux: Can we change the name of this function to `isNotEmpty`? The current name sounds to me like "only validate this input if it's not empty".

front/templates/main/login.hbs
```diff
@@ -1,3 +1,3 @@
-<a href='/wiki/Special:UserLogin' class='external login'>
+<a href='/join' class='external login'>
```
lizlux: This url should include the redirect location so we can pass it through the landing page to the login page.

server/views/_layouts/wikia-static.hbs
```diff
@@ -22,5 +22,8 @@
 		{{#unless hideFooter }}
 		<footer class="auth-footer splash splash-bottom">Don\'t have an account? <a href="#">Register now</a></footer>
 		{{/unless}}
-	</body>
+        {{#if loadScripts}}
+        <script src="/front/scripts/auth.js"></script>
```
lizlux: We'll probably want to set up a gulp tasks to concatenate and minify JS files. Also it would pull from `www/front`.

server/views/login.hbs
```diff
@@ -1,10 +1,17 @@
...
+        </div>
+        <div class="input-container">
+            <input class="sign-in-password" type="password" name="password" id="loginPassword" required />
+            <label class="floating-label" for="loginPassword">Password</label>
+        </div>
+        <a class="forgot-password" href="#">Forgot password?</a>
         <input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
         <input type="checkbox" name="remember"> Remember me<br>
-        <button id="loginSubmit" type="submit">Log in</button>
+        <button id="loginSubmit" type="submit" disabled=true>Sign in</button>
```
lizlux: Should just be `disabled`. (See http://www.w3.org/TR/html5/forms.html#enabling-and-disabling-form-controls:-the-disabled-attribute and http://www.w3.org/TR/html5/infrastructure.html#boolean-attribute)

bkoval: Makes sense, but I'll be setting it to true later anyway. Hence the explicit declaration :)

server/views/_layouts/wikia-static.hbs
bkoval: Agreed.

front/scripts/auth/auth.ts
bkoval: Sure, let's do so.

server/views/login.hbs
```diff
@@ -1,10 +1,17 @@
...
+            <input type="text" name="username" id="loginUsername" required />
+            <label class="floating-label" for="loginUsername">Username</label>
+        </div>
+        <div class="input-container">
+            <input class="sign-in-password" type="password" name="password" id="loginPassword" required />
+            <label class="floating-label" for="loginPassword">Password</label>
+        </div>
+        <a class="forgot-password" href="#">Forgot password?</a>
         <input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
         <input type="checkbox" name="remember"> Remember me<br>
```
lizlux: let's wrap this input and text in a `<label>` tag so the text will be clickable and get rid of the `<br>` tag.

front/styles/auth/base/_input.scss
```diff
@@ -0,0 +1,26 @@
...
+	border: none;
+	border-bottom: 1px solid $color-blue-gray;
+	border-radius: 0;
+	box-shadow: none;
+	height: auto;
+	margin-top: 0;
+	padding: 1.25rem 0 0.5rem;
+}
+
+input[type='text']:focus  + .floating-label,
```
lizlux: I'd probably put all the styles for the floating label feature into `module/floating-label`, since that seems like a feature/module as opposed to base styles for all inputs. Possibly including L12 where the padding is specific to this feature. That will create a clear distinction between a regular input and a floating-label input.

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,16 @@
+.input-container {
+	position: relative;
+}
+
+input.sign-in-password {
```
lizlux: I feel like this doesn't belong here. Maybe in themes?

front/styles/auth/component/_input.scss
```diff
@@ -0,0 +1,16 @@
...
+	margin-bottom: rem-calc(5px);
```
lizlux: Why use `rem-calc` here?

lizlux:
https://github.com/Wikia/mercury/pull/621 is merged now.

front/scripts/auth/auth.ts
```diff
@@ -0,0 +1,48 @@
...
+		return !!this.loginUsername.value.length &&
```
lizlux: Looks like typescript wants you to put parentheses around this statement. 

From the console log during startup:
`front/scripts/auth/auth.ts(33,10): 2322 Type 'number' is not assignable to type 'boolean'.`

`return !!(this.loginUsername.value.length && this.loginPassword.value.length);` fixes it.

front/styles/auth/component/_input.scss
kenkouot: I really wish UX didn't change the base font on breakpoints :x:

front/styles/auth/base/_input.scss
bkoval: Hey, what do you mean by L12?

front/scripts/auth/auth.ts
bkoval: I added it, noticed typescript complaining about it after I broke the line before committing :)

front/scripts/auth/auth.ts
```diff
@@ -0,0 +1,48 @@
...
+		this.loginPassword = <HTMLInputElement> window.document.querySelector('#loginPassword');
+		this.loginSubmit = <HTMLButtonElement> window.document.querySelector('#loginSubmit');
+		this.onInput();
+		window.document.querySelector('#loginForm').addEventListener('input', this.onInput);
+	}
+
+	/**
+	 * Activates / deactivates submit button in the login form
+	 */
+	onInput = ():void => {
```
kenkouot: actually you can just declare it like this:
```typescript
onInput (): void {

Comment (kenkouot) [inline front/scripts/auth/auth.ts]: I think the way you are doing it now (assignment without var) declares a global

File: front/scripts/auth/auth.ts [hunk 19]
```diff
@@ -0,0 +1,48 @@
...
+			this.deactivateSubmit();
+		}
+	};
+
+	/**
+	 * Checks if both username and password fields are not empty
+	 * @returns {boolean}
+	 */
+	isNotEmpty = ():boolean => {
+		return !!(this.loginUsername.value.length &&
```
kenkouot: this is fine on oneline imo

front/styles/auth/_settings.scss
```diff
@@ -1,3 +1,9 @@
+$auth-content-background-color: $color-white;
```
kenkouot: can you sort these keys

front/styles/auth/base/_anchors.scss
```diff
@@ -1,3 +1,12 @@
 a {
 	color: inherit;
 }
+
+.forgotten-password {
```
kenkouot: class selectors don't belong in `base/`, move this to a module

front/styles/auth/base/_input.scss
kenkouot: Line 12

front/styles/auth/module/_input.scss
```diff
@@ -0,0 +1,16 @@
+.input-container {
+	position: relative;
+}
+
+input.sign-in-password {
+	margin-bottom: rem-calc(5px);
+}
+
+.floating-label {
```
kenkouot: No class selectors in `base/`

server/views/_layouts/wikia-static.hbs
kenkouot: Use the  tag

kenkouot:
Can you add a better PR desc and title?

front/scripts/auth/auth.ts
bkoval: I think it does not, by looking at the .js output. I used it to be sure that a correct scope is used when invoking these methods from the constructor. I'll verify your way though and check if it assigns the method to an object inside the constructor or it's prototype.

bkoval: Unfortunately, looks like the way it is implemented now (adding methods to the constructor) is the only way that works, so I would leave it for now :)

front/styles/auth/_settings.scss
bkoval: done!

bkoval:
Hey @lizlux I found out that our way of adding style-guide to the app.scss is breaking the gulp-sass build for production (gulp build --type=production), so I moved this import to settings.scss file.

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-567 - ticket for adding returnTo param to the url :)

lizlux:
@bkoval the gulp-sass issue has been fixed, so if you merge from dev up into your branch it will start working again. Sorry if I didn't communicate that properly. You should be able to put the style guide back into app.scss.

front/styles/auth/app.scss
```diff
@@ -1,6 +1,5 @@
-// compiled style guide css and scss variables
+// compiled style guide css
 @import '../../vendor/wikia-style-guide/dist/css/main.css';
-@import '../../vendor/wikia-style-guide/src/scss/lib/settings';
```
lizlux: This should be put pack.

server/views/login.hbs
```diff
@@ -1,10 +1,17 @@
...
+        </div>
+        <div class="input-container">
+            <input class="sign-in-password" type="password" name="password" id="loginPassword" required />
+            <label class="floating-label" for="loginPassword">Password</label>
+        </div>
+        <a class="forgotten-password" href="#">Forgot password?</a>
         <input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
         <input type="checkbox" name="remember"> Remember me<br>
-        <button id="loginSubmit" type="submit">Log in</button>
+        <button id="loginSubmit" type="submit" disabled=true>Sign in</button>
```
lizlux: This should be just `disabled` instead of `disabled=true`

server/views/login.hbs
```diff
@@ -1,10 +1,17 @@
...
+            <input type="text" name="username" id="loginUsername" required />
+            <label class="floating-label" for="loginUsername">Username</label>
+        </div>
+        <div class="input-container">
+            <input class="sign-in-password" type="password" name="password" id="loginPassword" required />
+            <label class="floating-label" for="loginPassword">Password</label>
+        </div>
+        <a class="forgotten-password" href="#">Forgot password?</a>
         <input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
         <input type="checkbox" name="remember"> Remember me<br>
```
lizlux: Let's put a label around this to make the text "remember me" clickable so it will toggle the checkbox on click.

server/views/login.hbs
```diff
@@ -1,10 +1,17 @@
...
-        <input type="password" name="password" id="loginPassword" placeholder="Your pw" required />
+        <div class="input-container">
+            <input type="text" name="username" id="loginUsername" required />
+            <label class="floating-label" for="loginUsername">Username</label>
+        </div>
+        <div class="input-container">
+            <input class="sign-in-password" type="password" name="password" id="loginPassword" required />
+            <label class="floating-label" for="loginPassword">Password</label>
+        </div>
+        <a class="forgotten-password" href="#">Forgot password?</a>
```
lizlux: I like to put `tabindex="-1"` on links that are inside forms because users generally don't think of links as the type of thing you tab into. I'd like to be able to tab from the password input to the remember me checkbox without stopping on the forgot password link.

server/views/_layouts/wikia-static.hbs
```diff
@@ -22,5 +24,10 @@
 		{{#unless hideFooter }}
 		<footer class="auth-footer splash splash-bottom">Don\'t have an account? <a href="#">Register now</a></footer>
 		{{/unless}}
-	</body>
+        {{#if loadScripts}}
+        <!-- build:js scripts/auth.js -->
+        <script src="/front/scripts/auth.js"></script>
```
lizlux: When I run this in production mode, using `npm run start`, this file is 404.

front/templates/main/login.hbs
lizlux: Thanks for creating that ticket! https://wikia-inc.atlassian.net/browse/SOC-567

front/scripts/auth/auth.ts
kenkouot: Yeah, but do we have to use the assignment operator? can't you just write it without the equal sign?

bkoval: I tried that, it won't work. I think since the method needs to be available on the constructed object, it needs to be assigned somewhere with the "=" operator. Either like this or declared on the top of the class and assigned inside the constructor. @kenkouot would you be satisfied with a different option?

server/views/_layouts/wikia-static.hbs
```diff
@@ -5,7 +5,9 @@
 		<meta name="viewport" content="width=device-width">
 
 		<title>{{{title}}}</title>
-		<link rel="stylesheet" href="/front/styles/auth/app.css">
+        <!-- build:css /front/styles/auth/app.css -->
```
rogatty: Use tabs and not spaces. In WebStorm it's:
_Settings / Editor / Code Style -> Default Indent Options -> Use tab character_

front/scripts/auth/auth.ts
lizlux: I played around with this and the following worked fine: 

```javascript
	onInput (): void {
		if (this.isNotEmpty()) {
			this.activateSubmit();
		} else {
			this.deactivateSubmit();
		}
	}
```
The JS output was the same either way though.

lizlux:
LGTM after the whitespace is adjusted in wikia-static.hbs. Also it would be good to update auth.ts based on Ken's suggestion.

PR #626, opened 2015-03-20
Handle authentication and user state on Ember front-end side

mklucsarits:


front/scripts/main/User.ts
```diff
@@ -0,0 +1,26 @@
...
+		var userId = this.get('userId');
+		if (userId !== null) {
+			App.UserModel.find({userId: userId}).then((result: any) => {
+				this.set('model', result);
+			}, function () {});
+		}
+		this._super();
+	},
+
+	isAuthenticated: function (): boolean {
```
hakubo: Ember.computed.bool

front/templates/main/components/site-head.hbs
```diff
@@ -1,7 +1,13 @@
 <div {{action 'expandSideNav'}} class='side-nav-toggle'>
     {{svg 'nav' viewBox='0 0 24 14' class='icon nav'}}
 </div>
-{{render 'login'}}
+{{#if user.isAuthenticated}}
+	<figure class='avatar'>
+		<img {{bind-attr src=user.model.avatarPath}} {{bind-attr alt=user.model.name}}>
```
hakubo:
```hbs
<img src={{user.model.avatarPath}} alt={{user.model.name}}>
```

front/scripts/main/User.ts
```diff
@@ -0,0 +1,26 @@
+/// <reference path="./app.ts" />
+/// <reference path="./models/UserModel.ts" />
+'use strict';
+
+App.User = Em.Object.extend({
```
hakubo: Why do we have User and UserModel separately?

Can't User just be model?

front/scripts/main/User.ts
```diff
@@ -0,0 +1,26 @@
...
+	model: null,
+
+	init: function () {
+		var userId = this.get('userId');
+		if (userId !== null) {
+			App.UserModel.find({userId: userId}).then((result: any) => {
+				this.set('model', result);
+			}, function () {});
```
hakubo: You implemented error handling in UserModel but here you don't use it. Can you explain what is the reason?

front/scripts/main/User.ts
mklucsarits: The reasoning was that the User object is for the current user only, while UserModel could be for any user, and utilized for such cases as profiles, messaging, etc. -- any place user data is needed.

hakubo: so how about
```
App.CurrentUser
```

front/scripts/main/User.ts
mklucsarits: I think I meant to add something here but never did. I wasn't sure what should happen if the user data wasn't retrieved as expected. Any ideas @kenkouot ?

mklucsarits:
This branch needs to be merged into the dev branch now, so I'm closing this PR.

PR #628, opened 2015-03-23
We need to normalize it for DT before they get updated

hakubo:
Before DT gets updated we need to normalise the name for build


Warkot:
LGTM :heavy_check_mark:

Warkot:
Reminder: we're waiting for [this](https://github.com/Wikia/deploytools-plugins/pull/44/files) to get released.

PR #629, opened 2015-03-23
add build task to package.json

hakubo:
Having this script
we can teach DTools to run
`npm run build`
instead of
`gulp build --type prod`

so it does not have to know the 'internals' of the service

It is mostly 1:1 mapping to what DT used to do
having it here will free us to tweak what build do, without having to deploy DT


hakubo:
Please review: @Warkot @bognix

Warkot:
LGTM :ok_hand:

bognix:
LGTM

PR #630, opened 2015-03-23
ADEN-1834 Interstitial ads on mercury

nandy-andy:
We implemented interstitial ads on Oasis and we're using there `UiFactory` and its regular modals. We also wanted to have the interstitial product on Mercury and changes below allows us to have it.

The changes includes:
* new simple ads lightbox,
* exposed `openMercuryAdLightbox()` function which opens *only* the ads lightbox.

This pull request is connected to https://github.com/Wikia/app/pull/6758.

To test if it's work open a mercury page in your browser and execute in JavaScript console following code:
```js
window.loadCustomAd && window.loadCustomAd({
  type: 'modal',
  width: 300,
  height: 250,
  code: 'An ad placeholder'
});
```

### DIFF
--- a/front/scripts/main/controllers/AdsLightboxController.ts
+++ b/front/scripts/main/controllers/AdsLightboxController.ts
@@ -11,7 +11,7 @@ App.AdsLightboxController = App.LightboxController.extend({
 
 	footerHidden: true,
 
-	header: 'Advertisment',
+	header: 'Advertisement',
 
 	init: function (): void {
 		this._super();
--- a/front/scripts/main/controllers/AdsLightboxController.ts
+++ b/front/scripts/main/controllers/AdsLightboxController.ts
@@ -4,13 +4,8 @@ App.AdsLightboxController = App.LightboxController.extend({
 	data: {
 		contents: null
 	},
-
-	contents: Em.computed.alias(
-		'data.contents'
-	),
-
+	contents: Em.computed.alias('data.contents'),
 	footerHidden: true,
-
 	header: 'Advertisement',
 
 	init: function (): void {
--- a/front/scripts/main/controllers/AdsLightboxController.ts
+++ b/front/scripts/main/controllers/AdsLightboxController.ts
@@ -1,0 +1,27 @@
+/// <reference path="./LightboxController.ts" />
+
+App.AdsLightboxController = App.LightboxController.extend({
+	data: {
+		contents: null
+	},
+
+	contents: Em.computed.alias(
+		'data.contents'
+	),
+
+	footerHidden: true,
+
+	header: 'Advertisment',
+
+	init: function (): void {
+		this._super();
+	},
+
+	reset: function (): void {
+		this.setProperties({
+			data: {}
+		});
+
+		this._super();
+	}
+});
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -8,6 +8,14 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 		return params;
 	},
 
+	activate: function() {
+		window.openMercuryAdLightbox = (content): void => {
+			// TODO: WIP: right now display first media on the page in lightbox but later on we want to display an ad
+			console.log('content: ', content);
+			this.send('openLightbox', 'media-lightbox', {mediaRef: 0});
+		}
+	},
+
 	actions: {
 		loading: function (): void {
 			this.controller.showLoader();
--- a/front/scripts/main/routes/ApplicationRoute.ts
+++ b/front/scripts/main/routes/ApplicationRoute.ts
@@ -15,7 +15,7 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
 		 * If there is and it's in a form of prestitial/interstitial the ad server calls our exposed JS function to
 		 * display the ad in a form of modal. The ticket connected to the changes: ADEN-1834.
 		 */
-		window.openMercuryAdLightbox = (contents): void => {
+		window.openMercuryAdLightbox = (contents: any): void => {
 			this.send('openLightbox', 'ads-lightbox', {contents: contents});
 		}
 	},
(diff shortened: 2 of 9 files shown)


front/scripts/main/controllers/AdsLightboxController.ts
```diff
@@ -0,0 +1,27 @@
...
+		contents: null
+	},
+
+	contents: Em.computed.alias(
+		'data.contents'
+	),
+
+	footerHidden: true,
+
+	header: 'Advertisment',
```
rogatty: Shouldn't it be localized? If not it should be `Advertisement`

front/scripts/main/controllers/AdsLightboxController.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference path="./LightboxController.ts" />
+
+App.AdsLightboxController = App.LightboxController.extend({
+	data: {
+		contents: null
+	},
+
+	contents: Em.computed.alias(
+		'data.contents'
+	),
```
rogatty: No need for splitting it into multiple lines.

front/scripts/main/controllers/AdsLightboxController.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference path="./LightboxController.ts" />
+
+App.AdsLightboxController = App.LightboxController.extend({
+	data: {
+		contents: null
+	},
+
+	contents: Em.computed.alias(
```
rogatty: This could be `Em.computed.oneWay`
http://emberjs.com/api/#method_computed_oneWay

front/scripts/main/controllers/AdsLightboxController.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference path="./LightboxController.ts" />
+
+App.AdsLightboxController = App.LightboxController.extend({
+	data: {
+		contents: null
+	},
```
rogatty: I think that `data` property is not needed at all. In `reset()` you can just set `contents` to null.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -8,6 +8,18 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
 	},
 
+	activate: function (): void {
+		/**
+		 * This global function is being used by our AdEngine code to provide prestitial/interstitial ads
+		 * It works in similar way on Oasis: we call ads server (DFP) to check if there is targeted ad unit for a user.
+		 * If there is and it's in a form of prestitial/interstitial the ad server calls our exposed JS function to
+		 * display the ad in a form of modal. The ticket connected to the changes: ADEN-1834.
+		 */
+		window.openMercuryAdLightbox = (contents): void => {
```
rogatty: Can we somehow put it into `Mercury` scope instead of a `window` one?

front/scripts/main/views/AdsLightboxView.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference path="../app.ts" />
+/// <reference path="./LightboxView.ts" />
+'use strict';
+
+App.AdsLightboxView = App.LightboxView.extend({
+	classNames: ['ads-lightbox'],
+	status: 'opening',
+
+	didInsertElement: function (): void {
+		this.set('status', 'open');
```
rogatty: Looks like something that could be extracted to a mixin. Now we have 3 views doing the same thing:
- `AdsLightboxView`
- `MapLightboxView`
- `MediaLightboxView`

Not necessarily in this PR ;)

nandy-andy: Yeah, I was thinking about similar thing. I was thinking of changing lightboxs to components instead of mixins. But I'm not too "fluent" with Ember and I'm not really sure what is a mixin :dash:

front/scripts/main/controllers/AdsLightboxController.ts
nandy-andy: Two reasons for not localizing it:
* it's an experiment,
* I wanted to keep it consistent with Wikia/app - we didn't translate it there either; once we do it there we'll do it here as well because it would mean the experiment is finished and we want a perfect, finished product ;)

Thanks for finding a typo! ;)

front/scripts/main/controllers/AdsLightboxController.ts
nandy-andy: Got it! Thanks :bow:

front/scripts/main/controllers/AdsLightboxController.ts
nandy-andy: Yeah, but it means we'll complicate the logic in `ApplicationRoute::openLightbox()`, won't we? Do we want it?

front/scripts/main/routes/ApplicationRoute.ts
nandy-andy: It sounds good. I managed to put it to extend `Mercury.Modules.Ads`. Unfortunately it creates TS error: `front/scripts/main/routes/ApplicationRoute.ts(18,23): 2339 Property 'openLightbox' does not exist on type 'typeof Ads'.` Where is the typing file I should modify to fix it?

front/scripts/main/controllers/AdsLightboxController.ts
rogatty: Oh, I see. You're right, let's leave it like that.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -8,6 +8,18 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, {
...
 	},
 
+	activate: function (): void {
+		/**
+		 * This global function is being used by our AdEngine code to provide prestitial/interstitial ads
+		 * It works in similar way on Oasis: we call ads server (DFP) to check if there is targeted ad unit for a user.
+		 * If there is and it's in a form of prestitial/interstitial the ad server calls our exposed JS function to
+		 * display the ad in a form of modal. The ticket connected to the changes: ADEN-1834.
+		 */
+		Mercury.Modules.Ads.openLightbox = (contents: any): void => {
```
rogatty: How about adding `public openLightbox (contents: any): void {}` in Ads.ts? Will it fix TS warning?
Put there some explanatory comment too :)

nandy-andy: Yeah, I tried since I didn't find anything connected to Ads module in `mercury.d.ts`. Unfortunately, it didn't fix the error. I can double-check it once I'm done with the stuff I'm working now.

rogatty:
LGTM

PR #634, opened 2015-03-26
Fixed variable clobbering and use Boom to create error object

kenkouot:
This fixes a tslint warning that was introduced with merging #618. This fixes the variable clobbering and handles the error from the getWikiVariables request.


kenkouot:
@gabrys

rogatty:
LGTM

PR #637, opened 2015-03-26
SOC-475 New Login Validation Errors

bkoval:
@lizlux @kenkouot Please, take a look.

Ticket: https://wikia-inc.atlassian.net/browse/SOC-475


front/styles/auth/base/_input.scss
```diff
@@ -0,0 +1,12 @@
+input[type='text'],
+input[type='text']:focus,
+input[type='password'],
+input[type='password']:focus {
+	background-color: #fff;
+	border: none;
+	border-bottom: 1px solid $label-color-inactive;
+	border-radius: 0;
+	box-shadow: 0 0 0 1000px #fff inset !important;
```
bkoval: I know it's ugly and I tried with input:-webkit-autofill, but it didn't work. !important on bg doesn't work for me, this is the only way I could think of to get rid of the yellow background around auto-suggested field.

front/styles/auth/base/_input.scss
```diff
@@ -6,7 +6,7 @@ input[type='password']:focus {
 	border: none;
 	border-bottom: 1px solid $label-color-inactive;
 	border-radius: 0;
-	box-shadow: none;
+	box-shadow: 0 0 0 1000px $auth-content-background-color inset !important;
```
bkoval: I'll ad a descr in a comment above I guess, so everyone knows this is fixing -webkit-autofill behavior.

lizlux: What about:

```css
input:-webkit-autofill {
-webkit-box-shadow: 0 0 0px 1000px #fff inset;
}
```

That's more self-documenting code and would have less possible unintended consequences on other browsers. A comment wouldn't hurt though too.

front/styles/auth/_settings.scss
```diff
@@ -3,6 +3,7 @@ $body-background-color: $color-blue-dark;
 $body-text-color: $color-gray-dark;
 $footer-text-color: $color-white;
 $forgotten-password-link-color: $color-blue-light;
+$form-error-color: #d0021b;
```
lizlux: Can we use the style guide for the for error styles? I realize that they don't match up exactly b/c the style guide has the whole input outlined and our mockups have just the bottom border colored red, but there's still things like the error message color that we can take from the style guide. 

https://github.com/Wikia/style-guide/blob/dev/src/scss/lib/components/forms.scss#L230
http://wikia.github.io/style-guide/section/forms/

server/facets/auth/login.ts
```diff
@@ -70,6 +69,19 @@ function authenticate (username: string, password: string, callback: AuthCallbac
...
+/**
+ * Obtains a proper message to display in Front-End based on Helios response
+ * @param {number} statusCode
+ * @returns {string}
+ */
+function getFormError (statusCode: number): String {
+	//TODO i18n
+	if (statusCode === 401) {
+		return 'Hm, we don\'t recognize these credentials. Please try again or register a new account.';
+	} else {
```
lizlux: I think the code looks cleaner when you get rid of the `else` clause if the `if` clause has a return statement.

server/facets/auth/login.ts
```diff
@@ -70,6 +69,19 @@ function authenticate (username: string, password: string, callback: AuthCallbac
 	});
 }
 
+/**
+ * Obtains a proper message to display in Front-End based on Helios response
+ * @param {number} statusCode
+ * @returns {string}
+ */
+function getFormError (statusCode: number): String {
+	//TODO i18n
```
lizlux: We should be able to i18n now since https://github.com/Wikia/mercury/pull/615 was merged, right? Is there anything blocking us at this point?

front/styles/auth/module/_form-error.scss
```diff
@@ -0,0 +1,5 @@
+.form-error {
```
mklucsarits: Let's not create a new module file just an error message. I would suggest creating a forms.scss, or better yet, use the style guide.

front/styles/auth/base/_input.scss
bkoval: Yeah, I tried that before experimenting with !important. Unfortunately, it doesn't work at all ;(

front/styles/auth/_settings.scss
bkoval: Yeah, I'll give it a try. Although, New Login form uses much different input styles so let's verify these will match with the error styling from style-guide.

bkoval: Done.

bkoval:
@kenkouot @lizlux I tried 18n-ing the thing:
1) I've set up a new file, auth.json
2) I made 2 sections there, "login" and "common"
3) When I try accessing a key like : 'auth:common.server-error' the translation does not appear, I can still see the key being displayed
4) When I replace it with a translation from a main namespace, e.g 'main:app.article-empty-label' it works fine

I commited it anyway for you to see how it looks, maybe you guys know where the problem is?

front/styles/auth/base/_input.scss
```diff
@@ -6,7 +6,8 @@ input[type='password']:focus {
 	border: none;
 	border-bottom: 1px solid $label-color-inactive;
 	border-radius: 0;
-	box-shadow: none;
+	//Fix for yellow background for autofill in webkit browsers
+	box-shadow: 0 0 0 1000px $auth-content-background-color inset;
```
bkoval: @lizlux I tested :-webkit-autofill pseudoselector and it didn't work, but the box-shadow thing is working without !important now. Can't figure out more elegant solution for now.

server/facets/auth/login.ts
```diff
@@ -70,6 +69,17 @@ function authenticate (username: string, password: string, callback: AuthCallbac
 	});
 }
 
+/**
+ * Obtains a proper message to display in Front-End based on Helios response
+ * @param {number} statusCode
```
lizlux: @kenkouot @bkoval what do you guys think about leaving off these doc block hints now since typescript makes these types more obvious? Just wondering if there's a convention around it.

server/app.ts
```diff
@@ -56,7 +56,10 @@ var plugins = [
 		options: {
 			i18nextOptions: {
 				resGetPath: path.join(__dirname, '..', 'front/locales/__lng__/__ns__.json'),
-				ns: 'main',
+				ns: {
```
kenkouot: @bkoval Pull dev and your messages should work now. As you can see in the changes below this line, the plugin simply wasn't configured for multiple namespaces. Now it is and everything works dandily!

server/facets/auth/login.ts
```diff
@@ -70,6 +69,17 @@ function authenticate (username: string, password: string, callback: AuthCallbac
...
+ * @returns {string}
+ */
+function getFormError (statusCode: number): String {
```
kenkouot: I think this logic is needless here. Just pass the statusCode to the template and in the template, check the value of the statusCode and show the message declaratively there. EG:

```handlebars
{{#if code !== 200}}
  {{#if code === 401}}
      {{i18n 'auth:...'}}
  {{else}}
      {{i18n 'auth:...'}}
  {{/if}}
{{/if}}
```

kenkouot:
Great work! Make sure jenkins is clean

server/facets/auth/login.ts
lizlux: Yeah, now that Ken got the i18n working, I can see that this method of passing in the i18n key dynamically doesn't work. So it sounds like Ken's suggestion above is the way to go. 
screen shot 2015-03-30 at 5 56 17 pm

server/app.ts
bkoval: Awesome, thanks! :dancers:

server/facets/auth/login.ts
bkoval: OK, I'll just leave the method description then :)

server/facets/auth/login.ts
bkoval: Hey @lizlux, I just tested it and passing keys through context object works well:

zrzut ekranu 2015-03-31 13 16 10

I'm trying to decide which way is more appropriate for us here. I'm slightly leaning towards leaving the logic producing i18n keys in the logical (ts) portion of the backend as this way seems more maintanable to me. I'm just thinking of a situation where Helios team adds more verbose messages to their error responses and then we'd be able to create more accurate error messages, but it would require a bit bigger logic to be wrote. This way it would be easier to change things around.

My guess is that this way we keep bigger separation between our logic and presentation layers. I'd feel more comfortable to perform statusCode mapping in the "controller" layer.

What do you think?

bkoval:
Thanks for updating the branch and quick fix @kenkouot !!! I addressed the comment about removing docs and left my answer here replying to the i18n rendering comment. I'll move this ticket to QA since this shouldn't be a blocker, but let's keep the discussion going :dancer:

PR #643, opened 2015-03-26
SOC-567 Set New Login redirect on Mercury's Front-End side

bkoval:
@kenkouot @lizlux @rogatty This should make the user come back to where he/she was before clicking CTA to log in.

https://wikia-inc.atlassian.net/browse/SOC-567


front/scripts/main/components/SiteHeadComponent.ts
```diff
@@ -7,6 +7,9 @@ App.SiteHeadComponent = Em.Component.extend(App.TrackClickMixin, {
 	classNames: ['site-head'],
 	tagName: 'nav',
 	headroom: null,
+	joinPathContext: {
```
lizlux: Would it make more sense to put this in something like a LoginComponent.ts file? So the context doesn't have to be passed to through site-head to login?

rogatty:
LGTM

kenkouot:
This looks good to me, @mklucsarits have a look since you have Ember auth work too.

mklucsarits:
LGTM

kenkouot:
PS Don't merge this yet. We're gonna merge hapi-auth tomorrow and we'll wait to merge this later.

PR #645, opened 2015-03-26
Fixing gulp sass error

lizlux:
We'll upgrade gulp-sass with https://wikia-inc.atlassian.net/browse/HG-573, but this is the workaround until then to make sure tests pass.

@kenkouot @bkoval


bkoval:
LGTM

PR #649, opened 2015-03-27
CONCF-126 Support articles with special chars in title

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-126
https://wikia-inc.atlassian.net/browse/CONCF-74

We shouldn't try to decode titles which weren't encoded in the first place. This PR cleans it up and fixes search results.

### DIFF
--- a/front/scripts/main/routes/ArticleRoute.ts
+++ b/front/scripts/main/routes/ArticleRoute.ts
@@ -19,7 +19,7 @@ App.ArticleRoute = Em.Route.extend({
 		}
 
 		if (!transition.params.article.title) {
-			this.transitionTo('article', Mercury.wiki.mainPageTitle);
+			this.transitionTo('article', M.String.titleToUri(Mercury.wiki.mainPageTitle));
 		}
 	},
 
--- a/front/scripts/main/routes/ArticleRoute.ts
+++ b/front/scripts/main/routes/ArticleRoute.ts
@@ -26,7 +26,7 @@ App.ArticleRoute = Em.Route.extend({
 		// `Transition #1: detected abort.`
 		// This is caused by the transition below but doesn't mean any additional requests.
 		this.transitionTo('article',
-			Mercury.Utils.String.sanitize(title)
+			M.String.sanitize(title)
 		);
 	},
 
--- a/front/scripts/main/routes/ArticleRoute.ts
+++ b/front/scripts/main/routes/ArticleRoute.ts
@@ -19,7 +19,7 @@ App.ArticleRoute = Em.Route.extend({
 		}
 
 		if (!transition.params.article.title) {
-			this.transitionTo('article', M.String.titleToUri(Mercury.wiki.mainPageTitle));
+			this.transitionTo('article', Mercury.wiki.mainPageTitle);
 		}
 	},
 
--- a/front/scripts/main/routes/ArticleRoute.ts
+++ b/front/scripts/main/routes/ArticleRoute.ts
@@ -14,13 +14,20 @@ App.ArticleRoute = Em.Route.extend({
 	},
 
 	beforeModel: function (transition: EmberStates.Transition) {
+		var title = transition.params.article.title ?
+			transition.params.article.title.replace('wiki/', ''):
+			Mercury.wiki.mainPageTitle;
+
 		if (Mercury.error) {
 			transition.abort();
 		}
 
-		if (!transition.params.article.title) {
-			this.transitionTo('article', Mercury.wiki.mainPageTitle);
-		}
+		// If you try to access article with not-yet-sanitized title you can see in logs:
+		// `Transition #1: detected abort.`
+		// This is caused by the transition below but doesn't mean any additional requests.
+		this.transitionTo('article',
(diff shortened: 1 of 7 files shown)


rogatty:
Needs fixes for articles with ? in title.

front/scripts/mercury/utils/string.ts
```diff
@@ -2,23 +2,23 @@
...
+	export function normalize (str: string = ''): string {
+		return str
 			.replace(/_/g, ' ')
 			.replace(/\s+/g, ' ');
 	}
+
+	export function titleToUri (title: string = '', encodeQuestionMarks = false): string {
+		var uri = encodeURI(
+			title
+				.replace(/\s/g, '_')
```
Warkot: I know that's not exactly in the scope of this ticket, but `\s` stands for any whitespace—my question is, can there be any other whitespace character than space? And that's an actual question, 'cause I don't know, but I'd like to. ;)

Warkot: Therefore, do we need to use a regexp?

test/specs/front/scripts/mercury/utils/string.js
```diff
@@ -1,54 +1,54 @@
 /* global App, resetMercuryBaseline */
 QUnit.module('String util functions');
 
-QUnit.test('sanitizing URL', function () {
+QUnit.test('sanitizing title', function () {
```
Warkot: Maybe use 'unifying' if you drop sanitizing everywhere?

front/scripts/mercury/utils/string.ts
rogatty: http://stackoverflow.com/questions/1144783/replacing-all-occurrences-of-a-string-in-javascript

We have to use `/g` to replace *all* whitespaces. `titleToUri` is currently used to create search results URIs only. I'm not sure if MediaWiki search API can return whitespace other than space. Better to be safe :)

test/specs/front/scripts/mercury/utils/string.js
rogatty: This is a group of tests for `normalize` and `titleToUri` methods.
I think that _sanitizing_ describes their purpose better than _unifying_. We're sanitizing titles for specific purpose, right? Maybe I misinterpret this word :speak_no_evil:

Warkot: I just wanted the name to be consistent with the test cases names. And it's fine. :+1:

front/scripts/mercury/utils/string.ts
Warkot: :+1:

hakubo:
LGTM

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -14,26 +14,19 @@ App.ArticleRoute = Em.Route.extend({
...
+		if (!transition.params.article.title) {
+			this.transitionTo('article', Mercury.wiki.mainPageTitle);
+		}
 	},
 
 	model: function (params: any) {
 		return App.ArticleModel.find({
 			basePath: Mercury.wiki.basePath,
-			title: Mercury.Utils.String.sanitize(params.title),
+			title: params.title,
```
dianafa: we have to sanitize/normalize this title. Otherwise we'll find in URL addresses like:
miss_____Piggy
miss%20Piggy etc.
which is ugly.

front/scripts/mercury/utils/articleLink.ts
```diff
@@ -25,10 +25,10 @@ module Mercury.Utils {
 	 * go to the link directly. NOTE: url might be a jumplink. Do with that what you will.
 	 */
 	export function getLinkInfo(basePath: string, title: string, hash: string, uri: string): LinkInfo {
-		var localPathMatch = decodeURI(uri).match('^' + window.location.origin + '(.*)$');
+		var localPathMatch = uri.match('^' + window.location.origin + '(.*)$');
```
dianafa: we cannot remove decodeURI because categories like категория:Global are not treated like categories and are not redirected to wikiamobile.

dianafa:
@rogatty @hakubo @Warkot 
Igor, I needed to bring back some of things you've removed because they caused some of the bugs mentioned by @tomasznapieralski (https://wikia-inc.atlassian.net/browse/CONCF-126) i.e:

- decodeURI (line 28 in front/scripts/mercury/utils/articleLink.ts)was needed to support category names with special chars (russian, japanese etc)
- sanitize function is responsible for handling links with multiple " " , "%20" and "______" in order to make them "_". We cannot use for this purpose titleToUri() because it uses encodeURI(), neither normalize() because it is used also in category names and we don't want category names like "Sesame_Street" in article.

Warkot:
As long as the articles with '?' (`%3F`) or '%' (`%25`) work, it's fine with me to do it. :+1:

rogatty:
It's now impossible to access _MercuryPercent for 100%_ article from search results.

rogatty:
LGTM

tomasznapieralski:
LGTM

PR #650, opened 2015-03-30
CONCF-80 | enable caching in Mercury

Warkot:


mklucsarits:
This is not directly related to your change, but I don't understand why we have this same object in two different places. It goes against the DRY principle.

Warkot:
They should probably have different names, as the one in `showArticle.ts` is for the first Mercury response and the one in `article.ts` is for the API responses. By the way, these sames aren't too descriptive.

hakubo:
LGTM

hakubo:
@Warkot is it still WIP than?

Warkot:
I wasn't able to make it work on sandbox env, but I heard, it *should* be working on prod. Something also has to be done to the random article caching, but it might need another pull request/ticket/story. @bognix, what do you think?

Warkot:
One `LGTM` from @hakubo, anyone is eager to give some more? @bognix @rogatty

rogatty:
LGTM

bognix:
LGTM

kenkouot:
LGTM

vforge:
LGTM

rogatty:
:hand: **Don't merge before https://github.com/Wikia/app/pull/6892 hits production!** :hand:

hakubo:
@Warkot mkay https://github.com/Wikia/app/pull/6892 has landed.
Should we release that tomorrow?

Warkot:
@hakubo I've checked it today (well, yesterday) and I did it once again just now. We're running release-285 on production (http://www.wikia.com/Special:Version) and here https://github.com/Wikia/app/blob/release-285/extensions/wikia/MercuryApi/MercuryApiHooks.class.php#L89 you can see that there's only one URL purged. We probably have to wait until Tuesday. :snail:

Warkot:
Branch merged, because it was already ported to the last release to prevent flooding Mecrury machines over the weekend—we are at 100% traffic at this point.

PR #651, opened 2015-03-30
Style guide CSS integration

mklucsarits:
These changes include all the fixes I found necessary for using the style guide CSS within Mercury.

The related changes for style-guide can be found here: https://github.com/Wikia/style-guide/pull/46


server/views/layout.hbs
```diff
@@ -20,6 +20,7 @@
 		{{else}}
 			<!-- build:css /front/styles/main/app.css -->
 			<link rel="stylesheet" href="/front/styles/main/app.css">
+			<link rel="stylesheet" href="/front/vendor/wikia-style-guide/dist/css/main.css">
```
mklucsarits: I think the order of these stylesheets needs to be reversed...

front/templates/main/components/collapsible-menu.hbs
```diff
@@ -1,10 +1,10 @@
-<button {{action 'toggleMenu'}} {{bind-attr class=':title isCollapsed:collapsed'}}>
+<a {{action 'toggleMenu'}} {{bind-attr class=':title isCollapsed:collapsed'}} href="#">
```
rogatty: Are you sure this won't cause problems with `handleLink` methods?

mklucsarits: I haven't seen any problems yet during my own testing of links and menus.

server/views/layout.hbs
```diff
@@ -13,6 +13,7 @@
 		<meta name="theme-color" content="{{themeColor}}">
 		{{/if}}
 
+		<link rel="stylesheet" href="/front/vendor/wikia-style-guide/dist/css/main.css">
```
mklucsarits: This will be removed and changed to use @import at the SCSS file level. I need to wait until the authentication styles are merged to dev first, though.

front/templates/main/components/collapsible-menu.hbs
rogatty: I've tried it and after clicking on TOC I can see in the console:
```
Handling link with href: http://127.0.0.1:8000/wiki/TOC_test01#
Attempting transition to article
```
It means that `handleLink` method is being called. I don't think we want that and this is why I changed `a` element to `div` for _Random Page_ in https://github.com/Wikia/mercury/commit/58de2b8ecff462c9c12eb832bacccb720900aac3#diff-2618f71361ceb7bf63b2e4ba5895ee34R17

I'm not sure which element should be used but pretty sure that `handleLink` shouldn't be called when there is no actual link to handle. If you make some decision consider updating the `.random-article-link` too.

front/styles/main/theme/_overrides.scss
```diff
@@ -4,3 +4,10 @@
 .navbox:not(> .article-table) {
 	display: none;
 }
+
+/* Polls don't display correctly with style attributes stripped out
+   for mobile experience. Additionally, poll submission is broken in Mercury.
+*/
+.ajax-poll {
+	display: none;
```
rogatty: Was Chris involved in making decision to hide polls? Should he?
BTW we already have a ticket to make them work https://wikia-inc.atlassian.net/browse/CONCF-123

mklucsarits: I only consulted Zach H. about it. I don't see the harm in hiding them until they are fixed, but I'll check with Chris as well.

mklucsarits:
I believe this is ready for CR, but it may conflict with style changes for the auth pages that Social is working on.

mklucsarits:
NOTE: This will pass Jenkins front-end CI test unless https://github.com/Wikia/style-guide/pull/46 is merged first.

hakubo:
LGTM, what are the next steps for it?

mklucsarits:
All that's left is to merge, but we need to be careful of and resolve any conflicts with styles introduced by @bkoval and @lizlux .

front/styles/main/theme/_overrides.scss
mklucsarits: For the record, Chris was OK with this change.

lizlux:
The auth page styles LGTM.

lizlux:
Let's merge this when the CI checks finish.

PR #652, opened 2015-03-30
Google pagespeed adjustments

kenkouot:
Defers the loading of our vendor and application scripts. Also defers loading the optimizely script as well since Ember fires after optimizely.

Also reorders inlined scripts to be after external resources in the head.


mklucsarits:
LGTM

rogatty:
LGTM

PR #654, opened 2015-03-31
CONCF-374 disable cache on random article inside jquery.ajax

bognix:
Response from /api/v1/article?random&titleOnly was cached for 60s on Varnish even though max-age and s-maxage are set to 0.

I decided to use $.ajax({cache:false}) which appends timestamp to URL.

ping @Warkot @rogatty


rogatty:
LGTM

hakubo:
LGTM

tomasznapieralski:
LGTM

PR #655, opened 2015-04-01
SOC-489 Front-End for the New SignUp Page

bkoval:
@kenkouot @lizlux Here's where all the work to make Front-End shiny for the New SignUp happens.


front/scripts/auth/common/floatingLabels.ts
```diff
@@ -0,0 +1,38 @@
...
+		var element = <HTMLInputElement> event.target,
+			nextElement = <HTMLElement> element.nextElementSibling,
+			parentElement = <HTMLInputElement> element.parentElement;
+
+		if (parentElement.className.match('input-container')) {
+			nextElement.classList.add('active');
+		}
+	}
+
+	private onBlur (input: Event): void {
```
mklucsarits: Perhaps you should keep the parameter names for this and onFocus the same, for consistency.

front/styles/auth/module/_forms.scss
```diff
@@ -0,0 +1,69 @@
...
+
+input[type='text'],
+input[type='text']:focus,
+input[type='password'],
+input[type='password']:focus,
+input[type='email'],
+input[type='email']:focus,
+input[type='date'],
+input[type='date']:focus {
+	padding: 1.25rem 0 0.5rem;
```
mklucsarits: Can this be added in front/styles/auth/base/_input.scss ?

server/views/_layouts/wikia-static.hbs
```diff
@@ -22,7 +22,7 @@
 		{{{content}}}
 
 		{{#unless hideFooter }}
-		<footer class="auth-footer splash splash-bottom">Don\'t have an account? <a href="#">Register now</a></footer>
+		<footer class="auth-footer splash splash-bottom">Don\'t have an account? <a href="/signup">Register now</a></footer>
```
mklucsarits: Be sure to replace this English string later with i18n helper.

bkoval: Yeah, we have a seperate ticket about i18n-ing the whole flow, thanks!

front/locales/en/auth.json
```diff
@@ -1,24 +1,35 @@
...
-    "connect-with-facebook": "Connect with Facebook",
-    "sign-up-with-email": "Sign up with email"
-  }
+	"common": {
+		"password": "Password",
+		"username": "Username",
+		"server-error": "We're sorry, there was an error processing your request. Please try again later."
+	},
+	"login": {
+		"footer": "Don't have an account? <a href=\"/signup?redirect=__redirectRoute__\">Register now</a>",
```
bkoval: Decided to combine callout and anchor translations but I'm not sure if they can be sent to crowdin like this(?). This way it's easier to use our logic for rendering footer and header that have different contents for /login and /signup (we can pass i18n keys and redirect route to the .hbs through context object).

server/views/signup.hbs
```diff
@@ -0,0 +1,33 @@
...
+			<label class="floating-label" for="signupUsername">{{i18n 'auth:common.username'}}</label>
+			<img class='dices' src="/front/images/dice.svg">
+	</div>
+		<div class="input-container">
+			<input type="text" name="password" id="signupPassword" required />
+			<label class="floating-label" for="signupPassword">{{i18n 'auth:common.password'}}</label>
+			<img class='password-toggler' src="/front/images/eye.svg">
+		</div>
+		<div class="input-container">
+			<input type="date" name="date" id="signupBirthDate" required />
```
bkoval: Just a placeholder for what we'll be done in the "Date field localization" ticket.

front/locales/en/auth.json
lizlux: I don't this we should do this, for a few reasons:
- I don't like having a hard-coded url like this in a translation string. If it changes we'll have to update all the translations
- I'd like to avoid HTML in messages as much as possible. Translators shouldn't deal with code
- I'm not sure I understand the problem with the way it was before, other than possibly that the link text wouldn't be fully in context with the preceding text. Usually we would use qqq messages to explain this. I'll check in with Noelle to get her thoughts. 

My intent was for the `wikia-static.hbs` file to have variables (instead of message keys) to pass into the i18n helper, then each view file would set the value of that variable ([example here](https://github.com/Wikia/mercury/blob/hapi-auth/server/views/_layouts/wikia-static.hbs#L25)). Let me know if that works for you @bkoval.

lizlux: I spoke with @gbenson-wikia today and got my answer for this. We should avoid markup (beyond variables) and use multiple messages when we need formatting and linking. So for example, messages like `Here's my <a href=''>link text</a>.` should actually be written as two messages: `Here's my __varInSomeForm__.` and `link text`, and then associated with one another via qqq. I say "var in some form" b/c it might be `__var__

The services team is actually going to start working on an i18n service in about a month which will likely replace much of what's currently in place for mercury. In the mean time it's good for us to follow the best practices they put in place in order to minimize the duplication of effort going forward.

kenkouot: Yeah, have to agree here. Markup should never be in translations as it makes maintenance not obvious.

front/scripts/auth/main.ts
```diff
@@ -0,0 +1,8 @@
+window.document.addEventListener('DOMContentLoaded', function ():void {
```
kenkouot: Pretty sure if you put scripts at before the closing bottom body tag, you don't need the DOMContentLoaded event listener. I'd just wrap everything in a SIAF to prevent leaking variables, but that should be good enough.

front/scripts/auth/common/Form.ts
```diff
@@ -0,0 +1,61 @@
...
+
+	private onClick (event: Event): void {
+		var element = <HTMLInputElement> event.target,
+			parentElement: HTMLElement,
+			input: HTMLInputElement;
+		if (element.className.match('password-toggler')) {
+			parentElement = <HTMLElement> element.parentElement;
+			input = <HTMLInputElement> parentElement.querySelector('input');
+			this.togglePasswordInput(input, element);
+		} else if (element.className.match('dices')) {
```
lizlux: Not sure if this is on purpose, but the plural of "dice" is "dice". The singular is actually "die", but people say "dice" too.

front/scripts/auth/main.ts
```diff
@@ -0,0 +1,8 @@
...
+	new Form().watch();
+
+	//TODO Figure out a nicer way of diving javascripts / recognizing a page type
```
lizlux: Do you plan on handling this `to do` in this PR?

server/views/signup.hbs
```diff
@@ -0,0 +1,33 @@
+<section class="auth-content">
+	<form id="signupForm" action="" method="POST">
+		<div class="input-container">
+			<input type="email" name="email" id="signupEmail" required />
+			<label class="floating-label" for="signupEmail">{{i18n 'auth:signup.email'}}</label>
+		</div>
+		<div class="input-container">
+			<input type="text" name="username" id="signupUsername" required />
+			<label class="floating-label" for="signupUsername">{{i18n 'auth:common.username'}}</label>
+			<img class='dices' src="/front/images/dice.svg">
```
kenkouot: Use single quotes in markup, this file mixes single and double!

front/styles/auth/module/_forms.scss
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+.dices {
+	&.on {
+		animation: spin 1s linear infinite;
+	}
+}
+
+@keyframes spin {
```
lizlux: Haha, this is cute!

front/styles/auth/module/_forms.scss
```diff
@@ -0,0 +1,81 @@
...
+	100% {
+		transform: rotate(360deg);
+	}
+}
+
+.inline-checkbox-wrapper {
```
lizlux: Was the formatting off here? Wondering why this is needed. I doesn't look good in IE11 as is. 

screen shot 2015-04-07 at 4 46 23 pm

Image also shows misaligned birthday label.

server/facets/auth/login.ts
```diff
@@ -30,22 +30,20 @@ interface HeliosResponse {
...
-	footerCalloutLink : string;
-	hideHeader?       : boolean;
-	hideFooter?       : boolean;
-	exitTo?           : string;
-	bodyClasses?      : string;
-	formErrorKey?     : string;
+	header?      : string;
+	footer?      : string;
+	bodyClasses? : string;
+	i18nContext? : string;
```
lizlux: remove this line

server/facets/auth/signup.ts
```diff
@@ -0,0 +1,33 @@
+interface SignupViewContext {
+	title : string;
+	header?  : string;
+	footer?  : string;
+	exitTo?      : string;
+	bodyClasses? : string;
+	loadScripts? : boolean;
```
lizlux: needs `i18nContext`

lizlux: https://github.com/Wikia/guidelines/pull/74

server/facets/auth/login.ts
```diff
@@ -109,14 +115,23 @@ export function post (request: Hapi.Request, reply: any): void {
 	var credentials: any = request.payload,
 		requestedWithHeader: string = request.headers['x-requested-with'],
 		isAJAX: boolean = requestedWithHeader && !!requestedWithHeader.match('XMLHttpRequest'),
+		authRedirect: string,
+		error: any = {},
```
lizlux: These vars don't appear to be called anywhere.

front/scripts/auth/common/Form.ts
```diff
@@ -0,0 +1,61 @@
...
+			element.classList.toggle('on');
+		}
+	}
+
+	/**
+	 * Starts continuous checking for new input
+	 */
+	public watch (): void {
+		var form = window.document.querySelector('form');
```
lizlux: We might want to pass this form element into the constructor to make this class worth instantiating, as opposed to being more static, like it is now. Otherwise we could make all these methods static. I'd go with the first option though, in case there ends up being more than on form on a page.

front/locales/en/auth.json
bkoval: Agreed. I had one more doubt, about languages that might want to put the anchor in the other place in the sentence (e.g. not at the end, but in the middle of it) but yeah, I guess we can do it without mixing markup with the translated text, sorry for that and thanks for claryfying this :)

front/styles/auth/module/_forms.scss
bkoval: Yeah, there's something wrong there. I'll take a look at the situation with checkbox. I added wrapper for the label to be inlined with checkbox even when the text exceeds page's width.

server/facets/auth/login.ts
bkoval: Good catch. Again, leftovers, sorry for that.

front/scripts/auth/main.ts
bkoval: Yeah, but the problem is that if we want to add methods to prototypes as typescript does, they won't be available right away if this code will trigger before assigning all the methods to the Form constructor's prototype. 

Using event we don't add a big delay on a small page and we ensure there are no race conditions if the .ts files are compiled / concatenated in the mixed order.

front/scripts/auth/common/Form.ts
bkoval: :laughing: Sorry for that.

server/facets/auth/signup.ts
bkoval: Agreed. Let's stick to what we've been using so far and once the PR to the guidelines repo gets merged, we'll address that.

front/scripts/auth/main.ts
bkoval: Let's handle it here: https://wikia-inc.atlassian.net/browse/SOC-626

front/styles/auth/module/_forms.scss
bkoval: ;)

bkoval:
@lizlux Checkboxes should work better in IE now, I used .inline class provided by style-guide, just removing the margin under the checkbox so there are consistent line-heights between all lines of the label.

I think it's not necessary to address the date field issue as you will reimplement it as a set of number inputs in SOC-480 (number inputs work fine with floating labels) and this input[type=date] is just a placeholder for that. :palm_tree:

front/locales/en/auth.json
```diff
@@ -1,24 +1,37 @@
...
+		"register-with": "Register with",
+		"connect-with-facebook": "Connect with Facebook"
+	},
+	"signup": {
+		"sign-up-with-email": "Sign Up With Email",
+		"email-label": "Email",
+		"footer-callout-text": "Already have an account?",
+		"footer-callout-link": "Log in",
+		"register-action": "Register",
+		"page-title": "Sign Up with Email",
```
lizlux: I think this should be "title" instead of "page-title" since the title of the page is currently "undefined"

front/scripts/auth/common/Form.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+class Form {
+	form: HTMLFormElement;
+
+	constructor (form: Element) {
+		this.form = <HTMLFormElement> form;
+	}
+
+	private onFocus (event: Event): void {
+		var element = <HTMLInputElement> event.target,
```
lizlux: I think it would be clearer if instead of "element" and "nextElement" we name these vars something more descriptive like "input" and "label". Same with "parentElement", maybe "wrapper"? 

I came across this when working with the birthdate input field, which may require us to update the way we select these elements. Making this work with that special case can be handled separately, but maybe updating the var names can happen in this PR. Let me know what you think.

front/scripts/auth/main.ts
lizlux: Yeah I think the main issue is the order in which the files are concatenated. Right now, `main.ts` is included before the `Form` class is defined. Is there a way to control that order? If not, waiting for the event to fire seems like an okay workaround for now. It would be nice to be able to control that order though.

front/styles/auth/module/_forms.scss
```diff
@@ -0,0 +1,73 @@
...
+	width: 20px;
+	position: absolute;
+	right: 5px;
+
+	&.on {
+		opacity: 0.5;
+	}
+}
+
+.dice {
```
lizlux: I wonder if this should be it's own component. Could it exist outside a form?

front/svg/images/dice.svg
```diff
@@ -0,0 +1 @@
...63.8 320.1 562.9 321 561.8 321 L561.8 321 Z" sketch:type="MSShapeGroup"/><path d="M589.2 342.6 L578.1 331.4 C577.1 330.5 575.6 330.5 574.6 331.4 L563.4 342.6 C562.5 343.6 562.5 345.1 563.4 346.1 L574.6 357.2 C575.6 358.2 577.1 358.2 578.1 357.2 L589.2 346.1 C590.2 345.1 590.2 343.6 589.2 342.6 L589.2 342.6 ZM577.8 335.2 C578.6 336 578.6 337.3 577.8 338.1 C577 338.9 575.7 338.9 574.9 338.1 C574.1 337.3 574.1 336 574.9 335.2 C575.7 334.4 577 334.4 577.8 335.2 L577.8 335.2 ZM567.2 345.8 C566.4 345 566.4 343.7 567.2 342.9 C568 342.1 569.3 342.1 570.1 342.9 C570.9 343.7 570.9 345 570.1 345.8 C569.3 346.6 568 346.6 567.2 345.8 L567.2 345.8 ZM574.9 345.8 C574.1 345 574.1 343.7 574.9 342.9 C575.7 342.1 577 342.1 577.8 342.9 C578.6 343.7 578.6 345 577.8 345.8 C577 346.6 575.7 346.6 574.9 345.8 L574.9 345.8 ZM574.9 353.5 C574.1 352.7 574.1 351.4 574.9 350.6 C575.7 349.8 577 349.8 577.8 350.6 C578.6 351.4 578.6 352.7 577.8 353.5 C577 354.2 575.7 354.2 574.9 353.5 L574.9 353.5 ZM582.6 345.8 C581.8 345 581.8 343.7 582.6 342.9 C583.4 342.1 584.7 342.1 585.5 342.9 C586.2 343.7 586.2 345 585.5 345.8 C584.7 346.6 583.4 346.6 582.6 345.8 L582.6 345.8 Z" sketch:type="MSShapeGroup"/></g></g></g></svg>
```
lizlux: probably want to get rid of sketch add-ons in these SVGs too

lizlux:
@bkoval, can you update the description for this PR now that it's CR ready?

server/facets/auth/join.ts
```diff
@@ -19,8 +20,9 @@ function get (request: Hapi.Request, reply: any): void {
 	}
 
 	context = {
-		title: 'Join Wikia',
+		pageTitleKey: 'auth:landing-page.page-title',
```
lizlux: This key doesn't exist, nor does the `pageTitleKey`mustache var.

server/views/_layouts/wikia-static.hbs
```diff
@@ -14,15 +14,18 @@
 		<a class="close" href="{{exitTo}}"></a>
 		{{/if}}
 
-		{{#unless hideHeader }}
+		{{#unless hideHeader}}
 		<header class="auth-header splash splash-top">{{i18n headerText}}</header>
```
lizlux: I think we should either rename this file to `auth.hbs` or remove anything that has to do with auth, like the `auth-header` class. I'd vote for the former, since this is the only instance of a wikia static page and we have no indication that other static pages are going to use a similar layout. I guess it's worth following up with design before we make that decision. Also, not a blocker for this PR.

front/locales/en/auth.json
```diff
@@ -1,24 +1,37 @@
...
+		"forgot-password": "Forgot password?",
+		"submit-text": "Sign In"
+	},
+	"join": {
+		"title": "Join Wikia",
+		"register-with": "Register with",
+		"connect-with-facebook": "Connect with Facebook"
+	},
+	"signup": {
+		"sign-up-with-email": "Sign Up With Email",
```
lizlux: Should this go under `common` since it's used in the `join` and `signup` pages?

server/views/login.hbs
```diff
@@ -2,16 +2,20 @@
...
+			<input class="sign-in-password" type="text" name="password" id="loginPassword" required />
+			<label class="floating-label" for="loginPassword">{{i18n "auth:login.password-label"}}</label>
+			<img class="password-toggler" src="/front/images/eye.svg">
 		</div>
-		<a class="forgotten-password" href="#" tabindex="-1">{{i18n 'auth:login.forgot-password'}}</a>
+		<a class="forgotten-password" href="#" tabindex="-1">{{i18n "auth:login.forgot-password"}}</a>
 		<input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
-		<label><input type="checkbox" name="remember" id="loginRemember"> {{i18n 'auth:login.remember-me-label'}}</label>
-		<button id="loginSubmit" type="submit" disabled>{{i18n 'auth:login.submit-text'}}</button>
+		<label for="loginRemember" class="inline">
```
lizlux: Turns out we're getting rid of "remember me" anyway. Oh well!

server/views/_helpers/i18n.ts
```diff
@@ -3,12 +3,11 @@ function i18nHelper (key: string, options: any): string {
 		params: {[key: string]: string} = {},
 		namespace = '',
 		instance = this.i18n.getInstance();
-
-	Object.keys(options.hash).forEach((key: string) => {
```
lizlux: Why the change here?

server/views/signup.hbs
```diff
@@ -0,0 +1,34 @@
+<section class="auth-content">
+	<form id="signupForm" action="" method="POST">
+		<div class="input-container">
+			<input type="email" name="email" id="signupEmail" required />
+			<label class="floating-label" for="signupEmail">{{i18n "auth:signup.email-label"}}</label>
+		</div>
+		<div class="input-container">
+			<input type="text" name="username" id="signupUsername" required />
+			<label class="floating-label" for="signupUsername">{{i18n "auth:common.username-label"}}</label>
+			<img class="dice" src="/front/images/dice.svg">
```
lizlux: Ken's comment was to use single quotes in markup, not double quotes. @kenkouot, is this guideline documented somewhere?

server/views/_helpers/i18n.ts
lizlux: I did some more research into this and that `hash` property is how handlebars expresses items passed into the helper. So, for example: 
```handlebars
{{i18n 'auth:common.my-message' foo='bar'}}
```
`auth.js` can include something like:
```json
common: {
    "my-message": "some __foo__"
}
```
and `options.hash` will be: 
```javascript
{foo: 'bar'}
```
More info on handlebars helpers here http://handlebarsjs.com/block_helpers.html#hash-arguments. The current implementation in dev isn't working correctly so I'm going to submit a PR to fix it.

lizlux: https://github.com/Wikia/mercury/pull/692

server/views/login.hbs
bkoval: Haha, I love it, no more problems with inlining for IE :hammer:

front/scripts/auth/common/Form.ts
bkoval: I'll do it here :dancer:

server/views/signup.hbs
bkoval: Yeah, I wanted to be compliant with ember-main.hbs which is a core html layout for the main Mercury app.

front/scripts/auth/main.ts
bkoval: Pinging @kenkouot and @hakubo to answer your question :)

front/svg/images/dice.svg
bkoval: Yeah, how about we do it here? https://wikia-inc.atlassian.net/browse/HG-588

server/views/_helpers/i18n.ts
bkoval: Yeah, I just intended to submit this change along with signup view, makes sense to extract it :+1:

front/svg/images/dice.svg
lizlux: :+1:

server/views/signup.hbs
lizlux: I'd rather see double quotes in HTML anyway, so as long as it's consistent it's fine with me. Especially if there's no documented rule.

lizlux:
LGTM!

PR #656, opened 2015-04-01
ADEN-1780: Added new high impact slot

nandy-andy:
We don't have high impact slot like `INVISIBLE_SKIN` for mobile (Mercury). This is first try to have it :smile_cat:

### DIFF
--- a/front/styles/main/module/_ads.scss
+++ b/front/styles/main/module/_ads.scss
@@ -23,3 +23,7 @@
 .wikia-ad-iframe {
 	border: 0;
 }
+
+#INVISIBLE_HIGH_IMPACT {
+	display: none;
+}
--- a/front/styles/main/module/_ads.scss
+++ b/front/styles/main/module/_ads.scss
@@ -6,6 +6,10 @@
 	}
 }
 
+.ad-slot[id^="INVISIBLE_"] {
+	display: none !important;
+}
+
 .mobile-top-leaderboard {
 	min-height: 50px;
 	margin-top: 60px;
@@ -22,8 +26,4 @@
 
 .wikia-ad-iframe {
 	border: 0;
-}
-
-#INVISIBLE_HIGH_IMPACT {
-	display: none;
 }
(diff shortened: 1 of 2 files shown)


front/styles/main/module/_ads.scss
```diff
@@ -6,6 +6,10 @@
 	}
 }
 
+.ad-slot[id^="INVISIBLE_"] {
```
mklucsarits: I don't think this style declaration is a good idea. How about creating a new attribute, "hidden" or something", for the ad slot component that adds the CSS class `hide` to that ad? Then it doesn't matter what the ad's name or id is, and we wouldn't need such specific styles.

nandy-andy: Hi Matt! :smile_cat: I'm glad you asked! :sunny: 

This is a pattern copied from Oasis and I agree with the statement we shouldn't copy bad patterns but I think this was the only way here - it's "clean" and explicit. AdEngine removes `hidden` classes on slots once they're populated with an ad. Invisible slots are special case when we really do not want to make the slot visible (even if it contains an ad / a code from DFP).

gabrys:
LGTM. Thanks

front/styles/main/module/_ads.scss
mklucsarits: Thanks for the explanation.

front/templates/main/application.hbs
```diff
@@ -10,6 +10,7 @@
 	{{smart-banner isVisible=view.smartBannerVisible}}
 	{{site-head smartBannerVisible=view.smartBannerVisible sideNavCollapsed=view.sideNavCollapsed trackClick='trackClick'}}
 	{{side-nav isCollapsed=view.sideNavCollapsed loadRandomArticle='loadRandomArticle'}}
+	{{ad-slot name='INVISIBLE_HIGH_IMPACT' noAds=noAds}}
```
hakubo: I really like how we can add additional slots now :)

hakubo:
LGTM

PR #657, opened 2015-04-02
Upgrade gulp-sass (for real this time)

kenkouot:
After merging the hapi-auth branch, we noticed that something odd was happening with the sass compilation. The settings file for `styles/main/app.scss` was getting overwritten by another settings file from `styles/auth/`.

To make a long story short, the version of libsass used by the gulp-sass plugin was getting confused when imports with two separate files in different directions with the same name. This was fixed in a later version of libsass so I upgraded our gulp-sass and also fixed all the regressions that resulted from the last time we tried to upgrade. Likely, they weren't regressions in the way we thought, but rather, the libsass version we were using was actually failing to correctly compile the mixins from Foundation. Upgrading gulp-sass revealed our mistakes and this PR fixes those:
* Fixed footer being too wide
* Fixed comments being too wide

### DIFF
--- a/front/styles/main/module/_comments.scss
+++ b/front/styles/main/module/_comments.scss
@@ -5,9 +5,12 @@
 
 	background: rgb(250,250,250);
 	color: $anchor-color-normal;
-	margin-bottom: 20px;
 	padding: 10px;
 	width: 100%;
+}
+
+.show-reply-btn {
+	margin-bottom: 20px;
 }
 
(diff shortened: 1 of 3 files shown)


kenkouot:
Just finished testing this and looks good. It fixes a bug on dev now so going to merge it!

rogatty:
Great to see it's working now, thanks @kenkouot!

PR #662, opened 2015-04-03
Set auth mode to false for unauthenticated routes

kenkouot:
Simply turns auth mode to false for unauthenticated routes, squelching errors from Kibana.


rogatty:
LGTM

PR #664, opened 2015-04-04
Do not try to access value on null

kenkouot:
@mklucsarits found a bug where if you opened up comments on a page with comments, and tried to transition using a `{{link-to}}` helper, like in the content recommendation area at the foot of an article, you would fail the transition because of trying to access a value on a null object. This adds a check that prevents that and allows the transition to proceed safely.

### DIFF
--- a/front/scripts/main/components/ArticleCommentComponent.ts
+++ b/front/scripts/main/components/ArticleCommentComponent.ts
@@ -23,8 +23,9 @@ App.ArticleCommentComponent = Em.Component.extend({
 
 	user: function () {
 		var users = this.get('users');
-
-		return users[this.get('comment.userName')] || {};
+		if (users) {
+			return users[this.get('comment.userName')] || {};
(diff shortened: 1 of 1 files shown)


mklucsarits:
Code looks fine and LGTM on sandbox-mercury

rogatty:
@kenkouot @mklucsarits this looks like bypassing a problem instead of solving it. This component shouldn't be used if there are no comments (and no `users`).

Can you provide URL where it was happening? I'd like to investigate what's the source of this issue.

mklucsarits:
@rogatty Steps to reproduce:
1) Go to any article with comments.
2) Expand/load the comments.
3) Click on an article link in "Read More" down below.
4) There will be an error in the browser console, and the article transition won't complete.

hakubo:
@mklucsarits @kenkouot should we remove it since Igor found a real root cause?

PR #667, opened 2015-04-07
CONCF-208 Spaces in gallery when no caption

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-208

Only small CSS change adding margin between images also when no caption present.


rogatty:
LGTM

mklucsarits:
LGTM

PR #670, opened 2015-04-08
INT-48: (WIP) Initial version of sharing feature in Mercury

inez:


front/scripts/main/controllers/ApplicationController.ts
```diff
@@ -18,7 +18,8 @@ App.ApplicationController = Em.Controller.extend({
 			mainPageTitle: Em.get(Mercury, 'wiki.mainPageTitle'),
 			siteMessage: Em.get(Mercury, 'wiki.siteMessage'),
 			siteName: Em.getWithDefault(Mercury, 'wiki.siteName', 'Wikia'),
-			editorPreview: Em.get(Mercury, 'article.preview')
+			editorPreview: Em.get(Mercury, 'article.preview'),
+			shareFeatureAvailable: true /* Em.get(Mercury, 'wiki.language.content') === 'ja' */
```
SebastianMarzjan: I know it's a comment only at this stage, though you might want to sync with @bognix and Chris on the approach to determining set of sharing functionalities for Japanese market - there were some options considered including geo, wiki language, user language and user browser language and if I recall correctly the last one is current working assumption for desktop.

bognix: yes we are using browser language right now. We use an AJAX call to get social icons for given language, controller's method returns HTML which is parsed mustache template . Maybe you can use it in Mercury so we don't have to duplicate the logic.

inez:
@kenkouot @mklucsarits

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -38,6 +38,10 @@ App.ArticleRoute = Em.Route.extend({
 		});
 	},
 
+	afterModel: function (model) {
```
hakubo: could this live under article route so we don't have to 'leak' it?

front/styles/main/component/_share-feature.scss
```diff
@@ -0,0 +1,25 @@
...
+	padding-top: 1px;
+	text-align: right;
+	transition: transform .5s;
+
+	&.pinned {
+		transform: translateY(-100%);
+	}
+}
+
+.share-feature .fb {
```
hakubo: you could nest it as well

front/styles/main/component/_share-feature.scss
```diff
@@ -0,0 +1,25 @@
...
+	height: 46px;
```
hakubo: soulds like a good number to 'abstract' ? does it have to be redefined per platform?

front/templates/main/components/share-feature.hbs
```diff
@@ -0,0 +1,3 @@
+<a {{action 'trackClick' 'share' 'line'}} target="_blank" {{bind-attr href=lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
```
hakubo: with ember 1.11 you can:
```hbs
<a href={{yourValue}}>
```

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -0,0 +1,67 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../mercury/utils/browser.ts" />
+/// <reference path="../../main/mixins/TrackClickMixin.ts" />
+'use strict';
+
+App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, {
```
hakubo: maybe just App.SharePageComponent
as you are sharing page
not feature

but I might be nit picky :)

hakubo:
Thanks! @inez 
How does Ember feel to you?

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -0,0 +1,67 @@
...
+
+	options: {
+		// keep it consistent with values in _wikia-variables.scss
+		smartBannerHeight: {
+			android: 66,
+			ios: 83
+		}
+	},
+
+	offset: function (): number {
```
mklucsarits: Because this smart banner and Headroom code is duplicated from SiteHeadComponent, maybe it can be moved to a new mixin, if time permits? Or maybe at least make the height offset a property of SmartBannerComponent rather than a property of SiteHead/ShareFeature.

front/scripts/main/routes/ArticleRoute.ts
inez: This is the solution that @kenkouot suggested. Title is needed in the ShareFeature which is referenced in application.hbs, but application itself does not know about title :(

front/scripts/main/controllers/ApplicationController.ts
inez: Thanks for your feedback.
For now we decided to enable this feature only for Japanese content - so we can get it out the door and start collecting some numbers.
I will sync with Chris and with @bognix later on to see if we can re-use the same logic.

front/scripts/main/components/ShareFeatureComponent.ts
inez: Ok, I will reach out directly to you to find out how to do it.

front/scripts/main/components/ShareFeatureComponent.ts
kenkouot: I'd actually just call it ShareComponent :smile:

front/scripts/main/routes/ArticleRoute.ts
hakubo: mhm, 
so how about moving this
```hbs
{{#if shareFeatureAvailable}}
	{{share-feature smartBannerVisible=view.smartBannerVisible title=currentTitle}}
{{/if}}
```

somewhere in article.hbs
so applications still does not have to know it?

front/scripts/main/components/ShareFeatureComponent.ts
inez: That's how I originally wanted to call it but Ember component naming convention does not allow for it: "Note: Components must have a dash in their name. So blog-post is an acceptable name, but post is not. This prevents clashes with current or future HTML element names, and ensures Ember picks up the components automatically."

front/scripts/main/routes/ArticleRoute.ts
inez: @hakubo It seem like a reasonable thing for application to know the title of displayed article, doesn't it?
Probably it could be moved somewhere else in HTML, but where it currently is seems to make a lot of sense (next to site-head) - because it works in a way very similar to site head. @mklucsarits let me know that you guys are going to re-think and re-do site head area of the page, so I decided to not over think it.

front/scripts/main/components/ShareFeatureComponent.ts
inez: @mklucsarits Do you consider it as a blocker for merging or not? If not then could you merge so I can release today? Thanks.

front/scripts/main/components/ShareFeatureComponent.ts
kenkouot: Let's call it ShareHeaderComponent then, less generic!

kenkouot:
Just one comment on naming then g2g

front/scripts/main/routes/ArticleRoute.ts
mklucsarits: The `model` param here needs a type, and we usually just use `any` for a model. So it would be just:
`model: any`

front/scripts/main/components/ShareFeatureComponent.ts
inez: Ok, discussed it with @kenkouot and agreed to keep it the way it is. Merging, OMG!!

PR #672, opened 2015-04-08
Fix content recommendations links

rogatty:
@mklucsarits found a bug which @kenkouot bypassed with https://github.com/Wikia/mercury/pull/664.

I've found that the source of problem was `bubbles=false` on content recommendation links. Because of it the method [`ApplicationRoute.handleLink`](https://github.com/Wikia/mercury/blob/dev/front/scripts/main/routes/ApplicationRoute.ts#L39) was never called on click, therefore [`ArticleController.changePage`](https://github.com/Wikia/mercury/blob/dev/front/scripts/main/controllers/ArticleController.ts#L21) was never called which caused `commentsPage` to not be cleared and we had a bug.

Also, tracking for related pages was broken (it's in `handleLink`).

This PR removes `bubbles=false` and fixes some minor issues in `ArticleCommentComponent.ts`


hakubo:
LGTM

mklucsarits:
LGTM. Thanks for looking into this further.

kenkouot:
Good job on this, yeah the original bug was discovered/fixed at like 6 on Friday. This does it justice, good work!

kenkouot:
Can we get the conflicts resolved and merge this through?

PR #666, opened 2015-04-06
Soc 526 enabling query string and cookies for setting lang

lizlux:
Fixing all hard-coded strings in the mercury authentication pages. 

This PR also adds support for setting language via query string ~~, once https://github.com/kenkouot/hapi-i18next/pull/2 is merged and pulled in to the mercury project~~. 

Also, `?uselang=qqx` will show the raw message keys once that PR is merged.

PR #675, opened 2015-04-08
SOC-526 removing buggy qqx functionality

lizlux:
In response to https://github.com/Wikia/mercury/pull/666/files#r27969261. Looks like we don't get to have `?uselang=qqx` functionality after all, but it's okay b/c we'll have an i18n service in the next couple months, built by the services team. 

The problem was that falling back to qqx meant that languages where strings didn't exist would show the raw message key instead of the english string. 

@bkoval @kenkouot


bkoval:
LGTM

mklucsarits:
LGTM

PR #678, opened 2015-04-09
Make footer not overlap content for small viewports

bkoval:
@lizlux @rogatty This just makes sure footer will not overlap form fields when the form is long / keyboard appears.


bkoval:
By default flex-basis is set to auto, which means the elements will take as much space as their flex-grow value let's them. In this case, auth-content can take as much space as it can between header and footer and both header and footer cannot exceed their natural height.

hakubo:
LGTM

lizlux:
LGTM.

PR #682, opened 2015-04-09
Update ember-performance-sender version

kenkouot:
Discovered a bug today that was revealed (but not necessarily caused) by the recent Ember upgrade. It was fixed here: https://github.com/Wikia/ember-performance-sender/pull/3 and this updates our packages to reflect it. It was never on production because we haven't pushed the chef changes to run EPS/Weppy yet.


hakubo:
mkay, lgtm

PR #683, opened 2015-04-09
Don't use the TrackClick mixin to send an action, but actually use it to...

kenkouot:
... track a click. @inez was having some issues with GA tracking and while looking into it, I discovered that TrackClickMixin was actually just a global proxy for ApplicationRoute's trackClick action. This seemed roundabout and confusing to me so I moved the tracking into TrackClickMixin and simply made ApplicationRoute a consumer of that mixin as well.


mklucsarits:
Totally makes sense, and LGTM. I blame my newness to Mercury/Ember when I did this for that inefficiency.

PR #686, opened 2015-04-14
add logging for crumb and result undefined errors

kenkouot:
temporary logging to help track down the root cause of some low volume errors on prod


mklucsarits:
LGTM

PR #690, opened 2015-04-14
Move crowdin config to multiple files, for now

mklucsarits:
https://wikia-inc.atlassian.net/browse/HG-589

The Mercury codebase is changing to use multiple JSON files for localization strings, separated by need and/or feature. However, multiple source files are not supported by the i18n CrowdIn API scripts. Therefore, until that happens, we need to use separate config files, one for each JSON file.


mklucsarits:
@Grunny Please have a look at this.

PR #692, opened 2015-04-14
Allowing i18n helper to accept inline params

lizlux:
Previously, `{{i18n 'my-key' foo='bar'}}` wouldn't work b/c `options.hasOwnProperty(key)` was always evaluating to `false`. One fix would have been to change it to `options.hash.hasOwnProperty(key)`, but `Objects.keys` already filters out inherited properties so I removed that check altogether. 

I also added a comment and tried to make the code a bit easier to read for someone new to handlebars helpers.

### DIFF
--- a/server/views/_helpers/i18n.ts
+++ b/server/views/_helpers/i18n.ts
@@ -2,13 +2,15 @@ function i18nHelper (key: string, options: any): string {
 	var translateWithCache: Function = this.i18n.translateWithCache,
 		params: {[key: string]: string} = {},
 		namespace = '',
-		instance = this.i18n.getInstance();
+		instance = this.i18n.getInstance(),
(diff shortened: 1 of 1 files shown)


lizlux:
@kenkouot @bkoval

bkoval:
LGTM. Just not sure about the comment, as context vars passed to i18n helper won't actually work in this case without the "=" sign. (Like {{i18n 'something' myVar}}).

lizlux:
@bkoval, it should work if `myVar == {hash: {foo: 'bar'}}`, which is not ideal, but that `hash` property is needed for handlebars to accept the tag parameters as well. We might be able to tweak the helper to accept both with and without the hash. Maybe we can create a ticket to add that optimization.

PR #699, opened 2015-04-16
Update type definitions

mklucsarits:
https://wikia-inc.atlassian.net/browse/HG-587


tsd.json
```diff
@@ -5,29 +5,98 @@
...
+    "fastclick/fastclick.d.ts": {
+      "commit": "7ca8add069411b3bde47fad7208706801cde8656"
+    },
+    "glob/glob.d.ts": {
+      "commit": "7ca8add069411b3bde47fad7208706801cde8656"
+    },
+    "google.analytics/ga.d.ts": {
+      "commit": "6e8f579f26c48e743f5490f6882fa1f0067ea855"
+    },
+    "gulp/gulp.d.ts": {
```
hakubo: since we have all gulp tasks written in JS,
what is the advantage of having those definitions in our repo?

mklucsarits: Ah you're right. Wasn't thinking of that.

typings/node/node.d.ts
```diff
@@ -1,11 +1,11 @@
-// Type definitions for Node.js v0.10.1
```
rogatty: Is it a good time to update it? We're not using 0.12.0 yet.

rogatty:
Check Jenkins or run tests locally to see errors related to type definitions.

PR #703, opened 2015-04-20
SOC-474 Input Errors Front-End for New Login / Signup

bkoval:
@mklucsarits Submitting PR to signup branch as it depends on it. After we merge this one into SOC-489, let's QA both features together.

This adds red messages under inputs to use when something goes wrong with the input validation on Front-End / Helios.

https://wikia-inc.atlassian.net/browse/SOC-474


front/styles/auth/module/_forms.scss
```diff
@@ -1,5 +1,19 @@
 .input-container {
+	padding-bottom: 1rem;
 	position: relative;
+
+	&.has-error {
```
mklucsarits: I don't think this class is necessary for styling the input. If you use the style guide CSS, all you have to do is add the 'error' class to the input element, and it will color the border for you.

bkoval:
@mklucsarits Hey, I used .error class on input to be compliant with the style-guide semantics, but we cannot use style-guide styling for the input fields yet (it differs from the new input styling. Once the UX team is happy with the new inputs we can move their styling to style-guide, but right now they want to test it.

mklucsarits:
LGTM

PR #704, opened 2015-04-20
SOC-633 Redirect Users to New Login by Login-Icon component

bkoval:
@mklucsarits @rogatty (I know this was already commited before, but we withdrawn because of it blocking Mercury 'dev' from being released. Let's bring it back on new-login-mvp feature branch as requested in https://wikia-inc.atlassian.net/browse/SOC-633


rogatty:
LGTM

mklucsarits:
LGTM

hakubo:
LGTM

lizlux:
Since we're going to do a staged rollout of the new login page on a per-wikia basis, it would be good to include logic here to show the /login route for enabled wikias, and the /Special:UserLogin route for non-enabled wikias. That would also make it so we wouldn't need the MVP feature branch, as we could enable /login on one or two test wikias. 

@kenkouot mentioned during our checkin meeting on Friday that we can create an array of enabled wikias, either in a config file or in the view itself, just for the rollout, unless you can think of a better way, since we don't really do WF variables in mercury.

bkoval:
Agreed. This should definitely be handled in a seperate ticket. Once we implement it and merge it to the feature branch, we'll be able to merge it to dev.

PR #706, opened 2015-04-20
Render not-found template when we have 404

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-466

1. Fix for displaying not-found template after transition to non-existent article
2. Fix for a regression created when upgrading to Ember 1.11


hakubo:
LGTM

bognix:
LGTM

PR #708, opened 2015-04-20
Set the auth cookie TTL

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-657

Even though the UI option for "remember me" was removed, we still need to set the cookie expiration.


kenkouot:
LGTM

lizlux:
:+1:

PR #709, opened 2015-04-20
HG-603 refactor prototype extensions

kenkouot:
Early on in the Mercury project, we had a discussion about whether or not to use Ember's prototype extensions. The arguments against them were that it broke best practices, and it generally felt icky, even though it offered syntactic sugar. Well, fast forward a year and it seems that the Ember Core team is having that same discussion and making some points that we hadn't considered. For instance, building in prototype extensions blocks native upgrade paths. An example of this is Ember's `Function#property` vs ES7's `Function#property`, which are of course, not compliant to each other.

After a few weeks of watching the discussion unfold, it looks like the Ember team is actively moving away from prototype extensions and moving toward a direction that supports ES7 decorator syntax. I've gone ahead and updated all instances of our prototype extensions as well. Please avoid them from here on out.

If you'd like to follow in on the thread over at the Ember repo, the following PR's are enlightening:
https://github.com/emberjs/ember.js/issues/9269
https://github.com/emberjs/guides/pull/110


front/scripts/main/components/ImageMediaComponent.ts
```diff
@@ -19,20 +19,20 @@ App.ImageMediaComponent = App.MediaComponent.extend(App.ArticleContentMixin, {
...
-	}.property('width', 'height'),
+	}),
 
 	/**
 	 * used to set proper height to img tag before it loads
 	 * so we have less content jumping around due to lazy loading images
 	 * @return number
 	 */
-	computedHeight: function (): number {
+	computedHeight: Em.computed(function (): number {
```
rogatty: I think you forgot `'width', 'height', 'articleContent.width'`

front/scripts/main/views/MediaLightboxView.ts
```diff
@@ -103,7 +120,7 @@ App.MediaLightboxView = App.LightboxView.extend(App.ArticleContentMixin, App.Lig
 	 * value = -5, max = -6, return -5
 	 * value = -6, max = -3, return -3
 	 */
-	limit: function (value: number, max: number): number {
```
rogatty: Why did you remove `function`?

front/scripts/main/mixins/ArticleContentMixin.ts
```diff
@@ -14,13 +16,13 @@ App.ArticleContentMixin = Em.Mixin.create({
 		width: null
 	},
 
-	onInit: function (): void {
+	onInit: Em.observer(function (): void {
```
rogatty: Why `observer`? Shouldn't it be `Em.on('init', ...),`?
http://emberjs.com/api/classes/Ember.html#method_on

hakubo: hmm or just
```js
init: function () {}
```

front/scripts/main/views/MediaLightboxView.ts
hakubo: I like this syntax
Does it actually work?

kenkouot: Yes, if you remove the `: function` in an object literal declaration or in a class definition. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

rogatty: @kenkouot you've changed it only for two methods in this file. IMO we should revert it. If we want to use this convention we should create a new PR which will update all of our code.

hakubo: I agree with Igor, this should come as a separate PR

lizlux: I don't think we should hold up this PR though by reverting code that we want to move towards anyway. Can we merge as is?

lizlux:
LGTM

PR #712, opened 2015-04-22
SOC-480 Internationalized birthdate input for the Login MVP

lizlux:
This adds a birthdate field that can be localized to the a user's language. Since the user is logged out, language will be gathered from the Accepts-Language header, "lang" cookie, or "uselang" param in the query string. It includes an auto-tab feature that will tab to the next input when a set max characters for an input is reached. 

Notes: 
* This doesn't include validation for when a user enters an invalid value. 
* The birthdate value is ultimately stored and the "birthdate" form input, so on submit, that is the value that should be stored in the DB. 
* The "floating label" interaction will be done with a separate ticket
* There's some issues with the way hapi-i18next handles cookies so it's best to test this feature wtih `?uselang=xx` in the querystring rather than updating your cookies or Accepts-Language header, at least until https://github.com/kenkouot/hapi-i18next/pull/4 is merged. See `/front/locales/xx/auth.json` for supported languages, including different versions of English. 

To do: 
* [ ] Create ticket for floating label interaction
* [ ] Respond to feedback and merge https://github.com/kenkouot/hapi-i18next/pull/4, then updated dependency in mercury repo. 
* [ ] Double check for typescript errors


front/scripts/auth/login/SubmitValidator.ts
```diff
@@ -29,13 +29,9 @@ class SubmitValidator {
 	 * @returns {boolean}
 	 */
 	private areAllFieldsFilled ():boolean {
-		var i;
-		for (i = 0; i < this.inputFields.length; i++) {
-			if (!this.inputFields[i].value) {
```
lizlux: There were a couple typescript errors in this code so I refactored it a bit to fix them.

front/locales/qqq/auth.json
```diff
@@ -0,0 +1,11 @@
+{
+	"date": {
+		"endian": "Choices are \"Little\", \"Middle\", and \"Big\". Please only choose one of those three, and don't translate this string. Little Endian means the day comes first, followed by the month and the year (DD/MM/YYYY). Big Endian means the year comes first, followed by the month and the day (YYYY/MM/DD). Middle means the Month comes first, followed by the day and the year (MM/DD/YYYY).",
+		"month-format": "This is the palceholder text for the birthday month input field. It indicates the expected value is two digits.",
```
mklucsarits: typo here --> "palceholder"

front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	public init(): void {
+		this.initFocus();
+		this.initAutoTab();
+		this.initBirthdateValue();
+	}
+
+	private initFocus(): void {
+		var firstInput: HTMLInputElement = <HTMLInputElement> this.fakeInputs[0];
```
mklucsarits: It seems strange that you'd need to cast the value here. In which case is that needed?

front/scripts/auth/component/AutoTab.ts
```diff
@@ -0,0 +1,58 @@
...
+	form: HTMLFormElement;
+	max: number;
+
+	constructor(input: HTMLInputElement) {
+		this.input = input;
+		this.form = input.form;
+		this.max = parseInt(input.getAttribute('data-max-length'));
+	}
+
+	public init(): void {
```
mklucsarits: I feel like this init is unnecessary and adds an extra step. Is there a reason not to add the event listener as soon as the object is constructed?

front/scripts/auth/component/AutoTab.ts
```diff
@@ -0,0 +1,58 @@
...
+		this.input.addEventListener('input', this.onInput.bind(this));
+	}
+
+	private onInput() {
+		var nextVisibleInput = this.getNextVisibleInput(),
+			length = this.input.value.length;
+
+		if (!this.input.validity.valid) {
+			// TODO: show an error message?
```
mklucsarits: In my opinion, let one class do one thing. AutoTab shouldn't worry about validity, a validator should worry about that.

server/facets/auth/BirthdateInput.ts
```diff
@@ -0,0 +1,95 @@
...
+	endian: string;
+	i18n: any;
+
+	/**
+	 * A date endian is the order in which year, month, and day are displayed
+	 * @see http://en.wikipedia.org/wiki/Date_format_by_country
+	 * @see http://grammarpartyblog.com/2011/07/17/one-little-endian-two-little-endians-formatting-dates-across-the-globe/
+	 */
+	endians: DateEndian = {
+		Big: 'Big',
```
mklucsarits: Do the values need to be capitalized? You're just comparing them with toLowerCase() anyway.

front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -0,0 +1,62 @@
+interface FormElements {
+	birthdate: HTMLInputElement;
+	day: HTMLInputElement;
+	month: HTMLInputElement;
+	year: HTMLInputElement;
+}
+
+class BirthdateInput {
+	el: HTMLElement;
```
bkoval: Just wondering if this var could have a more descriptive name :)

bkoval:
Hey @lizlux - this is lookin' good! Although I got an error while running dev build on this branch:

[14:23:29] Starting 'scripts-front'...
front/scripts/auth/common/BirthdateInput.ts(20,32): 2352 Neither type 'HTMLCollection' nor type 'FormElements' is assignable to the other.
  Property 'birthdate' is missing in type 'HTMLCollection'.

front/scripts/auth/component/AutoTab.ts
bkoval: Agreed, we could create SignupValidator class for that :)

front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	private initAutoTab(): void {
+		Array.prototype.forEach.call(this.fakeInputs, function (input: HTMLInputElement) {
+			new AutoTab(input).init();
+		});
+	}
+
+	private initBirthdateValue(): void {
+		this.el.addEventListener('input', this.setRealValue.bind(this));
```
bkoval: Do we have to do it on every input? Maybe it would just be enough to perform it just before form submission :)

front/scripts/auth/component/AutoTab.ts
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		if (
+			length >= this.max &&
```
bkoval: Yeah, I guess we could think if we should let users paste numbers longer than maxlengh or we should use .js to trim them :) But it's a different story, let's consider it an action point for the future.

Also, I think both checks fit nicely into one line, what do you think?

front/styles/auth/module/_birthdate-field.scss
```diff
@@ -0,0 +1,14 @@
+.birthdate-container {
+	padding-bottom: 1px;
```
bkoval: Hey, maybe just 0 here? :)

front/scripts/auth/component/AutoTab.ts
lizlux: The idea is to keep side-affects out of constructors. Just googled around for an explanation that would be better than what I can put to words here, and found http://blog.millermedeiros.com/constructors-should-not-cause-side-effects/

front/scripts/auth/common/BirthdateInput.ts
lizlux: The problem was that fakeInputs is just a NodeList, not necessarily a list of input elements, so calling `focus()` on line 42 was erroring if I didn't specify that the element was an input element.

front/scripts/auth/common/BirthdateInput.ts
lizlux: I have it this way so the submit button will live update (enable/disable) on key-press. The hidden field is a required field in order to keep the label color updating correctly, and also b/c it's ultimately the field that gets used upon form submission.

server/facets/auth/BirthdateInput.ts
lizlux: Lower case would be in keeping with variable naming. I'll switch it.

lizlux:
New PR is https://github.com/Wikia/mercury/pull/720

front/scripts/auth/component/AutoTab.ts
bkoval: I agree with that, we should seperate instantiation of the object and any background actions to be taken by it.

PR #713, opened 2015-04-23
HG-595 fixing access_token cookie setting and removing unused cookie

lizlux:
The `access_token` cookie wasn't actually being set upon login. Now it is.

### DIFF
--- a/server/facets/auth/login.ts
+++ b/server/facets/auth/login.ts
@@ -129,7 +129,6 @@ export function post (request: Hapi.Request, reply: any): void {
 		}
 
 		request.auth.session.set({
-			'user_id'       : response.user_id,
 			'access_token'  : response.access_token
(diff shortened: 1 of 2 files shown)


kenkouot:
LGTM :closed_lock_with_key:

server/facets/auth/login.ts
```diff
@@ -129,7 +129,6 @@ export function post (request: Hapi.Request, reply: any): void {
 		}
 
 		request.auth.session.set({
-			'user_id'       : response.user_id,
```
mklucsarits: If I understand correctly, this change will create a cookie called access_token, which then contains an encoded object with a property called access_token. The cookie seems inappropriately named, in this case. It doesn't expressly contains the access token in plain text.

PR #714, opened 2015-04-24
Set hammerOptions.cssProps.touchCallout to 'default'

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-480

This fixes context menu on iOS 8.3. No idea why it was working in the previous versions, probably some bug in Safari?


bognix:
LGTM

Archbold:
Verified fixed 30-Apr-15 iPhone 6 iOS 8.3 
http://sandbox-qa01.harrypotter.wikia.com/wiki/Main_Page (pulled from CONCF-480)

PR #715, opened 2015-04-25
CONCF-461 Basic collection view

dianafa:
https://wikia-inc.atlassian.net/browse/CONCF-461

Collection Component uses ImageMedia components to display collection elements, as it uses lazy loading to load pics.

@hakubo @rogatty @Warkot


Warkot:
LGTM

front/styles/main/component/_collection.scss
```diff
@@ -0,0 +1,22 @@
+$collection-view-image-size: 100px;
```
rogatty: Is it done only by CSS? Thumbnailer module should be used to get image URL with appropriate size. It's possible that ImageMedia does it by default but I think we have to pass some param.

front/scripts/main/models/ArticleModel.ts
```diff
@@ -168,7 +170,8 @@ App.ArticleModel.reopenClass({
 					media: App.MediaModel.create({
 						media: article.media
 					}),
-					categories: article.categories
+					categories: article.categories,
+					collections: article.collections
```
rogatty: What if it doesn't exist?

front/styles/main/component/_collection.scss
```diff
@@ -0,0 +1,22 @@
...
+
+.collection-component {
+	text-align: center;
+
+	.article-image.visible,
+	.media-component {
```
rogatty: Isn't `.media-component` enough?

front/templates/main/components/collection-component.hbs
```diff
@@ -0,0 +1,6 @@
+<p> {{{header}}} </p>
```
rogatty: Spaces.

front/templates/main/components/collection-component.hbs
```diff
@@ -0,0 +1,6 @@
...
+{{#each item in items}}
+	<a href={{item.link}}>
+		{{image-media url=item.img caption=item.description}}
+	</a>
```
rogatty: Let's keep it in one line.

front/styles/main/component/_collection.scss
dianafa: it provides only height resize. I would implement also width if this code would hit the production. For now this solution seems to be ok.

kenkouot:
May I ask what the context/use case for this is?

dianafa:
@ken This **was** the part of our work in Modular Main Pages project. We wanted to create a new `<collection>` tag in the wikitext and then send it together with all other aticle parts i.e. media, comments, users etc. 
However, we've decided to abandon this solution and switch to Curated Content approach. So for now it won't be used.

PR #720, opened 2015-04-27
Soc 480 Internationalized birthdate input

lizlux:
Old PR was https://github.com/Wikia/mercury/pull/712. Please see old PR for description.


bkoval:
Awesome work in this, LGTM! :cake:

server/facets/auth/BirthdateInput.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		this.inputData = this.orderInputDataByLocale(inputData);
+	}
+
+	private orderInputDataByLocale(data: DateElements): Array<InputData> {
+		var inputsArr: Array<InputData> = [],
+			endian = this.endian.toLowerCase();
+
+		switch (endian) {
+			case this.endians.big.toLowerCase():
```
mklucsarits: You can remove these calls to toLowerCase() here.

lizlux: I just left them in there for safety, in case a translator uses upper case for some reason.

mklucsarits:
LGTM

lizlux:
@bkoval, the css changes LGTM!

mklucsarits:
Still looks good to me, but it can't be merged right now.

server/facets/auth/BirthdateInput.ts
kvas-damian: @lizlux 
What do you think about storing info about date format in other place than translation files?
I'm not sure if it's the right place for locale specific settings. I'm thinking about some setup file for localization module (in future it can include hour format, currency, etc). For me it makes more sense there, additionally it'd be available globally instead of having specific place for storing date format only for auth pages.

lizlux: @kvas-damian, that's an interesting idea. I think the reason I wanted to put this in the i18n file was b/c there's actual strings that needed to be translated, like "MM", and the date separators. For example, Japan and China do it very differently from US https://github.com/Wikia/mercury/blob/SOC-480/front/locales/ja/auth.json#L3-L9. 

That said, if you want to store the actual format/endian somewhere else, I could see that making sense, since it's not a translated string, it's more of a localization settings (hence the confusion for the translators!). There is something to be said for keeping this information in one place though.

kvas-damian: I'm not sure if it needs to be done while working on this ticket, but I would like to clarify which team should take care of that. Is it Social or Mercury?

lizlux: Login and registration is owned by social, so this is social team work. 

On that note though, the mercury team is working on defining areas of responsibility for mercury, which will make questions like this more easy to answer. So stay tuned and thanks for asking!

garthwebb: I agree, I don't think this config belongs in the localization directory.  The localization required for this is not specific to Wikia and I don't think we would expect translators to know how to change these values anyway.

As feedback to a Hg pull request, I think this is work Hg should be doing.  If storing date formats was out of scope for this pull request and Damian suggested adding it, I think that would be fair game for a separate ticket which could be owned by whichever team holds responsibility over the domain.

hakubo:
Do we plan to display date in more places maybe we should consider http://momentjs.com/

hakubo:
I see no unit tests here, I think core Mercury team should set a nice example on how to write them :)

lizlux:
@hakubo, agreed 100% about unit tests, and I should have written them before creating this PR.  However, the social team owns this code (I wrote it just before switching to HG), so ideally they would write the tests.

hakubo:
@lizlux this sounds to me like passing the buck. Engineers should be responsible for the code and quality of it as a whole. If we'll be doing such things (passing stuff around without care) all the time we'll never have clean, understandable and tested code.

kvas-damian:
@hakubo I'm not sure how http://momentjs.com/ can help us with date input localization. See mocks:
sign up with email - default filled

What we want to achieve is input field that is split into sections for day, month and year localized correctly.

lizlux:
@hakubo, I created https://wikia-inc.atlassian.net/browse/HG-749 for tests. I will talk to the rest of the HG team about adding it to the queue for next week.

hakubo:
@kvas-damian ok, understood, thanks
@lizlux thank you! :+1:

front/scripts/auth/component/AutoTab.ts
```diff
@@ -0,0 +1,51 @@
+class AutoTab {
+	input: HTMLInputElement;
+	form: HTMLFormElement;
+	max: number;
+
+	constructor(input: HTMLInputElement) {
+		this.input = input;
+		this.form = input.form;
+		this.max = parseInt(input.getAttribute('maxlength'));
```
rafalkalinski: how about a second parameter, 10?

front/styles/auth/module/_birthdate-field.scss
```diff
@@ -0,0 +1,24 @@
+.birthdate-container {
+	display: flex;
+	flex-direction: row;
+	height: 3rem;
+	justify-content: center;
+	margin-bottom: 1rem;
+	padding-bottom: 0;
+
+	input,
+	input:focus {
```
rafalkalinski: Do we need that line to override some different :focus styles?

kvas-damian: Yes, there are at least couple styles that need to be overridden: https://github.com/Wikia/mercury/blob/dev/front/styles/auth/base/_input.scss

front/scripts/auth/component/AutoTab.ts
```diff
@@ -0,0 +1,51 @@
...
+	max: number;
+
+	constructor(input: HTMLInputElement) {
+		this.input = input;
+		this.form = input.form;
+		this.max = parseInt(input.getAttribute('maxlength'), 10);
+	}
+
+	public init(): void {
+		this.input.addEventListener('input', this.onInput.bind(this));
```
kvas-damian: I'm not sure if ```input``` is best event to listen on. We already got some problems with it on login pages.
@bkoval what do you think about it?

bkoval: I added a 'change' event handler to make AutoTab react on browser autocompletion :)

rafalkalinski:
LGTM

kvas-damian:
LGTM

server/facets/auth/signup.ts
```diff
@@ -1,3 +1,9 @@
+/// <reference path='../../../typings/hapi/hapi.d.ts' />
+/// <reference path='./BirthdateInput.ts' />
+
+var BirthdateInput = require('./BirthdateInput');
+var dateUtils = require('../../lib/DateUtils');
```
rogatty: One `var` per scope, see https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#variables

PR #721, opened 2015-04-28
OPS-6210: Jenkins - apply updates to dependencies for tests

ljagiello:
Ticket: https://wikia-inc.atlassian.net/browse/OPS-6210

The idea is to keep md5 sum of a file and based on that sum remove or not node_modules directory (keep modules or rebuild dependencies).

Code below should be a part of "Execute shell" in Jenkins (before (re)?build dependencies step):
```
#!/usr/bin/env bash

set -e

compute_md5() {
  local output="$(openssl md5)"
  echo "${output##* }"
}

compare_file() {
  local file=$1
  local md5="$(compute_md5 < ${file} )"
  if [ ! -f ".${file}.md5" ];
    then
    echo "No information about previous version of ${file}. Removing node_modules directory..."
    rm -rf node_modules
    echo ${md5} > ".${file}.md5"
  elif [[ ${md5} != `cat ".${file}.md5"` ]];
  then
    echo "New version of ${file}. Removing old node_modules directory..."
    rm -rf node_modules
    echo ${md5} > ".${file}.md5"
  fi
}

bower="bower.json"
package="package.json"

compare_file "${bower}"
compare_file "${package}"
```
/cc @lizlux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ test/coverage
 /* fixtures */
 test/fixtures
 
+/* Jenkins Cache */
+.*.md5
+
 .idea
 /docs
 *.sublime-*


lizlux:
LGTM

PR #723, opened 2015-04-28
SOC-694 UI Fixes After Design Review

bkoval:
@mklucsarits @lizlux https://wikia-inc.atlassian.net/browse/SOC-694


lizlux:
LGTM

front/styles/auth/module/_buttons.scss
```diff
@@ -1,4 +1,14 @@
-button, a.button {
+button,
+a.button,
+button[disabled] {
+	background-color: $button-color;
+	border-color: $button-color;
```
mklucsarits: These two color styles should not be needed. They're already set by the style guide scss. When I use your branch and remove these styles in developer tools, I see that the buttons are still colored correctly. Also, I don't think the $button-color definition is needed then, either.

front/svg/images/eye-crossed.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sketch="http://www.bohemiancoding.com/sketch/ns" width="41px" height="41px" viewBox="0 0 41 41" version="1.1"><title>icon - hide</title><desc>Created with Sketch.</desc><defs/><g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage"><g sketch:type="MSArtboardGroup" transform="translate(-548.000000, -453.000000)"><g sketch:type="MSLayerGroup" transform="translate(49.500000, 414.000000)"><g sketch:type="MSShapeGroup"><path d="M518.9 48 C507.9 48 498.5 59.2 498.5 59.2 L498.5 59.6 C498.5 59.6 507.8 70.8 518.9 70.8 C529.9 70.8 539.3 59.7 539.3 59.7 L539.3 59.2 C539.3 59.2 530 48 518.9 48 L518.9 48 ZM518.9 66.4 C515.1 66.4 512.1 63.3 512.1 59.6 C512.1 55.8 515.1 52.7 518.9 52.7 C522.6 52.7 525.6 55.8 525.6 59.6 C525.6 63.3 522.6 66.4 518.9 66.4 L518.9 66.4 Z" fill="#1A5EB8"/><path d="M518.6 56 C516.9 56 515.5 57.4 515.5 59.1 C515.5 60.9 516.9 62.3 518.6 62.3 C520.3 62.3 521.7 60.9 521.7 59.1 C521.7 57.4 520.3 56 518.6 56 " fill="#1A5EB8"/><path d="M500.5 78.1 L538 40.6 " stroke="#1A5EB8" stroke-width="4"/></g></g></g></g></svg>
```
mklucsarits: This file has some unneeded markup added from Sketch. Can it be removed?

bkoval: I already removed everything I could do manually and by SVG optimizer. Deleting any of the remaining tags was causing the svg to fail to render. However, we can get back to it here: https://wikia-inc.atlassian.net/browse/HG-588 if there's any other way to optimize it.

front/styles/auth/module/_buttons.scss
bkoval: You're right. I checked it and bower probably cached one of the older versions of SG package, which uses dardker color as the default button color.

bkoval:
@mklucsarits I moved styles you were talking about to style just disabled button. However, they need to stay there until https://github.com/Wikia/style-guide/issues/32 gets implemented.

mklucsarits:
LGTM. I removed the extra markup from the SVG file and confirmed it still appears correctly.

PR #725, opened 2015-04-28
enables internal varnish for mercury

kenkouot:
Sources`mediawikiDomain` in config from environment variables.


kenkouot:
Verified on sandbox.

PR #727, opened 2015-04-29
SOC-695 Adding whitelisting for redirects to the new login flow

bkoval:
@rogatty @mklucsarits We should redirect to the new login flow if:

 - a wiki is on the whitelist
 - a wiki is japanese

This should handle it. Ticket: https://wikia-inc.atlassian.net/browse/SOC-695


front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -5,7 +5,49 @@ App.LoginIconComponent = Em.Component.extend({
...
 	click: function (): void {
-		window.location.href = '/join?redirect=' + encodeURIComponent(window.location.href);
+		if (this.shouldRedirectToNewLogin()) {
+			window.location.href = '/join?redirect=' + encodeURIComponent(window.location.href);
+		} else {
+			window.location.href = '/Special:UserLogin';
+		}
+	},
+
+	isJapanese: function (): void {
```
bkoval: Cannot reuse logic from ShareFeatureComponent, I copied it. Any ideas for a place where we could abstract it and use in both places?

rogatty: Mixin would be the simplest.

front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -5,7 +5,49 @@ App.LoginIconComponent = Em.Component.extend({
...
+		}
+		return lang === 'ja';
+	},
+
+	/**
+	 * Redirects to new login flow if a wiki is japanese / whitelisted above
+	 * @returns {boolean}
+	 */
+	shouldRedirectToNewLogin: function (): boolean {
+		var hostname = window.location.hostname,
```
rogatty: It's not used anywhere.

front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -5,7 +5,49 @@ App.LoginIconComponent = Em.Component.extend({
...
+			shouldRedirect = false,
+			dbName = Mercury.wiki.dbName;
+
+		this.newLoginWhitelist.forEach((whitelistedDBName) => {
+			shouldRedirect = whitelistedDBName === dbName;
```
rogatty: This will iterate over all elements and `shouldRedirect` might be set to false even if it should be true. Use `some` instead of `forEach`.
http://stackoverflow.com/a/2641374/1050577

front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -5,7 +5,49 @@ App.LoginIconComponent = Em.Component.extend({
...
+		});
+
+		if (this.isJapanese()) {
+			shouldRedirect = true;
+		}
+
+		return shouldRedirect;
+
```
rogatty: No need for `\n`

front/scripts/main/components/LoginIconComponent.ts
mklucsarits: It would be great if this function could also be generalized to take any language code as a parameter, but I don't know if that can be done and still use it in a template like ShareFeature does.

dianafa:
LGTM!

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -0,0 +1,15 @@
...
+
+	isJapanese: function (): boolean {
+		var lang = navigator.language || navigator.browserLanguage;
+		if (lang) {
+			lang = lang.substr(0, 2);
+		} else {
+			lang = this.get('language.content');
+		}
+		return lang === 'ja';
+	}.property()
```
kenkouot: We're not using the prototype extensions from Ember anymore. You can rewrite this using decorator syntax:
```js
// replace .property with:
Ember.computed(function () ...);
```

bkoval:
Closing this one, please use https://github.com/Wikia/mercury/pull/740

PR #730, opened 2015-04-29
HG-615 adding properties to the list-reset scss placeholder to override ...

lizlux:
...those set with style guide lists

This mixin is also used in the contributors list and recommendations ("read more") list, both of which look fine after the change.


mklucsarits:
LGTM

PR #732, opened 2015-04-30
HG-613 removing redundant button mixins

lizlux:
This is actually related to https://wikia-inc.atlassian.net/browse/HG-614, but I chose the parent ticket as a branch name b/c I closed HG-613 already. 

This PR removes the button mixins since they are redundant now with the button styles we brought in from the style guide. 

@mklucsarits


dianafa:
LGTM

PR #736, opened 2015-04-30
HG 604 Set custom var 8 at track page view time and also check if it is a main page or not

kenkouot:
See title.

### DIFF
--- a/front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
+++ b/front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
@@ -54,7 +54,6 @@ module Mercury.Modules.Trackers {
 				[prefix + '_setCustomVar', 4, 'Skin', 'mercury', 3],
 				// TODO: Currently the only login status is 'anon', in the future 'user' may be an option
 				[prefix + '_setCustomVar', 5, 'LoginStatus', 'anon', 3],
-				[prefix + '_setCustomVar', 8, 'PageType', 'article', 3],
 				[prefix + '_setCustomVar', 9, 'CityId', String(Mercury.wiki.id), 3],
 				[prefix + '_setCustomVar', 15, 'IsCorporatePage', 'No', 3],
 				// TODO: Krux segmenting not implemented in Mercury https://wikia-inc.atlassian.net/browse/HG-456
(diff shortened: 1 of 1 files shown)


front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
```diff
@@ -54,7 +54,6 @@ module Mercury.Modules.Trackers {
 				[prefix + '_setCustomVar', 4, 'Skin', 'mercury', 3],
 				// TODO: Currently the only login status is 'anon', in the future 'user' may be an option
 				[prefix + '_setCustomVar', 5, 'LoginStatus', 'anon', 3],
```
lizlux: @kenkouot, is this line something we need to make sure gets changed when we enable logged in state on mercury?

front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
```diff
@@ -105,6 +105,16 @@ module Mercury.Modules.Trackers {
 		 * Tracks the current page view
 		 */
 		trackPageView (): void {
+			var mainPageTitle = Mercury.wiki.mainPageTitle,
+				isMainPage = window.location.pathname.split('/').indexOf(mainPageTitle),
+				pageType = ['_setCustomVar', 8, 'PageType', isMainPage ? 'home' : 'article', 3];
```
lizlux: are we sure we don't need a prefix before `_setCustomVar`?

front/scripts/mercury/modules/Trackers/GoogleAnalytics.ts
```diff
@@ -105,6 +105,16 @@ module Mercury.Modules.Trackers {
...
+
+			this.queue.push(pageType),
+
+			// Set custom var in ad account as well
+			pageType[0] = this.accounts[this.accountAds].prefix + '.' + pageType[0];
```
lizlux: is there a reason we're not using the same prefix getting scheme as above? (just wondering, I'm new to this code)

kenkouot: Answered this offline!

lizlux:
LGTM

PR #737, opened 2015-04-30
ADEN-1993 Fixed ads lightbox styling in Mercury

nandy-andy:
After recent lightbox refactoring the ads lightbox styling got broken.

As a quick fix and workaround I've added a new wrapper in ads lightbox so the styling is not broken.

### DIFF
--- a/front/templates/main/ads-lightbox.hbs
+++ b/front/templates/main/ads-lightbox.hbs
@@ -1,1 +1,3 @@
-{{contents}}
+<div class="lightbox-content-inner">
+	{{contents}}
+</div>


kenkouot:
What's the difference between this and the long fix?

nandy-andy:
The right way of doing that would probably be changing `AdsLightboxView` to `AdsLightboxComponent` or something like that - basically finish the "refactoring" of lightboxes, I guess. 

The idea of ads lightbox was to use as many common parts as possible with "native" Mercury lightboxes, so it doesn't feel so different. We wanted it to change together with its "parent" and from Ad Engineering perspective just "feed" it with different content.

kenkouot:
Thanks for the great answer @nandy-andy! LGTM.

PR #738, opened 2015-05-01
HG-625 setting helios config vars

lizlux:
We need the private helios config values to be available to mercury settings. Previously, we were overriding these vars in our local settings, but that solution doesn't work for production environments. 

Corresponding chef PR to make these vars available to the mercury environment: https://github.com/Wikia/chef-repo/pull/5065.

@kenkouot


kenkouot:
goo to me

PR #740, opened 2015-05-04
SOC-695 Adding whitelisting for redirects to the new login flow

bkoval:
Original Pull Request with comments: https://github.com/Wikia/mercury/pull/727

### DIFF
--- a/front/scripts/main/mixins/LanguagesMixin.ts
+++ b/front/scripts/main/mixins/LanguagesMixin.ts
@@ -3,7 +3,7 @@
 
 App.LanguagesMixin = Em.Mixin.create({
 
-	isJapanese: function (): boolean {
+	isJapanese: Ember.computed(function (): boolean {
 		var lang = navigator.language || navigator.browserLanguage;
 		if (lang) {
 			lang = lang.substr(0, 2);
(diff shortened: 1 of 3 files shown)


rogatty:
LGTM

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -0,0 +1,14 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.LanguagesMixin = Em.Mixin.create({
+	isJapanese: Ember.computed(function (): boolean {
+		var lang = navigator.language || navigator.browserLanguage;
```
lizlux: I'm just wondering if we want to use our i18n implementation for determining the language, since it takes into account headers, cookies, and query string as opposed to just the language set in the browser. Not a big deal since this is temporary and probably works for our purposes. It might be easier to test if we could set language via query string though.

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -0,0 +1,14 @@
...
+		if (lang) {
+			lang = lang.substr(0, 2);
+		} else {
+			lang = this.get('language.content');
```
lizlux: ah, this is probably what I was talking about above (haven't actually used this feature on the front end yet, just the hapi-i18next version). Well then my question is, what's the purpose of checking both browser lang and `language.content`?

PR #744, opened 2015-05-05
HG-631 setting and unsetting unencrypted access_token cookie

lizlux:
Previously the access_token cookie's value was encrypted, so MW wouldn't have been able to read it. Now we're setting two cookies, one for hapi to manage an authenticated session, `sid`, and one that tags along and acts as the bridge between MW and Mercury, `access_token`. 

I created https://wikia-inc.atlassian.net/browse/HG-632 to make sure we document which cookies are used where in mercury authentication.


mklucsarits:
LGTM

kenkouot:
LGTM

PR #747, opened 2015-05-06
Fixing cookie domain

lizlux:
Previously, when we'd set a cookie from (sandbox.)www.wikia.com/login, the cookie's domain was being set to (sandbox.)www.wikia.com. This changes it so it is set to a `cookieDomain` var, where `cookieDomain` is either a variable set in `localSettings.ts` or `.wikia.com`. We should only have to update `localSettings.ts` for local environments and devboxes. Sandboxes, staging servers and prod should be fine with `.wikia.com`. 

Also, /join and /signup weren't getting the hapi-auth-cookie settings imported into the view context, so features like redirecting those page to the home page when logged in weren't working. They should work now. 

@bkoval @mklucsarits @kenkouot

### DIFF
--- a/config/localSettings.d.ts
+++ b/config/localSettings.d.ts
@@ -27,7 +27,7 @@ interface LocalSettings {
 	domain: string;
 	mediawikiDomain?: string;
 	devboxDomain?: string;
-	cookieDomain?: string;
+	authCookieDomain?: string;
 	environment: any;
 	helios: {
 		host: string;
--- a/config/localSettings.d.ts
+++ b/config/localSettings.d.ts
@@ -27,6 +27,7 @@ interface LocalSettings {
 	domain: string;
 	mediawikiDomain?: string;
 	devboxDomain?: string;
+	cookieDomain?: string;
 	environment: any;
 	helios: {
 		host: string;
--- a/server/app.ts
+++ b/server/app.ts
@@ -79,6 +79,7 @@ server.register(plugins, (err: any) => {
 		cookie         : 'sid',
 		isSecure       : false,
 		password       : localSettings.ironSecret,
+		domain         : localSettings.cookieDomain || '.wikia.com',
 		redirectTo     : '/login'
 	});
 });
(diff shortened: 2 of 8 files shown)


config/localSettings.d.ts
```diff
@@ -27,6 +27,7 @@ interface LocalSettings {
 	domain: string;
 	mediawikiDomain?: string;
 	devboxDomain?: string;
+	cookieDomain?: string;
```
rogatty: This name might be confusing as in the app there is `wgCookieDomain` which is either `.wikia.com` or a custom domain like `.wowwiki.com`.

bkoval:
LGTM

server/app.ts
```diff
@@ -21,6 +21,7 @@ import fs            = require('fs');
 var counter = 1,
 	isDevbox: boolean = localSettings.environment === Utils.Environment.Dev,
 	plugins: any,
+	cookieDomain: string = localSettings.cookieDomain || '.wikia.com',
```
mklucsarits: I know I'm in the minority when it comes to this opinion, but I don't think domain names should exist as strings anywhere in code, unless it's defined as a constant to be used in multiple places, or is in a config file.

lizlux: yeah, I thought about this after the fact, I'll put it in localSettingsBase.

config/localSettings.d.ts
lizlux: You're talking about `wgCookieDomain` in the MW app, right? I actually modeled this after the MW version, since they are basically doing the same thing. 

It's true though, that I didn't add consideration for the custom domains. Since they've been deprecated though, I'm hesitant to add more code to support them. Maybe a comment will do? We're not launching auth pages on custom domains, which is what this is being used for.

lizlux: I changed the variable name to `authCookieDomain` so there wouldn't be any confusion as to what it actually provides.

rogatty:
LGTM

PR #749, opened 2015-05-07
SOC-750 Password field should be masked by default

bkoval:
@mklucsarits https://wikia-inc.atlassian.net/browse/SOC-750


mklucsarits:
LGTM

PR #750, opened 2015-05-07
New login to dev integration

bkoval:
This puts new login under the flag. This way we can easily release the new login, by setting newLoginEnabled flag in localSettings to true. Also this removes new login flow from Japanese Wikias (even with the main flag set) as requested by the Product Team.

@lizlux @mklucsarits @kenkouot @rogatty

This should allow us to merge the whole new-login-mvp branch to dev and be independent from the Mercury releases.

### DIFF
--- a/front/scripts/main/components/LoginIconComponent.ts
+++ b/front/scripts/main/components/LoginIconComponent.ts
@@ -1,12 +1,13 @@
-/// <reference path="../../main/mixins/LanguagesMixin.ts" />
+/// <reference path="../../baseline/mercury.ts" />
+
 'use strict';
 // This was disabled for now and should be re-enabled with https://wikia-inc.atlassian.net/browse/SOC-633 when
 // we're ready to launch the new auth pages.
-App.LoginIconComponent = Em.Component.extend(App.LanguagesMixin, {
+App.LoginIconComponent = Em.Component.extend({
 	tagName: 'a',
 	classNames: ['external', 'login'],
 	//Let's remove this flag once we're good to go with the new login flow
-	newLoginEnabled: false,
+	newLoginEnabled: M.prop('newLoginEnabled'),
 
 	newLoginWhitelist: [
 		'clashofclans',
@@ -24,7 +25,7 @@ App.LoginIconComponent = Em.Component.extend(App.LanguagesMixin, {
 	},
 
 	/**
-	 * Redirects to new login flow if a wiki is japanese / whitelisted above
+	 * Redirects to new login flow if a wiki is whitelisted above
 	 * @returns {boolean}
 	 */
 	shouldRedirectToNewLogin: function (): boolean {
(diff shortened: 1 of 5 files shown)


server/facets/showArticle.ts
```diff
@@ -61,6 +61,7 @@ function onArticleResponse (request: Hapi.Request, reply: any, error: any, resul
 		reply.redirect(localSettings.redirectUrlOnNoData);
 	} else {
 		Tracking.handleResponse(result, request);
+		result.newLoginEnabled = localSettings.newLoginEnabled;
```
bkoval: This doesn't seem like the best place to export it, but I couldn't locate a place where apiBase and other properties form localSettings are being exported to ember-main template.

server/views/_layouts/ember-main.hbs
```diff
@@ -42,6 +42,7 @@
 			M.prop('translations', {{{json (translations wiki.language.user ns='main')}}});
 			M.prop('weppyConfig', {{{json weppyConfig}}});
 			M.prop('wikiaLanguage', '{{{wiki.language.user}}}');
+			M.prop('newLoginEnabled', {{{newLoginEnabled}}});
```
lizlux: Since `newLoginEnabled` is a boolean and doesn't need to be unescaped we can use two brackets instead of three.

front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -14,7 +17,7 @@ App.LoginIconComponent = Em.Component.extend(App.LanguagesMixin, {
 	],
 
 	click: function (): void {
-		if (this.shouldRedirectToNewLogin()) {
+		if (this.newLoginEnabled && this.shouldRedirectToNewLogin()) {
```
rogatty: I'd put it in the `shouldRedirectToNewLogin()`

lizlux:
One minor comment but LGTM

rogatty:
I'm not sure about keeping the flag in `localSettings` and polluting `M` object.

@kenkouot what would be the best approach to have a feature hidden behind a flag?

bkoval:
Yeah, but Weppy's config is still getting to Front-End through M:
"M.prop('weppyConfig', {{{json weppyConfig}}});"

aquilax:
Can't you pass the setting from `WikiFactory` or I'm missing the point? :point_left:

bkoval:
We had an offline discussion with @rogatty and decided that moving this check to WikiFactory would be an overkill since this flag will be removed anyway after we ship the new login.

server/facets/showArticle.ts
```diff
@@ -74,7 +74,6 @@ function onArticleResponse (request: Hapi.Request, reply: any, error: any, resul
 			(localSettings.environment === Utils.Environment.Prod ?
 				localSettings.optimizely.account : localSettings.optimizely.devAccount) + '.js';
 		}
-
```
rogatty: I like this `\n`, give it back!

rogatty:
Return the `\n` and it'll LGTM.

PR #752, opened 2015-05-07
SOC-745 Changing hero unit for /join page

bkoval:
@mklucsarits https://wikia-inc.atlassian.net/browse/SOC-745


front/locales/en/auth.json
```diff
@@ -21,7 +21,7 @@
 	},
 	"join": {
 		"title": "Join Wikia",
-		"register-with": "Register with",
+		"join-wikia-today": "Join Wikia Today",
```
bkoval: Any idea on a better name for this thing? ;)

aquilax: How about `join-text`?

mklucsarits: I agree with our guest star commenter.

bkoval: Wowow, So glad to hear from you @aquilax !!! :dancer:

PR #756, opened 2015-05-07
HG-612 Decouple Application and Article

kenkouot:
This PR addresses some ugly technical debt from Mercury's prototype days. We were doing all sorts of weird rendering stuff, and also failed to correctly isolate the application and article dependencies from one another.


rogatty:
Let's run automation tests before merging this one.

rogatty:
@kenkouot tests are failing with
`front/scripts/main/routes/ApplicationRoute.ts(50,5): 7005 Variable 'title' implicitly has an 'any' type.`

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -21,13 +21,6 @@ App.ArticleRoute = Em.Route.extend({
 		if (Mercury.error) {
 			transition.abort();
 		}
-
-		// If you try to access article with not-yet-sanitized title you can see in logs:
-		// `Transition #1: detected abort.`
-		// This is caused by the transition below but doesn't mean any additional requests.
-		this.transitionTo('article',
```
kenkouot: @dianafa I remove this code because it seemed to be extraneous. Can you offer some some context as to what cases it covered?

dianafa: Do not remove pls :P This part of code is responsible for handling urls like ".../wiki/Miss Piggy" and displaying them like ".../wiki/Miss_Piggy" instead of ".../wiki/Miss%20Piggy". So basically it is the fake transition for the 'next article': the one with sanitized url.

kenkouot: Can you explain what situation causes those type of URLs? Can we sanitize them before the route is triggered? Maybe you could give me a few reproduction cases because in testing the URLs with spaces in them seemed fine :smile:

dianafa: Yes, the usage is quite common: go to http://elderscrolls.127.0.0.1.xip.io:8000/wiki/The_Elder_Scrolls_Wiki and choose random page. If the name of article will contain for example spaces, instead of: `wiki/Sodres_Nerethi` you will see `wiki/Sodres%20Nerethi`. Moreover, when you type manually, from memory, `wiki/Sodres Nerethi` you will also see ugly URL. 

I don't think sanitization before the route is triggered is possoble, as this is the first place when we can get the url/article.

kenkouot: Talked offline w/ @dianafa about this. We're going to add this code back in for now, but will work towards not using an Ember transition to handle this case.

front/scripts/main/views/ArticleView.ts
```diff
@@ -23,10 +23,17 @@ App.ArticleView = Em.View.extend(App.AdsMixin, {
...
 	willInsertElement: function (): void {
-		Em.addObserver(this.get('controller'), 'model.article', this, this.onArticleChange);
-		// Trigger an article change once on insertion because the first insertion happens after article
-		// state has changed
-		this.get('controller').notifyPropertyChange('model.article');
+		this.scheduleArticleTransforms();
+	},
+
+	onModelChange: Em.observer('controller.model.article', function (): void {
+		if (this.$()) {
```
inez: What is this checking?

kenkouot: It checks if there is a valid element on this view. Events were firing when views were getting trashed as well which was causing errors, this is making sure that this observer is only called when a view has a valid element.

rogatty: @kenkouot can you add this explanation as a comment?

kenkouot: Good suggestion

front/scripts/main/router.ts
```diff
@@ -16,6 +16,8 @@ App.Router.map(function () {
 		path: '/'
 	});
 
+	this.route('test');
```
lizlux: is this debug code?

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -40,6 +44,11 @@ App.ArticleRoute = Em.Route.extend({
 
 	afterModel: function (model: any) {
 		this.controllerFor('application').set('currentTitle', model.get('title'));
+		App.VisibilityStateManager.reset();
+
+		// Reset query parameters
+		model.set('commentsPage', null);
+		model.set('file', null);
```
lizlux: do we need to add `cb` to handled query params here?

lizlux:
Seems legit, but this is all pretty new to me.

front/scripts/main/router.ts
rogatty: Looks like it.
@kenkouot wanna remove it?

front/scripts/main/views/ArticleView.ts
```diff
@@ -23,10 +23,19 @@ App.ArticleView = Em.View.extend(App.AdsMixin, {
...
 	 */
 	willInsertElement: function (): void {
-		Em.addObserver(this.get('controller'), 'model.article', this, this.onArticleChange);
-		// Trigger an article change once on insertion because the first insertion happens after article
-		// state has changed
-		this.get('controller').notifyPropertyChange('model.article');
+		this.scheduleArticleTransforms();
+	},
+
+	onModelChange: Em.observer('controller.model.article', function (): void {
```
rogatty: I'd call it `modelArticleObserver` or something else with `Observer` suffix to keep it consistent with the rest of code.

front/scripts/main/views/ArticleView.ts
```diff
@@ -23,10 +23,19 @@ App.ArticleView = Em.View.extend(App.AdsMixin, {
...
+		// This check is here because this observer will actually be called for views wherein the state is actually
+		// not valid, IE, the view is in the process of preRender
+		if (this.get('_state') === 'inDOM') {
+			this.scheduleArticleTransforms();
+		}
+	}),
+
+	scheduleArticleTransforms: function (): void {
+		Ember.run.scheduleOnce('afterRender', this, this.onArticleChange);
```
rogatty: Use `Em`

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -25,9 +25,13 @@ App.ArticleRoute = Em.Route.extend({
...
-		this.transitionTo('article',
-			M.String.sanitize(title)
-		);
+		// TODO: This could be improved upon by not using an Ember transition to 'rewrite' the URL
+		// Ticket here: https://wikia-inc.atlassian.net/browse/HG-641
+		if (title.match(/\s/)) {
+			this.transitionTo('article',
+				M.String.sanitize(title)
+			);
+		}
```
dianafa: this `if()` is actually unnecessary. Transition will not fire if there will be no change between current and passed title.

kenkouot: I prefer the explicitness of not even attempting the transition if it doesn't have spaces

PR #757, opened 2015-05-08
Hg 612 fixing jump-to links in TOC with cb param in query string

lizlux:
This PR designates which content we need to apply special click logic to and which content we can let Ember handle on it's own. This should help simplify click handler logic by isolating it to only UGC. 

To do: 
- [ ] document this new convention in the github wiki pages. 
- [ ] run selenium tests


kenkouot:
LGTM!

PR #768, opened 2015-05-11
DAT-2781 adjust infobox height according to device width

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2781

For big pics infobox was looking like a normal image. This PR adjusts infobox height basing on the device screen width.

ping @rogatty @Warkot @hakubo


front/scripts/main/views/ArticleView.ts
```diff
@@ -162,12 +162,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, {
...
 	 * @desc handles expanding portable infoboxes
+	 * The minimumHeight took from 9/16 proportions of screen (width * 16 / 9 + 100px). We want to always
+	 * show the image AND some other infobox informations to show that this is infobox, not only an ordinary image.
 	 * @todo we should figure out if we can somehow merge this method and handleInfoboxes method
 	 */
 	handlePortableInfoboxes: function (): void {
 		var collapsedClass = 'collapsed',
 			expandButtonClass = 'portable-infobox-expand-button',
-			minimumHeight = 450,
+			deviceWidth = (window.innerWidth > 0) ? window.innerWidth : screen.width,
```
hakubo: this will work but
I think more common used pattern is:
`deviceWidth = window.innerWidth || screen.width,`

hakubo:
LGTM

front/scripts/main/views/ArticleView.ts
rogatty: Use [ViewportMixin](https://github.com/Wikia/mercury/blob/dev/front/scripts/main/mixins/ViewportMixin.ts) instead. It reacts to resizing too.

Warkot:
It LGTM, but take @rogatty's comment about the mixin.

front/scripts/main/views/ArticleView.ts
dianafa: well, do we really need a mixin for this simple check?

rogatty: That's what this mixin is for. Let's not repeat the code as it makes things harder to debug.

RafLeszczynski:
LGTM

PR #769, opened 2015-05-11
HG-647 adding cb value to url upon login

lizlux:
MediaWiki needs a cachebuster value added to the URL when users log in. Without this we could see cached logged out pages. 

@bkoval @kenkouot


server/facets/auth/login.ts
```diff
@@ -115,8 +115,13 @@ export function post (request: Hapi.Request, reply: any): void {
 		requestedWithHeader: string = request.headers['x-requested-with'],
 		isAJAX: boolean = requestedWithHeader && !!requestedWithHeader.match('XMLHttpRequest'),
 		redirect: string = request.query.redirect || '/',
+		successRedirect: string,
 		context: LoginViewContext = getLoginContext(redirect),
-		ttl = 1.57785e10; // 6 months,
+		ttl = 1.57785e10, // 6 months
+		cb: number = Math.floor(Math.random() * 10000);
```
rogatty: Would timestamp be a better choice here?

server/facets/auth/login.ts
```diff
@@ -115,8 +115,13 @@ export function post (request: Hapi.Request, reply: any): void {
...
+
+	// add cache buster value to the URL upon successful login
+	successRedirect = redirect + (redirect.indexOf('?') > -1 ? '&' : '?') + 'cb=' + cb;
```
bkoval: This LGTM, but we might consider creating a cachebusting method in https://github.com/Wikia/mercury/blob/new-login-mvp/server/lib/AuthUtils.ts and using URL.parse() for obtaining the url object, this way you don't need to worry about checking if the querystring exists.

bkoval:
Hey, LGTM, just one question left. I know Mercury team is using changelog with names of the pull requests so maybe you could tweak the title a bit? :dart:

server/facets/auth/login.ts
lizlux: It's a good question, but I'm inclined to say that this method is good enough. It's what we've been using on MW forever, and adding a 13 digit cb value to the URL might be overkill from a UX perspective.

bkoval:
@rogatty @lizlux feel free to take a look :)

rogatty:
LGTM

server/lib/AuthUtils.ts
```diff
@@ -26,3 +28,17 @@ export function getLoginUrlFromRedirect(redirect: string): string {
 	forgotPasswordUrlObj.pathname = wikiaLoginPathname;
 	return url.format(forgotPasswordUrlObj);
 }
+
+export function getCacheBustedUrl(url: string): string {
+	var cacheBustedUrlObj = url.parse(redirect),
+	cachebuster = Math.floor(Math.random() * 10000);
+
+	if (cacheBustedUrlObj.search) {
+		cacheBustedUrlObj.search += '&cb=';
```
hakubo: In node you can yous
`query`
that is a `querystring` object

that will simplify this code a bit

https://nodejs.org/api/querystring.html

bkoval: Right, thanks!

lizlux:
LGTM, thanks @bkoval!

PR #770, opened 2015-05-11
CONCF-489 Curated Main Page tracking

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-489

@bognix: I've changed tracking categories so they're the same for all main page modules.
@rogatty: I've added the support for interactive/non-interactive events


front/scripts/main/components/FeaturedContentComponent.ts
```diff
@@ -12,7 +12,7 @@ interface FeaturedContentItem {
 	article_local_url: string;
 }
 
-App.FeaturedContentComponent = Em.Component.extend({
+App.FeaturedContentComponent = Em.Component.extend(App.TrackClickMixin, {
```
bognix: don't you need to add reference path?

front/scripts/main/components/FeaturedContentComponent.ts
```diff
@@ -36,6 +36,10 @@ App.FeaturedContentComponent = Em.Component.extend({
 		},
 	},
 
+	click: function (): void {
```
bognix: why not doing this on the template side using action 'trackClick'?

front/scripts/mercury/utils/track.ts
```diff
@@ -105,10 +106,11 @@ module Mercury.Utils {
...
 		var trackingMethod: string = params.trackingMethod || 'both',
-		    action: string = params.action,
-		    category: string = params.category ? 'mercury-' + params.category : null,
-		    label: string = params.label || '',
-		    value: number = params.value || 0,
+			action: string = params.action,
+			category: string = params.category ? 'mercury-' + params.category : null,
+			label: string = params.label || '',
+			value: number = params.value || 0,
+			isNonInteractive: boolean = !params.isInteractive,
```
rogatty: Let's not do this. Instead make it `params.nonInteractive` to keep consistency with GA API. Also add `true` as the default.

front/scripts/main/components/FeaturedContentComponent.ts
rogatty: Because I want to track all the clicks on the container element of this component. There is no other access to it other than here.

rogatty:
LGTM but I've written half of it :P

bognix:
LGTM but I've written second half of it :P

hakubo:
LGTM and I have nothing to do with this :construction_worker:

Warkot:
LGTM :water_buffalo:

PR #773, opened 2015-05-12
HG-648 temporary fix for footer links

lizlux:
This is a temporary fix to include footer links in the list of content areas that should be given special link handling. I'm going to keep this ticket open (or maybe open a new one) so we can find a better solution. For now we just need to unblock the release.


rogatty:
LGTM

vforge:
:heavy_check_mark:

lizlux:
Created https://wikia-inc.atlassian.net/browse/HG-654 to follow up with this.

PR #775, opened 2015-05-12
HG-653 escape query params before inserting into document

kenkouot:
@Grunny

### DIFF
--- a/server/views/_helpers/escapeQueryParams.ts
+++ b/server/views/_helpers/escapeQueryParams.ts
@@ -1,6 +1,6 @@
 /// <reference path='../../../typings/hoek/hoek.d.ts' />
 import Hoek = require('hoek');
-function decodeQueryParams (query: any): any {
+function escapeQueryParams(query: any): any {
 	var keys = Object.keys(query),
 		obj: {[idx: string]: any} = {};
 
@@ -10,4 +10,4 @@ function decodeQueryParams (query: any): any {
 
 	return obj;
 }
-export = decodeQueryParams;
+export = escapeQueryParams;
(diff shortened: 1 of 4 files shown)


rogatty:
Wouldn't a whitelist be a better solution? For sure it would be safer.

kenkouot:
It would be a better solution and is handled natively in Hapi's validation if we went that path, but @hobbs seems to think that the variability of our query parameters (some have random ids appended to the keys) makes that a non-starter/

hakubo:
Right we also have to think if we still need that part of code that reflects all queryparams.

It's LGTM for the fix, but I would think if we could remove that part at all.

rogatty:
Currently we use `M.prop('queryParams')` in only two places which are:
https://github.com/Wikia/mercury/blob/dev/front/scripts/mercury/utils/track.ts#L116 and https://github.com/Wikia/mercury/blob/dev/front/scripts/mercury/utils/track.ts#L156 both looking like this:
```
if (M.prop('queryParams.noexternals')) {
	return;
}
```

Maybe this should be solved in another way?

rogatty:
LGTM for a quick P2 fix.

server/views/_helpers/escapeQueryParams.ts
```diff
@@ -0,0 +1,13 @@
...
+	var keys = Object.keys(query),
+		obj = {};
+
+	keys.forEach((key) => {
+		obj[Hoek.escapeHtml(key)] = Hoek.escapeHtml(query[key].toString());
+	});
+
+	return obj;
+}
+export = decodeQueryParams;
```
rogatty: File name is `escapeQueryParams.ts`. Should it be consistent?

Grunny:
LGTM as a quick fix. I'm not familiar enough with Mercury yet, but would it be possible to never store query params in a script tag within the document anyway as it is user controlled data? Can they just be retrieved either on demand or retrieved at the start client-side?

kenkouot:
I created HG-655 as a followup to this

kenkouot:
Gonna merge this as a quick fix once tests pass, the root cause should be addressed in our next "sprint"

PR #777, opened 2015-05-13
CONCF-541 | Use Optimizely test to disable CuratedContent

vforge:
Ping @rogatty @hakubo @bognix

### DIFF
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -82,7 +82,7 @@ App.initializer({
 
 App.initializer({
 	name: 'performanceMonitoring',
-	after: 'optimizelyCuratedMainPageLoader',
+	after: 'preload',
 	initialize () {
 		if (typeof EmPerfSender === 'undefined') {
 			return;
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -15,6 +15,7 @@ interface Window {
 
 declare var i18n: I18nextStatic;
 declare var EmPerfSender: any;
+declare var optimizely: any;
 
 var App: any = Em.Application.create({
 	// We specify a rootElement, otherwise Ember appends to the <body> element and Google PageSpeed thinks we are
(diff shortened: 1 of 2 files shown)


front/scripts/main/app.ts
```diff
@@ -70,8 +71,18 @@ App.initializer({
...
+	initialize: () => {
+		// 2870342045 -> Experiment ID in Optimizely / Mercury (production)
+		// 1 -> Variation ID that should have CuratedMainPages Enabled
+		M.prop('optimizelyCuratedMainPage', window.optimizely && optimizely.variationMap[2870342045] == 1);
+	}
+});
+
+App.initializer({
+	name: 'performanceMonitoring',
+	after: 'optimizelyCuratedMainPageLoader',
```
hakubo: since this does not depend on optimizelyCuratedMainPageLoader does it have to run after it?

vforge: Good point.

hakubo:
LgTM

front/scripts/main/app.ts
```diff
@@ -70,6 +71,16 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'optimizelyCuratedMainPageLoader',
+	after: 'preload',
+	initialize: () => {
+		// 2870342045 -> Experiment ID in Optimizely / Mercury (production)
+		// 1 -> Variation ID that should have CuratedMainPages Enabled
+		M.prop('optimizelyCuratedMainPage', window.optimizely && optimizely.variationMap[2870342045] == 1);
```
rogatty: Should we just call it `isCuratedMainPage` as @bognix does?

front/scripts/main/app.ts
```diff
@@ -70,6 +71,16 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'optimizelyCuratedMainPageLoader',
+	after: 'preload',
```
rogatty: Why did you do it like that? `preload` initializer removes `preload` class so doing this will cause content flash, right? From our discussion with @kenkouot I've understood that we should run this code before `preload`.

vforge: It's not dependant on either of initializers and can be run totally separate. I've just wanted it to be run after `preload`, because it's less important.

@kenkouot Can you express your opinion on this?

front/scripts/main/app.ts
vforge: I would rather leave it, because it's totally different from what @bognix does. Her code enables/disables the feature too, but mine is dependant on optimizely only and I wanted to have that explicitly in variable's name - that's i'm using `optimizelyCuratedMainPage` as property's name.

front/scripts/main/app.ts
vforge: Accoring to http://nerdyworm.com/blog/2013/04/03/ember-initializers/ initializers are run before {{App.ready}} so we're good.

rogatty:
LGTM

bognix:
LGTM

PR #786, opened 2015-05-14
DAT-2788: For TOC creation only select H2s that have section attribute specified

inez:
In article HTML H2s (and other header tags) can comeout from various of wikitext - not only wikitext sections - as a part of INT-58 I modified ArticleAsJson to add section attrbiute to all header tags which are actually wikitext sections. More info: https://wikia-inc.atlassian.net/browse/HG-600


hakubo:
LGTM

rogatty:
LGTM

hakubo:
@rogatty hasn't the match started ? :)

PR #787, opened 2015-05-14
(INT-67) Resize textarea when window size is changed

Grunny:
Resize textarea when window size or orientation is changed.

/cc @inez

### DIFF
--- a/front/scripts/main/views/EditView.ts
+++ b/front/scripts/main/views/EditView.ts
@@ -8,7 +8,16 @@ App.EditView = Em.View.extend({
 		Em.run.scheduleOnce('afterRender', this, () => {
 			this.adjustTextareaHeight();
 		});
-		// TODO: Call adjustTextareaHeight after resize event and make sure to unsubscribe from that event when destroying the view
+	},
+
+	willInsertElement: function (): void {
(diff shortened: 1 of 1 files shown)


front/scripts/main/views/EditView.ts
```diff
@@ -8,10 +8,19 @@ App.EditView = Em.View.extend({
...
+			this.adjustTextareaHeight();
+		});
+	},
+
+	willDestroyElement: function(): void {
+		Em.$(window).off('resize.editor');
 	},
 
 	adjustTextareaHeight: function(): void {
 		Em.$('textarea').css('height', Em.$(window).height() - Em.$('.edit-head').outerHeight());
```
hakubo: I know it is not part of the PR
but Em here could be replaced with `this`

`this.$()` - limits search to current view

hakubo:
LGTM

front/scripts/main/views/EditView.ts
Grunny: Hmm, it looks like one of the times `this.adjustTextareaHeight` is called when the init method is executed results in `this.$('textarea')` being undefined (when transitioning from the article page, it seems to be called three times?). I'll look at that later. :)

PR #788, opened 2015-05-14
Category links were not calling to correct context, this makes them c…

kenkouot:
…lickable for now


rogatty:
@kenkouot: we already have release-61 on production and are preparing release-62, you're too slow ;)

See https://github.com/Wikia/mercury/pull/793

PR #793, opened 2015-05-14
HG-658 Fix category links

rogatty:
https://wikia-inc.atlassian.net/browse/HG-658

Category links were not calling to correct context, this makes them clickable for now


hakubo:
LGTM

PR #794, opened 2015-05-14
Remove obsolete trackClick='trackClick'

rogatty:
This is no longer needed and actually breaks footer links.


rogatty:
@hakubo says LGTM but is currently offline :P

rogatty:
@hakubo would say LGTM but he is a little bit shy.

PR #799, opened 2015-05-15
HG-649 Don't add adsContext if it's an empty array

rogatty:
https://wikia-inc.atlassian.net/browse/HG-649


vforge:
LGTM

hakubo:
LGTM

hakubo:
Ads code:
- ads slots to DOM (so we can debug it better) 
- does not load ads

nandy-andy:
LGTM and thanks! :)

PR #803, opened 2015-05-16
SOC-811 Fixing caching issues for i18n on /join page

bkoval:
@rogatty @mklucsarits


rogatty:
LGTM

server/lib/AuthUtils.ts
```diff
@@ -39,3 +39,7 @@ export function getCacheBusterUrl(redirect: string): string {
 
 	return url.format(cacheBustedUrlObj);
 }
+
+export function disableCache (response: Hapi.Response): void {
```
hakubo: There is Caching.ts how about moving it there?

mklucsarits:
I agree with hakubo's comment. Otherwise it looks good to me.

server/lib/AuthUtils.ts
```diff
@@ -1,5 +1,5 @@
 /// <reference path='../../typings/node/node.d.ts' />
-
+/// <reference path='../../typings/hapi/hapi.d.ts' />
```
mklucsarits: I know it's not harming anything, but maybe this line can be removed or this file just reverted.

mklucsarits:
LGTM

PR #804, opened 2015-05-16
Add baseline performance metrics to send to Weppy

kenkouot:


front/scripts/main/app.ts
```diff
@@ -88,6 +88,41 @@ App.initializer({
...
+					})
+
+					events.push({
+						module: 'App',
+						name: 'domInteractive',
+						type: 'timer',
+						value: times.domInteractive - times.domLoading
+					})
+
+					events.forEach((event: any) => {
```
rogatty: `:void`

front/scripts/main/app.ts
```diff
@@ -88,6 +88,41 @@ App.initializer({
 			return;
 		}
 
+		if (window.performance && window.performance.timing) {
+			var times: any = window.performance.timing,
+				events: any[] = [];
+
+			document.onreadystatechange = () => {
```
rogatty: `:void`

front/scripts/main/app.ts
hakubo: events.forEach(M.trackPerf);

front/scripts/main/app.ts
```diff
@@ -88,6 +88,41 @@ App.initializer({
...
+				if (document.readyState === 'complete') {
+					events.push({
```
hakubo: You could save some repetition by creating a factory here:

```
interface TrackEvent {
  module: string;
  name: string;
  type: string;
  value: number;
}
createEvent: function (name: string, value: number): TrackEvent {
return {
  module: 'App',
  name: name,
  type: 'timer',
  value: value
}
}

createEvent(name, value);
```

rogatty:
LGTM

front/scripts/main/app.ts
```diff
@@ -88,6 +88,34 @@ App.initializer({
...
+			function createEvent (name: string, value: number): PerfTrackerParams {
+				return {
+					module: 'App',
+					name: name,
+					type: 'timer',
+					value: value
+				};
+			}
+
+			document.onreadystatechange = () => {
```
hakubo: why not document.addEventListener
This way if someone adds this as well this won't stop working

front/scripts/main/app.ts
```diff
@@ -88,6 +88,34 @@ App.initializer({
...
+				if (document.readyState === 'complete') {
```
hakubo: what is the benefit of using this instead of
$(function)
or
document.addEventListener('load',...

aren't those equivalent?

PR #805, opened 2015-05-16
Views do not get recreated and are long lived

hakubo:
that seems to cause articles to be appended to a previous article


front/templates/main/article.hbs
```diff
@@ -33,13 +33,13 @@
 					{{/each}}
 				{{/collapsible-menu}}
 			{{/if}}
-			{{#view 'article-content' class='mw-content'}}
+			{{#article-content class='mw-content'}}
```
hakubo: @kenkouot 
Actually to be honest I have no idea what is the benefit of having this as  a component/view?
Can't this be just a <div class='mw-content'> ?

kenkouot: Agreed, there might have been a point some time ago, but I vote for simplicity!

rogatty:
@hakubo: how is this one going? Want to remove article-content?

hakubo:
I am going to close this now.
I would go for removing article content component/view

PR #808, opened 2015-05-18
New translations

bkoval:
@mklucsarits @Warkot Could you guys take a look? Let's merge it asap


front/locales/de/auth.json
```diff
@@ -1,5 +1,7 @@
 {
   "common":{
+    "password-label":"Passwort",
```
Warkot:
```
"username-label":"Benutzername",
"password-label":"Passwort",
```
These two are already in lines 13 and 14.

front/locales/de/auth.json
```diff
@@ -8,17 +10,31 @@
...
+    "footer-callout-link":"Anmelden",
+    "register-action":"Registrieren",
+    "page-title":"Mit deiner E-Mail-Adresse registrieren",
+    "terms-of-use-text":"Durch Erstellen eines Kontos stimmst du",
+    "terms-of-use-link":"den Nutzungsbedingungen von Wikia zu",
+    "wikia-newsletter-prompt":"Ich möchte E-Mails zu News und Events von Wikia erhalten"
   }
-}
\ No newline at end of file
+}
```
Warkot: Adding newlines seems a pretty good idea, but Crodwin always removes them in the non-EN files. The same goes with swapping tabs for four spaces. I would just leave it as it is, and forget about it in the future OR modify the `crowdin` script to handle it automatically.

hakubo:
Nice, thanks!

front/locales/es-es/auth.json
```diff
@@ -0,0 +1,40 @@
+{
+  "common":{
+    "password-label":"Contraseña",
```
Warkot: The same as for German — these two labels are already included below in this file. The other languages have the same issue. IMHO, as they are the same, we should decide which namespace to use "common" or "login" for these labels and adjust the Mercury code appropriately.

front/locales/de/auth.json
bkoval: I'll let the i18n-tools people know that we could be adding newlines, I think it's a good practice. Thanks!

front/locales/fi/auth.json
```diff
@@ -0,0 +1,40 @@
+{
```
mklucsarits: Are these files with just the English text (because they weren't yet translated) needed? Will Mercury fall back to English if the file doesn't exist for the language?

front/locales/de/auth.json
Warkot: Ideally we could ask Crowdin guys to handle this.

front/locales/de/auth.json
mklucsarits: It looks like this is the case for most (all?) of the auth files.

Warkot: Indeed, I wrote about it later here: https://github.com/Wikia/mercury/pull/808/files#r30529319 :wink:

front/locales/pl/main.json
```diff
@@ -3,9 +3,7 @@
     "article-empty-label":"Niestety ten artykuł wydaje się być pusty!",
     "article-toc-label":"Spis treści",
     "article-comments-label":"__count__ komentarz",
-    "article-comments-label_plural_0":"__count__ komentarzy",
-    "article-comments-label_plural_2":"__count__ komentarze",
```
Warkot: We could manually leave the paucal forms for Polish. The problem is within the Crowdin tool that it can't handle that (like adding new keys for specific languages) while i18next can. So the translations look like xxx(-y). It's not cool. :(

front/locales/de/auth.json
bkoval: Yeah, I added a ticket of clieaning it up in the code: https://wikia-inc.atlassian.net/browse/SOC-814

front/locales/pl/main.json
bkoval: What's the action point here? Should I exclude these lines from the commit?

front/locales/fi/auth.json
bkoval: Yes, it falls back to english language.

front/locales/pl/main.json
Warkot: I would do so, but be cautious about `"username-anonymous"` and `"wikia-map-view-map"`, they were not translated in the old diff, they are in the new one.

bkoval:
OK, obsolete labels removed, untranslated files removed, polish paucal forms restored :)

hakubo:
LGTM

Warkot:
LGTM :water_buffalo:

front/locales/de/auth.json
```diff
@@ -10,15 +10,27 @@
     "welcome-back":"Schön, dass du wieder hier bist!",
     "register-callout":"Du hast noch kein Benutzerkonto?",
     "register-now":"Registriere dich jetzt",
-    "username-label":"Benutzername",
```
mklucsarits: Looks like DE doesn't have any username-label or password-label after removing these.

mklucsarits:
\o/ LGTM

rogatty:
Indentation is incorrect but whatever ;)

PR #809, opened 2015-05-18
Downgrade gulp-develop-server

rogatty:
`gulp-develop-server` version that we use has a problem with multiple restarts, see https://github.com/narirou/gulp-develop-server/issues/14

This causes frequent crashes of server. Reverting the upgrade seems to fix the issue (haven't yet crashed for me).

/cc @lizlux


lizlux:
LGTM

PR #811, opened 2015-05-18
Use the returned value from wrapResult method

rogatty:
Using `wrapResult` without using the value it returns caused exceptions when there was no `result`. For me it happened when MW returned 503 error.

This is the fix and a naming clean up.


hakubo:
LGTM

PR #812, opened 2015-05-18
SOC-797 Removing encrypted session cookie

bkoval:
@mklucsarits @kenkouot This removes encrypted access token, we do not need to set the encoded cookie with it, setting it in a raw form is enough (mediawiki needs it in it's raw form to create a user object and it sets the unencrypted cookie as well).

If we'd leave this cookie here, Mercury would think that the user is logged in even after logging out through Mediawiki.


server/facets/auth/login.ts
```diff
@@ -138,18 +139,16 @@ export function post (request: Hapi.Request, reply: any): void {
...
+		} else {
+			ttl = defaultTTL;
+		}
+
 		// set unencrypted cookie that can be read by all apps (i.e. MW and Mercury) HG-631
 		reply.state('access_token', response.access_token, {ttl: ttl});
 
-		// set session cookie via hapi-auth-cookie
-		request.auth.session.set({
-			'sid'  : response.access_token
```
mklucsarits: If we are getting rid of this cookie, this implies we're no longer using the hapi-auth-cookie plugin, and the code here should be removed as well https://github.com/Wikia/mercury/blob/dev/server/app.ts#L76

There may be additional related code that can be removed. Check with Liz or Ken.

PR #813, opened 2015-05-18
i18n-update | bulked translations downloaded from Crowdin

Warkot:
I downloaded new translations from Crowdin and they concern more then one feature, therefore this PR, which incorporates them all.


rogatty:
Where do we use `wikia-top-contributors-label`? IIRC we currently use only the `article-top-contributors-label` and if you remove it then all non-english wikis will fallback to english. Are you sure you want to do that?

Warkot:
Yes, I'm sure. It's incorrect for all languages.

Warkot:
We have to invalidate the translations by removing the key `article-top-contributors-label`. If it's not removed as it was before, anyone who upload a Crowdin config file with `-t` will mark the faulty translations are OK ones.

rogatty:
Is there somebody who made the decision that english fallback is better than incorrect label? Yeah, I've seen https://wikia-inc.atlassian.net/browse/CONCF-300 but I don't see anyone saying that.

Warkot:
In my humble opinion, it's better. I found the root cause why the thing wasn't able to update, fixed it, but we don't need to merge it right now.

rogatty:
Ok, so that's you who made the decision, don't be shy ;)
LGTM but I'd confirm with someone else too.

Also I think that we should remove `wikia-top-contributors-label` at all. It's not being used anywhere.

Warkot:
It would be a pity to throw out so many good translations. :cry: But on the other hand, it's possible to pull them back from git/github histrory.

hakubo:
Have you already uploaded the change to crowdin?
If yes lets wait a day or two for major languages at least

also please create a ticket for thing like this also

Warkot:
I uploaded the changes to Crowdin.
I reopened and commented on [CONCF-300](https://wikia-inc.atlassian.net/browse/CONCF-300).
I created [LOC-66](https://wikia-inc.atlassian.net/browse/LOC-66).

kvas-damian:
LGTM

PR #814, opened 2015-05-18
CONCF-549

vforge:
/cc @igor @hakubo


front/scripts/main/mixins/ThirdsClickMixin.ts
```diff
@@ -0,0 +1,60 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.ThirdsClickMixin = Em.Mixin.create({
+	leftClickHandler: undefined,
+	rightClickHandler: undefined,
+	centerClickHandler: undefined,
+
+	viewportSize: Em.computed(function () {
```
rogatty: Add a returning type.

front/scripts/main/mixins/ThirdsClickMixin.ts
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * @desc Checks on which area on the screen an event took place
+	 * and use proper handler
+	 *
+	 * @param {Touch} event
+	 */
+	thirdsClick: function (event: Touch): void {
+		var screenArea = this.getScreenArea(event);
+
+		if (screenArea === this.screenAreas.right && this.rightClickHandler) {
```
rogatty: Should we check if `rightClickHandler` is a function?

front/scripts/main/components/FeaturedContentComponent.ts
```diff
@@ -28,6 +29,16 @@ App.FeaturedContentComponent = Em.Component.extend(App.TrackClickMixin, {
 		return this.getWithDefault('model', []).length - 1;
 	}),
 
+	rightClickHandler: function(): void {
+		this.nextItem();
```
rogatty: I'm not sure if we should return false here to prevent event from bubbling up and loading a link. Does it work properly with current implementation?

front/scripts/main/mixins/ThirdsClickMixin.ts
```diff
@@ -0,0 +1,60 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.ThirdsClickMixin = Em.Mixin.create({
+	leftClickHandler: undefined,
```
hakubo: this could be `Em.K`
as per: 
http://code418.com/blog/2012/03/08/useful-emberjs-functions/

front/scripts/main/components/FeaturedContentComponent.ts
vforge: Yes, it's working fine.

front/scripts/main/mixins/ThirdsClickMixin.ts
```diff
@@ -0,0 +1,60 @@
...
+	},
+
+	/**
+	 * @desc Checks on which area on the screen an event took place
+	 * and use proper handler
+	 *
+	 * @param {Touch} event
+	 */
+	thirdsClick: function (event: Touch): void {
+		var screenArea = this.getScreenArea(event);
```
hakubo: since we are not using getScreenArea anywhere
how about simplifying to
`this.callClickHandler(event)`
that would already call appropriate tap handler

this way we don't need screenAreas object
and callClickHandler
could just call 
this.rightClickHandler(event)

and since by default it can be
Em.K
you would not have to check if it is a function or not

front/scripts/main/components/FeaturedContentComponent.ts
rogatty: I've just checked and clicking on the edge changes current item AND opens the link. Simply adding `return false;` doesn't help.

front/scripts/main/mixins/ThirdsClickMixin.ts
```diff
@@ -0,0 +1,41 @@
...
+	}),
+
+	/**
+	 * @desc Checks on which area on the screen an event took place
+	 * and calls proper handler
+	 *
+	 * @param {CancellableClickEvent} event
+	 * @param {boolean} allowDefault
+	 */
+	callClickHandler: function (event: CancellableClickEvent, allowDefault: boolean = true): void {
```
rogatty: Why revert the logic? How about `preventDefault: boolean = false`?

front/scripts/main/mixins/ThirdsClickMixin.ts
```diff
@@ -0,0 +1,41 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+interface CancellableClickEvent extends MouseEvent, Touch {
```
rogatty: `PreventableClickEvent`?

test/specs/front/scripts/main/views/MediaLightboxView.js
```diff
@@ -45,31 +45,3 @@ test('if isCurrentMediaType method recognizes media', function () {
 	isMediaTypeImage = mediaLightboxView.isCurrentMediaType('image');
 	equal(isMediaTypeImage, true);
 });
-
-test('if getScreenArea correctly returns screen areas', function () {
```
rogatty: This logic is still there, just in another form. Let's remove it now but remember about rewriting it for phase 3.

vforge: :+1:

rogatty:
LGTM, tests are passing locally.

PR #816, opened 2015-05-18
Add check for call-signature to tslint and fix existing signatures

rogatty:
I'm tired of writing comments like "Add :void". This will enforce correct typing with tslint.
See https://www.npmjs.com/package/tslint#supported-rules

/cc @hakubo @kenkouot


hakubo:
very very nice,
Thanks
LGTM

rogatty:
@kenkouot do you agree that we should enable this check?

hakubo:
@lizlux do you agree that we should enable this check?

lizlux:
LGTM. If we're going to have that rule it's great to be able to enforce it with TSLint. Thanks @rogatty!

kenkouot:
Yeah this is good!

rogatty:
This rule doesn't actually fail the tests, just displays warning. Still, better than nothing!

PR #817, opened 2015-05-19
(INT-69) Add simple success notification on publish

Grunny:
This adds a simple alert notification on a successful publish using the
Foundation alert styling.

We did find some libraries like https://github.com/aexmachina/ember-notify but this is
a simple solution that avoids the dependency issues in those for now. I'll open a ticket
for the Mercury team to look at the best option for alert notifications when use cases
demand it.

/cc @inez @lizlux @kenkouot


front/scripts/main/components/AlertNotificationsComponent.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+interface AlertNotification {
+	type: string;
+	message: string;
+}
+
+App.AlertNotificationsComponent = Em.Component.extend({
```
hakubo: Do you think you could add a unit test to this?

Grunny: Added a unit test for closing notifications on the component in https://github.com/Wikia/mercury/commit/2029b1b9b :)

hakubo: :bow: thanks, how does it feel :)
I mean testing in Ember and qunit here?

kenkouot:
Was there any design/product work for this?

front/templates/main/application.hbs
```diff
@@ -19,6 +19,7 @@
 			{{outlet}}
 		</section>
 	</div>
+	{{alert-notifications alerts=view.alertNotifications}}
```
inez: Perhaps alert-notifications should be handled similarly to loading-spinner - (if present) then visible always - in fullPage mode and not in fullPage mode. Practical use case: displaying saving error on edit page (which is a full page).

inez:
Currently I noticed that notifications are not removed even if transition to different route happened. So for instance in this user flow: article -> edit -> save -> article -> edit -> save -> article - I will be seeing two notifications (one for each edit) - perhaps it's not good user experience.
I think that notifications should automatically disappear after certain amount of time, and should not persist between transitions.

inez:
Given the fact that "alert" has certain meaning in JavaScript I suggest renaming this component to maybe SimpleNotifications or ModalNotifications. (I would suggest just Notifications but that would not comply with components naming rules.)

inez:
Given that this is MVP version of notifications for MVP version of editing I will assume that everyone is ok with merging it as it is now (and handle provided feedback later on). I will proceed with merging later today - if it's not ok please speak up.

front/styles/main/component/_alert-notifications.scss
```diff
@@ -0,0 +1,6 @@
+.alert-notifications {
+	bottom: 0;
+	position: fixed;
+	width: 100%;
+	z-index: 10000;
```
lizlux: Let's define this z-index here https://github.com/Wikia/mercury/blob/dev/front/styles/main/_wikia-variables.scss#L6 in hopes of keeping z-index from getting out of control.

front/templates/main/application.hbs
Grunny: Yep, I've done that in my branch for INT-70 anyway, so it'll be in that PR.

Grunny: Actually, done in this PR. :)

front/styles/main/component/_alert-notifications.scss
Grunny: Done in https://github.com/Wikia/mercury/commit/846f4f5e34. :)

Grunny:
@inez I moved individual alert notifications to their own component and have set a 10 second timeout as a quick fix to make them clear on their own. We can tweak the timeout or even make them configurable on a per notification basis in the future. I thought this might do for now and we can work on clearing on transitions later.

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -0,0 +1,35 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.AlertNotificationComponent = Em.Component.extend({
+	classNames: ['alert-notification', 'alert-box'],
+	tagName: ['div'],
+
+	classNameBindings: ['alert.type'],
+
+	alerts: null,
```
inez: So it seems like Notification knows about Notifications - and perhaps it shouldn't. Maybe this should be handled within events?

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -0,0 +1,35 @@
...
+	alert: null,
+
+	notificationExpiry: 10000,
+	timeout: null,
+
+	didInsertElement: function (): void {
+		this.set('timeout', setTimeout(() => {
```
rogatty: http://emberjs.com/api/classes/Ember.run.html#method_later
>You should use this method whenever you need to run some action after a period of time instead of using setTimeout(). This method will ensure that items that expire during the same script execution cycle all execute together, which is often more efficient than using a real setTimeout.

Grunny: Ah, good to know, thanks! :)

PR #818, opened 2015-05-19
(INT-71) Add i18n for edit page heading and publish button

Grunny:
/cc @inez

### DIFF
--- a/front/templates/main/edit.hbs
+++ b/front/templates/main/edit.hbs
@@ -2,10 +2,11 @@
 
 	{{#link-to 'article' model.title}}{{svg 'arrow-left' role='img' class='icon arrow-left'}}{{/link-to}}
 
-	<button {{action 'publish'}} class="large" style="float: right;">Publish</button>
+	<button {{action 'publish'}} class="large" style="float: right;">{{i18n 'app.edit-page-publish'}}</button>
 
 	<h5>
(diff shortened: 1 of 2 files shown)


front/templates/main/edit.hbs
```diff
@@ -2,10 +2,11 @@
 
 	{{#link-to 'article' model.title}}{{svg 'arrow-left' role='img' class='icon arrow-left'}}{{/link-to}}
 
-	<button {{action 'publish'}} class="large" style="float: right;">Publish</button>
+	<button {{action 'publish'}} class="large" style="float: right;">{{i18n 'app.edit-page-publish'}}</button>
 
 	<h5>
-		Editing: <span style="font-weight: normal;">{{model.title}}</span>
+		{{! TODO: Ideally the title would be passed to the message so translators can choose the order (see HG-663) }}
+		{{i18n 'app.edit-page-heading'}} <span style="font-weight: normal;">{{model.title}}</span>
```
hakubo: can we resist the temptation of inline styles?

inez: I will create a follow up ticket for this - not addressing here since this is not PR that introduced it.

hakubo: kk, thanks

PR #820, opened 2015-05-19
Include timezone in changelog

gabrys:


tasks/changelog-update.sh
```diff
@@ -24,7 +24,7 @@ fi
 /bin/bash tasks/changelog-view.sh 1> new
 
 WORDC=$(wc -w new | tr -d '[:alpha:][:blank:][:punct:]/-')
-DATE=$(date -u +"%Y-%m-%d %H:%M")
+DATE=$(date -u +"%Y-%m-%d %H:%M %Z")
 
 if [ $WORDC -gt 0 ]; then
 	echo "## "$RELEASE" ("$DATE" UTC)" > temp
```
hakubo: so should we remove this hardcoded UTC here?

rogatty: From http://unixhelp.ed.ac.uk/CGI/man-cgi?date:
```
  -u, --utc, --universal
	      print or set Coordinated Universal Time
```
Timezone is always UTC so it doesn't matter if it's hardcoded or added with `%Z`

rogatty:
@gabrys do you agree that this PR is not needed?

gabrys:
I thought I already closed it. Sorry about it

PR #822, opened 2015-05-19
DAT-2812 style image caption inside portable infoboxes

RafLeszczynski:


hakubo:
LGTM,

It'd be nice to think maybe how we can do media-component easier to reuse?
Or is it fine as it is now?

RafLeszczynski:
@hakubo let's talk about it when you're be back in Poznan

PR #826, opened 2015-05-19
Feature/hg 656 tracking

kenkouot:
Abstracts UA dimensions from the class declaration and moves it to a developer controlled initialization. The only prerequisite is tracking dimensions must be set before any invocations of the tracking function.

Adds a new static function `UniversalAnalytics#setDimensions` with a signature of `(dimensions: typeof UniversalAnalytics.dimensions, overwrite?: boolean): boolean`. Properly invoke this function before using `M.track` anywhere or you will have a bad time.

@bkoval

### DIFF
--- a/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
+++ b/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
@@ -43,8 +43,8 @@ module Mercury.Modules.Trackers {
 		/**
 		 * @static
 		 * @description Synchronously sets the UA dimensional context
-		 * @param dimensions {Array} array of dimensions to set, may be strings or functions
-		 * @param overwrite {boolean} set to true to overwrite all preexisting dimensions and unset ones not declared
+		 * @param {Array} dimensions  array of dimensions to set, may be strings or functions
+		 * @param {boolean} overwrite  set to true to overwrite all preexisting dimensions and unset ones not declared
 		 * @returns {boolean} true if dimensions were successfully set
 		 */
 		public static setDimensions (dimensions: typeof UniversalAnalytics.dimensions, overwrite?: boolean): boolean {
@@ -63,6 +63,7 @@ module Mercury.Modules.Trackers {
 
 		/**
 		 * @private
+		 * @param {number} index of dimension
 		 * @description Retrieves string value or invokes function for value
 		 * @returns {string}
 		 */
@@ -163,7 +164,7 @@ module Mercury.Modules.Trackers {
 			var pageType = this.getDimension(8);
 
 			if (!pageType) {
-				throw new Error('mission page type dimension (#8)');
+				throw new Error('missing page type dimension (#8)');
 			}
 
 			ga('set', 'dimension8', pageType, 3);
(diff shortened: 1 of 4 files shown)


mklucsarits:
Hey Ken, I don't quite understand how these changes relate to HG-656. The needs for HG-656 are abstracting out the GA tracking code to make it usable on the new login (or any non-EmberJS) pages. Is this for a different ticket, or a first step, or something else?

kenkouot:
Talked to @mklucsarits about this in Slack, but might as well clarify here. The only thing blocking the use of the code now, eg the coupling piece was the **setting of dimensions**. This removes that responsibility out of the class and into the invocation context, freeing it up to be used wherever.

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -10,37 +10,76 @@ interface TrackerOptions {
...
+			this.initAccount(this.accountPrimary, domain);
+
+			this.initAccount(this.accountAds, domain);
+		}
+
+
+		/**
+		 * @static
+		 * @description Synchronously sets the UA dimensional context
+		 * @param dimensions {Array} array of dimensions to set, may be strings or functions
```
mklucsarits: FYI these ```@param```s are malformed. The type goes before the parameter name.

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -142,13 +160,15 @@ module Mercury.Modules.Trackers {
 		 * Tracks the current page view
 		 */
 		trackPageView (): void {
-			var mainPageTitle = Mercury.wiki.mainPageTitle,
-				isMainPage = window.location.pathname.split('/').indexOf(mainPageTitle);
+			var pageType = this.getDimension(8);
 
-			ga('set', 'dimension8', isMainPage >= 0 ? 'home' : 'article', 3);
+			if (!pageType) {
+				throw new Error('mission page type dimension (#8)');
```
mklucsarits: Typo here "mission" --> "missing"

mklucsarits:
LGTM

bkoval:
LGTM

rogatty:
LGTM

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference path="../../../../../typings/google.analytics/ga.d.ts" />
-/// <reference path="../../modules/Ads.ts" />
 /// <reference path="../../../baseline/mercury.ts" />
+/// <reference path="../../../baseline/mercury.d.ts" />
```
mklucsarits: Why is this needed? This reference causes a problem when trying to build this module with the Hapi back-end:
```
/mercury/front/scripts/baseline/mercury.ts(5,13): 2403 Subsequent variable declarations must have the same type.  Variable 'require' must be of type '{ (id: string): any; resolve(id: string): string; cache: any; extensions: any; main: any; }', but here has type '(deps: string[], func: Function) => void'.
```
The problem is Mercury's version of ```require``` is incompatible with Node's ```require```.

mklucsarits: Can you explain why this reference is needed?

PR #827, opened 2015-05-20
Fixing forgot-password link

bkoval:
Turns out Mercury supports only wikias with /wiki/ in the article path. This should be safe this way, tested on sandbox.

@mklucsarits

### DIFF
--- a/server/lib/AuthUtils.ts
+++ b/server/lib/AuthUtils.ts
@@ -7,8 +7,8 @@
 import url = require('url');
 import querystring = require('querystring');
 
-var wikiaSignupPathname: string = '/Special:UserSignup',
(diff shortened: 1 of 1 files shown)


mklucsarits:
LGTM

server/lib/AuthUtils.ts
```diff
@@ -7,8 +7,8 @@
 import url = require('url');
 import querystring = require('querystring');
 
-var wikiaSignupPathname: string = '/Special:UserSignup',
-	wikiaLoginPathname: string = '/Special:UserLogin',
+var wikiaSignupPathname: string = 'wiki/Special:UserSignup',
```
hakubo: don't you need leading /
as you had before?

PR #828, opened 2015-05-20
CONCF-552 Categories as Curated Content sections

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-552

### DIFF
--- a/front/scripts/main/components/CuratedContentComponent.ts
+++ b/front/scripts/main/components/CuratedContentComponent.ts
@@ -16,10 +16,14 @@ App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.T
 		return this.get('sectionsStack.lastObject');
 	}),
 
-	didInsertElement: function(): void {
+	didInsertElement: function (): void {
 		this.set('model', App.CuratedContentModel.create());
 		this.set('spinnerDelay', 50);
 		this.createTopLevelSection();
+	},
+
+	willDestroyElement: function (): void {
+		this.sectionsStack.clear();
 	},
 
 	actions: {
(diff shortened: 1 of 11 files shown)


front/templates/main/components/curated-content-item.hbs
```diff
@@ -1,18 +1,21 @@
...
 	<figcaption class='item-figcaption'>
 		<div class='item-caption-row'>
-		<span class='item-caption label'>{{label}}</span>
-		<span class='item-caption grid-icon'>
-			<div class='grid-icon-container'>
-				{{svg 'category-thumb-icon' role='img' class='category-icon' viewBox='0 0 17 17'}}
+			<span class='item-caption label'>{{model.label}}</span>
+			<div class='item-caption grid-icon'>
+				<div class='grid-icon-container'>
+					{{#if isArticle}}
```
hakubo: how about adding a property
'iconName' or something like that
so we could
`{{svg (iconName) role='img' viewBox='0 0 17 17'}}`
would that work?

front/scripts/main/components/CuratedContentComponent.ts
```diff
@@ -1,38 +1,100 @@
 /// <reference path="../app.ts" />
 ///<reference path="../mixins/LoadingSpinnerMixin.ts"/>
 ///<reference path="../mixins/TrackClickMixin.ts"/>
+///<reference path="../models/CuratedContentModel.ts"/>
 'use strict';
 
 App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.TrackClickMixin, {
```
hakubo: Could you add some unit tests?
That would be :100: points

hakubo:
Looking good! Thanks

front/templates/main/components/curated-content-item.hbs
rogatty: I've tried, it doesn't work. Surely it's possible but I'd have to modify `svgHelper` and I don't want to do this now - deadlines and all ;)

front/scripts/main/components/CuratedContentComponent.ts
rogatty: Good idea but how about we do this after our little experiment succeeds?

front/templates/main/components/curated-content-item.hbs
hakubo: mkay

front/scripts/main/components/CuratedContentComponent.ts
hakubo: We are already working on stuff that are nice to haves, 
I think we do have time for some unit tests :)

front/templates/main/components/curated-content-item.hbs
vforge: @rogatty can you create ticket for modifying `svgHelper` then?

front/scripts/main/components/CuratedContentComponent.ts
vforge: @hakubo what should we postpone and wait for unit tests? can we QA in the meantime?

hakubo: I'd wait for unit test, they might lead to some refactors and then we would have to test it again

PR #829, opened 2015-05-20
(INT-68) Activate publish button after content change

Grunny:
Simple event on text change for the MVP.

/cc @inez

### DIFF
--- a/front/scripts/main/models/EditModel.ts
+++ b/front/scripts/main/models/EditModel.ts
@@ -10,7 +10,7 @@ App.EditModel = Em.Object.extend({
 	sectionIndex: null,
 	isNotDirty: Em.computed('content', 'originalContent', function (): boolean {
 		return this.get('content') === this.get('originalContent');
-	});
+	})
 });
 
 App.EditModel.reopenClass({
--- a/front/scripts/main/models/EditModel.ts
+++ b/front/scripts/main/models/EditModel.ts
@@ -4,9 +4,13 @@
 
 App.EditModel = Em.Object.extend({
 	content: null,
+	originalContent: null,
 	timestamp: null,
 	title: null,
-	sectionIndex: null
+	sectionIndex: null,
+	isNotDirty: function () {
+		return this.get('content') === this.get('originalContent');
+	}.property('content', 'originalContent')
 });
 
 App.EditModel.reopenClass({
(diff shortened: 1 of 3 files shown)


front/scripts/main/views/EditView.ts
```diff
@@ -16,6 +16,12 @@ App.EditView = Em.View.extend({
 		});
 	},
 
+	didInsertElement: function (): void {
+		this.$('textarea').on('input propertychange', () => {
```
inez: Use "one" instead of "on"

front/scripts/main/views/EditView.ts
```diff
@@ -16,6 +16,12 @@ App.EditView = Em.View.extend({
...
+			this.$('.edit-head').find('button').prop('disabled', false);
```
hakubo: could this be a a bound property? so you don't have to update it manually?

front/scripts/main/views/EditView.ts
hakubo: isin't value bound?
so you could listen to change on model.content

instead of manually listen to changes on textarea

inez:
After brief discussion with @kenkouot, @rogatty and @hakubo I decided to change it to more "Ember way" (rather than "jQuery way").

front/scripts/main/models/EditModel.ts
```diff
@@ -4,9 +4,13 @@
 
 App.EditModel = Em.Object.extend({
 	content: null,
+	originalContent: null,
 	timestamp: null,
 	title: null,
-	sectionIndex: null
+	sectionIndex: null,
+	isNotDirty: function () {
```
rogatty: Use `Em.computed('content', 'originalContent', function (): string { ... })`.
See https://github.com/Wikia/mercury/pull/709

front/scripts/main/models/EditModel.ts
```diff
@@ -88,8 +87,14 @@ App.EditModel.reopenClass({
 					return page.revisions[0];
 				} )[0];
 
-				model.set('content', revision['*']);
-				model.set('timestamp', revision.timestamp);
+
+				var model = App.EditModel.create({
```
rogatty: One `var` per scope.

PR #660, opened 2015-04-02
Remove 100% GA account

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-410

PR #832, opened 2015-05-21
Add check for GA special wiki

kenkouot:
Adds a check for `isGASpecialWiki`. This can be defined in two places, `Mercury.wiki.isGASpecialWiki` or `M.prop('isGASpecialWiki')`. The preferred place to declare values as these is in `M.prop`, but the Mercury app currently still has legacy coupling with the globals, `Mercury.wiki`.

### DIFF
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -63,6 +63,11 @@ var localSettings: LocalSettings = {
 				prefix: 'ads',
 				id: 'UA-32129071-1',
 				sampleRate: 100
+			},
+			special: {
+				prefix: 'special',
+				id: 'UA-32132943-1',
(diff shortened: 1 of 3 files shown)


rogatty:
Should we add some extra handling that was removed in #660?

- `track` - https://github.com/Wikia/mercury/pull/660/files#diff-a76be64ed2e86aa0e73ed9c466b8fe1eL115
- `trackPageView` - https://github.com/Wikia/mercury/pull/660/files#diff-a76be64ed2e86aa0e73ed9c466b8fe1eL139

If not then it LGTM. Would be nice if you adjust unit tests but IMO it's not a blocker for releasing this fix.

PR #834, opened 2015-05-21
(INT-70) Handle errors on publishing an edit

Grunny:
Simple error handling on failed attempts at publishing. We handle a couple of common
errors that would not be resolvable by the anon editing with more specific messages
and otherwise fallback to a generic error message.

/cc @inez

### DIFF
--- a/front/locales/en/main.json
+++ b/front/locales/en/main.json
@@ -63,6 +63,12 @@
 		"edit-page-heading": "Editing:",
 		"edit-page-publish": "Publish",
 		"edit-success": "Successfully edited page __pageTitle__!",
-		"edit-load-error": "Something went wrong."
+		"edit-load-error": "Something went wrong.",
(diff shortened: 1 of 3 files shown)


front/locales/en/main.json
```diff
@@ -63,6 +63,12 @@
 		"edit-page-heading": "Editing:",
 		"edit-page-publish": "Publish",
 		"edit-success": "Successfully edited page __pageTitle__!",
-		"edit-load-error": "Something went wrong."
+		"edit-load-error": "Sorry, it looks like something went wrong. Please try again.",
+		"edit-publish-error": "Sorry, it looks like something went wrong. Please try again.",
```
Grunny: We can probably just unify these into a `edit-error-try-again` but was going to double check if there should be any wording tweaks on the publish one.

PR #835, opened 2015-05-21
CONCF-582 | Tell Optimizely to use UA Dimensions

vforge:
https://wikia-inc.atlassian.net/browse/CONCF-582

/cc @rogatty @Warkot @hakubo

### DIFF
--- a/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
+++ b/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
@@ -1,7 +1,6 @@
 /// <reference path="../../../../../typings/google.analytics/ga.d.ts" />
 /// <reference path="../../../baseline/mercury.ts" />
 /// <reference path="../../../baseline/mercury.d.ts" />
-/// <reference path="../../utils/variantTesting.ts" />
 
 interface TrackerOptions {
 	name: string;
@@ -44,8 +43,6 @@ module Mercury.Modules.Trackers {
 			if (isSpecialWiki) {
 				this.initAccount(this.accountSpecial, domain);
 			}
-
-			Mercury.Utils.VariantTesting.initUA();
 		}
 
 
@@ -139,7 +136,7 @@ module Mercury.Modules.Trackers {
 					eventCategory: category,
 					eventAction: action,
 					eventLabel: label,
-					eventValue:value,
+					eventValue: value,
 					nonInteraction: nonInteractive
 				}
 			);
@@ -162,7 +159,7 @@ module Mercury.Modules.Trackers {
 					eventCategory: category,
 					eventAction: action,
 					eventLabel: label,
-					eventValue:value,
+					eventValue: value,
 					nonInteraction: nonInteractive
 				}
 			);
@@ -183,7 +180,6 @@ module Mercury.Modules.Trackers {
 				ga(`${prefix}set`, 'dimension8', pageType, 3);
 				ga(`${prefix}send`, 'pageview');
 			});
-
 		}
 	}
 }
--- a/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
+++ b/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
@@ -1,6 +1,7 @@
 /// <reference path="../../../../../typings/google.analytics/ga.d.ts" />
 /// <reference path="../../../baseline/mercury.ts" />
 /// <reference path="../../../baseline/mercury.d.ts" />
+/// <reference path="../../utils/variantTesting.ts" />
 
 interface TrackerOptions {
 	name: string;
@@ -38,11 +39,7 @@ module Mercury.Modules.Trackers {
 
 			this.initAccount(this.accountAds, domain);
 
-			// Optimizely Universal Analytics Integration Code
-			// as per Optimizely suggestion in this article:
-			// https://help.optimizely.com/hc/en-us/articles/200039995-Integrating-Optimizely-with-Google-Universal-Analytics?flash_digest=e09d4e784c1ef44e4f0048f405f9c92396bd65a2
-			window.optimizely = window.optimizely || [];
-			window.optimizely.push('activateUniversalAnalytics');
+			Mercury.Utils.VariantTesting.initUA();
 		}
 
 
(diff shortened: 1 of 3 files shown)


front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -37,6 +37,12 @@ module Mercury.Modules.Trackers {
 			this.initAccount(this.accountPrimary, domain);
 
 			this.initAccount(this.accountAds, domain);
+
```
mklucsarits: We should try to keep all the Optimizely code de-coupled and in one place. Currently it lives in https://github.com/Wikia/mercury/blob/dev/front/scripts/mercury/utils/variantTesting.ts. Can this be moved there?

vforge: Yeah, thanks for noticing!

rogatty:
@vforge: this is Optimizely related so let's fix tests too as they're failing for everybody now.
```
PhantomJS 1.9.8 (Linux) Article Comments Model url creates valid url to a resource FAILED[39m
	afterEach failed on url creates valid url to a resource: 'undefined' is not an object (evaluating 'optimizely.variationMap[2870342045]')
	TypeError: 'undefined' is not an object (evaluating 'optimizely.variationMap[2870342045]')
```

rogatty:
LGTM, just resolve the conflicts.

mklucsarits:
Thanks, V! LGTM also.

rogatty:
LGTM

front/scripts/mercury/utils/variantTesting.ts
```diff
@@ -36,5 +36,37 @@ module Mercury.Utils.VariantTesting {
...
+	 * @returns {[]}
+	 */
+	export function integrateOptimizelyWithUA (dimensions: any[]): any[] {
+		var optimizely = window.optimizely,
+			experimentId: string,
+			dimension: number,
+			experimentName: string,
+			variationName: string;
+
+		if (optimizely) {
```
rogatty: If we're already so cautious (look at all those checks!) let's add `optimizely.allExperiments` here too.

front/scripts/mercury/utils/variantTesting.ts
```diff
@@ -36,5 +36,37 @@ module Mercury.Utils.VariantTesting {
...
+				if (
+					optimizely.allExperiments.hasOwnProperty(experimentId) &&
+					typeof optimizely.allExperiments[experimentId].universal_analytics === 'object' &&
+					optimizely.allExperiments[experimentId].enabled
+				) {
+					dimension = optimizely.allExperiments[experimentId].universal_analytics.slot;
+					experimentName = optimizely.allExperiments[experimentId].name;
+					variationName = optimizely.variationNamesMap[experimentId];
+
+					dimensions[dimension] = `Optimizely ${experimentName} (${experimentId}): ${variationName}`;
```
rogatty: I'd like this format to be confirmed by Chris before merging.

hakubo: Does it really matter?
As long as we can distinguish them that should be fine I think.

front/scripts/mercury/utils/variantTesting.ts
Warkot: Nah, it won't throw an error. :sunglasses: Take a look:
```
for (i in foo) {
	console.log(i);
}
```
throws `Uncaught ReferenceError: foo is not defined (…)`.

While 
```
for (i in window.foo) {…}
```
or even
```
for (i in undefined) {…}
```
is fine. It just doesn't run.


It is different for the check in line 57:
```
typeof optimizely.allExperiments[experimentId].universal_analytics === 'object'
```
because later we read `slot` property of this object
```
dimension = optimizely.allExperiments[experimentId].universal_analytics.slot;
```
So, while this is fine
```
console.log(window.foo);
```
and just returns `undefined`, this is not
```
console.log(window.foo.bar);
```
and throws `Uncaught TypeError: Cannot read property 'bar' of undefined`.

front/scripts/mercury/utils/variantTesting.ts
Warkot: It would be nice to agree on one format that satisfies everyone, so we can keep it consistent through time, IMO. And Chris is the one that will be using UA most extensively among us.

vforge:
@Warkot Overall good job with that implementation!

front/scripts/mercury/utils/variantTesting.ts
rogatty: There is also matter of optimization. We'll send this string with every UA tracking request, right?

front/scripts/mercury/utils/variantTesting.ts
rogatty: M-m-m-m-magic! :star2: 
Couldn't find any docs on this behaviour but I believe you ;)

hakubo:
:shipit: LGTM,

We could thogh send an email to them asking how can this be done on their side

front/scripts/mercury/utils/variantTesting.ts
Warkot: Indeed.

front/scripts/mercury/utils/variantTesting.ts
Warkot: Yeah, magic :stars:, sparkles :sparkler: and some typing in console :construction:. :wink:

Warkot:
Thanks, @v! :smile_cat:
@hakubo we can ask them, but I believe that our page is too complex for their solution according to documentation.

vforge:
@hakubo Optimizely's explanation is given in original ticket. Our page is too complex I think.

front/scripts/mercury/utils/variantTesting.ts
vforge: As the same format is default for Optimizely around the web (and quite possibly will be the same for Oasis/Desktop implementation) I would strongly recommend to leave it as-is.

Warkot:
Exactly. We may say 'our page' but this is still a set of pages (using different trackers) under a common domain.

rogatty:
LGTM

vforge:
Tested, merged.

PR #839, opened 2015-05-21
SOC-811 Fixing caching issues for i18n on /join page

bkoval:
https://github.com/Wikia/mercury/pull/803 is an original PR


server/facets/auth/join.ts
```diff
@@ -33,13 +35,16 @@ function get (request: Hapi.Request, reply: any): void {
...
-	return reply.view(
+	response = reply.view(
 		'auth-landing-page',
 		context,
 		{
 			layout: 'auth'
 		}
 	);
+
+	caching.disableCache(response);
```
lizlux: do we need to add this on `/login` and `/signup` too? If so, is it possible to do it at a higher level?

bkoval: OK, let's address it in the feature branch :)

PR #844, opened 2015-05-22
(INT-82) Don't display edit buttons on main pages or non-Japanese wikias

Grunny:
For the MVP, we don't want the edit buttons to show on main pages, and to
only display them on Japanese-language wikias.

/cc @inez

### DIFF
--- a/front/scripts/main/models/ArticleModel.ts
+++ b/front/scripts/main/models/ArticleModel.ts
@@ -194,6 +194,8 @@ App.ArticleModel.reopenClass({
 				data.topContributors = source.topContributors;
 			}
 
+			data.isMainPage = (model.title === Mercury.wiki.mainPageTitle);
+
 			if (source.mainPageData && M.prop('optimizelyCuratedMainPage')) {
 				data.mainPageData = source.mainPageData;
 				data.isCuratedMainPage = true;
--- a/front/scripts/main/models/ArticleModel.ts
+++ b/front/scripts/main/models/ArticleModel.ts
@@ -194,7 +194,9 @@ App.ArticleModel.reopenClass({
 				data.topContributors = source.topContributors;
 			}
 
-			data.isMainPage = (model.title === Mercury.wiki.mainPageTitle);
+			if (source.isMainPage) {
+				data.isMainPage = source.isMainPage;
+			}
 
 			if (source.mainPageData && M.prop('optimizelyCuratedMainPage')) {
 				data.mainPageData = source.mainPageData;
--- a/front/scripts/main/models/ArticleModel.ts
+++ b/front/scripts/main/models/ArticleModel.ts
@@ -28,6 +28,7 @@ interface Response {
 			users: any;
 			categories: any[];
 		};
+		isMainPage: boolean;
 		mainPageData: any[];
 		relatedPages: any[];
 		topContributors: any[];
@@ -41,6 +42,7 @@ App.ArticleModel = Em.Object.extend({
 	categories: [],
 	cleanTitle: null,
 	comments: 0,
+	isMainPage: false,
 	mainPageData: null,
 	media: [],
 	mediaUsers: [],
--- a/front/scripts/main/views/ArticleView.ts
+++ b/front/scripts/main/views/ArticleView.ts
@@ -16,12 +16,11 @@ interface HTMLElement {
 	scrollIntoViewIfNeeded: () => void
 }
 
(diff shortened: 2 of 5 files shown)


front/scripts/main/models/ArticleModel.ts
```diff
@@ -194,6 +194,8 @@ App.ArticleModel.reopenClass({
 				data.topContributors = source.topContributors;
 			}
 
+			data.isMainPage = (model.title === Mercury.wiki.mainPageTitle);
```
rogatty: If we want to add this property to the model maybe we should do this on the API side?

front/scripts/main/views/ArticleView.ts
```diff
@@ -19,6 +19,11 @@ interface HTMLElement {
 App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, {
 	classNames: ['article-wrapper'],
 
+	editButtonsVisible: Em.computed('controller.model.isMainPage', function (): boolean {
+		var contentLanguage: string = Em.get(Mercury, 'wiki.language.content');
+		return !this.get('controller.model.isMainPage') && contentLanguage === 'ja';
```
rogatty: How about using `App.LanguagesMixin` and its method `isJapanese()`?

inez: That's a good idea.
A little context about "isJapanese": it is very specialized method (as it checks only for that one particular language) and one may think why there are no methods for other languages - this is because Japanese language are sort of out playground for now and for some foreseeable future. There will be more and more (hopefully) features that we will be releasing for Japanese first and later (if feature performs well and it makes sense) release to other communities.

front/scripts/main/models/ArticleModel.ts
inez: Why would we do that? ArticleModel is not 1:1 representation of the API response anyways.

front/scripts/main/views/ArticleView.ts
Grunny: Doesn't `isJapanese` also check the browser language? So, it would also show for Japanese language users on non-Japanese language wikias. We can switch to it, but just need to double check we're OK with that for the MVP as it was my understanding it was Japanese language wikias only, not potentially Japanese-language users on all wikias.

front/scripts/main/models/ArticleModel.ts
rogatty: In MediaWiki we can use `$title->isMainPage()` method which surely is more trustworthy than comparing two strings.

It would be as easy as modifying this bit: https://github.com/Wikia/app/blob/dev/extensions/wikia/MercuryApi/MercuryApiController.class.php#L295

It's just an idea :)

front/scripts/main/views/ArticleView.ts
inez: @Grunny Good catch!! I wouldn't change "requirements" last second. My suggestion is to have two methods in mixin isJapaneseWiki and isJapaneseBrowser (something like that). Pros:
* we are not changing product requirements last minute
* methods are where they belong (language mixin)
* names are less confusing - so next time we will avoid situation like this - that we overlooked that it is checking "different language" (browser vs wiki)

front/scripts/main/models/ArticleModel.ts
inez: Oh yeah :) That other method is relying on Title->equals which compares only namespace - not an issue for now for Mercury (only MAIN namespace) supported - but might bite us in the future.
Yeah, I think it's a good suggestion.

Grunny: Yep, sounds good. It will definitely be more reliable in the long run. :)

Grunny: Alright, added in https://github.com/Wikia/app/pull/7311 We can modify this after that is merged.

Grunny: ...and done. :)

rogatty:
LGTM

front/scripts/main/models/ArticleModel.ts
```diff
@@ -194,6 +194,10 @@ App.ArticleModel.reopenClass({
 				data.topContributors = source.topContributors;
 			}
 
+			if (source.isMainPage) {
+				data.isMainPage = source.isMainPage;
```
inez: I think inMainPage should be added to the interfacer at the beginning of this file as well.

Grunny: Done. :)

PR #852, opened 2015-05-27
Fix curated content invalid typing

rogatty:
`isTopSection` property was missing from the `CuratedContentSection` type.

https://github.com/Wikia/mercury/blob/dev/front/scripts/main/components/CuratedContentComponent.ts#L56 and https://github.com/Wikia/mercury/blob/dev/front/scripts/main/components/CuratedContentComponent.ts#L92 were highligthed by WebStorm because of invalid type.

/cc @Warkot @vforge


vforge:
LGTM :shipit:

PR #855, opened 2015-05-27
add exception to changelog-view

tomasznapieralski:


tomasznapieralski:
Please CR

tasks/changelog-view.sh
```diff
@@ -1,19 +1,21 @@
...
   TO=origin/master
 fi
 
-git --no-pager log $TO..$FROM --merges --pretty=format:'* %s: %b' |
+git --no-pager log $TO..$FROM --merges --pretty=tformat:'* %s: %b' |
 grep 'Merge pull' |
 sed -e 's/* .*from Wikia\//* /' \
     -e 's/^* \([A-Z]\{2,5\}-[0-9]*\)/* [\1](https:\/\/wikia-inc\.atlassian\.net\/browse\/\1)/' \
-	-e 's/).*:/)/'
+	-e 's/).*:/)/' |
```
Warkot: tab vs spaces

tasks/changelog-view.sh
```diff
@@ -1,19 +1,21 @@
...
 
 if [ "$2" != "" ]; then
-  TO=origin/$2
+  TO=$2
 else
   TO=origin/master
 fi
 
-git --no-pager log $TO..$FROM --merges --pretty=format:'* %s: %b' |
+git --no-pager log $TO..$FROM --merges --pretty=tformat:'* %s: %b' |
```
Warkot: **t**format?

tasks/changelog-view.sh
```diff
@@ -1,19 +1,21 @@
...
 grep 'Merge pull' |
 sed -e 's/* .*from Wikia\//* /' \
     -e 's/^* \([A-Z]\{2,5\}-[0-9]*\)/* [\1](https:\/\/wikia-inc\.atlassian\.net\/browse\/\1)/' \
-	-e 's/).*:/)/'
+	-e 's/).*:/)/' |
+grep -vi 'release-'
```
rogatty: Will it remove merges with `release-` string in the name from changelog? It seems too restrictive, what if there is a branch named `fix-release-blocker` or something alike?

tasks/changelog-view.sh
tomasznapieralski: The 'tformat:' format works exactly like 'format:', except that it
provides "terminator" semantics instead of "separator" semantics. In
other words, each commit has the message terminator character (usually a
newline) appended, rather than a separator placed between entries.
This means that the final entry of a single-line format will be properly
terminated with a new line, just as the "oneline" format does.

tomasznapieralski: http://opensource.apple.com/source/Git/Git-19/src/git-htmldocs/pretty-formats.txt

tasks/changelog-view.sh
tomasznapieralski: i don't know how to override that. Maybe "release" should be restricted?

Warkot: How it goes with regexes? http://www.robelle.com/smugbook/regexpr.html

PR #861, opened 2015-05-27
CONCF-601 | added fixed translations for article-top-contributors-label

Warkot:
Added fixed translations for `article-top-contributors-label` and removed the faulty ones.
It's just a part of #813 that concerns CONCF-601 / CONCF-300.
@hakubo @rogatty @vforge


Warkot:
Key changes already merged in #813.

PR #862, opened 2015-05-27
SOC-753 Track events for log-in and sign-up

mklucsarits:
These changes add event tracking in UA for various events on the new log-in page.

https://wikia-inc.atlassian.net/browse/SOC-753

@bkoval @kenkouot


front/scripts/auth/main.ts
```diff
@@ -1,5 +1,70 @@
+/// <reference path='../baseline/mercury.ts' />
+/// <reference path='../mercury/utils/track.ts' />
+
 window.document.addEventListener('DOMContentLoaded', function ():void {
 	var formElement = window.document.querySelector('form');
 	new Form(formElement).watch();
 	new SubmitValidator(formElement).watch();
+
+	var UA = Mercury.Modules.Trackers.UniversalAnalytics,
```
bkoval: We could move these two definitions to the top of the methods body to have just one var statement per the whole method :)

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,70 @@
...
+		dimensions: (string|Function)[] = [];
+
+	dimensions[4] = 'mercury'; // Skin
```
bkoval: What do the guidelines say about the inline comments? Should we be using them or should we just be placing them in the line above the code they're describing?

server/views/_layouts/auth.hbs
```diff
@@ -27,6 +29,19 @@
 		{{/unless}}
 
 		{{#unless noScripts}}
+		<script src="/front/vendor/jquery/dist/jquery.js"></script>
+		<script src="/front/vendor/fastclick/lib/fastclick.js"></script>
+		<!-- build:js /front/scripts/baseline.js -->
+		<script src="/front/scripts/baseline.js"></script>
+		<!-- endbuild -->
+		<!-- build:js /front/scripts/mercury.js -->
+		<script src="/front/scripts/mercury.js"></script>
```
bkoval: Do we really need all these dependencies? The whole login was supposed to use as little javascript as possible, hence it moved away from the main Ember stack.

There's no jQuery used in the new code here, maybe we could extract the sole functionality we need from Mercury.ts and make Mercury and our module dependent on it?

Separate question - do we need fastlick here, is there any reason for using it here except removing the 300ms delay?

server/views/_layouts/auth.hbs
```diff
@@ -27,6 +29,19 @@
 		{{/unless}}
 
 		{{#unless noScripts}}
+		<script src="/front/vendor/jquery/dist/jquery.js"></script>
```
bkoval: I think all the JS here should be put between build tags, so it's combined into one in the production mode.

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,70 @@
...
+	Mercury.Utils.track({
+		trackingMethod: 'ga',
+		action: Mercury.Utils.trackActions.click,
+		category: 'user-login-mobile',
+		label: 'login-submit'
+	});
+});
+
+// Click X to "close" log-in form
+window.document.querySelector('a.close').addEventListener('click', function (): void {
```
bkoval: .close should be fine here :)

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,70 @@
...
+	Mercury.Utils.track({
+		trackingMethod: 'ga',
+		action: Mercury.Utils.trackActions.close,
+		category: 'user-login-mobile',
+		label: 'login-modal'
+	});
+});
+
+// Click "Forgot Password" link
+window.document.querySelector('a.forgotten-password').addEventListener('click', function (): void {
```
bkoval: .forgotten-password would be enough :)

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,70 @@
...
+	Mercury.Utils.track({
+		trackingMethod: 'ga',
+		action: Mercury.Utils.trackActions.click,
+		category: 'user-login-mobile',
+		label: 'forgot-password-link'
+	});
+});
+
+// Click "Register Now" link
+window.document.querySelector('footer .footer-callout-link').addEventListener('click', function (): void {
```
bkoval: .footer-callout-link is enough :)

front/scripts/auth/main.ts
kenkouot: Line above!

mklucsarits: I was copying and pasting from elsewhere in the Mercury code. I'll fix these comments.

server/views/_layouts/auth.hbs
mklucsarits: Unfortunately these were all the dependencies required for using GA/UA on the auth pages. Otherwise compile errors are thrown.

front/scripts/main/app.ts
```diff
@@ -18,6 +18,7 @@ interface Window {
 declare var i18n: I18nextStatic;
 declare var EmPerfSender: any;
 declare var optimizely: any;
+declare var FastClick: any;
```
mklucsarits: Moved to this file to decouple Fastclick and jQuery from util modules.

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -55,7 +54,11 @@ module Mercury.Modules.Trackers {
 			if (overwrite) {
 				this.dimensions = dimensions;
 			} else {
-				$.extend(this.dimensions, dimensions);
+				for (var key in dimensions) {
```
mklucsarits: Removing jQuery dependency.

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -55,7 +54,11 @@ module Mercury.Modules.Trackers {
 			if (overwrite) {
 				this.dimensions = dimensions;
 			} else {
-				$.extend(this.dimensions, dimensions);
```
mklucsarits: Removing jQuery dependency.

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
```diff
@@ -55,7 +54,11 @@ module Mercury.Modules.Trackers {
...
+				for (var key in dimensions) {
+					if (dimensions.hasOwnProperty(key)) {
```
mklucsarits: @kenkouot Should this overwrite existing properties or only append? The original ```$.extend``` overwrote properties in ```this.dimensions``` but is that desired?

server/views/_layouts/auth.hbs
mklucsarits: I removed the Fastclick and jQuery dependencies here. Further decoupling the utility modules, like tracking, from the Ember-Mercury core is a big task (removing all instances of M.prop, for example) that the core Mercury team would need to consider. @kenkouot @lizlux

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,72 @@
...
-	var formElement = window.document.querySelector('form');
+	var formElement = window.document.querySelector('form'),
+		UA = Mercury.Modules.Trackers.UniversalAnalytics,
+		dimensions: (string|Function)[] = [];
+
 	new Form(formElement).watch();
 	new SubmitValidator(formElement).watch();
+
+	// Skin
+	dimensions[4] = 'mercury';
```
lizlux: can we separate this into it's own function?

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,72 @@
...
+			category: 'user-login-mobile',
+			label: 'login'
+		});
+	}
 });
+
+// Event Tracking
+
+// Click "Sign In" button
+window.document.querySelector('#loginSubmit').addEventListener('click', function (): void {
```
lizlux: `window.document` seems like overkill. Can we just use `document`? We don't have any rule about being this explicit do we?

front/scripts/auth/main.ts
mklucsarits: I can. Do you have any suggestions for namespacing it so it's not a global function? But aside from the creation of the array, it kind of already is its own function (UA.setDimensions).

front/scripts/auth/main.ts
mklucsarits: Yeah, I can change it to ```document```. I was following the lead of ```window.document``` use already in the file.

front/scripts/auth/main.ts
lizlux: This whole file could be wrapped in an IIFE

server/views/_layouts/auth.hbs
bkoval: I'm fine with leaving them for now and creating a ticket for abstracting reusable utils in the HG project. Thanks for removing the deps, this will save us a lot of kBs!

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
hakubo: why?
are we moving it away somewhere?

why do we spend time to write stuff that has already been written?

mklucsarits:
I had to re-add back the jQuery dependency as there were several other places it was needed for $.extend within the /mercury utils directory.

kvas-damian:
I'm still not sure about using jQuery as a dependency for registration and login pages. We plan to load them via iframe on Oasis and I'm sure we use different versions (and URLs) so this library won't be fetched from browser cache. For AJAX calls, most major browsers fully support XHR and code isn't that bad. I already implemented ajax register here without jQuery: https://github.com/Wikia/mercury/compare/SOC-684?expand=1#diff-f82ce89c093e08b6a3fc530263679c2cR88

Having that in mind I still suggest not loading jQuery on login and signup pages. @mklucsarits what do you think?

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,76 @@
-window.document.addEventListener('DOMContentLoaded', function ():void {
-	var formElement = window.document.querySelector('form');
+/// <reference path='../baseline/mercury.ts' />
+/// <reference path='../mercury/utils/track.ts' />
+
+(function () {
+document.addEventListener('DOMContentLoaded', function ():void {
```
kvas-damian: Please correct indentation.

front/scripts/auth/main.ts
```diff
@@ -1,5 +1,76 @@
...
+	var formElement = document.querySelector('form');
 	new Form(formElement).watch();
 	new SubmitValidator(formElement).watch();
+
+	setTrackingDimensions();
```
kvas-damian: What do you think about moving tracking code into separate file?

front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
kvas-damian: @hakubo We would like to use tracking on login pages where we try to avoid jQuery. You can find more info here: https://github.com/Wikia/mercury/pull/862#issuecomment-110033876
Feel free to ask if something is not clear.

mklucsarits:
@kvas-damian I agree with you that jQuery should not be a dependency. However, as the Mercury code is currently organized, the tracking module is tightly coupled with Mercury.Utils, which requires Mercury core files. There is also no easy way to specify adding just the tracking module without adding all modules in /front/scripts/mercury. The reason this is a problem is that there are about a dozen uses of jQuery within the /front/scripts/mercury files. So as I see it, to avoid using jQuery on the login page would require either (1) replacing all instances of ```$``` functions in /front/scripts/mercury with native JS, or (2) decoupling the tracking module from Mercury core and utils to allow use of modules in /front/scripts/mercury without dependencies on other parts of Mercury. Either of these increase the scope of what code this ticket touches, # 2 much more so than # 1. So I'm not sure what we should do to handle this. Maybe @kenkouot can also provide an opinion.

rogatty:
@mklucsarits: if I understand correctly then this should be enough to decouple tracking module:

I've added new task to `gulp/tasks/scripts-front.js` that compiles `front/scripts/mercury/modules/Trackers` folder to `www/front/scripts/Trackers.js`:

```js
gulp.task('scripts-front-tracker', function () {
	var tsProject = ts.createProject(options),
		src = 'front/scripts/mercury/modules',
		folder = 'Trackers';

	return gulp.src([
		'!' + path.join(src, folder, paths.dFiles),
		path.join(src, folder, paths.files)
	])
		.pipe(newer(path.join(paths.dest, folder + '.js')))
		.pipe(ts(tsProject)).js
		.on('error', function () {
			console.log('kszsz');
			if (gutil.env.testing && environment.isProduction) {
				console.error('Build contains some typescript errors/warnings');
				process.exit(1);
			}
		})
		.pipe(concat(folder + '.js'))
		.pipe(gulpif(environment.isProduction, uglify()))
		.pipe(gulp.dest(paths.dest));
});
```

Try adding it, running with `gulp scripts-front-tracker` and check if this is the code you need. If yes then it obviously needs some rethinking and cleaning up to be reusable.

kvas-damian:
@rogatty thanks for your input. It looks very promising :+1: 
@mklucsarits Can you verify if it works for login pages?

kenkouot:
I don't like the idea of just compiling out so many different packages, but I also don't think we should refactor all usages of jQuery since organizationally, we tend to use jQuery everywhere. But if we had to, I'd rather refactor the jQuery usage out of Mercury Utils instead, since I don't think it's an unreasonable expectation to include Mercury Utilities as an standard common lib.

hakubo:
on the other hand to use Ember they say you have to load jQuery as well.
It seems reasonable to me to make jQuery a hard dependency of libs that we use not to rewrite all the stuff once again...

Short, clean code is more important than premature optimisations IMHO.

kvas-damian:
My random thoughts - not directly connected to current problems with dependencies.
I wasn't involved in the initial discussion where and how we want to develop login and signup flow in Wikia, so I could be missing something. Please correct me if I write something wrong.
* we decided to implement login out of Ember to increase it's performance.
* we wanted to serve the same login pages on mobile and desktop 
* we decided to use mercury to serve those pages - and here is my question: Was that because we wanted increase speed of initial work like setup tools and environment? Are there any additional reasons why we made that choice? I'm asking because it looks that currently login flow grows as a hack on mercury stack - it the same repo but using completely different approach, we put our code here and there and there are some problems with different things including dependencies.  How I perceive it now: we're growing another monolith. 

What I was thinking about is putting login and signup flow frontend as a separate service, on a separate server instances, using it's own stack and maybe reusing code that could be shared between it and mercury. By doing that we could benefit from all advantages of having SOA.

I don't think it's the best place start this discussion but we should have that talk.

Finally: This is out of scope of this ticket. As a short term solution we can include those dependencies but we should fix it before we roll out registration on desktop (because users will have to download a lot of code that they not use on Oasis).

ping @kenkouot @lizlux @hakubo @rogatty @rafalkalinski @garthwebb @mklucsarits @bkoval and maybe @cmiller-wikia

lizlux:
I like the idea of keeping jQuery out of the login pages, while still making sure we maintain a nice utils collection to stay DRY. The functionality of the auth pages is very limited in scope and things like XHR and `$.extend` are easy enough to shim without loading the whole library. 

As for the decision to build the login pages inside the Mercury repo, I still see the advantages. For example, we utilize the same i18n libraries, the same baseline JS (like `M.prop`, although I'm not sure if we're actually using it at this point, but we should), and the same gulp infrastructure (including typescript, build, tslint, etc.). Also, we're growing to have a general familiarity with the Mercury setup and this speeds up development, deployment, VCL logic, etc. 

Finally, I think sharing some of the same utilities between two different apps in the same repo is a great way to encourage us to write modular code so we can save bytes when sharing code across apps.

hakubo:
I think you have a really good idea @kvas-damian. Those should be two separate services.

They may live in one repo to easy reuse of code, but we should be able to build and deploy them separately.

@lizlux I don't think Damian meant to make separate repos, just as Services team have decided to try one repo for all of them.
It's just a matter of correct folder structure and build steps

in other words instead of building something like this:
```
app
  server
    mercuryandauth.ts
  front
    mercury.ts
    auth.ts
  gulp
    mercuryandauth.ts
```

we should think of a structure that is closer to this
```
app
  commons
    gulp
      commonbuild.ts
    libs
      somecommonfile.ts
  auth
    server
      auth.ts
    front
      auth.ts
    gulp
      build.ts
  mercury
    server
      auth.ts
    front
      auth.ts
    gulp
      build.ts
```

so we could build them separately, deploy them separately but they could still reuse some code without the need of separate repositories and 'complex' dependency building.

lizlux:
@hakubo ah I see. I would be in support of a change like that for sure. Thanks for clarifying.

kvas-damian:
To be clear, initially I was thinking about separate repo, but after your comments it makes more sense to keep using one repo.

Lets stop offtopic here, I'll move this discussion into email so we can decide if we want to go this way or not and how.

kvas-damian:
@mklucsarits can you address last comment: https://github.com/Wikia/mercury/pull/862/files#r31346746
So we can finally move this ticket to QA?

front/scripts/main/app.ts
```diff
@@ -34,6 +35,12 @@ window.emberHammerOptions = {
 	}
 };
 
+$(function() {
+	// FastClick disables the 300ms delay on iOS and some Android devices. It also uses clicks so that
+	// elements have access to :hover state
+	FastClick.attach(document.body);
```
rogatty: Should this be in an intiliazer? I don't know how it would impact Hammer.js, though.
/cc @kenkouot @hakubo

front/scripts/auth/tracking.ts
```diff
@@ -0,0 +1,70 @@
+/// <reference path='../baseline/mercury.ts' />
+/// <reference path='../mercury/utils/track.ts' />
+
+(function () {
+	document.addEventListener('DOMContentLoaded', (): void => {
+		setTrackingDimensions();
+
+		if (document.querySelector('small.error') !== null) {
```
RafLeszczynski: I don't have full context but it seams that login errors comes from backend and are rendered inside `<small class="error">`. It brings a risk of loosing tracking if the template markup will change. Do you have a test for this so you will catch the case when error tracking stops because of the change in template. 

Alternatively you could consider tracking errors before rendering (on backend) and do not depend on checking if particular DOM element exist.

front/scripts/auth/tracking.ts
```diff
@@ -0,0 +1,70 @@
...
+		// LoginStatus
+		dimensions[5] = 'anon';
+		// IsCorporatePage
+		dimensions[15] = 'No';
+		Mercury.Modules.Trackers.UniversalAnalytics.setDimensions(dimensions);
+	}
+
+	// Click "Sign In" button
+	document.querySelector('#loginSubmit').addEventListener('click', (): void => {
+		Mercury.Utils.track({
```
RafLeszczynski: The call to `Mercury.Utils.track() is done multiple times in this closure and many arguments are repeated in each call. Please consider creating a helper function which accepts 2 arguments - type of action and label. This will save you a lot of typing.

front/scripts/main/app.ts
RafLeszczynski: Do we have `d.ts` for these external libraries? If not it would be good to create them and avoid setting type to `any`

front/scripts/auth/tracking.ts
mklucsarits: I agree with your point, but I'm not sure it's worth messing with right now. There's a change coming to move login requests to AJAX (https://github.com/Wikia/mercury/pull/905), which will require the logic for this tracking call to change anyway. I can make the suggested change at that point, either as part of SOC-698 or right after.

bkoval: Good point, @RafLeszczynski ! For now we have only 2 possible errors for login form and they both result in an unsuccessful login. Adding the fact that small.error is a generic place for displaying errors for style-guide we should be safe to keep it this way for now. After moving to AJAX the problem will be solved anyway, as we'll have a direct access to the errors from XHR responses, so @mklucsarits is right :)

front/scripts/auth/tracking.ts
bkoval: Agreed!

bkoval:
LGTM after addressing @RafLeszczynski 's comments!

front/scripts/main/app.ts
mklucsarits: We don't have a type file for Fastclick that I can see. Anyway that's outside of the scope of this PR. I only moved this code here from another file.

front/scripts/auth/tracking.ts
mklucsarits: I don't see the value in that. Although it might be a better idea to assign the common strings to variables.

bkoval:
LGTM :dancer:

kvas-damian:
LGTM

bkoval:
Closing since I fixed unit tests in SOC-753 and we'll test both tickets at the same time.

PR #864, opened 2015-05-27
HG-605 Description meta tag now updates on article change

katanka:
https://wikia-inc.atlassian.net/browse/HG-605


front/scripts/main/views/ArticleView.ts
```diff
@@ -108,6 +108,8 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, {
 		if (model) {
 			var title = model.get('cleanTitle');
 			document.title = title + ' - ' + Mercury.wiki.siteName;
+			var description = model.get('description');
```
rogatty: One `var` per scope, see https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#variables

server/views/_layouts/ember-main.hbs
```diff
@@ -6,7 +6,7 @@
 		<title>{{#if displayTitle}}{{displayTitle}} - {{/if}}{{wiki.siteName}}</title>
 		<meta name="keywords" content="{{#if wiki.siteMessage}}{{wiki.siteMessage}},{{/if}}{{wiki.siteName}},{{wiki.dbName}},{{displayTitle}}">
 		{{#if article.article.description}}
```
rogatty: `$('#meta-description').attr('content', description);` won't work if this condition isn't met (the initial article doesn't have description).
Is it possible? Should we care? Maybe this condition shouldn't be here at all?

rogatty:
This feels hacky but setting `document.title` does too and it's already there :see_no_evil:

server/views/_layouts/ember-main.hbs
```diff
@@ -6,7 +6,7 @@
...
-			<meta name="description" content="{{article.article.description}}">
+			<meta name="description" id="meta-description" content="{{article.article.description}}">
```
kenkouot: id's should be `camelCased` https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#class-and-id-naming-conventions

rogatty:
@katanka https://github.com/Wikia/mercury/commit/2894a703cfc03d0ade9999e568d25cee0464e25c doesn't really address what I meant by https://github.com/Wikia/mercury/pull/864#discussion_r31215463. If `{{#if article.article.description}}` returns false then we won't render `<meta name="description" id="metaDescription" content="{{article.article.description}}">` and because of that `$('#metaDescription').attr('content' ...` won't change anything.

Right?

katanka:
@rogatty Ok, I see what you mean. I added a check to create and append the description tag to head in the correct place if it wasn't already there

front/scripts/main/views/ArticleView.ts
```diff
@@ -119,6 +119,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, App.LanguagesM
 		if (model) {
 			var title = model.get('cleanTitle');
 			document.title = title + ' - ' + Mercury.wiki.siteName;
+
+
+			if ($('meta[name="description"]').length == 0) {
```
kenkouot: Use `===` strict equality operator in JS

front/scripts/main/views/ArticleView.ts
```diff
@@ -119,6 +119,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, App.LanguagesM
...
+				var metaTag = $(document.createElement('meta'));
```
kenkouot: jQuery already wraps `document.createElement`, just do `$('<meta></meta')` for a new element.

katanka: $(document.createElement) is significantly faster than just jQuery
Source: https://jsperf.com/jquery-vs-createelement

kenkouot: :+1:

front/scripts/main/views/ArticleView.ts
```diff
@@ -119,6 +119,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, App.LanguagesM
 		if (model) {
 			var title = model.get('cleanTitle');
 			document.title = title + ' - ' + Mercury.wiki.siteName;
+
+
```
rogatty: Too many `\n`

front/scripts/main/views/ArticleView.ts
rogatty: Use one `var` per scope.

front/scripts/main/views/ArticleView.ts
rogatty: Maybe instead we should change https://github.com/Wikia/mercury/blob/dev/server/views/_layouts/ember-main.hbs#L8 and always render this element, just leaving it empty when there is no description? It would be consistent with what you're doing below in https://github.com/Wikia/mercury/pull/864/files#diff-ad202095518d9272071f81ee71c9f8f7R130 (you're not removing the element, just setting value as empty).

PR #867, opened 2015-05-28
INT-87: Use relative URL to MW api.php. It will get it to work properly on sandbox environment and production, but will not work on localhost

inez:


rogatty:
You would be able to use full URL (that would work everywhere) if https://wikia-inc.atlassian.net/browse/HG-712 was fixed.

PR #868, opened 2015-05-28
SOC-850 & SOC-845 New Login Fixes to cooperate with autofill

bkoval:
Both Android and iOS have different autofill mechanisms. We need to gear up as much as we can to let them do their work, not affecting our validation. 

 - iOS has a "Password autofill" feature on login input that will autofill the password and login for the user, not "input" or "focus/blur" events on the way. We only have the "change" event on the whole form )with capturing flag switched on) to intercept that

 - Android's Chrome will autofill the fields after DOMContentLoaded handlers finish executing and it will spontaneously trigger "change" event for either one or more fields (not deterministically)

 - Both autofill mechanisms can be used "on the fly" as the user types in and selects username from a dropdown / clicks "password autofill" above the keyboard

 - Password in Android's Chrome will dissapear when a username that it was associated with changes in the username field and it will not fire a "change" event

This all means we need to iterate through fields to see what's the actual state of the values whenever one of the fields changes.

### DIFF
--- a/front/scripts/auth/common/Form.ts
+++ b/front/scripts/auth/common/Form.ts
@@ -3,9 +3,11 @@
  */
 class Form {
 	form: HTMLFormElement;
+	inputs: NodeList;
 
 	constructor (form: Element) {
 		this.form = <HTMLFormElement> form;
+		this.inputs = this.form.querySelectorAll('input[type=text], input[type=password], input[type=email]');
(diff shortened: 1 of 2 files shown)


rafalkalinski:
LGTM

kvas-damian:
LGTM

PR #869, opened 2015-05-28
CONCF-512 Refactor lightboxes

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-512

This one is fun! Big refactor of lightboxes as we want to use components instead of controllers and views.

I've used the occassion to clean up ordering of properties in ArticleView too. Sorry for all the conflicts :see_no_evil: See https://github.com/Wikia/guidelines/tree/master/Ember#declaring-ember-objects

~~Work in progress, to do:~~
- [x] fix keyboard navigation
- [x] fix that the first item in gallery is opened, no matter which one you click
- [x] fix clicking on the video, shouldn't bubble up
- [x] check how `LightboxAdsComponent` is doing
- [x] handle query params
  - [x] open media lightbox when `?file` is present
  - [x] change `?file` when current media changes
  - [x] open map lightbox when `?map` is present
  - [x] ~~handle back & forward buttons~~ out of scope, will be handled as another task
- [x] clean up displaying `Media not found` error
- [x] clean up `lightbox-content` and `lightbox-content-inner` classes
- [x] clean up handling clicks, `ApplicationView` vs `ArticleView`
- [x] transition opacity when opening lightbox
- [x] fix unit tests
- [x] check if automation tests need fixes

/cc @kenkouot @lizlux @hakubo


front/scripts/main/mixins/ClickHandlerMixin.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	/**
+	 * Determine if we have to apply special logic to the click handler for MediaWiki / UGC content
+	 */
+	shouldHandleClick: function (target: EventTarget): boolean {
+		var $target = $(target);
+
+		return (
+			$target.closest('.mw-content').length &&
+				// ignore polldaddy content
```
inez: indentation

kenkouot:
Can we find another solution than binding the `style` attribute part of this PR?

front/scripts/main/components/LightboxMediaComponent.ts
```diff
@@ -0,0 +1,202 @@
...
+			return value;
+		}
+	}),
+
+	galleryLength: Em.computed('isGallery', 'current', function (): number {
+		if (this.get('isGallery')) {
+			return this.get('current').length;
+		} else {
+			return -1;
+		}
```
dianafa: do we need to write `else` here to return `-1`?

front/scripts/main/views/ArticleView.ts
```diff
@@ -293,6 +316,38 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, App.LanguagesM
...
+	handleMedia: function (target: HTMLElement): void {
+		var $target = $(target),
+			galleryRef = $target.closest('[data-gallery-ref]').data('gallery-ref'),
+			$mediaElement = $target.closest('[data-ref]'),
+			mediaRef = $mediaElement.data('ref'),
+			media: typeof App.MediaModel;
+
+		if (mediaRef >= 0) {
+			Em.Logger.debug('Handling media:', mediaRef, 'gallery:', galleryRef);
+
```
dianafa: as far as I remember this was mixed I mean, not intuitive for me, explanation:
gallery0:
image0, image1, image2, image3
gallery1:
image4, image5, image6

When you click on image2 it shows you: `Handling media: 0 gallery: 2` and IMO should be  `Handling media: 2 gallery: 0` :)

rogatty: Media model is filled with images or arrays of images (galleries). `mediaRef` is an index of top level item (image or gallery) and `galleryRef` is an index of gallery item.

We can't do it like you've suggested (`Handling media: 2 gallery: 0`) because not always there is a gallery. Would message `Handling media with mediaRef: 0 and galleryRef: 2` would be more clear for you?

Example of model:
```
media0: image0
media1: [image1, image2, image3]
```
image3 has `mediaRef === 1` and `galleryRef === 2`.

rogatty:
@kenkouot
RE: https://github.com/Wikia/mercury/pull/869#issuecomment-106957440
How would you like to modify width and height of elements from JS level without binding style (using templates like now or `attributeBindings`)? We can use jQuery but it would be worse IMO.

Also this task is already taking me too much time. Would be nice if you create a separate story with proposed solution. Thanks.

front/scripts/main/components/LightboxAdsComponent.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.LightboxAdsComponent = Em.Component.extend({
+	classNames: ['lightbox-ads', 'lightbox-content-inner'],
```
rogatty: @nandy-andy because of this change a CSS selector in tests will need a little update (`ads-lightbox` -> `lightbox-ads`).
See https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/helpers/ModalSelectorsHelper.java#L7

nandy-andy: Cool! Thanks for checking it and... let us know when it's released. We'll merge the changes then:
https://github.com/Wikia/selenium-tests/compare/aden-2082-mercury-lightbox-class-change?expand=1

nandy-andy: @rogatty At the end we came up with solution: https://github.com/Wikia/selenium-tests/pull/664
You can run the tests on your devbox if you want - the command is above.
Once you release the changes we've got a follow-up task to remove old class: [ADEN-2083](https://wikia-inc.atlassian.net/browse/ADEN-2083)

rogatty: Thanks, will let you know.

front/templates/main/components/lightbox-map.hbs
```diff
@@ -0,0 +1,5 @@
+{{#if model.url}}
+	<iframe class='current' src='{{model.url}}'></iframe>
+{{else}}
+	{{i18n 'app.map-lightbox-error'}}
```
hakubo: should this also have: <p class="lightbox-error">

front/scripts/main/components/LightboxImageComponent.ts
```diff
@@ -4,26 +4,33 @@
...
-	classNames: ['lightbox-content-inner'],
+App.LightboxImageComponent = Em.Component.extend(App.ArticleContentMixin, App.LoadingSpinnerMixin, {
+	classNames: ['lightbox-image', 'lightbox-content-inner'],
 	maxZoom: 5,
 	lastX: 0,
 	lastY: 0,
 	lastScale: 1,
+
+	//Easy to port if we find a way to use enum here
+	screenAreas:  {
```
hakubo: this should be removed I think

and all code that is responsible for 'thirds' moved to ThirdsClickMixin

front/templates/main/application.hbs
```diff
@@ -5,14 +5,15 @@
 		</section>
 	</div>
 	{{outlet 'lightbox'}}
-{{else if fullPage}}
+{{elseif fullPage}}
```
hakubo: hmm in docs I can see
else if

http://handlebarsjs.com/block_helpers.html

are you sure that this should be elseif?

rogatty: You're right. I just did a test and `elseif` doesn't work properly. I changed it because WebStorm marks this file as invalid with message "Expected Open End Block {{/" and when changed to `elseif` it's fine.

Reverting now.

front/templates/main/components/lightbox-map.hbs
rogatty: Nope, I removed all instances of `<p class="lightbox-error">` instead.

front/scripts/main/components/LightboxImageComponent.ts
rogatty: It's not so simple. [doubleTap action](https://github.com/Wikia/mercury/pull/869/files#diff-9f4e88df642d30e97e5d22e5c4de0292R182) checks if it's a gallery OR if it's the center of screen. ThirdsClickMixin doesn't support cases like that. For that exact reason @vforge didn't change this file when creating the mixin.

IMO we should think about rewriting the mixin in a more universal way and then use it here but not as a part of this PR.

@hakubo is it fine?

hakubo: ok, procede

hakubo:
Ok, lets test it!

tomasznapieralski:
LGTM

PR #880, opened 2015-05-29
ADEN-2065 Fix styling for Floor Adhesion unit with transparency image

nandy-andy:
When an ad in Floor Adhesion unit uses image with transparency with `height` bigger than the unit's background the close button of the unit is displayed wrongly. This is because of incorrect styling and using `height: 100%` for the close button.

Changes below fixes the issue and address some issues found by scss-lint tool.


rogatty:
Fix indentation, please.

PR #882, opened 2015-05-29
HG-712 mercury_ads_js now pulling from correct wiki domain

katanka:
problems with undefined url when running on sandbox-mercury

### DIFF
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -32,7 +32,8 @@ var localSettings: LocalSettings = {
 	// By default send logs to local syslog only. Possible targets are [syslog, console, default]
 	// The value represent the minimum logging level
 	loggers: {
-		syslog: 'debug'
+		syslog: 'debug',
+		console: 'debug'
 	},
 	devboxDomain: Utils.stripDevboxDomain(process.env.HOST || process.env.LOGNAME),
 	// auth pages aren't supported on custom domains, so this value should only be used for auth features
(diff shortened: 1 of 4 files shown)


config/localSettings.base.ts
```diff
@@ -32,7 +32,8 @@ var localSettings: LocalSettings = {
 	// By default send logs to local syslog only. Possible targets are [syslog, console, default]
 	// The value represent the minimum logging level
 	loggers: {
-		syslog: 'debug'
+		syslog: 'debug',
+		console: 'debug'
```
kenkouot: Console should be removed, otherwise this will turn on console logging in prod, which we don't need.

server/lib/Logger.ts
```diff
@@ -13,7 +13,7 @@ interface AvailableTargets {
 	[key: string]: CreateBunyanLoggerStream;
 }
 /**
- * Loogger interface
```
kenkouot: hahaha

server/lib/Article.ts
```diff
@@ -23,9 +23,9 @@ interface ServerData {
  *
  * @returns ServerData
  */
-function createServerData (): ServerData {
+function createServerData (wikiDomain: string): ServerData {
```
kenkouot: Are there any instances other instances of `createServerData` that would be affected by this new required argument? We could make it optional. What do you think?

katanka:
@kenkouot Effectively, it is optional. Not passing wikiDomain has no effect, because it is handled in Utils.

PR #885, opened 2015-06-01
SOC-705 Drop-down menu for logged-in users

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-705

@bkoval @lizlux


front/styles/main/component/_user-menu.scss
```diff
@@ -0,0 +1,63 @@
+.user-menu {
```
lizlux: Just curious why we weren't able to reuse styles from the existing slide-out, like the overlay and drawer.

mklucsarits: I think maybe the overlay could be extracted to a placeholder (but wouldn't that duplicate the CSS anyway?). However, the drawer here is actually pretty different from the slide-out style of the side nav.

lizlux: Ah okay, just checked out the code and ran it. Didn't realize we were adding a new type of menu.

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,21 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.UserMenuComponent = Em.Component.extend({
+	tagName: 'div',
```
rogatty: This is by default, no need to add it.

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,21 @@
...
+		return '//community.wikia.com/wiki/User:' + this.get('userName');
+	}),
+
+	actions: {
+		hide: function (): void {
+			this.set('isCollapsed', true);
+		}
+	}
+});
+
```
rogatty: This `\n` is not needed.

front/locales/en/main.json
```diff
@@ -71,6 +71,8 @@
 		"edit-publish-error-blocked": "You have been blocked from editing.",
 		"edit-publish-error-noedit": "You don't have permission to edit pages on this wikia.",
 		"edit-publish-error-noedit-anon": "Anonymous users can't edit pages.",
-		"edit-publish-error-protectedpage": "You don't have permission to edit this page."
+		"edit-publish-error-protectedpage": "You don't have permission to edit this page.",
+		"user-menu-profile": "Profile",
+		"user-menu-log-out": "Log out"
```
rogatty: Before uploading to Crowdin, please download the english main.json from there and check if there are any translation keys that are not present in your version.

If you don't do that then uploading this file to Crowdin will effectively delete all of those non existent keys (and translations). This process really needs to be improved.

front/scripts/main/components/UserMenuComponent.ts
mklucsarits: This is what I was referring to during our chat on Slack the other day. Our guidelines for JS indicate a blank line at the end of each file: https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#white-space-guidelines (last bullet point).

kvas-damian: But here you have 2 blank lines ;)

front/templates/main/components/user-menu.hbs
```diff
@@ -0,0 +1,9 @@
+<div {{action 'hide'}} class='overlay'></div>
+<ul class='drawer'>
+	<li>
+		<a href="{{profileUrl}}" class="external">{{i18n 'app.user-menu-profile'}}</a>
+	</li>
+	<li>
+		<a href="/logout" class="external">{{i18n 'app.user-menu-log-out'}}</a>
```
kvas-damian: What do you think about moving URLs out from template?

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,20 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.UserMenuComponent = Em.Component.extend({
+	classNames: ['user-menu'],
+	classNameBindings: ['isCollapsed:collapsed:visible'],
+
+	isCollapsed: true,
+
+	profileUrl: Em.computed(function (): string {
```
hakubo: Em.computed('userName' ...

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,22 @@
...
+		return '//community.wikia.com/wiki/User:' + Em.computed('userName');
```
jsutterfield: Does JS or TS have a way to make this a constant somewhere? I'm not super familiar with the languages, but this seems like a good candidate for that.

front/styles/main/component/_user-menu.scss
```diff
@@ -0,0 +1,63 @@
...
+		transition: height $transition-length;
+		width: 100%;
+		z-index: $z-side-nav;
+
+		li {
+			background-color: #fff;
+			border-bottom: rem-calc(1) solid #d0d0d0;
+		}
+
+		a {
```
jsutterfield: should the `li{}` and `a{}` declarations be switched to keep it alphabetical?

front/templates/main/components/site-head.hbs
```diff
@@ -1,9 +1,8 @@
 <div {{action 'expandSideNav'}} class='side-nav-toggle'>
     {{svg 'nav' viewBox='0 0 24 14' class='icon nav'}}
 </div>
-{{!-- This conditional block should be removed for SOC-705 --}}
 {{#if currentUser.isAuthenticated}}
```
jsutterfield: We decided not to remove the conditional like the removed comment said we would?

mklucsarits: I decided not to just because it didn't make sense for the avatar ```figure``` to be its own component. It would only exist to call another component (user menu). I also didn't want to merge it with the login-icon component because of the difference in functionality and purpose.

front/scripts/main/components/UserMenuComponent.ts
mklucsarits: I agree this should be a constant somewhere. For a different ticket I created a method that constructs a Wikia URL from a given wiki name, but that's not available to use yet, so I'll do something in the interim here.

rogatty: It should be `this.get('userName')`

front/templates/main/components/user-menu.hbs
```diff
@@ -0,0 +1,9 @@
+<div {{action 'hide'}} class='overlay'></div>
```
rogatty: Use double quotes for HTML attributes, see https://github.com/Wikia/guidelines/tree/master/Handlebars#quotes

front/scripts/main/components/UserMenuComponent.ts
mklucsarits: @hakubo up above said to use Em.computed. Is there there a good reason for one or the other?

hakubo: oh, maybe I should have added more context:

http://emberjs.com/api/classes/Ember.computed.html

rogatty: `Em.computed` is a helper which "returns a new property descriptor that wraps the passed computed property function".

`this.get` is an object getter that you should use inside computed property functions.

See http://emberjs.com/api/classes/Ember.computed.html for examples.

front/templates/main/components/user-menu.hbs
```diff
@@ -0,0 +1,9 @@
+<div {{action 'hide'}} class="overlay"></div>
+<ul class="drawer">
+	<li>
```
hakubo: those two look almost the same, do you think we could iterate over some array/object to create it?

front/templates/main/components/site-head.hbs
```diff
@@ -1,9 +1,8 @@
...
-	<figure class='avatar'>
+	<figure {{action 'showUserMenu'}} class='avatar'>
```
rogatty: Use double quotes for HTML attributes, see https://github.com/Wikia/guidelines/tree/master/Handlebars#quotes

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,22 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.UserMenuComponent = Em.Component.extend({
```
hakubo: can you write unit test for this component?

kenkouot:
buildUrl should have tests

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+			domain = Mercury.wiki.basePath.match(/(?!\w+)\.([\w.-]+\.\w{2,3})($|\/)/)[1];
+		} else {
+			domain = Mercury.wiki.basePath.match(/([\w-]+\.\w{2,3})($|\/)/)[1];
+		}
+
+		url += urlParams.wiki ? urlParams.wiki : 'www';
+		url += '.' + domain;		
+
+		if (urlParams.title) {
+			url += '/wiki/' + (urlParams.namespace ? urlParams.namespace + ':' : '') + urlParams.title;
```
rogatty: Should we already assume that `/wiki/` will always be there? Everywhere else we're using `wiki.articlePath`.

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		if (urlParams.path) {
+			url += urlParams.path;
+		}
+
+		if (urlParams.query) {
+			url += '?';
+			url += Object.keys(urlParams.query).map((key: string) =>
```
rogatty: `: string`

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+	 *   {path: '/login', query: {redirect: '/somepage'}}
+	 *   ...returns '//www.wikia.com/login?redirect=%2Fsomepage'
+	 *
+	 *   {wiki: 'glee', title: 'Jeff'}
+	 *   ...returns '//glee.wikia.com/wiki/Jeff'
+	 *
+	 *   {wiki: 'community', namespace: 'User', title: 'JaneDoe', path: '/preferences'}
+	 *   ...returns '//community.wikia.com/wiki/User:JaneDoe/preferences'
+	 */
+	export function buildUrl (urlParams: UrlParams = {}) {
```
rogatty: `: string`

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+		// Domain is extracted from basePath wiki variable, removing the *first* subdomain.
+
+		var domain: string,
+			url = '//';
```
rogatty: Look at comment from @lizlux about missing protocol: https://github.com/Wikia/mercury/pull/898#discussion_r32071729

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+			url += '?';
+			url += Object.keys(urlParams.query).map((key: string) =>
+				`${encodeURIComponent(key)}=${encodeURIComponent(urlParams.query[key])}`
+			).join('&');
+		}
+
+		return url; 
+	}
+}
+
```
rogatty: This `\n` is not needed.
Maybe take a look at your vim configuration or just don't add new lines at the end as they're probably added automatically for you.

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+	export function buildUrl (urlParams: UrlParams = {}) {
+		// Domain is extracted from basePath wiki variable, removing the *first* subdomain.
+
+		var domain: string,
+			url = '//';
+		if (/wikia-dev\.\w{2,3}($|\/)/.test(Mercury.wiki.basePath)) {
+			domain = Mercury.wiki.basePath.match(/(?!\w+)\.([\w.-]+\.\w{2,3})($|\/)/)[1];
+		} else {
+			domain = Mercury.wiki.basePath.match(/([\w-]+\.\w{2,3})($|\/)/)[1];
+		}
```
rogatty: Does it support local environment (`mediawiki119.127.0.0.1.xip.io:8000` kind of URLs)? Would be nice if it did, there is some logic written already https://github.com/Wikia/mercury/blob/dev/server/lib/Utils.ts#L147

front/scripts/baseline/mercury/utils/buildUrl.ts
kvas-damian: Can we add documentation with descriptions for all available params?

front/scripts/baseline/mercury/utils/buildUrl.ts
kvas-damian: and staging environment?

mklucsarits:
@kenkouot It will! I didn't have time to finish them yet.

front/scripts/baseline/mercury/utils/buildUrl.ts
mklucsarits: I'm not trying to be argumentative, but is there a better explanation against using it? The blog Liz links to only says:
```
Links or Canonical URLs

If we use protocol relative URLs in these cases, they can cause duplicate content issues between the http and https versions of the website.
```
I don't understand the point the author is trying to make. (Maybe someone can explain it better.) As far as I understand it, protocol-relative URLs are good for traversing links within your own site -- because if you were on https but linked to http, that may cause a problem -- but external links to other sites should be explicit in the protocol, since the protocol used on your site has no bearing on the other site.

front/scripts/baseline/mercury/utils/buildUrl.ts
mklucsarits: I'll work on a catch-all solution for all possible hosts.

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,67 @@
...
+	 * @config {object} [query] Querystring data, which is converted to a string and properly escaped
+	 * @config {string} [title] Article title
+	 * @config {string} [wiki] Wiki name, as it would be used as a subdomain
+	 * @returns {string}
+	 */
+	export function buildUrl (urlParams: UrlParams = {}): string {
+		// Domain is extracted from basePath wiki variable, removing the *first* subdomain.
+
+		var domain: string,
+			url = '//';
```
lizlux: My point is that the protocol of the page you're on has no bearing on that of the page you're linking to. For assets (JS, images, etc) relative protocols make sense, but for links, they don't - there's just no concrete relationship there.

My suggestion is that we allow a protocol option to be passed in, and set a smart default to https. Our current [VCL logic](https://github.com/Wikia/wikia-vcl/blob/master/wikia.vcl#L40-L45) will redirect from https to http where SSL isn't supported, so we can use this as a general policy while rolling out https to all traffic.

front/scripts/baseline/mercury/utils/buildUrl.ts
kvas-damian: I think @mklucsarits is right. In our case we additionally enforces protocol by redirect on varnish.

front/scripts/baseline/mercury/utils/buildUrl.ts
kvas-damian: I don't totally agree on https as default. Currently most of pages uses HTTP because we cannot easily introduce HTTPS everywhere. We'll end up with redirecting user back to http on every link that was generated by this module. This means additional round trip to our server, which could affect performance - specially on slow mobile connections.
Additionally when we finally introduce SSL everywhere our code will be still valid.

As @mklucsarits wrote in https://github.com/Wikia/mercury/pull/885#discussion_r32253011 we don't see the real reason why we should add protocol here by default. OPS Team already introduces solution for content duplication on Varnish, which is redirect to correct protocol of the page.

kvas-damian:
@mklucsarits Please resolve merge conflicts

front/scripts/baseline/mercury/utils/buildUrl.ts
rogatty: I think that the main reason to avoid relative URLs when linking to other pages is SEO. For example take a look at http://www.dirigodev.com/blog/seo-web-best-practices/relative-vs-absolute-urls-seo/

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,93 @@
...
+			host = host.replace(match[1] + '.' + match[2], match[1] + '.' + urlParams.wiki);
+		// (2) Production wikia.com
+		} else if ((match = host.match(/^(.+?)\.wikia\.com($|\/|:)/)) !== null) {
+			// Domain is specified here in case host subdomain is actually "wiki", "com", etc.
+			host = host.replace(match[1] + '.wikia.com', urlParams.wiki + '.wikia.com');
+		// (3) Devbox hosted on wikia-dev.com, wikia-dev.us, wikia-dev.pl, etc.
+		} else if ((match = host.match(/^(.+)\.(.+?)\.wikia-dev.\w{2,3}($|\/|:)/)) !== null) {
+			host = host.replace(match[1] + '.' + match[2], urlParams.wiki + '.' + match[2]);
+		// (4) Local environment using 127.0.0.1.xip.io
+		} else if ((match = host.match(/^(.+)\.127\.0\.0\.1\.xip\.io($|\/|:)/)) !== null) {
```
kvas-damian: Can we find a rule that matches all xip.io URLs?
Sometimes I test my code on my phone and then I need to provide different IP.

rogatty:
What @kvas-damian said and it LGTM.

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,93 @@
...
+	 * @returns {string}
+	 */
+	export function buildUrl (urlParams: UrlParams = {}, context: any = window): string {
+		var domain: string,
+			host: string = context.location.host,
+			match: Array<string>,
+			url: string;
+
+		if (!urlParams.protocol) {
+			urlParams.protocol = 'https';
```
mklucsarits: After discussion it's sounding like this should be http instead.

lizlux: :+1:

kvas-damian:
Please resolve conflicts and fix unit tests.

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,93 @@
...
+module Mercury.Utils {
+	/**
+	 * This function constructs a URL given pieces of a typical Wikia URL. All URL
+	 * parts are optional. Passing in empty params will result in the site home page,
+	 * ie. www.wikia.com.
+	 *
+	 * Some example parameters and results:
+	 *
+	 *   {path: '/login', query: {redirect: '/somepage'}}
+	 *   ...returns '//www.wikia.com/login?redirect=%2Fsomepage'
```
kvas-damian: Please update examples to reflect default protocol.

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,93 @@
...
+		} else if ((match = host.match(/^(.+?)\.wikia\.com($|\/|:)/)) !== null) {
+			// Domain is specified here in case subdomain is actually "wiki", "com", etc.
+			host = host.replace(match[1] + '.wikia.com', urlParams.wiki + '.wikia.com');
+		// (3) Devbox hosted on wikia-dev.com, wikia-dev.us, wikia-dev.pl, etc.
+		} else if ((match = host.match(/^(.+)\.(.+?)\.wikia-dev.\w{2,3}($|\/|:)/)) !== null) {
+			host = host.replace(match[1] + '.' + match[2], urlParams.wiki + '.' + match[2]);
+		// (4) Environment using xip.io
+		} else if ((match = host.match(/^(.+)\.(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\.xip\.io($|\/|:)/)) !== null) {
+			host = host.replace(match[1] + '.' + match[2] + '.xip.io', urlParams.wiki + '.' + match[2] + '.xip.io');
+		}
```
kvas-damian: Can we extract environment related code to separate method?

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,35 @@
...
+	isCollapsed: true,
+
+	links: Em.computed(function (): Array<any> {
+		return [
+			{
+				href: M.buildUrl({
+					wiki: 'community',
+					namespace: 'User',
+					title: this.get('userName')
+				}),
```
kvas-damian: Why are we linking to community instead of current wiki?

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -0,0 +1,93 @@
...
+	 * @config {string} [namespace] MediaWiki article namespace
+	 * @config {string} [path] Additional URL path appended to the end of the URL before the querystring
+	 * @config {string} [protocol] Protocol
+	 * @config {object} [query] Querystring data, which is converted to a string and properly escaped
+	 * @config {string} [title] Article title
+	 * @config {string} [wiki] Wiki name, as it would be used as a subdomain
+	 * @param {object} context Window context
+	 * @returns {string}
+	 */
+	export function buildUrl (urlParams: UrlParams = {}, context: any = window): string {
```
kvas-damian: In future we can think about extracting this method to localUrl and fullUrl.
Until we don't have a use case for that it's ok to keep it as it is.

bkoval:
Unit tests failed for the last commit:

"[1A[2KPhantomJS 1.9.8 (Linux 0.0.0) Mercury.Utils.buildUrl helper function (loaded with baseline) Wiki subdomain is correctly replaced for each environment host FAILED"

front/scripts/main/components/UserMenuComponent.ts
mklucsarits: If we link to the current wiki, the namespace for "User" may be in another language, and I thought the URL would be incorrect using the English namespace for that wiki, leading to a 404. However, I see that if you use the English namespace, it will redirect to the proper namespace in the wiki's language, so I will change this.

For example, see http://es.walkingdead.wikia.com/wiki/User:Mklucsarits

kvas-damian: :+1:

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,35 @@
...
+	classNames: ['user-menu'],
+	classNameBindings: ['isCollapsed:collapsed:visible'],
+
+	isCollapsed: true,
+
+	links: Em.computed(function (): Array<any> {
+		return [
+			{
+				href: M.buildUrl({
+					wiki: Mercury.wiki.dbName,
```
kvas-damian: ```dbName``` is not equivalent for wiki domain, see ```http://spolecznosc.wikia.com``` vs ```plwikia``` as a dbName

kvas-damian: BTW: we should use relative URLs when targeting current wiki

front/scripts/main/components/UserMenuComponent.ts
```diff
@@ -0,0 +1,35 @@
...
+					namespace: 'User',
+					title: this.get('userName')
+				}),
+				textKey: 'user-menu-profile'
+			},
+			{
+				href: M.buildUrl({
+					path: '/logout'
```
bkoval: I don't think we have a /logout path ready at this point. For now we should use /wiki/Special:UserLogout

front/styles/main/component/_user-menu.scss
```diff
@@ -0,0 +1,63 @@
...
+			display: block;
+			overflow: hidden;
+			padding: rem-calc(10) $side-nav-item-gutter;
+			text-overflow: ellipsis;
+			width: 100%;
+		}
+
+		li {
+			background-color: #fff;
+			border-bottom: rem-calc(1) solid #d0d0d0;
```
bkoval: Maybe these colors could be moved to settings file?

front/styles/main/component/_user-menu.scss
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	&.collapsed {
+		.drawer {
+			height: 0;
+		}
+
+		.overlay {
+			opacity: 0;
+			transition: visibility 0s $transition-length, opacity $transition-length;
```
bkoval: What's the point of adding "0s" here?

front/styles/main/component/_user-menu.scss
mklucsarits: Well, the settings file would be the style guide settings file. I don't think we should do that, because there's no standard yet for either this border color or this menu as a component. Also, this menu and its styles are temporary until the new global nav is finished, which will replace this.

front/styles/main/component/_user-menu.scss
mklucsarits: It's the transition-delay value. https://developer.mozilla.org/en-US/docs/Web/CSS/transition

These styles were copied from _side-nav.scss and are temporary until the new global nav is finished, which will replace this.

kvas-damian:
LGTM

front/styles/main/component/_user-menu.scss
bkoval: But isn't transition-delay implicitly set to 0s if you just omit this value?

bkoval:
LGTM

front/styles/main/component/_user-menu.scss
mklucsarits: Yeah I would guess so.

PR #887, opened 2015-06-01
HG-686 removed namespace specifier from random page query for community wiki

katanka:
https://wikia-inc.atlassian.net/browse/HG-686

### DIFF
--- a/server/lib/MediaWiki.ts
+++ b/server/lib/MediaWiki.ts
@@ -156,7 +156,6 @@ export class ArticleRequest {
 		var url = createUrl(this.wikiDomain, 'api.php', {
 			action: 'query',
 			generator: 'random',
-			grnnamespace: 0,
 			format: 'json'
 		});
 
--- a/server/lib/MediaWiki.ts
+++ b/server/lib/MediaWiki.ts
@@ -153,11 +153,21 @@ export class ArticleRequest {
 	 * @return {Promise<any>}
 	 */
 	randomTitle (): Promise<any> {
-		var url = createUrl(this.wikiDomain, 'api.php', {
-			action: 'query',
-			generator: 'random',
-			format: 'json'
-		});
+		var url = "";
+		if (this.wikiDomain.indexOf('community') > -1) {
+			url = createUrl(this.wikiDomain, 'api.php', {
+				action: 'query',
+				generator: 'random',
+				format: 'json'
+			});
+		} else {
+			url = createUrl(this.wikiDomain, 'api.php', {
+				action: 'query',
+				generator: 'random',
+				grnnamespace: 0,
+				format: 'json'
+			});
+		}
 
 		return fetch(url, this.wikiDomain);
 	}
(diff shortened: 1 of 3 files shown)


server/lib/MediaWiki.ts
```diff
@@ -153,12 +153,21 @@ export class ArticleRequest {
 	 * @return {Promise<any>}
 	 */
 	randomTitle (): Promise<any> {
-		var url = createUrl(this.wikiDomain, 'api.php', {
-			action: 'query',
-			generator: 'random',
-			grnnamespace: 0,
-			format: 'json'
-		});
+		var url = "";
```
rogatty: Use single quotes.

rogatty:
@katanka can you explain this change? How does removing the namespace fix the bug? Won't it return objects that Mercury doesn't support (non-articles)?

katanka:
@rogatty For community.wikia.com, that is the correct functionality. If you try the random page link on the desktop (non-mercury) site, it will go to all of the namespaces. So by removing the namespace specifier for just community, it makes the functionality identical on both platforms.

rogatty:
But will it work? What happens when non-article is returned? I believe that somewhere in the code we assume it will always be an article.

kenkouot:
Would it return the full namespace + title? EG: "Help:Foo"?

server/lib/MediaWiki.ts
```diff
@@ -153,12 +153,21 @@ export class ArticleRequest {
...
+		if (this.wikiDomain.indexOf('community') > -1) {
```
hakubo: also can we restrain from doing special cases for specific communities?
maybe we can do 'global' fix? 

We have this pattern in many places and it's hard to reason what happens where after some time.

katanka: Yeah, I agree it isn't pretty. However, it sorta is a special case: we need behavior on the Community wiki that is different from the other wikis. Is there a better way to implement it?

katanka:
@kenkouot Yes, it returns [namespace]:[title]

@rogatty As far as I can tell, that bug was caused by the fact that random was only returning pages in the Main namespace, which only has ~60 pages in it. So, it was very likely to get repeats. I was not able to replicate it returning the same page over and over again. 

By opening it up to all the namespaces, like the desktop site does, we only get repeats at a very small rate.

rogatty:
@katanka:
Open http://gameofthrones.wikia.com/api.php?action=query&generator=random&grnnamespace=0&format=json in your browser and refresh multiple times. Each time I do it I get a different article.
Now open http://gameofthrones.wikia.com/ with Mercury and try getting the random article. I clicked ~20 times and got two articles alternately.

server/lib/MediaWiki.ts
rogatty: tslint warning: `(quotemark) lib/MediaWiki.ts[156, 13]: " should be '`
@katanka please look at the console before doing commits.

rogatty:
@kenkouot why did you merge it? Now on Community after clicking on random page I get comments, user pages and a lot of "Sorry, this article seems to be empty!". Those pages aren't supported by Mercury and we shouldn't even try to load them.

I'm reverting the merge because it introduces regression - broken pages.

Please start doing QA.

/cc @hakubo @hobbs @jkuner

rogatty:
@katanka can you create a proper PR with change that was described in the ticket (fixing cache issue)? It's this one: https://github.com/Wikia/mercury/pull/887/files#diff-826bec53995755be79d0d75d93dd48e3R38

Thanks.

PR #890, opened 2015-06-03
HG-731 removed random page function

katanka:
https://wikia-inc.atlassian.net/browse/HG-731

It hasn't been decided yet if we're actually going to remove it, but if we do, its ready to go.


kenkouot:
Merge conflict!

rogatty:
@kenkouot please don't merge it before there is a good reason stated in the ticket. So far, there is none.

kenkouot:
Closing this for now.

PR #891, opened 2015-06-04
Hg 654 remove 'mw-content' from non-mediawiki content wrappers

lizlux:
This undoes a hack we were using to make sure links that were not processed by the `ApplicationRoute` `handleLink` action were still followed. Turns out, by default, ember action helpers will prevent the default action of an event, such as click. Adding `preventDefault=false` to the action helper fixes this.  

This PR also ensures the tracking call is still sent properly before navigating to external links, and I updated the hbs to match [new guidelines](https://github.com/Wikia/guidelines/tree/master/Handlebars) re: quotes. 

@inez @rogatty 

I think there's still more work to do to clean up the `handleLink` function, but that can be done in a separate PR.

### DIFF
--- a/front/scripts/main/mixins/TrackClickMixin.ts
+++ b/front/scripts/main/mixins/TrackClickMixin.ts
@@ -16,12 +16,21 @@ App.TrackClickMixin = Em.Mixin.create({
 		trackClick: function (category: string, label: string = '', isNonInteractive: boolean = true): void {
 			this.trackClick(category, label, isNonInteractive);
 		},
-		trackClickAndFollow: function (url: string, category: string, label: string = '', isNonInteractive: boolean = true): void {
+		trackClickAndFollow: function (
+			url: string, category: string,
+			label: string = '',
+			isNonInteractive: boolean = true,
+			newTab: boolean = false
+		): void {
 			this.trackClick(category, label, isNonInteractive);
 			if (!url) {
 				throw new Error('trackClickAndFollow needs an href to follow')
 			}
-			window.location.assign(url);
+			if (newTab) {
+				window.open(url);
+			} else {
+				window.location.assign(url);
+			}
 		}
 	}
 });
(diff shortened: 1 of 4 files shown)


kenkouot:
LGTM

front/scripts/main/mixins/TrackClickMixin.ts
```diff
@@ -15,6 +15,22 @@ App.TrackClickMixin = Em.Mixin.create({
...
+		},
+		trackClickAndFollow: function (
+			url: string, category: string,
+			label: string = '',
+			isNonInteractive: boolean = true,
+			newTab: boolean = false
+		): void {
+			this.trackClick(category, label, isNonInteractive);
+			if (!url) {
+				throw new Error('trackClickAndFollow needs an href to follow')
```
rogatty: Do we really want to throw an error or maybe just use `Em.Logger.error()`? Also, missing semicolon at the end.

front/templates/main/components/share-feature.hbs
```diff
@@ -1,8 +1,8 @@
 {{#if isJapaneseBrowser}}
-	<a {{action 'trackClick' 'share' 'line'}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
+	<a {{action 'trackClickAndFollow' lineShare 'share' 'line' true true}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
```
rogatty: `lineShare 'share' 'line' true true` doesn't say much what is going on.

@lizlux what do you think about explicitly using options like `nonInteractive=true newTab=true`? I know this is longer to write but so much easier to copy from another file and understand immediately without searching for action definition.

It would need some refactoring though, to pass options hash to the method, instead of arguments.

lizlux:
Found a better solution and updated the description. Also created a pull request to update the ember docs to include this functionality - https://github.com/emberjs/guides/pull/339.

rogatty:
@lizlux: two things:
1. Is `true` in `{{action ... true ...}}` needed?
2. Are you sure that introducing new lines inside `a` elements won't break any styling? If they're `inline` or `inline-block` it might make a difference. Would be better to check it.

lizlux:
@rogatty good catch on the `true` param. It wasn't needed. The white space change is fine, I checked.

PR #894, opened 2015-06-08
change default destination

tomasznapieralski:
Looking for changes will be now from last release tag instead of origin/master


tomasznapieralski:
@rogatty @Warkot please CR

rogatty:
LGTM

PR #898, opened 2015-06-08
HG-732 Wikia logo in header now links to http://www.wikia.com

katanka:
https://wikia-inc.atlassian.net/browse/HG-732

### DIFF
--- a/front/templates/main/components/share-feature.hbs
+++ b/front/templates/main/components/share-feature.hbs
@@ -1,8 +1,16 @@
 {{#if isJapaneseBrowser}}
-	<a {{action 'trackClickAndFollow' lineShare 'share' 'line' true true}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
+	<a {{action 'trackClick' 'share' 'line' preventDefault=false}} href="{{lineShare}}" target="_blank" >
+		{{svg 'line' role='img' class='icon line'}}
+	</a>
 {{/if}}
-<a {{action 'trackClickAndFollow' facebookShare 'share' 'fb' true true}} href={{facebookShare}}>{{svg 'fb' role='img' class='icon fb'}}</a>
-<a {{action 'trackClickAndFollow' twitterShare 'share' 'twitter' true true}} href={{twitterShare}}>{{svg 'twitter' role='img' class='icon twitter'}}</a>
+<a {{action 'trackClick' 'share' 'fb' preventDefault=false}} href="{{facebookShare}}" target="_blank">
+	{{svg 'fb' role='img' class='icon fb'}}
+</a>
+<a {{action 'trackClick' 'share' 'twitter' preventDefault=false}} href="{{twitterShare}}" target="_blank">
+	{{svg 'twitter' role='img' class='icon twitter'}}
+</a>
 {{#if isJapaneseBrowser}}
-	<a {{action 'trackClickAndFollow' googleShare 'share' 'google' true true}} href={{googleShare}}>{{svg 'google' role='img' class='icon google'}}</a>
+	<a {{action 'trackClick' 'share' 'google' preventDefault=false}} href="{{googleShare}}" target="_blank">
+		{{svg 'google' role='img' class='icon google'}}
+	</a>
 {{/if}}
--- a/front/templates/main/components/share-feature.hbs
+++ b/front/templates/main/components/share-feature.hbs
@@ -1,8 +1,8 @@
 {{#if isJapaneseBrowser}}
-	<a {{action 'trackClick' 'share' 'line'}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
+	<a {{action 'trackClickAndFollow' lineShare 'share' 'line' true true}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
 {{/if}}
-<a {{action 'trackClick' 'share' 'fb'}} href={{facebookShare}}>{{svg 'fb' role='img' class='icon fb'}}</a>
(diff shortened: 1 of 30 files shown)


front/templates/main/components/share-feature.hbs
```diff
@@ -1,8 +1,8 @@
 {{#if isJapaneseBrowser}}
-	<a {{action 'trackClick' 'share' 'line'}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
+	<a {{action 'trackClickAndFollow' lineShare 'share' 'line' true true}} href={{lineShare}}>{{svg 'line' role='img' class='icon line'}}</a>
```
rogatty: `trackClickAndFollow` is no longer a thing. See https://github.com/Wikia/mercury/commit/50ee2963435830ed86ef5937d1460c2e28b4e4fc

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -44,6 +44,7 @@ App.ArticleRoute = Em.Route.extend({
 
 	afterModel: function (model: any) {
 		this.controllerFor('application').set('currentTitle', model.get('title'));
+		this.controllerFor('application').set('wikiaHomePath', model.get('wikiaHomePath'));
```
rogatty: `this.controllerFor('application').setProperties({...})`

rogatty:
Why do you set property on every article load if it's just a hardcoded string? How is it better than putting it straight in the `site-head.hbs`?

front/scripts/main/models/ArticleModel.ts
```diff
@@ -51,6 +51,7 @@ App.ArticleModel = Em.Object.extend({
 	user: null,
 	users: [],
 	wiki: null,
+	wikiaHomePath: '//wikia.com',
```
hakubo: I don't think this link should be a part of ArticleModel
Have we considered creating WikiModel for such data?

hakubo: maybe consider storing this data
into M.prop via ember-main template?

not perfect either but at least we omit this circle 
ArticleModel -> ApplicationController -> SiteHead

server/views/_layouts/ember-main.hbs
```diff
@@ -42,6 +42,7 @@
 			M.prop('translations', {{{json (translations wiki.language.user ns='main')}}});
 			M.prop('weppyConfig', {{{json weppyConfig}}});
 			M.prop('wikiaLanguage', '{{{wiki.language.user}}}');
+			M.prop('wikiaHomepageUrl', '//wikia.com');
```
lizlux: We should include the protocol here since this is a link, not an asset we're trying to load. https://www.virendrachandak.com/techtalk/protocol-relative-urls/

front/templates/main/components/site-head.hbs
```diff
@@ -9,6 +9,8 @@
 {{else}}
 	{{login-icon}}
 {{/if}}
-<div class='site-logo' {{action 'trackClick' 'wordmark'}}>
-    {{svg 'wikia-logo' role='img' class='icon wikia-logo'}}
+<div class='site-logo' >
+	<a {{action 'trackClick' 'wordmark' preventDefault=false}} href={{wikiaHomepageUrl}}>
```
rogatty: Why not `href="wikia.com"`? What's the point of passing it through multiple levels if it's a constant? It only introduces complexity.

hakubo: actually that makes sense to me as well

since we don't use it in any other place
we might as well put it in template as well

and when need arises then abstract it away

katanka: I wanted to avoid a "magic number" type situation. I agree it doesn't have immediate use right now, but I figured I'd put it with the other constants in case it gets used somewhere else in the future.

rogatty: http://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it & http://en.wikipedia.org/wiki/KISS_principle ;)

I would simplify it until abstraction is actually needed.

front/templates/main/components/site-head.hbs
```diff
@@ -9,6 +9,8 @@
 {{else}}
 	{{login-icon}}
 {{/if}}
-<div class='site-logo' {{action 'trackClick' 'wordmark'}}>
-    {{svg 'wikia-logo' role='img' class='icon wikia-logo'}}
+<div class='site-logo' >
```
lizlux: There's extra white space on this line (before the `>`) and the whitespace on the lines below is inconsistent. Should be tab indented.

mklucsarits: I think we're also using double quotes now for HTML attributes in Handlebars templates. Can the single quotes be updated? (But leave single quotes for helper parameters.)

lizlux: Ah yes, https://github.com/Wikia/guidelines/tree/master/Handlebars#quotes. Thanks @mklucsarits

front/templates/main/components/site-head.hbs
```diff
@@ -1,14 +1,16 @@
...
+	<figure class="avatar">
 		<img src={{currentUser.model.avatarPath}} alt={{currentUser.model.name}}>
 	</figure>
 {{else}}
 	{{login-icon}}
 {{/if}}
-<div class='site-logo' {{action 'trackClick' 'wordmark'}}>
-    {{svg 'wikia-logo' role='img' class='icon wikia-logo'}}
+<div class="site-logo">
+	<a {{action 'trackClick' 'wordmark' preventDefault=false}} href="http://wikia.com">
```
rogatty: Should this be `http://www.wikia.com`?
/cc @tedgill

katanka:
Ok, I added www for consistency.

lizlux:
LGTM!

PR #900, opened 2015-06-09
SOC-684 | Registration screen

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-684
* Basic logic behind registration screen
* Registration request uses Helios directly via ajax (XHR)
* Email validation ready
* General form validation ready
* Some UI changes to be able to fit in error messages on small screen
* [not ready] - account language selection will be implemented in separate ticket

@mklucsarits @rafalkalinski Please CR


server/views/_layouts/auth.hbs
```diff
@@ -27,6 +27,11 @@
 		{{/unless}}
 
 		{{#unless noScripts}}
+		<script>
+			var translations = {{{json (translations language ns='auth-front')}}},
```
kvas-damian: Is there better way to pass those variables to frontend JS?
I'd like to avoid loading all mercury scripts to make that happen.

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	private urlEncode(object: Object): string {
+		var encodedString: string = '';
+		for (var prop in object) {
+			if (object.hasOwnProperty(prop)) {
+				if (encodedString.length > 0) {
+					encodedString += '&';
+				}
+				encodedString += encodeURI(prop + '=' + object[prop]);
```
hakubo: I think you want to use encodeURIComponent here
we also prefer I think Object.keys

how about:
```
return Object.keys(object).map((key: string) =>
	`${encodeURIComponent(key) }=${encodeURIComponent(object[key]) }`
).join('&');

File: front/scripts/auth/signup/SignupForm.ts [hunk 3]
```diff
@@ -0,0 +1,138 @@
...
+			}
+		}
+		return encodedString;
+	}
+
+	private clearValidationErrors(): void {
+		var errorNodes: NodeList = this.form.querySelectorAll('.error');
+
+		Array.prototype.forEach.call( errorNodes, function( node: HTMLElement ) {
+			if (node.tagName == 'INPUT') {
```
hakubo: ===

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+			data: HeliosRegisterInput = {
+				username: (<HTMLInputElement> formElements.namedItem('username')).value,
+				password: (<HTMLInputElement> formElements.namedItem('password')).value,
+				email: (<HTMLInputElement> formElements.namedItem('email')).value,
+				birthdate: (<HTMLInputElement> formElements.namedItem('birthdate')).value
+				// TODO add langCode
+			},
+			submitButton: HTMLElement = <HTMLElement> this.form.querySelector('button');
+
+		submitButton.setAttribute('disabled', 'disabled');
```
hakubo: shouldn't that be set to true?
```
submitButton.disabled = true;
```

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			if (xhr.status !== 200) {
+				this.displayGeneralError();
+				return;
+			}
+
+			// all is well
+			if(xhr.readyState === 4) {
+				alert('signed in correctly');
```
hakubo: alert, alert !

front/styles/auth/module/_forms.scss
```diff
@@ -1,12 +1,16 @@
...
+		line-height: .75rem;
+		margin-bottom: 0;
+		padding-left: 0;
+		padding-right: 0;
 	}
 }
 
 .input-container {
-	padding-bottom: 1rem;
+	min-height: 62px;
```
hakubo: why 62?
can we even out numbers?
or can we add some comments?

server/views/_layouts/auth.hbs
hakubo: do you have any ideas on how to do it better?
Maybe we could implement such solution?

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+				encodedString += encodeURI(prop + '=' + object[prop]);
+			}
+		}
+		return encodedString;
+	}
+
+	private clearValidationErrors(): void {
+		var errorNodes: NodeList = this.form.querySelectorAll('.error');
+
+		Array.prototype.forEach.call( errorNodes, function( node: HTMLElement ) {
```
mklucsarits: Remove extra whitespace in parentheses.

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+				node.classList.remove('error');
+			} else {
+				node.parentNode.removeChild( node );
+			}
+		});
+		this.generalErrorShown = false;
+	}
+
+	private displayValidationErrors(errors: Array<HeliosError>): void {
+		Array.prototype.forEach.call( errors, (function( err: HeliosError ) {
```
mklucsarits: Remove extra whitespace.

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+		var errorNode: HTMLElement = this.createValidationErrorHTMLNode(err.description),
+			input: HTMLFormElement = <HTMLFormElement> this.form.elements[err.additional.field];
+		input.parentNode.appendChild(errorNode);
+		input.classList.add('error');
+	}
+
+	private displayGeneralError(): void {
+		if (!this.generalErrorShown) {
+			var errorNode: HTMLElement = this.createValidationErrorHTMLNode('registration_error');
+			this.form.insertBefore(errorNode, this.form.querySelector('#signupNewsletter').parentNode);
```
mklucsarits: Let's stick with getElementById for id selectors, because it's faster.

front/scripts/auth/signup/SignupForm.ts
mklucsarits: Yes I agree, this ^ will set the property of the element, and not just the attribute, which can cause problems.

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+				birthdate: (<HTMLInputElement> formElements.namedItem('birthdate')).value
+				// TODO add langCode
+			},
+			submitButton: HTMLElement = <HTMLElement> this.form.querySelector('button');
+
+		submitButton.setAttribute('disabled', 'disabled');
+		submitButton.classList.add('on');
+		this.clearValidationErrors();
+
+		xhr.onreadystatechange = (function() {
```
mklucsarits: Can XMLHttpRequest.onload be used instead? Then we don't need to check ready state at all.

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			// all is well
+			if(xhr.readyState === 4) {
+				alert('signed in correctly');
+			}
+		}).bind(this);
+
+		xhr.open('POST', this.form.action, true);
+		xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
```
mklucsarits: I'm not sure if it's needed, because the browser might be smart enough to handle it, but we might need ```; charset=UTF-8``` on the end of this header?

server/facets/auth/login.ts
```diff
@@ -43,12 +43,13 @@ interface LoginViewContext {
...
 }
 
-function getLoginContext (redirect: string): LoginViewContext {
+function getLoginContext (request: Hapi.Request, redirect: string): LoginViewContext {
 	return <LoginViewContext> {
 		title: 'auth:login.login-title',
 		headerText: 'auth:login.welcome-back',
 		footerCallout: 'auth:login.register-callout',
 		footerCalloutLink: 'auth:login.register-now',
+		language: request.server.methods.i18n.getInstance().lng(),
```
mklucsarits: This property needs to be added to LoginViewContext interface.

server/views/signup.hbs
```diff
@@ -15,15 +15,14 @@
 			<span class="input-icon password-toggler"></span>
 		</div>
 		<div class="input-container">
-			<input type="date" name="date" id="signupBirthDate" required />
+			<input type="date" name="birthdate" id="signupBirthDate" required />
 			<label class="floating-label" for="signupBirthDate">Birthdate (MM / DD / YY)</label>
 		</div>
-		<input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
```
mklucsarits: Is crumb no longer needed?

hakubo:
I see loads of code added but no tests. Do you plan to write some?

server/views/_layouts/auth.hbs
kvas-damian: First solution that comes to my mind is parsing front JSes and access them by lang parameter in URL. Not sure if it isn't too hacky.

server/views/signup.hbs
kvas-damian: We make requests directly to Helios which doesn't accept ```crumb```, that's why I removed it. 
Additionally I don't see a reason to protect Register form against CSRF attacks.

front/scripts/auth/signup/SignupForm.ts
kvas-damian: This link suggest that it's not required: http://stackoverflow.com/a/24415321

front/scripts/auth/signup/SignupForm.ts
kvas-damian: Handling success response is out of scope for this ticket. Added TODO to not miss that in future.

server/facets/auth/login.ts
kvas-damian: Fixed!

kvas-damian:
@hakubo we plan to cover register flow with selenium test when functionality is ready. 
Do you think about any particular parts of code that could be covered by unit tests?

hakubo:
SignupForm could use some unit test I think

front/locales/de/auth-front.json
```diff
@@ -0,0 +1,7 @@
+{
```
mklucsarits: In my opinion I don't think there's a good reason to create an auth-front translation group. We're already moving, for example, the log-in handling purely to the front-end, so for all intents and purposes, auth translations already are for front only. I don't see the need for a separation.

kvas-damian: There're a lot of messages that we use only on backend - field labels and CTA messages. The reason why I split those files is bandwidth savings. We're going to introduce even more translations on reset password, facebook connect and other pages.

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,128 @@
...
+			data: HeliosRegisterInput = {
+				username: (<HTMLInputElement> formElements.namedItem('username')).value,
+				password: (<HTMLInputElement> formElements.namedItem('password')).value,
+				email: (<HTMLInputElement> formElements.namedItem('email')).value,
+				birthdate: (<HTMLInputElement> formElements.namedItem('birthdate')).value
+				// TODO add langCode
+			},
+			submitButton: HTMLElement = <HTMLElement> this.form.querySelector('button'),
+			enableSubmitButton = () => {
+				submitButton.disabled = false;
```
rafalkalinski: why not to remove disabled property?

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -0,0 +1,128 @@
...
+		return Object.keys(object).map((key: string) =>
+				`${encodeURIComponent(key)}=${encodeURIComponent(object[key])}`
+		).join('&');
+	}
+
+	private clearValidationErrors(): void {
+		var errorNodes: NodeList = this.form.querySelectorAll('.error');
+
+		Array.prototype.forEach.call( errorNodes, (node: HTMLElement): void => {
+			if (node.tagName === 'INPUT') {
```
rafalkalinski: How about other fields? Probably we will never need textarea there, but select is imaginable.

front/scripts/auth/signup/SignupForm.ts
kvas-damian: @rafalkalinski is there any difference on those solutions?

front/scripts/auth/signup/SignupForm.ts
kvas-damian: I agree on that, on the other hand we still don't know how we'll handle error handling on birthdate field. I'd postpone changing that until we find it useful in our use cases.

kvas-damian:
@hakubo added unit tests for SignupForm and UrlHelper

hakubo:
Thanks!

front/scripts/auth/common/UrlHelper.ts
```diff
@@ -0,0 +1,7 @@
+class UrlHelper {
+	public urlEncode(object: Object): string {
+		return Object.keys(object).map((key: string) =>
```
rogatty: `: string`

front/scripts/auth/signup/SignupForm.ts
rafalkalinski: ok

rafalkalinski:
LGTM

mklucsarits:
LGTM

PR #901, opened 2015-06-09
HG-688 - Rendering twice when changing routes

katanka:
I created a new test route for debugging purposes, needs to be removed before merging.


kenkouot:
Needs cleanup but I'm satisfied with the solution. Basically, Ember seems to lazily delete it's instances, and we need this one deleted when the view is actually removed. Can you remove my state check in the observer function as we discussed since this fixes that?

hakubo:
And why do we need it deleted?
This seems like hacking around not understanding how Ember works.

Maybe we should just do stuff differently?

rogatty:
Would https://github.com/Wikia/mercury/pull/805 fix this issue?

```
willDestroyElement: function (): void {
	this.destroy();
}
```
seems risky. This is obviously a hack and I wouldn't be surprised if it stops working in the next version of Ember or introduces some side effects.

kenkouot:
This actually cleans up a hack that I put in place to unblock INT a few weeks ago.

@hakubo I'm also not sure that this is actually hacky, but maybe you just don't like the implementation? This is from the Ember official docs (http://emberjs.com/api/classes/Ember.View.html#event_willDestroyElement):
```
Called when the element of the view is going to be destroyed. Override this function to do any teardown that requires an element, like removing event listeners.

Please note: any property changes made during this event will have no effect on object observers.
```

Furthermore, the (Ember.View#destroy)[http://emberjs.com/api/classes/Ember.View.html#method_destroy] method is there in the documentation and not discouraged from being used to manually (and specifically) clean up a view.

It seems these are valid uses of functions.

rogatty:
IMO it looks like we're intruding the natural lifecycle of Ember views. If Ember does _things_ between `willDestroyElement` event and actually destroying it then we will prevent them from happening, right?

kenkouot:
Excellent points, after doing more research I agree and think we can improve our usage of the API. Using the hook to unbind elements is correct. Using `destroy` is incorrect, even though killing the events is a side effect of explicitly destroying an element.

It looks like Ember marks a instance for deletion, then at a later time it's instance object & element are destroyed. Now, the `willDestroyElement` hook is explicitly provided as a safe, designated place for teardown before the destruction lifecycle is fully realized.

Now, @katanka is using `View#destroy` ultimately to unbind the event handlers, but really, we should just be unbinding the event handlers here, which is a valid use of the `willDestroyElement` hook based on the documentation.

I asked Erik & Stefen from the Ember Core team about this issue as well and they generally discourage use of the `destroy` method explicitly.

katanka:
@rogatty it looks like #805 fixes the issue, but I don't really understand why

front/templates/main/article.hbs
```diff
@@ -38,13 +38,13 @@
 					{{/each}}
 				{{/collapsible-menu}}
 			{{/if}}
-			{{#view 'article-content' class='mw-content'}}
+			{{#article-content class='mw-content'}}
```
hakubo: does it have to be a component? since all we need here is setting some class 
how about just use div for it?

rogatty:
`ArticleContentView.ts` isn't needed anymore, we should delete it. It was wrapping content in the `<article>` tag, should we put it manually in `article.hbs`? It probably has some meaning for SEO.

kenkouot:
Can you remove this check and verify that all the conditions still work after this:
https://github.com/Wikia/mercury/blob/dev/front/scripts/main/views/ArticleView.ts#L129

rogatty:
Should we wrap the content in `<article class="article-content mw-content"></article>` element? Because this PR removes it.

rogatty:
If what @kenkouot asked about is still working then it LGTM.

hakubo:
LGTM

katanka:
I'm looking to add some sort of automated test, because I think it might be easy to inadvertently re-create this bug in the future.

PR #902, opened 2015-06-09
HG-736 Add in support for forwarding headers

kenkouot:
This PR is part of a fix for HG-736. At a high level, there is a race between our purges in Celery, the MW edit API and refreshing/viewing an article in Mercury. In order to prevent a user who has just edited from getting stale content (and also for logged in users), we're going to use the same mechanism MW does and simply ignore cache for requests in which the `wikicities_session` cookie is present.

Previously, calls to the internal Varnish were not setting any cookies. This sets the user session cookie if it is available.

### DIFF
--- a/server/facets/api/article.ts
+++ b/server/facets/api/article.ts
@@ -30,21 +30,23 @@ function isRequestForRandomTitle (query: any): boolean {
  * @param result
  */
 export function get (request: Hapi.Request,  reply: any): void {
-	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host);
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host),
+		params = {
+			wikiDomain: wikiDomain,
+			title: request.params.articleTitle,
+			redirect: request.params.redirect
+		},
+		article = new Article.ArticleRequestHelper(params);
 
 	if (isRequestForRandomTitle(request.query)) {
-		Article.getArticleRandomTitle(wikiDomain, (error: any, result: any): void => {
+		article.getArticleRandomTitle((error: any, result: any): void => {
 			var wrappedResult = wrapResult(error, result);
 			Caching.setResponseCaching(reply(wrappedResult), randomTitleCachingTimes);
 		});
 		return;
 	}
 
-	Article.getData({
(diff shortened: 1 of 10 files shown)


lizlux:
`server/facets/api/article.ts(34,11): 2304 Cannot find name 'ArticleRequestParams'.
server/facets/showArticle.ts(21,11): 2304 Cannot find name 'ArticleRequestParams'.`

server/facets/api/article.ts
```diff
@@ -30,21 +30,31 @@ function isRequestForRandomTitle (query: any): boolean {
...
-	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host);
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host),
+		params: ArticleRequestParams = {
+			wikiDomain: wikiDomain,
+			title: request.params.articleTitle,
+			redirect: request.params.redirect
+		},
+		article: Article.ArticleRequestHelper;
+
+		if (request.state.wikicities_session) {
```
lizlux: I wish this code was more self-documenting, but I understand why it isn't - b/c it is taking this pattern from MW. Let's add a comment here so we know why we're adding this cookie header to the request.

lizlux:
This would probably be a good time to add some unit tests.

server/facets/api/article.ts
rogatty: Indentation is off here.

rogatty:
@lizlux: I don't get [these warnings](https://github.com/Wikia/mercury/pull/902#issuecomment-110528105). Do you, still?

`ArticleRequestParams` is defined in `mercury-server.d.ts`.
Both `server/facets/api/article.ts` and `server/facets/showArticle.ts` refer to `lib/Article.ts` which in turn refers to `mercury-server.d.ts`.

Maybe we should add this last reference to both `article.ts` and `showArticle.ts`? Would be nice to standarize approach and add it to the guidelines. I work on this code for some time now and I'm still not sure when do I have to add references :see_no_evil:

lizlux:
@rogatty, not seeing them in dev, so you're good. 

I'd be all for adding this to the guidelines.

PR #903, opened 2015-06-09
HG-686 disabled caching for random page querys

katanka:
This fixes this issue where the random page request returned the same page repeatedly. Varnish caches by URL, and the random page request url is the same for every call, so it was just returning one response for all requests. By disabling caching for random page responses, it now returns random pages.

### DIFF
--- a/server/facets/api/article.ts
+++ b/server/facets/api/article.ts
@@ -35,7 +35,7 @@ export function get (request: Hapi.Request,  reply: any): void {
 	if (isRequestForRandomTitle(request.query)) {
 		Article.getArticleRandomTitle(wikiDomain, (error: any, result: any): void => {
 			var wrappedResult = wrapResult(error, result);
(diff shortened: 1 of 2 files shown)


kenkouot:
LGTM!

server/facets/api/article.ts
```diff
@@ -35,7 +35,7 @@ export function get (request: Hapi.Request,  reply: any): void {
 	if (isRequestForRandomTitle(request.query)) {
 		Article.getArticleRandomTitle(wikiDomain, (error: any, result: any): void => {
 			var wrappedResult = wrapResult(error, result);
-			Caching.setResponseCaching(reply(wrappedResult), randomTitleCachingTimes);
```
rogatty: `randomTitleCachingTimes` is no longer used so it can be removed.

tomasznapieralski:
Unfortunately on prod and sandbox problem still occurs due to caching issue.

PR #905, opened 2015-06-10
SOC-698 AJAX login initial commit

mklucsarits:
This moves login from the back end to the front end, using an AJAX request directly to the log-in API.

SOC-515 is already merged here.
@kvas-damian @bkoval @kenkouot


front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
+interface LoginCredentials {
+	username: string;
+	password: string;
+}
+
+interface LoginResponse {
+	'user_id': string;
```
lizlux: Just a style thing - I think it's clearer without the quotes.

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+		this.form.addEventListener('submit', (event: Event): void => {
+			event.preventDefault();
+			this.onSubmit();
+		});
+	}
+
+	public onSubmit (): void {
+		var xhr = new XMLHttpRequest(),
+			credentials: LoginCredentials = this.getCredentials(),
+			_this = this;
```
lizlux: JS guidelines say to use `self` for referencing context var. https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#storing-context-in-a-local-variable

front/scripts/auth/login/Login.ts
mklucsarits: I copied and pasted this from the back end login.ts file. I'll remove the quotes unless @kenkouot added them for a certain reason.

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+		};
+
+		xhr.open('post', this.form.action, true);
+		xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
+		xhr.send('username=' + encodeURIComponent(credentials.username) + '&password=' + encodeURIComponent(credentials.password));
+	}
+
+	private getCredentials (): LoginCredentials {
+		return {
+			username: this.form.elements['username'].value,
```
lizlux: Is the reason you used this syntax b/c typescript complains about `this.form.elements.username`? The way I've worked around this in the past is by defining an interface for the possible form elements, and casting a variable to represent those elements. Example here: https://github.com/Wikia/mercury/blob/SOC-480/front/scripts/auth/common/BirthdateInput.ts#L20
I'd prefer this solution b/c it's more explicit and it follows our JS guidelines for dot notation https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#objects.

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+			password: this.form.elements['password'].value
+		};
+	}
+
+	private appendCachebuster (path: string): string {
```
lizlux: Is this function now repeated here and in auth utils?

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+			// Path already contains a cb param in the querystring
+			return path;
+		}
+
+		path += 'cb=' + Math.floor(Math.random() * 10000);
+		return path;
+	}
+
+	private displayError (messageKey: string): void {
+		var errorElement = <HTMLElement> this.form.querySelector('small.error');
```
lizlux: Is there a reason for including the tag name in the selector? My understanding is this is less performant than just `.error`. Didn't set up a test or anything though.

server/views/_layouts/auth.hbs
```diff
@@ -27,6 +27,11 @@
 		{{/unless}}
 
 		{{#unless noScripts}}
+		<script>
+			var translations = {{{json (translations language ns='auth')}}},
```
lizlux: Can we namespace these vars? Or better yet, use `M.prop`?

server/views/login.hbs
```diff
@@ -12,10 +12,10 @@
 		</div>
 		<input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
 		<button id="loginSubmit" type="submit" disabled>{{i18n 'auth:login.submit-text'}}</button>
-		{{#if formErrorKey}}
 		<small class="error">
-			{{i18n formErrorKey}}
+			{{#if formErrorKey}}
```
kvas-damian: While we're removing login handling from hapi I don't think we still need this code.

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+			}
+		};
+
+		xhr.onerror = function (): void {
+			_this.displayError('common.server-error');
+		};
+
+		xhr.open('post', this.form.action, true);
+		xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
+		xhr.send('username=' + encodeURIComponent(credentials.username) + '&password=' + encodeURIComponent(credentials.password));
```
kvas-damian: we could share encoding method from registration request: https://github.com/Wikia/mercury/pull/900/files#diff-f82ce89c093e08b6a3fc530263679c2cR26

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		query = path.match(/\?.+/);
+
+		if (query === null) {
+			path += '?';
+		} else if (query[0].match(/cb=/) === null) {
+			path += '&';
+		} else {
+			// Path already contains a cb param in the querystring
+			return path;
```
kvas-damian: Shouldn't we put new ```cb``` value in this case?

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+			var response: LoginResponse;
+
+			if (this.status !== 200) {
+				return _this.displayError(this.status === 401 ? 'login.wrong-credentials' : 'common.server-error');
+			}
+
+			response = JSON.parse(this.responseText);
+
+			if (response.error) {
+				// Helios may return an error even if the request returns a 200
```
kvas-damian: What kind of error is this? Shouldn't we fix that in Helios?

server/views/_layouts/auth.hbs
kvas-damian: @mklucsarits I know you have different opinion on that, but I'm more happy with splitting frontend and backend translations: https://github.com/Wikia/mercury/pull/900/files#r32215087

server/facets/auth/login.ts
```diff
@@ -110,52 +42,3 @@ export function get (request: Hapi.Request, reply: any): void {
...
-		// set session cookie via hapi-auth-cookie
-		request.auth.session.set({
-			'sid'  : response.access_token
-		});
-
-		// Set cookie TTL for "remember me" period of 6 months
-		// TODO: Helios service should control the length of auth session
-		request.auth.session.ttl(ttl);
-
-		reply.state('wikicitiesUserID', response.user_id, {ttl: ttl});
```
kvas-damian: Don't we need this cookie?

server/facets/auth/login.ts
```diff
@@ -110,52 +42,3 @@ export function get (request: Hapi.Request, reply: any): void {
...
-			// Always set the correct status code
-			}).code(err.output.statusCode);
-		}
-
-		// set unencrypted cookie that can be read by all apps (i.e. MW and Mercury) HG-631
-		reply.state('access_token', response.access_token, {ttl: ttl});
-
-		// set session cookie via hapi-auth-cookie
-		request.auth.session.set({
-			'sid'  : response.access_token
```
kvas-damian: does session still works when we remove ```sid``` or do we need to finish SOC-798?
I'm asking because we plan to release logged in state on few wikis when SOC-705 is mearged.
Maybe we should postpone merging this PR till we solve SOC-798, what do you think?

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+			} else {
+				window.location.href = _this.redirect;
+			}
+		};
+
+		xhr.onerror = function (): void {
+			_this.displayError('common.server-error');
+		};
+
+		xhr.open('post', this.form.action, true);
```
kvas-damian: While we set action to empty string in view I don't think we hit right backed:
https://github.com/Wikia/mercury/blob/SOC-698/server/views/login.hbs#L2

What do you think about putting helios URL in form action? This way we could fall back to helios standard request when js fails.

server/views/login.hbs
```diff
@@ -12,10 +12,10 @@
 		</div>
 		<input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
```
kvas-damian: Do we need this crumb when we hit Helios?

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+			this.onSubmit();
+		});
+	}
+
+	public onSubmit (): void {
+		var xhr = new XMLHttpRequest(),
+			credentials: LoginCredentials = this.getCredentials(),
+			_this = this;
+
+		xhr.onload = function (): void {
```
bkoval: I guess we could just use .bind(this) to get rid of the _this variable.

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+class Login {
+	credentials: LoginCredentials;
+	form: HTMLFormElement;
+	redirect: string;
+
+	constructor (form: Element, redirect: string = '') {
+		this.form = <HTMLFormElement> form;
+		this.redirect = this.appendCachebuster(redirect);
```
bkoval: Could we naming sth like getCacheBustedRedirectUrl or sth, so it's clear this method returns a cachebusted url?

front/scripts/auth/login/Login.ts
bkoval: Ping @ArturKlajnerok are 200s with errors still getting returned by Helios? If so, we should create a ticket to deal with this situation.

front/scripts/auth/login/Login.ts
bkoval: Also, I'm in for abstracting that into a seperate method. We could create a Helios class on the Front-End side as well, as we need to store urls and url builders somewhere :)

front/scripts/auth/login/Login.ts
bkoval: Yeah, I like making use of casting and interfaces here.

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	private appendCachebuster (path: string): string {
+		var query: Array<string>;
+
+		if (typeof path !== 'string' || path === '') {
+			path = '/';
+		}
+
+		query = path.match(/\?.+/);
```
bkoval: Let's move this regex to a named variable so we know what exactly it is used for :)

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		if (query === null) {
```
bkoval: How about using front/scripts/baseline/mercury/utils/buildUrl.ts for url handling?

front/scripts/auth/login/LoginValidator.ts
```diff
@@ -7,9 +7,9 @@ class LoginValidator {
 	loginSubmit: HTMLButtonElement;
 
 	constructor() {
-		this.loginUsername = <HTMLInputElement> window.document.querySelector('#loginUsername');
-		this.loginPassword = <HTMLInputElement> window.document.querySelector('#loginPassword');
-		this.loginSubmit = <HTMLButtonElement> window.document.querySelector('#loginSubmit');
+		this.loginUsername = <HTMLInputElement> window.document.getElementById('loginUsername');
```
bkoval: What's the reason of the change here?

server/views/login.hbs
bkoval: Yeah, this should just be required by hapi to prevent XSS, Helios should be just fine without it.

front/scripts/auth/login/Login.ts
mklucsarits: I can make this change, but I don't necessarily think it's a good idea. I think this is one of those cases where TypeScript works against us instead of for us.

lizlux: Agreed it's a bummer, but given that there's a work-around, I'd prefer sticking with our coding conventions.

front/scripts/auth/login/Login.ts
mklucsarits: Because "error" is a really generic class name that we may end up using on other elements, such as a container div. ```small.error``` corresponds to error messages as specified by the style guide and nothing else.

server/views/_layouts/auth.hbs
mklucsarits: @lizlux Yes, we could use M.prop once we include Mercury baseline as a dependency for auth pages.

@kvas-damian My opinion on that was based on the assumption that once we migrate log-in and sign-up code to the front end, to allow ajax requests, there would no longer be any back end translations for auth, making the distinction pointless. But I might be (and probably am) wrong.

kvas-damian: From my perspective translations on backend still have some advantages in that situation:
* performance - we don't need transfer and translate in the browser
* accessibility - I hope our auth pages will be able to login and register without js - user will submit form to helios, then helios authorizes that user. Backend translations allows us to display forms correctly even when js is disabled.

front/scripts/auth/login/Login.ts
mklucsarits: I'll add a comment instead. I don't think it's a good idea to set one-time use variables.

front/scripts/auth/login/Login.ts
mklucsarits: Maybe. I wrote this with the assumption that the existing ```cb``` given in the redirect is the one intended to be used.

front/scripts/auth/login/Login.ts
Grunny: You can also just the fat arrow to maintain the scope, i.e. `(): void => {`.

server/facets/auth/login.ts
mklucsarits: This will have to be set another way. This is an HTTP-only cookie, so it can't be set by JS on the front end.

server/facets/auth/login.ts
mklucsarits: Yeah, I think SOC-798 should be merged first.

server/facets/auth/login.ts
mklucsarits: This cookie was originally set for session interaction between Mercury and Wikia/app. However, it's no longer needed for that purpose, because sessions are being handled differently now by Helios.

But... this cookie's value was also being used to store the current user's ID between visits, for Mercury's logged-in state. The ```wikicitiesUserID``` cookie is/was HTTP-only, so it couldn't have been set by JS on the front-end. So we'll still need a cookie (or some way) to get the current user ID between visits to Mercury. I haven't thought of the best way to do that yet, whether it's with a new cookie storing only the user ID, or another method.

front/scripts/auth/login/Login.ts
kvas-damian: We're not manipulating changing ```cb``` value when linking to auth pages, so we can end up with redirect to the same URL that use came from (when user was on a page that had ```cb``` in URL). I'm not sure about current caching rules, but the same URL can cause serving anon page from browser cache.

rafalkalinski: timestamp is set as cb value, and it is set even if there was one before

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,106 @@
...
+					category: 'user-login-mobile',
+					label: 'login-success'
+				});
+
+				window.location.href = this.redirect;
+			}
+		};
+
+		xhr.onerror = (): void => {
+			M.track({
```
bkoval: 3 calls for tracking here, maybe we could abstract tracking? Or make tracking.ts public and abstract it there?

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,106 @@
...
+			});
+
+			this.displayError('common.server-error');
+		};
+
+		xhr.open('post', this.form.action, true);
+		xhr.setRequestHeader('Content-type', 'application/x-www-form-urlencoded');
+		xhr.send((new UrlHelper()).urlEncode(postData));
+
+		event.preventDefault();
```
bkoval: Would it be possible to move it up to the top of the handler? :)

server/views/login.hbs
```diff
@@ -10,12 +10,7 @@
...
 			<a class="forgotten-password" href="{{{forgotPasswordHref}}}" tabindex="-1">{{i18n "auth:login.forgot-password"}}</a>
 		</div>
-		<input type="hidden" name="crumb" id="crumb" value="{{crumb}}" />
 		<button id="loginSubmit" type="submit" disabled>{{i18n 'auth:login.submit-text'}}</button>
-		{{#if formErrorKey}}
-		<small class="error">
-			{{i18n formErrorKey}}
-		</small>
-		{{/if}}
+		<small class="error"></small>
```
bkoval: Just thinking if we should keep this span here or just add it if there's an error :)

bkoval:
LGTM :) :package:

server/facets/auth/login.ts
```diff
@@ -1,82 +1,14 @@
...
-	'token_type': string;
-	'expires_in': string;
-	'error'?: string;
-	'error_description'?: string;
-}
-
 interface LoginViewContext extends authView.AuthViewContext {
 	headerText: string;
 	forgotPasswordHref?: string;
 	formErrorKey?: string;
```
mklucsarits: Is ```formErrorKey``` still needed?

server/facets/operations/prepareArticleData.ts
```diff
@@ -47,7 +47,7 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 		result.weppyConfig.samplingRate = result.queryParams.buckySampling / 100;
 	}
 
-	result.userId = request.state.wikicitiesUserID ? request.state.wikicitiesUserID : 0;
+	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
```
mklucsarits: :+1:

server/routes.ts
```diff
@@ -108,39 +106,61 @@ unauthenticatedRoutes = [
...
 		method: 'POST',
-		path: '/login',
-		handler: require('./facets/auth/login').post
-	},
+		path: '/editorPreview',
+		handler: require('./facets/editorPreview')
+	}
+];
+
+authenticatedRoutes = [
```
mklucsarits: ```authenticatedRoutes``` implies the user is authenticated before accessing these. I believe that's how @kenkouot originally intended it to be used. But now it contains routes related to authenticating a user.

server/facets/auth/authView.ts
```diff
@@ -79,6 +79,18 @@ module authView {
...
 		};
 	}
+
+	export function validateRedirect (request: Hapi.Request, reply: any): any {
+		var queryRedirectUrl = authView.getRedirectUrl(request);
+
+		if (request.query.redirect && queryRedirectUrl !== request.query.redirect) {
+			request.url.query.redirect = queryRedirectUrl;
+			request.url.search = null;
+			return reply.redirect(request.url.format()).takeover();
```
mklucsarits: Originally we were going to add a new cachebuster to the redirect, even if the ```cb``` already existed and needed to be replaced. Is that no longer the case? Where is the cachebuster handled now?

front/scripts/auth/login/Login.ts
```diff
@@ -0,0 +1,122 @@
...
+		this.redirect = redirect || '/';
+	}
+
+	public onSubmit (event: Event): void {
+		var xhr = new XMLHttpRequest(),
+			postData: LoginCredentials = this.getCredentials(),
+			submitButton: HTMLElement = <HTMLElement> this.form.querySelector('button'),
+			enableSubmitButton = () => {
+				submitButton.disabled = false;
+				submitButton.classList.remove('on');
```
mklucsarits: This CSS class doesn't appear to serve any purpose. Was it supposed to? If it was intended for styling, the ```:disabled``` pseudoselector can be used instead.

server/facets/auth/authView.ts
kvas-damian: I don't see reason to add CB here. HTML is not cached in browser, on varnish we don't serve cached content when user have ```access_token``` cookie. Is there anything else I'm not aware of?

front/scripts/auth/login/Login.ts
kvas-damian: There are two states of disabled submit button. first is regular disabled button when fields are not filed in and second is when user submits form. The latter display throbber on a button. That's why we added this class here.

server/routes.ts
kvas-damian: Basically is a matter how you interpret those variables.
When I changed it, my way of thinking was following: I need to validate if user is authenticated or not on those routes, so I need to run authentication process here. In this case it makes sense to call them ```authenticatedRoutes```. 

BTW in future we'd need more complex system for validating permissions. We're not handling only anon vs logged in but also admins, founders and different levels of permissions.

server/facets/auth/authView.ts
lizlux: Are you sure the HTML isn't cached in the browser? The MediaWiki app sets a cache buster when users log in, and when I was testing the new login page a while back I had caching problems before I added the cache buster.

kvas-damian: Yes, I'm. We disabled browser cache when rolling out logged in state in mercury.
There were at least couple reasons why we shouldn't use it:
* user can't see fresh edits
* there is no way to display different state of the same page for logged in user while it's cached in browser

BTW - I don't think MW uses browser cache directly for HTML. We use ETAGs and we're validating on varnish if page was changed since last page view.

PR #906, opened 2015-06-10
SOC-805 Create CTA to New Login Flow on Wikiamobile

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-805


front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -1,46 +1,14 @@
...
-		if (!this.newLoginEnabled) {
-			return false;
-		}
-
-		shouldRedirect = this.newLoginWhitelist.some((whitelistedDBName: string): boolean => {
-			return whitelistedDBName === dbName;
-		});
+	classNames: ['external', 'login'],
+	attributeBindings: ['href'],
+	href: Mercury.Utils.prop('loginUrl')
```
rogatty: Use `M.prop` shorthand.

server/views/_layouts/ember-main.hbs
```diff
@@ -43,6 +43,7 @@
 			M.prop('weppyConfig', {{{json weppyConfig}}});
 			M.prop('wikiaLanguage', '{{{wiki.language.user}}}');
 			M.prop('userId', {{{userId}}});
+			M.prop('loginUrl', '{{{loginUrl}}}');
```
rogatty: Does it have to be inside three brackets? I wonder if this can be used as an attack vector by preparing `request.path` (https://github.com/Wikia/mercury/pull/906/files#diff-9a2ced1d5be5f50db94aab4fd4a0029fR44).

>Handlebars HTML-escapes values returned by a {{expression}}. If you don't want Handlebars to escape a value, use the "triple-stash", {{{

@Grunny what do you think?

Grunny: @rogatty @rafalkalinski Yes, this is vulnerable to XSS, try turning this on and going to http://127.0.0.1:8000/wiki/%27%29;alert%28%270 and so needs to be escaped. Worse, since the server is outputting it rather than the client, it bypasses Chrome's XSS filter. In general, I don't like the way we build these values in a script element in the head, it's just prone to developer error and we've had a couple of vulnerabilities from it already.

server/facets/operations/prepareArticleData.ts
```diff
@@ -42,6 +43,7 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 	result.queryParams = Utils.parseQueryParams(request.query);
 	result.weppyConfig = localSettings.weppy;
 	result.userId = request.state.wikicitiesUserID ? request.state.wikicitiesUserID : 0;
+	result.loginUrl = AuthUtils.getLoginUrl(request, result);
```
bkoval: Am I wrong or is ot being set on the first pageview only? If it is, then redirect will not follow the URL when it changes on the Front-End side (e.g. when user goes to another article by clicking a link).

server/lib/AuthUtils.ts
```diff
@@ -38,3 +38,9 @@ export function getCacheBusterUrl(redirect: string): string {
 
 	return url.format(cacheBustedUrlObj);
 }
+
+export function getLoginUrl(request: Hapi.Request, result: any): string {
+	return result.wiki.enableNewAuth ?
+		'/join?redirect=' + request.server.info.uri + request.path :
```
bkoval: What about querystring? Does it get lost here? Also, I would consider building this link using a Node's URL class.

kvas-damian: I think that ```redirect``` value should be encoded.
There is separate comment about escaping it from @rogatty and @Grunny

kvas-damian:
Please resolve merge conflicts

server/lib/AuthUtils.ts
```diff
@@ -38,3 +38,18 @@ export function getCacheBusterUrl(redirect: string): string {
 
 	return url.format(cacheBustedUrlObj);
 }
+
+export function getLoginUrl(request: Hapi.Request, result: any): string {
+	if (result.wiki.enableNewAuth) {
+		var loginUrlObj = url.parse('/join');
+
+		var redirect = url.parse(request.server.info.uri + request.path);
```
bkoval: Could you move this declaration after the first one using a comma, so there's only one var statement in this method? :)

server/lib/AuthUtils.ts
```diff
@@ -38,3 +38,18 @@ export function getCacheBusterUrl(redirect: string): string {
...
+		redirect.search = querystring.stringify(request.query);
+
+		loginUrlObj.search = '?redirect=' + encodeURIComponent(url.format(redirect));
```
bkoval: Could you make sure we need the question mark at the beginning? I'm sure URL class can make it for us while formatting the url :)

server/lib/AuthUtils.ts
```diff
@@ -38,3 +38,18 @@ export function getCacheBusterUrl(redirect: string): string {
...
+		return url.format(loginUrlObj);
+	}
+	else {
+		return '/wiki/Special:UserLogin';
```
bkoval: return wikiaLoginPathname; (it's already defined at the top of the file)

front/scripts/main/views/ArticleView.ts
```diff
@@ -94,6 +94,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		this.get('controller').send('articleRendered');
 	},
 
+	refreshLoginUrl: function (): void {
+		var loginUrl = '/join?redirect=' + encodeURIComponent(window.location.href);
+		document.querySelector('.login-page-link').setAttribute('href', loginUrl);
+	},
+
+	scheduleArticleTransforms: function (): void {
```
bkoval: Hey, this code is not easily understandable for me, maybe you could add a doc / comment explaining what does it do? :)

front/scripts/main/components/SiteHeadComponent.ts
```diff
@@ -7,6 +7,7 @@ App.SiteHeadComponent = Em.Component.extend(App.TrackClickMixin, {
 	classNames: ['site-head'],
 	tagName: 'nav',
 	headroom: null,
+	loginHref: Mercury.Utils.prop('loginUrl'),
```
rogatty: Use `M.prop` shorthand.

front/scripts/main/views/ArticleView.ts
rogatty: This change is caused by badly resolved merge conflict. Look at https://github.com/Wikia/mercury/blob/dev/front/scripts/main/views/ArticleView.ts and move methods back where they belong. Thanks and sorry for the mess (yeah, it was me that moved them around).

front/scripts/main/views/ArticleView.ts
```diff
@@ -94,6 +94,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		this.get('controller').send('articleRendered');
 	},
 
+	refreshLoginUrl: function (): void {
+		var loginUrl = '/join?redirect=' + encodeURIComponent(window.location.href);
+		document.querySelector('.login-page-link').setAttribute('href', loginUrl);
```
rogatty: This should be done the Ember way, not by modifying the DOM. You can send action up to ApplicationRoute, then to ApplicationController and there change the property that would be modified in SiteHeadComponent.

I know this looks like going around but this is how Ember works - Data Down, Actions Up.

For example look at the following chain:
1. https://github.com/Wikia/mercury/blob/dev/front/scripts/main/views/ArticleView.ts#L393 (automatically bubbles up to ArticleRoute, then ApplicationRoute, see http://guides.emberjs.com/v1.12.0/images/template-guide/action-bubbling.png)
2. https://github.com/Wikia/mercury/blob/dev/front/scripts/main/routes/ApplicationRoute.ts#L133
3. https://github.com/Wikia/mercury/blob/dev/front/scripts/main/controllers/ApplicationController.ts#L65
4. https://github.com/Wikia/mercury/blob/dev/front/templates/main/application.hbs#L28

front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -1,46 +0,0 @@
...
-		'clashofclans',
-		'creepypasta',
-		'castleclash',
-		'glee',
-		'mobileregressiontesting'
-	],
-
-	click: function (): void {
-		if (this.shouldRedirectToNewLogin()) {
-			window.location.href = '/join?redirect=' + encodeURIComponent(window.location.href);
```
rogatty: Is there a good reason that you don't want to do this anymore? We wouldn't need to pass window location through the app, only react to user action (click).

rafalkalinski:
I got back to the old Ember-way solution, where the login url is set up during the click event on the front end only.

rafalkalinski:
And thanks for the feedback @rogatty

kvas-damian:
LGTM

bkoval:
LGTM :palm_tree:

rogatty:
LGTM

PR #907, opened 2015-06-11
HG-727 adding loading graphic for A/B test

lizlux:
This is the basis for an A/B test to see if engagement increases when we provide visual feedback to users via a loading graphic, instead of showing a blank page, while ember is bootstrapping. 

The JS to include via optimizely for users in the test is: 
```javascript
// with jQuery
$('.ab-test-loading-overlay').show();

// without jQuery
document.querySelector('.ab-test-loading-overlay').style.display='block';
```

@kenkouot @katanka

### DIFF
--- a/front/scripts/main/views/ArticleView.ts
+++ b/front/scripts/main/views/ArticleView.ts
@@ -87,9 +87,6 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, App.LanguagesM
 			M.trackPageView(model.get('adsContext.targeting'));
 		}
 
-		// A/B test spinner (HG-727)
-		$('.ab-test-loading-overlay').hide();
-
 		return true;
 	},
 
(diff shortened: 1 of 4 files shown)


front/scripts/main/views/ArticleView.ts
```diff
@@ -87,6 +87,9 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.ViewportMixin, App.LanguagesM
 			M.trackPageView(model.get('adsContext.targeting'));
 		}
 
+		// A/B test spinner (HG-727)
+		$('.ab-test-loading-overlay').hide();
```
rogatty: Shouldn't this be done in https://github.com/Wikia/mercury/blob/HG-727/front/scripts/main/views/ApplicationView.ts#L41?

I didn't test if it works but should make sense, right? We want to hide this overlay only on the first load, not on every article change.

lizlux: Good call @rogatty. Moved and it works.

kenkouot:
LGTM

PR #909, opened 2015-06-11
Generate TS sourcemaps to allow direct TS debugging in Chrome devtools

perjg:
These changes add generation of TypeScript sourcemaps and copying of neccesary support files to allow direct debugging of TypeScript from inside Chrome devtools. Generated for dev config only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ test/fixtures
 .idea
 /docs
 *.sublime-*
+
+front/scripts/**/*.js
+front/scripts/**/*.js.map
--- a/gulp/tasks/copy-ts-source.js
+++ b/gulp/tasks/copy-ts-source.js
@@ -1,0 +1,14 @@
+var gulp = require('gulp'),
+	environment = require('../utils/environment'),
+	paths = require('../paths').scripts.front,
+	path = require('path');
+
+gulp.task('copy-ts-source', [ 'scripts-front' ], function() {
+	var src  = path.join(paths.src, '/**/*');
+	var dest = path.join(paths.dest, process.cwd());
+
+	if (!environment.isProduction) {
+		gulp.src([src])
+			.pipe(gulp.dest(dest));
+	}
+});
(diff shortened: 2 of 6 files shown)


kenkouot:
Very cool, are there any additional instructions that would be helpful for devs to take?

perjg:
According to the docs (https://developer.chrome.com/devtools/docs/javascript-debugging#source-maps)
Sourcemaps are enabled by default from Chrome 39.

If not, turn them on in dev tools settings: Enable CSS source maps

After that it should automagically pick it up when you go to sources in dev tools.
To test, set a breakpoint anywhere in main.js, and it should jump to the correct .ts file, and you can debug from there.

.gitignore
```diff
@@ -36,3 +36,6 @@ test/fixtures
 .idea
 /docs
 *.sublime-*
+
+front/scripts/**/*.js
+front/scripts/**/*.js.map
```
rogatty: Some kind of comment here would be nice.

gulp/tasks/copy-ts-source.js
```diff
@@ -0,0 +1,14 @@
+var gulp = require('gulp'),
+	environment = require('../utils/environment'),
+	paths = require('../paths').scripts.front,
+	path = require('path');
+
+gulp.task('copy-ts-source', [ 'scripts-front' ], function() {
+	var src  = path.join(paths.src, '/**/*');
+	var dest = path.join(paths.dest, process.cwd());
```
rogatty: We use single `var` per scope.

perjg:
I can add a comment to the .gitignore file.

For var, you mean formatting it like this?

var src  = path.join(paths.src, '/**/*'),
    dest = path.join(paths.dest, process.cwd());

rogatty:
Yes and yes, thanks :)
BTW https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#variables

perjg:
Hmm... I see /* */ being used as comments in .gitignore
Is that really correct? The docs seem to indicate that only # is supported:
http://git-scm.com/docs/gitignore

PR #911, opened 2015-06-12
HG-750 - added GA tracking to random article link

katanka:
it will show up in GA as 'randomArticle'


lizlux:
Was there a reason you decided not to go with the handlebars `action` helper?

katanka:
The link already had an action, and I couldn't figure out how to make it run two actions at once.

lizlux:
Ah I see. LGTM

PR #912, opened 2015-06-12
Int-102: Add photo to Japanese wikia

mari-mcmurtrie:
https://wikia-inc.atlassian.net/browse/INT-102

add photo feature in mobile web - "add photo" icon is added to section 1+, upload a photo, add a photo to wiki text, photo name change are roughly completed. Not handling error cases at all and no validation is placed.


front/styles/main/module/_addphoto.scss
```diff
@@ -0,0 +1,95 @@
+.preview-header {
```
inez: All properties in CSS should be in alphabetical order.

front/styles/main/module/_addphoto.scss
```diff
@@ -0,0 +1,95 @@
...
+	@extend .site-head;
+	text-align: inherit;
+
+	.arrow-left {
```
inez: This is copy/paste from edit functionality? Avoid copy/paste.

front/styles/main/module/_edit.scss
```diff
@@ -51,3 +51,4 @@ textarea {
 		top: 3.5rem;
 	}
 }
+
```
inez: Is this intentional?

front/styles/main/module/index.scss
```diff
@@ -16,4 +17,4 @@
 @import 'loader';
 @import 'site-footer';
 @import 'wikia-map';
-@import 'wiki/index';
+@import 'wiki/index';
```
inez: Is this intentional?

front/scripts/main/views/ArticleView.ts
```diff
@@ -204,6 +211,19 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		});
 	},
 
+	setupAddPhotoButtons: function (): void {
+		var photo = '<div class="upload_photo"><svg class="icon" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg><input class="file-input" type="file" accept="image/*" capture="camera"/></div>';
+		this.$(':header[section]').each((i: Number, item: any): void => {
```
inez: Ok, so it seems that edit buttons and add photo buttons are added in very similar way and under same conditions which means that code could be unified so avoiding firing same selector and same loop twice would be possible.

front/scripts/main/views/ArticleView.ts
```diff
@@ -66,6 +66,10 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		return !this.get('controller.model.isMainPage') && this.get('isJapaneseWikia');
 	}),
 
+	addPhotoButtonsVisible: function () {
+		return true; //!this.get('controller.model.isMainPage') && this.get('isJapaneseWikia');
```
inez: Seems like debug code.

front/scripts/main/views/AddPhotoView.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.AddPhotoView = Em.View.extend({
+	classNames: ['addphoto-view'],
+	init: function(): void {
+		this._super();
+		Em.run.scheduleOnce('afterRender', this, (): void => {
```
inez: What's the point of this code?

front/scripts/main/routes/AddPhotoRoute.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../../../typings/ember/ember.d.ts" />
+
+'use strict';
+
+App.AddPhotoRoute = Em.Route.extend({
+	model: function(params: any): Em.Object {
+		console.log("model will be loaded!!!", params);
```
inez: debug code?

front/scripts/main/models/EditModel.ts
```diff
@@ -18,7 +18,7 @@ App.EditModel.reopenClass({
 	getEditToken: function(title: string): Em.RSVP.Promise {
 		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
 			Em.$.ajax({
-				url: '/api.php',
+				url:  Mercury.wiki.basePath + '/api.php',
```
inez: Is this intentional change?

front/scripts/main/controllers/ArticleController.ts
```diff
@@ -35,6 +35,29 @@ App.ArticleController = Em.Controller.extend({
 			});
 		},
 
+		addPhoto: function (title: string, photoData: any, sectionIndex: number): void {
+			App.VisibilityStateManager.reset();
+			this.transitionToRoute('addPhoto', title, sectionIndex).then(function(newRoute) {
```
inez: Well, I'm pretty concerned about this approach of passing photoData like that - if it's ok for photoData why wouldn't you pass all data this way (title, sectionIndex)? It breaks isolation.
Well, I bet uploading file in ember is something that someone already had to deal with - did you find any interesting code for a reference?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,203 @@
...
+
+	upload: function(model: any): Em.RSVP.Promise {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			this.temporaryUpload(model)
+				.then((addmediatemporary: any): void => {
+					if (addmediatemporary.tempName === undefined) { //we already have the file!
+						resolve({'title':addmediatemporary.title});
+					}
+					model = this.handleNameChange(model);
+					var title = encodeURIComponent(model.name);
```
inez: Please use one var statement.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,203 @@
...
+					var tempName = encodeURIComponent(addmediatemporary.tempName);
```
inez: Have jQuery handle for you encoding parameters and preparing correct URL (so no concatenating with  "+")

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,203 @@
...
+							reject(err);
+						}
+					});
+				}, (err:any) => {
+					reject(err);
+				});
+		});
+	},
+
+	upload: function(model: any): Em.RSVP.Promise {
```
inez: This method got very nested - please try to isolate to smaller logical pieces with clear responsibilities.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,203 @@
...
+		});
+	},
+
+	createFormData: function(photoData: any) {
+		var form = new FormData( document.createElement( 'form' ) );
+		form.append( 'file', photoData );
+		return form;
+	},
+
+	handleNameChange: function(model: any): any {
```
inez: Please document - hard to tell what it does

front/locales/en/main.json
```diff
@@ -64,6 +64,10 @@
 		"wikia-map-view-map": "View",
 		"search-label": "Search",
 		"cancel-label": "Cancel",
+		"addphoto-page-heading": "Adding Photo to:",
```
rogatty: Why not `add-photo-page-heading`?

front/locales/en/main.json
```diff
@@ -64,6 +64,10 @@
...
+		"addphoto-page-add-button": "ADD",
```
rogatty: In the ticket description it's "Add".

front/locales/en/main.json
```diff
@@ -64,6 +64,10 @@
...
+		"addphoto-photo-name": "Photo Name",
+		"addphoto-success": "Successfully add a photo to page __pageTitle__!",
```
rogatty: In the ticket description it's "The photo has been added to this page.".

If the description is not up to date then "add" should be "added" anyway.

front/styles/main/module/_addphoto.scss
mari-mcmurtrie: I did not know such a rule here. Fixed.

front/styles/main/module/_addphoto.scss
mari-mcmurtrie: I still have not figured out how all files are organized in this code base. I ask you a question about how I should refactor css files.

front/styles/main/module/_edit.scss
mari-mcmurtrie: it is not intentional... I remove it.

front/styles/main/module/index.scss
mari-mcmurtrie: No, not intentional. Fixed.

front/scripts/main/routes/AddPhotoRoute.ts
mari-mcmurtrie: It is debug code! I should have been more careful to check diff before I committed my changes.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: That is good point! I am gonna change these 2 setup functions into 1.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Yeah, related to your comment above, the logic to show and set up edit and add_photo buttons are very similar. This function is for the show button flag. I would like to consolidate the logic to 2 buttons. Asking Yoko about product requirement whether or not we always show edit and add buttons as pair or not.

front/scripts/main/views/AddPhotoView.ts
mari-mcmurtrie: I put it as place-holder because I thought I might have to write "afterRender" for this view, but it turned out I do not need it at all. Removed.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: ok I didn't know the rule.Fixed.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Yoko said we always have "edit" and "add photo" features together, for now. I combine the logic to determine these 2 buttons into 1 function called "editFeatureEnabled".

front/scripts/main/models/EditModel.ts
mari-mcmurtrie: Since my front-end is 'localhost:8000", if I use relative url, then my ajax request becomes "localhost:8000/api.php" :(

inez: Hey @mklucsarits! You were working on some code for producing/building Mercury URLs. What's the status of it? Do you think it would be useful here?

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: ok!

front/scripts/main/models/EditModel.ts
mklucsarits: The new function can be seen here: https://github.com/Wikia/mercury/pull/885/files#diff-1 (Mercury.Utils.buildUrl) but it's not merged yet. As it is now, my function wouldn't fix the issue with two different hosts, because it uses the current hostname as a base. My intention with that function was to avoid needing to pass in the host. We could update that function to accept a host parameter, but that kind of defeats the purpose of avoiding typing out the full URL.

rogatty: @mklucsarits it would be perfect if `buildUrl` could use MediaWiki host. It's already available in `server.mediawikiDomain` (https://github.com/Wikia/mercury/blob/dev/server/views/_layouts/ember-main.hbs#L33).

front/locales/en/main.json
mari-mcmurtrie: The mock (I attached to the task) says "ADD", that is why. Not sure which one I should follow. I ask Nick/Yoko.

front/locales/en/main.json
mari-mcmurtrie: I suck naming things. Changed to "add-photo-page-heading".

front/locales/en/main.json
mari-mcmurtrie: Ok, I talked to Yoko/Nick. I updated texts according to the final spec. Ticket has been updated to reflect the final spec, too.

front/locales/en/main.json
mari-mcmurtrie: Texts updated.

front/locales/en/main.json
rogatty: Even if it should be "ADD" in the page you can apply `text-transform: uppercase` with CSS. If designer's concept changes to lowercase we would have to update all the translations.

It's just safer to keep styling in CSS.

front/styles/main/module/_addphoto.scss
mari-mcmurtrie: Ok, so I refactored!

front/styles/main/module/edit/_addphoto.scss
```diff
@@ -0,0 +1,63 @@
...
+		background-color: #DFDFDF;
+		border: 1px solid #8F8F8F;
+		display: block;
+		margin-top: 4.125rem;
+		margin-left: auto;
+		margin-right: auto;
+		position: relative;
+
+		.preview-photo {
+			max-width:19.80rem;
```
inez: Styling

front/scripts/main/views/ArticleView.ts
```diff
@@ -194,13 +195,20 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 		this.$(':header[section]').each((i: Number, item: any): void => {
 			var $sectionHeader = this.$(item),
-				$pencil = this.$(pencil).appendTo($sectionHeader);
+				$pencil = this.$(pencil).appendTo($sectionHeader),
+				$photo = this.$(photo).appendTo($sectionHeader);
 			$pencil.on('click', (): void => {
 				this.get('controller').send('edit', this.get('controller.model.cleanTitle'), $sectionHeader.attr('section'));
 			});
+			$photo.on('change', () => {
+				var photoData = this.$($sectionHeader.find('.file-input'))[0].files[0];
```
inez: Is that this.$ needed?

front/scripts/main/views/ArticleView.ts
```diff
@@ -194,13 +195,20 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 
 	setupEditButtons: function (): void {
```
inez: Well, it sets up both edit and upload buttons - should be perhaps called something else?

front/scripts/main/controllers/ArticleController.ts
```diff
@@ -35,6 +38,26 @@ App.ArticleController = Em.Controller.extend({
 			});
 		},
 
+		addPhoto: function (title: string, photoData: any, sectionIndex: number): void {
+			App.VisibilityStateManager.reset();		
+			var oFReader = new FileReader();
+			oFReader.readAsDataURL(photoData);
+			oFReader.onload = function(oFREvent: any) {
+				this.set('photoData', photoData);
```
inez: I think you can pass it all as a collection/object.

front/scripts/main/controllers/ArticleController.ts
```diff
@@ -35,6 +38,26 @@ App.ArticleController = Em.Controller.extend({
...
+				this.set('photoName', photoData.name);
+				this.set('photoImage', oFREvent.target.result);
+			}.bind(this);
+
+			this.transitionToRoute('addPhoto', title, sectionIndex);
```
inez: So, this transition to addPhoto route right away, but data needed for that route to work correctly (photoData,Name,Image) is set later (in asynchronous handler) is that desired?

front/scripts/main/controllers/ArticleController.ts
```diff
@@ -35,6 +38,26 @@ App.ArticleController = Em.Controller.extend({
 			});
 		},
 
+		addPhoto: function (title: string, photoData: any, sectionIndex: number): void {
+			App.VisibilityStateManager.reset();
```
inez: What is this call for here?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,197 @@
...
+
+App.AddPhotoModel = Em.Object.extend({
+	title: null,
+	sectionIndex: null
+});
+
+App.AddPhotoModel.reopenClass({
+	
+	load: function(title: string, sectionIndex: number) {
+		return App.AddPhotoModel.create({
```
inez: Formatting here is off

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,197 @@
...
+					title: title,
+					sectionIndex: sectionIndex
+				});
+	},
+
+	//TODO: This code is exactly the same as the one in EditModel. Refactor.
+	getEditToken: function(title: string): Em.RSVP.Promise {
```
inez: @mari-mcmurtrie We talked about not copying this method, right?

front/scripts/main/controllers/ArticleController.ts
```diff
@@ -35,6 +38,26 @@ App.ArticleController = Em.Controller.extend({
...
 		},
 
+		addPhoto: function (title: string, photoData: any, sectionIndex: number): void {
+			App.VisibilityStateManager.reset();		
+			var oFReader = new FileReader();
+			oFReader.readAsDataURL(photoData);
+			oFReader.onload = function(oFREvent: any) {
+				this.set('photoData', photoData);
+				this.set('photoName', photoData.name);
+				this.set('photoImage', oFREvent.target.result);
```
inez: It seems that you don't need photoData and photoImage - just one should be enough to preview it and later to upload it.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,197 @@
...
+					}
+					var newPhotoTitle = this.mayChangePhotoTitle(photoName, photoTitleNoExt),
+						params = $.param({action: 'addmediapermanent',
+										  format: 'json',
+										  title: newPhotoTitle,
+										  tempName: addmediatemporary.tempName}),	
+						requestUrl = Mercury.wiki.basePath + '/api.php?' + params + '&*';
+					Em.$.ajax({
+						url:  requestUrl,
+						method: 'GET',
```
Grunny: All state changing actions should use a POST (and the API endpoint should require it, which is a bug).

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,197 @@
...
+		});
+	},
+
+	createFormData: function(photoData: any) {
+		var form = new FormData();
+		form.append('file', photoData);
+		return form;
+	},
+
+	mayChangePhotoTitle: function(originalPhotoName: string, userInputPhotoTitle: string): any {
```
inez: @mari-mcmurtrie let's talk about this method tomorrow

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+
+App.AddPhotoController = Em.Controller.extend({
+	needs: ['application', 'article'],
+	photoData: Ember.computed.alias('controllers.article.photoData'),
+	photoName: Ember.computed.alias('controllers.article.photoName'),
+	photoImage: Ember.computed.alias('controllers.article.photoImage'),
+	photoTitleNoExt: function() {
+		var photoName = this.get('photoName');
+		return photoName ? photoName.substr(0, photoName.lastIndexOf('.')) : '';
+	}.property('photoName'),
```
rogatty: Please remove the prototype extension, see http://guides.emberjs.com/v1.12.0/configuring-ember/disabling-prototype-extensions/#toc_functions and use `Em.computed()` syntax.

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+	handleUploadError: function(error: any) {
+		//TODO: Error handling
+	},
+
+	actions: {
+		upload: function (): void {
+			this.get('controllers.application').showLoader();	
+			App.AddPhotoModel.upload(this.get('controllers.article.photoData'), 
+									 this.get('controllers.article.photoName'),
+									 this.get('photoTitleNoExt')).then(
```
rogatty: Indentation is off.

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+		back: function (): void {
+			this.transitionToRoute('article', this.get('model.title'));
+			M.track({
+				action: M.trackActions.click,
+				category: 'sectionaddphoto',
+				label: 'back'
+			});
+		}
+	},
+});
```
rogatty: `\n` is missing.

front/scripts/main/models/EditModel.ts
mklucsarits: @mari-mcmurtrie @inez The buildUrl() method can be used as soon as that #885 is merged, if you want to wait for it. Otherwise, I don't think it's unreasonable to change this to an absolute URL, since we deal with localhost during local testing. And also because going forward, using buildUrl(), all paths to API will be absolute paths anyway, using the remote Mediawiki host.

front/scripts/main/controllers/AddPhotoController.ts
mari-mcmurtrie: It looks fine in my editor. Fixed.

front/scripts/main/controllers/AddPhotoController.ts
mari-mcmurtrie: fixed.

front/scripts/main/controllers/AddPhotoController.ts
inez: Well, wouldn't something like this read better?
```
App.AddPhotoModel.upload(
	this.get('controllers.article.photoData'),
	this.get('controllers.article.photoName'),
	this.get('photoTitleNoExt')
).then(
	this.handleUploadSuccess.bind(this),
	this.handleUploadError.bind(this)
);
```

front/scripts/main/controllers/ArticleController.ts
mari-mcmurtrie: I added since edit function above also has it. Do not know what this actually do! I remove a line.

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../baseline/mercury.d.ts" />
+'use strict';
+
+App.AddPhotoController = Em.Controller.extend({
+	needs: ['application', 'article'],
+	photoData: Ember.computed.alias('controllers.article.photoData'),
+	photoName: Ember.computed.alias('controllers.article.photoName'),
+	photoImage: Ember.computed.alias('controllers.article.photoImage'),
+	photoTitleNoExt: function() {
```
inez: I'm putting here just a reminder of that idea to change how file name and file extension is stored - basically simplify: one var for filename without extension and one var for file extension. Then you just let user edit one (the filename without extension) and your method "mayChangePhotoTitle" wouldn't be even needed I think.

front/scripts/main/controllers/ArticleController.ts
mari-mcmurtrie: I improve code here, as we discussed the other day.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: Do you mean there is 1 empty line right above load function? If so, I am shocked. Anyways, I remove an empty line.

front/scripts/main/controllers/ArticleController.ts
inez: @mari-mcmurtrie it's important that you understand code that you write, if some things are not crystal clear - then there should be at least fixme or todo statement.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I totally agree with you @Grunny. This is not REST-ful.

inez: To clarify: @mari-mcmurtrie you still can (and perhaps should) use POST instead of GET.

front/scripts/main/models/AddPhotoModel.ts
inez: No, it wasn't about empty line - it was about indentation - but I think it's good now.

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+		var photoName = this.get('photoName');
+		return photoName ? photoName.substr(0, photoName.lastIndexOf('.')) : '';
+	}.property('photoName'),
+
+	handleAddContentSuccess: function(data: any) {
+		var title = this.get('model.title');
+		this.transitionToRoute('article', title).then((): void => {
+			this.get('controllers.application').addAlert('success', i18n.t('app.add-photo-success', {pageTitle: title}));
```
Grunny: The `add-photo-success` message does not currently have any parameters, so doesn't need the page title passed to it, unless you plan to add a page title parameter to the message?

front/scripts/main/controllers/ArticleController.ts
mari-mcmurtrie: I changed this code as we discussed. No more aync here; hence removed race condition. Thanks for your suggestion!

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: Maybe I remove "createFormData()" since it is kind of meaningless...

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,197 @@
...
+			errors.push( [ 'size', Math.round( maxUploadSize / 1048576 * 100 ) / 100 ] );
+		}
+		if ( !filetype ) {
+			errors.push( [ 'filetype',  fileExtensions.join( ', ' ) ] );
+		}
+
+		return errors;
+		*/
+	}
+});
```
mari-mcmurtrie: I added a new line at the end of file.

front/scripts/main/views/AddPhotoView.ts
```diff
@@ -0,0 +1,6 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.AddPhotoView = Em.View.extend({
+	classNames: ['addphoto-view'],
+});
```
mari-mcmurtrie: Added new line.

front/scripts/main/controllers/AddPhotoController.ts
mari-mcmurtrie: Thanks, I did not know that (new to Ember). I rewrote this as you pointed out.

front/scripts/main/models/EditModel.ts
mari-mcmurtrie: I talked to Matt. This should be ok as-is.

front/scripts/main/controllers/ArticleController.ts
inez: Well, it totally looks async to me, try putting console.log(46) above line 46 and console.log(51) above line 51. "51" will be printed before "46" - if not I buy you bottle of wine :)

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I removed. It is not needed at all.

front/styles/main/module/edit/_addphoto.scss
mari-mcmurtrie: indentation looked fine in my editor :(

front/styles/main/module/edit/_edit.scss
```diff
@@ -0,0 +1,21 @@
...
+.pencil {
+	float: right;
+	height: 2.2rem;
+	width: 2.2rem;
+	&.zero {
+		position: absolute;
+		right: 1.125rem;
+		top: 3.5rem;
+	}
+}
```
mari-mcmurtrie: I added new line there.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I changed the name from "edit" -> "contribution".

front/styles/main/module/edit/_edit-common.scss
```diff
@@ -28,26 +29,4 @@
...
-	float: right;
-	height: 2.2rem;
-	width: 2.2rem;
-	&.zero {
-		position: absolute;
-		right: 1.125rem;
-		top: 3.5rem;
-	}
-}
+}
```
mari-mcmurtrie: Added a new line there.

front/templates/main/article.hbs
```diff
@@ -11,7 +11,7 @@
 	{{#if model.isCuratedMainPage}}
 		{{main-page model=model.mainPageData}}
 	{{else}}
-		{{#if view.editButtonsVisible}}
+		{{#if view.editFeatureEnabled}}
 			<span {{action 'edit' model.cleanTitle '0'}}>
 				{{svg 'pencil' role='img' class='icon pencil zero'}}
 			</span>
```
mari-mcmurtrie: Changed name from "edit" -> "contribution"

front/scripts/main/controllers/ArticleController.ts
```diff
@@ -35,7 +35,17 @@ App.ArticleController = Em.Controller.extend({
 			});
 		},
 
-		articleRendered: function () {
+		addPhoto: function (title: string, photoData: any, sectionIndex: number): void {
```
inez: It's not very significant but wouldn't be better to keep the same order of parameters in both places? - title, sectionIndex, photoData?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,150 @@
...
+
+App.AddPhotoModel = Em.Object.extend({
+	title: null,
+	sectionIndex: null,
+	photoData: null,
+	photoImage: null,
+	photoName: null,
+	photoExtension: null,
+
+	photoNameExtensionUpdater: function(): void {
```
inez: Having observer here seems to me to be an overkill. Because "photoData" property for given instance of model is never going to change and is not meant to change - observer shouldn't be used.
Suggested solution:
* create a utility method that takes full filename and return filename without extension and extension
* use that method where you initialize instance of this model (App.AddPhotoModel.create...) or overwrite method create (make sure it is ok to do it) to use that utility method internally

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,150 @@
...
+				);
+			};
+			oFReader.onerror = function(OFREvent: any): void {
+				reject();
+			};
+		});
+	},
+
+	addToContent: function(uploadedPhotoTitle: any, model: any): Em.RSVP.Promise {
+		var photoWikiText = '[[File:' + uploadedPhotoTitle + '|thumb]]\n',
```
inez: If we always append shouldn't we have new line in front of what we append?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,150 @@
...
+					}
+					var newPhotoTitle = model.photoName.trim().length === 0 ? model.photoData.name : model.photoName.trim() + '.' + model.photoExtension,
+						params = $.param({action: 'addmediapermanent',
+										  format: 'json',
+										  title: newPhotoTitle,
+										  tempName: addmediatemporary.tempName}),
+						requestUrl = Mercury.wiki.basePath + '/api.php?' + params + '&*';
+					Em.$.ajax({
+						url:  requestUrl,
+						method: 'GET',
```
inez: Shouldn't POST be used here?

front/scripts/main/router.ts
```diff
@@ -23,6 +23,10 @@ App.Router.map(function () {
 		path: articlePath + 'edit/:title/:sectionIndex'
 	});
 
+	this.route('addPhoto', { // Symbolic link to AddPhotoController
+		path: articlePath + 'addPhoto/:title/:sectionIndex'
```
inez: I wonder if we still need :title and :sectionIndex - given that now we are just passing entire model to the route.

front/scripts/main/views/ArticleView.ts
```diff
@@ -200,15 +201,29 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 
-	setupEditButtons: function (): void {
+	setupContributionButtons: function (): void {
 		// TODO: There should be a helper for generating this HTML
-		var pencil = '<svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg>';
+		var pencil = '<svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg>',
+			photo = '<div class="upload_photo"><svg class="icon camera" role="img"><use xlink:href="#camera"></use></svg><input class="file-input" type="file" accept="image/*" capture="camera"/></div>',
+			$photoZero = this.$('.upload_photo');
+
+		$photoZero.on('change', () => {
```
inez: This handler for change method is almost identical to handler in line 223. Could it be just one handler?

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+	needs: ['application', 'article'],
+
+	errorCodeMap: {
+		'invalidtitle': 'app.add-photo-section-title-error'
+	},
+
+	handleAddContentSuccess: function(data: any): void{
+		var title = this.get('model.title');
+		this.transitionToRoute('article', title).then((): void => {
+			this.get('controllers.application').addAlert('success', i18n.t('app.add-photo-success', {pageTitle: title}));
```
Grunny: You don't need to pass the page title to the message, unless you're going to change the message to something like `"A photo has been added to __pageTitle__."`

front/locales/en/main.json
```diff
@@ -64,6 +64,13 @@
 		"wikia-map-view-map": "View",
 		"search-label": "Search",
 		"cancel-label": "Cancel",
+		"add-photo-page-heading": "Adding photo to:",
+		"add-photo-page-add-button": "Add",
+		"add-photo-photo-name": "Photo Name",
+		"add-photo-instruction": "The image will be added to the end of the section",
+		"add-photo-success": "A photo has been addded to the page.",
```
Grunny: Typo: "addded" -> "added"

mari-mcmurtrie: Thank you for catching it!

front/scripts/main/controllers/AddPhotoController.ts
mari-mcmurtrie: Fixed. thanks!

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: addmediatemporary is POST, addmediapermanent is GET.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: You are right. I was also thinking to append new line at the end. Let me try that and check if that affects formatting of the placement of images.

inez: Besides placement of the image the formatting of the wikitext is also important. Users like when wikitext is well formatted - still lots of them (especially hardcode users) edit wikitext directly.

front/scripts/main/models/AddPhotoModel.ts
Grunny: It's a state changing action, so you should still be using POST (even if the API module isn't requiring it, though it should and we can file a bug for that).

inez: Also that will save you from having to concatenate that URL and use $.param. You can pass your object just in data key in $.ajax settings.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I think new line in front of image looks better since it puts the image in the line below the sentence.

front/scripts/main/controllers/ArticleController.ts
mari-mcmurtrie: Agreed. Fixed.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I did the same as desktop visual editor uploading media (it sends GET), but I think the one is a bug. Let me change this to POST and also change the params to payload to be proper POST request. Thanks.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: You are right. I do static method to decompose photoName.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I refactored.

front/scripts/main/router.ts
mari-mcmurtrie: I think we need it for url?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,156 @@
...
+App.AddPhotoModel = Em.Object.extend({
+	title: null,
+	sectionIndex: null,
+	photoData: null,
+	photoImage: null,
+	photoName: null,
+	photoExtension: null
+});
+
+App.AddPhotoModel.processFileName = function(fileName: string): any {
```
inez: This method do a very specific thing to any file name - there is no need to give it confusing name and inside use "photo" prefixes - it would work as well for any type of file.
Please add documentation how it should behave it file have no extension etc, and having unit tests for it would be really good as well.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,156 @@
...
+	var photoName = fileName.substr(0, fileName.lastIndexOf('.')),
+	photoExtension = fileName.substr(fileName.lastIndexOf('.') + 1);
+	return { photoName: photoName, photoExtension: photoExtension };
+};
+
+App.AddPhotoModel.reopenClass(App.EditMixin, {
+	load: function(title: string, sectionIndex: number, photoData: any): Em.RSVP.Promise {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			var photoName = '',
```
inez: What's the point of declaring those variables here - in scope in which they are not used?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,156 @@
...
+			data = {
+				action: 'edit',
+				title: model.title,
+				section: model.sectionIndex,
+				format: 'json',
+				prependtext: '',
+				appendtext: '',
+				token: ''
+			};
+		data.appendtext = photoWikiText; //For now, we always add image file at the bottom of section.
```
inez: Please address it accordingly to conversation we had.

front/scripts/main/views/ArticleView.ts
```diff
@@ -200,18 +201,35 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 			$pencil.on('click', (): void => {
 				this.get('controller').send('edit', this.get('controller.model.cleanTitle'), $sectionHeader.attr('section'));
 			});
+			$photo.on('change', () => {
+				this.photoIconChange($sectionHeader, $sectionHeader.attr('section'));
+			});
 		});
 	},
 
+	photoIconChange: function(sectionHeader:any, sectionNumber:any): void {
```
inez: Please do rename from photoIconChange to onPhotoIconChange.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: done

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: done

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I thought I would declare variable all in 1 place....  :(

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: ok

front/scripts/main/models/AddPhotoModel.ts
inez: Btw. this is still not addressed.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,156 @@
...
+	},
+
+	upload: function(model: any): Em.RSVP.Promise {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			this.temporaryUpload(model.photoData)
+				.then((addmediatemporary: any): void => {
+					if (addmediatemporary.tempName === undefined) { //we already have the file.
+						resolve({'title': addmediatemporary.title});
+					}
+					var newPhotoTitle = model.photoName.trim().length === 0 ? model.photoData.name : model.photoName.trim() + '.' + model.photoExtension,
```
inez: What is the point of this code? If user clears the name and try to upload shouldn't error by displayed?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,156 @@
...
+
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			Em.$.ajax({
+				url:  Mercury.wiki.basePath + '/api.php?action=addmediatemporary&format=json',
+				method: 'POST',
+				data: formData,
+				cache: false,
+				xhrFields: {
+						withCredentials: true
+					},
```
inez: Too much indentation

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I asked product requirement for the use case when a user supplies empty image name. We decided that, in such a case, we silently replace empty name with the original file name.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: fixed.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I cannot separate this. This is not very nested.

front/styles/main/module/edit/_addphoto.scss
Grunny: This and the others are indented fine, but they need a [space after the colon](https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#spaceafterpropertycolon) per our conventions.

front/styles/main/module/edit/_addphoto.scss
```diff
@@ -0,0 +1,72 @@
...
+			max-height:19.80rem;
+			width: auto;
+			height: auto;
```
Grunny: We also [order properties alphabetically](https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#propertysortorder).

front/scripts/main/views/ArticleView.ts
```diff
@@ -196,18 +197,35 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		}
 	},
 
-	setupEditButtons: function (): void {
+	setupContributionButtons: function (): void {
 		// TODO: There should be a helper for generating this HTML
-		var pencil = '<svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg>';
+		var pencil = '<svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg>',
+			photo = '<div class="upload_photo"><svg class="icon camera" role="img"><use xlink:href="#camera"></use></svg><input class="file-input" type="file" accept="image/*" capture="camera"/></div>',
```
Grunny: The class on the div should use a dash rather than an underscore [per conventions](https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#class-and-id-naming-conventions).

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,158 @@
...
+App.AddPhotoModel = Em.Object.extend({
+	title: null,
+	sectionIndex: null,
+	photoData: null,
+	photoImage: null,
+	photoName: null,
+	photoExtension: null
+});
+
+App.AddPhotoModel.separateFileNameToNameAndExtension = function(fileName: string): any {
```
Grunny: This would be a good method to add some unit tests for. Also, maybe a slightly shorter name without losing meaning, maybe just `separateFileNameAndExtension`?

front/scripts/main/views/ArticleView.ts
```diff
@@ -197,18 +198,35 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 			$pencil.on('click', (): void => {
 				this.get('controller').send('edit', this.get('controller.model.cleanTitle'), $sectionHeader.attr('section'));
 			});
+			$photo.on('change', () => {
+				this.onPhotoIconChange($sectionHeader, $sectionHeader.attr('section'));
+			});
 		});
 	},
 
+	onPhotoIconChange: function(sectionHeader:any, sectionNumber:any): void {
```
Grunny: Need to have spaces between the colon and the type on the variables.

front/templates/main/addPhoto.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="edit-head">
+<span {{action 'back'}}>
```
Grunny: Should be indented to be consistent.

front/scripts/main/views/ArticleView.ts
```diff
@@ -197,18 +198,35 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		}
 	},
 
-	setupEditButtons: function (): void {
+	setupContributionButtons: function (): void {
 		// TODO: There should be a helper for generating this HTML
-		var pencil = '<svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg>';
+		var pencil = '<svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg>',
+			photo = '<div class="upload_photo"><svg class="icon camera" role="img"><use xlink:href="#camera"></use></svg><input class="file-input" type="file" accept="image/*" capture="camera"/></div>',
+			$photoZero = this.$('.upload_photo');
```
Grunny: This isn't just about the zero section photo button, so the variable should be renamed.

front/templates/main/article.hbs
```diff
@@ -11,7 +11,11 @@
 	{{#if model.isCuratedMainPage}}
 		{{main-page model=model.mainPageData openLightbox='openLightbox'}}
 	{{else}}
-		{{#if view.editButtonsVisible}}
+		{{#if view.contributionFeatureEnabled}}
+			<div class="upload_photo zero">
```
Grunny: Same here, the class name should have a dash not an underscore.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,158 @@
...
+	return { name: name, extension: extension };
+};
+
+App.AddPhotoModel.reopenClass(App.EditMixin, {
+	load: function(title: string, sectionIndex: number, photoData: any): Em.RSVP.Promise {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			var oFReader = new FileReader();
+			oFReader.readAsDataURL(photoData);
+			oFReader.onload = function(oFREvent: any): void {
+				var photoName = App.AddPhotoModel.separateFileNameToNameAndExtension(photoData.name).name,
```
inez: I think you should be calling separateFileNameToNameAndExtension just once. @Grunny what do you think?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,158 @@
...
+				});
+		});
+	},
+
+	upload: function(model: any): Em.RSVP.Promise {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			this.temporaryUpload(model.photoData)
+				.then((addmediatemporary: any): void => {
+					if (addmediatemporary.tempName === undefined) { //we already have the file.
+						resolve({'title': addmediatemporary.title});
```
inez: Reminder per out Friday conversation - should you have "return" here?

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,158 @@
...
+					}
+					//If a user inputs an empty image name, then we silently replace it with original file name.
+					var newPhotoTitle = model.photoName.trim().length === 0 ? model.photoData.name : model.photoName.trim() + '.' + model.photoExtension,
```
inez: Reminder per out Friday conversation - split into multiple lines.

front/scripts/main/models/AddPhotoModel.ts
inez: Let's clarify: you don't know how to separate it or you think it shouldn't be separated?
In the same way that temporaryUpload is a method responsible for "temporary upload" - there could be another one "permanentUpload" responsible for "permanent upload"  - and this current method (upload) would be just responsible for appropriately calling them (or not).

front/scripts/main/router.ts
inez: I recommend removing dynamic segments from code - as they are not being used anymore for anything.

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+'use strict';
+
+App.AddPhotoController = Em.Controller.extend({
+	needs: ['application', 'article'],
+
+	errorCodeMap: {
+		'invalidtitle': 'app.add-photo-section-title-error'
+	},
+
+	handleAddContentSuccess: function(data: any): void{
```
RafLeszczynski: You missed a space between `void` and `{`. Consider using tools to automate fixing spacing, newline and indentation issues. You can find the details and how to setup them in our Guid lines repo: https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#tools

front/scripts/main/models/AddPhotoModel.ts
Grunny: Yep, agreed.

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -0,0 +1,69 @@
...
+		var title = this.get('model.title');
+		this.transitionToRoute('article', title).then((): void => {
+			this.get('controllers.application').addAlert('success', i18n.t('app.add-photo-success'));
+		});
+		M.track({
```
RafLeszczynski: This is actually question to Mercury Platform. @ken @lizlux do we have any rules for how to handle tracking in Mercury? Should we add tracking inside business logic implementation (like below) or create a separate `.ts` file for tracking (like it was done for login https://github.com/Wikia/mercury/pull/972/files)?

front/scripts/main/mixins/EditMixin.ts
```diff
@@ -0,0 +1,37 @@
...
+					intoken: 'edit',
+					format: 'json'
+				},
+				dataType: 'json',
+				success: (resp: any): void => {
+					var edittoken: string,
+						pages: any = Em.get(resp, 'query.pages');
+					if (pages) {
+						// FIXME: MediaWiki API, seriously?
+						edittoken = pages[Object.keys(pages)[0]].edittoken;
```
RafLeszczynski: Are we sure that `pages` will always be `Object` and that `Object.keys(pages)[0]` will always return the `string` with expected property name?

front/scripts/main/models/AddPhotoModel.ts
RafLeszczynski: Consider more strict approach to defining function return type. For example: `: {name: string, extension: string}`

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,158 @@
...
+				.then((token: any): void => {
+					data.token = token;
+					Em.$.ajax({
+						url: M.buildUrl({path: '/api.php'}),
+						dataType: 'json',
+						method: 'POST',
+						data: data,
+						success: (resp: any): void => {
+							if (resp && resp.edit && resp.edit.result === 'Success') {
+								resolve();
```
RafLeszczynski: There is quite a lot of multi level nesting inside this function. Simplifying that would greatly increase code readability. Consider breaking this part of code into multiple smaller functions.  You can also find sme our guide lines about nested functions in the guidelines repo: https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#nested-functions

front/scripts/main/controllers/AddPhotoController.ts
lizlux: What Mari has done is the correct way to add tracking in the Ember app. The auth pages weren't built on Ember so they don't have these hooks automatically set up for them, which is probably why the social team handled tracking the way they did.

front/scripts/main/views/ArticleView.ts
```diff
@@ -197,18 +198,35 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+
+		$photoZero.on('change', () => {
+			this.onPhotoIconChange($photoZero, 0);
+		});
+
 		this.$(':header[section]').each((i: Number, item: any): void => {
 			var $sectionHeader = this.$(item),
+				$photo = this.$(photo).appendTo($sectionHeader),
```
RafLeszczynski: Please consider reducing the number of DOM operations to one instead of two, especially that this is done inside the loop.

front/scripts/main/views/ArticleView.ts
```diff
@@ -197,18 +198,35 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 				$pencil = this.$(pencil).appendTo($sectionHeader);
+				
 			$pencil.on('click', (): void => {
```
RafLeszczynski: If I understand correctly, you are attaching the same  listeners to multiple similar DOM elements inside the loop? Please consider moving this part of code outside the loop and use event delegation pattern to attach the listeners.

front/scripts/main/views/ArticleView.ts
RafLeszczynski: Consider avoiding the use of `: any` whenever you can. If I understand the code correctly `sectionHeader` is a jquery collection object and `sectionNumber:any` is a string?

front/styles/main/module/edit/_addphoto.scss
RafLeszczynski: consider using SCSS Linter for tracking all syntax formatting issues:
https://github.com/brigade/scss-lint 
In addition you can find all Wikia SCSS coding conventions here:
https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md

front/styles/main/module/edit/_addphoto.scss
```diff
@@ -0,0 +1,72 @@
+.preview-body {
+
+	.preview-container {
+		width: 20rem;
+		height: 20rem;
```
RafLeszczynski: Consider declaring magic number values as local variables - especially when they are used multiple times throughout the stylesheet.

front/styles/main/module/edit/_addphoto.scss
```diff
@@ -0,0 +1,72 @@
...
+		background-color: #DFDFDF;
```
RafLeszczynski: are these color values unique for this feature or they are some common colors used across mercury app?

front/svg/symbols/main/camera.svg
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 18.1.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
RafLeszczynski: consider removing unneeded code from all SVG files to reduce the file size - for example Illustrator comments

front/scripts/main/mixins/EditMixin.ts
mari-mcmurtrie: This code was originally in EditModel.ts. I just refactored into mixin since AddPhotoModel.ts also uses it.

front/scripts/main/controllers/AddPhotoController.ts
RafLeszczynski: Thanks @lizlux good to know :+1:

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: May I ask you why?

RafLeszczynski: Of course :-) For performance reasons. DOM manipulation are causing page reflows. These ones are done inside the loop so We could reduce the number of reflows by half which based on the number of `:header[section]` could be a lot.

front/styles/main/module/edit/_addphoto.scss
mari-mcmurtrie: I have to ask UX/UI person about it. I added this color before I learned about style guide.

mari-mcmurtrie: This is 1 off so I just hard-code the color as is.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Got it, thank you!

front/styles/main/module/edit/_addphoto.scss
RafLeszczynski: Some colors are defined in `_wikia-variables.scss`
https://github.com/Wikia/mercury/blob/74eb4c8a197c438bab104cf589cf8838e2a1503c/front/styles/main/_wikia-variables.scss#L39

They all comes from the style guide:
https://github.com/Wikia/style-guide/blob/dev/src/scss/lib/_settings.scss#L95

front/scripts/main/views/ArticleView.ts
RafLeszczynski: you're welcome :-)

front/scripts/main/router.ts
mari-mcmurtrie: I and Inez worked with this issue.

front/styles/main/module/edit/_addphoto.scss
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: Fixed (in next push). I did not know you can specify return type like { name: name, extension: extension }. I learned, thank you!

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Fixed (in next push). I am so new to TypeScript as well as Ember, in addition to Mercury code so I tend to miss details like this here and there. Thanks for pointing out, I will be better :)

front/templates/main/addPhoto.hbs
mari-mcmurtrie: Fixed (in next push).

front/templates/main/article.hbs
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/controllers/AddPhotoController.ts
mari-mcmurtrie: I have javascript formatter plugin in my sublime but I guess it does not cover this particular case. I may add more of those tools in order to avoid formatting errors.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: I refactored. (in next push)

front/svg/symbols/main/camera.svg
mari-mcmurtrie: removed.

front/scripts/main/views/ArticleView.ts
RafLeszczynski: You're welcome. If you didn't have a chance yet, check out type script handbook: http://www.typescriptlang.org/Handbook - there are some great examples.

front/styles/main/module/edit/_addphoto.scss
mari-mcmurtrie: Thanks! I did not know these pre-defined colors and how to use them. I learned again. I am gonna use $color-blue-gray-light since it looks very close to the mock color.

front/scripts/main/controllers/AddPhotoController.ts
RafLeszczynski: Try setting them up as github pre-commit hooks so they will run automatically before each commit.

front/styles/main/module/edit/_addphoto.scss
mari-mcmurtrie: Fixed (in next push).

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: So this one, I want to keep this as-is because pencil needs "click" event listener but Photo needs "change". These are not the same event listeners.

front/styles/main/module/edit/_addphoto.scss
RafLeszczynski: That's a great idea. When ever you find out custom color in the design spec consider asking designer if this is some special case or could we use color already defined in style guide.

front/scripts/main/views/ArticleView.ts
RafLeszczynski: consider something like this somewhere outside the loop:

```JS
$('.article-content')
    .on('click', '.pencil', (event): void => {
        var $sectionHeader = $(event.target).parent();
        this.get('controller').send('edit', this.get('controller.model.cleanTitle'), $sectionHeader.attr('section'));

})
    .on('change', '.upload_photo', (event): void => {
       var $sectionHeader = $(event.target).parent();
       this.onPhotoIconChange($sectionHeader, $sectionHeader.attr('section'));
});
```

Instead of `2n` listeners where `n = sections.length` you could have only two listeners with event delegation that handles all clicks on '.pencil' and changes in '.upload_photo'.

front/styles/main/module/edit/_addphoto.scss
```diff
@@ -0,0 +1,76 @@
...
+	.icon {
+		height: $upload-icon-size;
+		position:absolute;
+		right: 0;
+		top: 0;
+		width: $upload-icon-size;
+	}
+
+	input {	
+		filter: alpha(opacity=0);
```
vforge: As this is only needed for IE8, and we don't support it, please consider removing `filter`.
Also properties here should be formatted alphabetically according to https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#propertysortorder

front/svg/symbols/main/camera.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
vforge: Please consider minimizing SVG using https://petercollingridge.appspot.com/svg-editor

front/svg/symbols/main/pencil.svg
```diff
@@ -1,5 +1,17 @@
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" x="0" y="0" viewBox="0 0 40 40" enable-background="new 0 0 40 40" xml:space="preserve" preserveAspectRatio="xMidYMin">
-  <g fill="#1A5EB8">
-    <path d="M20 1C9.5 1 1 9.5 1 20s8.5 19 19 19 19-8.5 19-19S30.5 1 20 1zM16.2 26.1l-3.5 1.4c-0.2 0.1-0.2 0-0.2-0.2l1.4-3.5c0.1-0.2 0.2-0.2 0.3-0.1l2.1 2.1C16.4 25.9 16.4 26.1 16.2 26.1zM17.3 25.2l-2.5-2.5 8-8 2.5 2.5L17.3 25.2zM27.4 15.2l-0.3 0.3 -0.6 0.6 -0.3 0.3 -0.3-0.3L24 14.2l-0.3-0.3 0.3-0.3 0.6-0.6 0.3-0.3c0.2-0.2 0.5-0.2 0.6 0l1.9 1.9C27.5 14.7 27.5 15 27.4 15.2z"/>
-  </g>
-</svg>
+<?xml version="1.0" encoding="utf-8"?>
```
vforge: Please consider minimizing SVG using https://petercollingridge.appspot.com/svg-editor (check previous version of this file to see example minified version)

mari-mcmurtrie: Minified.

inez:
@mari-mcmurtrie What's the status of this PR? Are you waiting for CR from particular people? It can't be merged now - there is a conflict - if you want it merged please solve that conflict first.
(Remember that if we merge it - it's gonna go on production within hours.)

front/styles/main/module/edit/_edit-common.scss
```diff
@@ -16,7 +16,7 @@
 	.publish {
 		float: right;
 	}
-
+
```
inez: This shouldn't be here.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -0,0 +1,185 @@
...
+
+	upload: function(model: any): Em.RSVP.Promise {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			this.temporaryUpload(model.photoData)
+				.then((addmediatemporary: any): void => {
+					if (addmediatemporary.tempName === undefined) { //we already have the file. No need to upload.
+						return resolve({'title': addmediatemporary.title});
+					}
+					//If a user inputs an empty image name, then we silently replace it with original file name.
+					var newPhotoTitle = '';
```
inez: There is no point of initializing it with a value. Can you just change to: "var newPhotoTitle;"?

front/scripts/main/views/ArticleView.ts
```diff
@@ -81,7 +81,7 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			$('meta[name="description"]').attr('content', (typeof model.get('description') === 'undefined') ? '' : model.get('description'));
 		}
 	}),
-
+
```
inez: No need for this.

inez:
LGTM - except couple minor comments I just left - and making it mergeable to dev.

front/scripts/main/models/AddPhotoModel.ts
mari-mcmurtrie: yes I can thanks.

PR #919, opened 2015-06-15
HG-740 Upgrade to Ember 1.13

rogatty:
https://wikia-inc.atlassian.net/browse/HG-740
http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html

Known issues:
- [x] Hammer doesn't work - gestures don't work
- [x] Tapping on search results doesn't work
- [x] Navigating in sidenav doesn't work
- [x] Swiping in lightbox changes URL but not displayed media
- [x] Link to open side nav on 404 page doesn't work

Deprecations:
- [ ] Ember.View is deprecated. Consult the Deprecations Guide for a migration strategy. See http://emberjs.com/deprecations/v1.x/#toc_ember-view
  - [ ] AddPhotoView
  - [ ] ApplicationView
  - [x] ArticleView
  - [ ] EditView
- [x] Using the '{{#each item in model}}' form of the {{#each}} helper is deprecated. Please use the block param form instead ('{{#each model as |item|}}'). See http://emberjs.com/guides/deprecations/#toc_code-in-code-syntax-for-code-each-code
- [x] Using '{{input action="enter"}}' is deprecated. Please use '{{input enter="enter"}}' instead.
- [ ] Using `{{view}}` or any path based on it has been deprecated. See http://emberjs.com/deprecations/v1.x#toc_view-and-controller-template-keywords
- [ ] Using the `{{view "string"}}` helper is deprecated. See http://emberjs.com/deprecations/v1.x#toc_ember-view
- [ ] Using 'itemController' with '{{each}}' is deprecated. Please refactor to a component.
- [ ] You modified view.style twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 2.0
- [ ] You modified concat([imageStyle]; separator="") twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 2.0
- [ ] You modified header twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 2.0
- [ ] You attempted to access `...` from `(generated application controller)`, but object proxying is deprecated. Please use `...` instead.
- [ ] A property of <Ember.OutletView:ember399> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation.
- [ ] Controller#needs is deprecated, please use Ember.inject.controller() instead


hakubo:
Phew! just this, and we're good to go?
No deprecations?

LGTM

rogatty:
@hakubo ha, I wish! This is WIP, just the beginning. There is still a lot to fix, we should make a check list in the PR description.

rogatty:
ember-hammer library had to be fixed, see https://github.com/chriswessels/ember-hammer/pull/12

hakubo:
@rogatty awesome work! thanks
We should merge it ASAP, not to have to many conflicts
@kenkouot @lizlux - ideally we would be merging this this week. I don't think we can afford waiting longer with this.

rogatty:
One thing. It reintroduces https://wikia-inc.atlassian.net/browse/HG-772 that was fixed in https://github.com/Wikia/mercury/pull/1099. Seems like this fix wasn't a bullet proof solution (`ArticleContentComponent` exists in this branch).

kenkouot:
Great work! I'd love to merge it this week but let's not until we can guarantee no regressions related to HG-772 or the double render. That bug caused several P2s and if we regress for the same reason, those P2's would likely show up again.

front/scripts/main/controllers/ApplicationController.ts
```diff
@@ -18,13 +18,21 @@ App.ApplicationController = Em.Controller.extend(App.LoadingSpinnerMixin, App.Al
...
-	userMenuCollapsed: true,
+	sideNavVisible: false,
+	userMenuVisible: false,
 	noScroll: false,
 	fullPage: false,
 	lightboxType: null,
 	lightboxModel: null,
 
+	sideNavCollapsedObserver: Em.observer('sideNavVisible', function (): void {
+		if (this.get('sideNavVisible')) {
```
hakubo:
```
noScroll: Em.computed.alias('sideNavVisible')
```

tomasznapieralski:
LGTM

PR #921, opened 2015-06-15
Set cache control headers to private if session cookie is present

kenkouot:
Will make DRYer later.


kenkouot:
Whoops, wrong branch

PR #922, opened 2015-06-15
Set cache control headers to private if session cookie is present

kenkouot:
Will make DRYer later

### DIFF
--- a/server/facets/api/article.ts
+++ b/server/facets/api/article.ts
@@ -36,12 +36,14 @@ export function get (request: Hapi.Request,  reply: any): void {
 			title: request.params.articleTitle,
 			redirect: request.params.redirect
 		},
-		article: Article.ArticleRequestHelper;
+		article: Article.ArticleRequestHelper,
+		allowCache = true;
 
 	if (request.state.wikicities_session) {
 		params.headers = {
 			'Cookie': `wikicities_session=${request.state.wikicities_session}`
 		};
+		allowCache = false;
 	}
 
 	article = new Article.ArticleRequestHelper(params);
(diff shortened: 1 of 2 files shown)


lizlux:
LGTM, seems to do what is requested in the ticket.

server/facets/api/article.ts
```diff
@@ -56,7 +58,13 @@ export function get (request: Hapi.Request,  reply: any): void {
 
 	article.getData((error: any, result: any): void => {
 		// TODO: Consider normalizing all error handling to Boom
-		var wrappedResult = wrapResult(error, result);
-		Caching.setResponseCaching(reply(wrappedResult).code(wrappedResult.status), cachingTimes);
+		var wrappedResult = wrapResult(error, result),
+			response = reply(wrappedResult).code(wrappedResult.status);
+
+		if (allowCache) {
+			return Caching.setResponseCaching(response, cachingTimes);
```
rogatty: This method (the anonymous one) has `void` return type. `setResponseCaching` returns `void` too but it looks confusing anyway. Don't you think that `if () { } else { }` would look better here? No need to return anything.

Same goes for https://github.com/Wikia/mercury/pull/922/files#diff-d278af34843da7ac020d3e5bd3213c39R103

server/facets/showArticle.ts
```diff
@@ -60,7 +62,7 @@ function showArticle (request: Hapi.Request, reply: Hapi.Response): void {
  * @param error
  * @param result
  */
-function onArticleResponse (request: Hapi.Request, reply: any, error: any, result: any = {}): void {
+function onArticleResponse (request: Hapi.Request, reply: any, error: any, result: any = {}, allowCache: boolean = true): void {
```
rogatty: `(max-line-length) facets/showArticle.ts[65, 1]: exceeds maximum line length of 120`

PR #927, opened 2015-06-16
Tasks update

tomasznapieralski:
I moved some of Jenkins scripts here:
* git-create.sh - creates git release branch and pull request
* git-merge.sh - merges pull request in specific number
* deploy.sh - lock/prep/push to environment mercury, app and config (all 3 are optional)


tomasznapieralski:
@Wikia/x-wing please CR

tasks/changelog-update.sh
```diff
@@ -32,11 +32,14 @@ if [ $WORDC -gt 0 ]; then
 	echo "" >> temp
 	cat CHANGELOG.md >> temp
 	mv temp CHANGELOG.md	
-	rm new
-	echo "Changelog updated"
+	echo
+	echo "Changelog updated with:"
```
Warkot: Instead of writing empty `echo` commands for newlines, you could use `echo -e` and put `\n` in the string. What do you think?

PR #928, opened 2015-06-16
HG-686 - added cache-buster for internal cache for random page request

katanka:
Tested on devbox with cookies disabled, and it works (unique pages). If you run dev, you get the same page every request.


server/lib/MediaWiki.ts
```diff
@@ -157,6 +157,7 @@ export class ArticleRequest {
 			action: 'query',
 			generator: 'random',
 			grnnamespace: 0,
+			CB: Date.now(),
```
rogatty: Would `cb` work? Right now it looks odd.

PR #929, opened 2015-06-16
INT-113: Fix problem by using cleanTitle property of ArticleModel instead of just title

inez:


inez:
@kenkouot Can you take a look at this?

inez:
@kenkouot Can you take a look at this?

kenkouot:
LGTM

PR #930, opened 2015-06-16
HG-757 - made lazy load image processing async

katanka:
Tested locally on http://starwars.wikia.com/wiki/Darth_Maleval - DOMContentLoaded improved from ~ 3.5 sec to ~2.5 sec


front/scripts/main/views/ArticleView.ts
```diff
@@ -151,7 +151,8 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.handleInfoboxes();
 			this.handlePortableInfoboxes();
-			this.lazyLoadMedia(model.get('media'));
+			// Make lazy load image processing async so it doesn't hold up DOMContentLoaded
+			setTimeout(this.lazyLoadMedia(model.get('media')), 0);
```
rogatty: `setTimeout` is a big no no in Ember, http://emberjs.com/api/classes/Ember.run.html#method_later

rogatty: Please rename `lazyLoadMedia` method to `replaceMediaPlaceholdersWithMediaComponents` because I can see it's confusing.

rogatty:
Proper way to fix this issue would be to investigate what really takes so much time when running `this.$(element).replaceWith(this.createMediaComponent(element, model));` in https://github.com/Wikia/mercury/blob/dev/front/scripts/main/views/ArticleView.ts#L191.

Would changing:
```
lazyImages.each((index: number, element: HTMLImageElement): void => {
	this.$(element).replaceWith(this.createMediaComponent(element, model));
});
```
to:
```
lazyImages.each(function (): void => {
	$(this).replaceWith(this.createMediaComponent(element, model));
});
```
make any difference? I think it removes the need for looking through DOM on each element.

Anyway, I believe that we should make data driven decisions when working on something so important like Mercury performance. Measure it, then decide how to fix it.

Postponing this heavy operation doesn't seem like a good solution, we should fix it to not be so heavy.

lizlux:
@rogatty's suggestions sound good. Removing the DOM lookup should help performance a lot. 

The one thing I'd add is that in cases of articles with tons of images it might just be too expensive and unnecessary to process all of them at once, regardless of how well you optimize the code. I'd be interested to see a solution where we instantiate one (or a few) media component object(s) right away and then actually lazy load the rest of them.

katanka:
I'm still working on it, but it seems like a mix of synchronous and async is good. I tested it for processing the first 4 images synchronously, and then the rest after two seconds. I'm currently working on making it short-circuit the 2 second wait if the user scrolls.

Here's my data: https://docs.google.com/spreadsheets/d/1Ks95-avwVCDZOdKo_qFlqQFa8jWjSVziilV_iHwrTNI/edit#gid=0

front/scripts/main/views/ArticleView.ts
```diff
@@ -151,7 +151,11 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.handleInfoboxes();
 			this.handlePortableInfoboxes();
-			this.lazyLoadMedia(model.get('media'));
+			// Make lazy load image processing async so it doesn't hold up DOMContentLoaded
+			this.replaceMediaPlaceholdersWithMediaComponents(model.get('media'), 4); // Process the first few images synchronously
+			Ember.run.later(this, function() {
+			 	this.replaceMediaPlaceholdersWithMediaComponents(model.get('media'), -1); // Process the rest async
```
rogatty: I'm not sure if `model` isn't lost in this context. Is it?
I'm sure that something like that would work:
```
Ember.run.later(this, (): void => {
  this.replaceMediaPlaceholdersWithMediaComponents(this.get('controller.model.media'), -1);
}, 2000);
```

Or maybe even better - just do `this.get('controller.model.media')` in the `replaceMediaPlaceholdersWithMediaComponents` method.

front/scripts/main/views/ArticleView.ts
```diff
@@ -184,12 +188,16 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 		var lazyImages = this.$('.article-media');
 
-		lazyImages.each((index: number, element: HTMLImageElement): void => {
-			this.$(element).replaceWith(this.createMediaComponent(element, model));
-		});
+		if(endIndex == -1 || endIndex > lazyImages.length){
+			endIndex = lazyImages.length;
+		}
+
+		for (var i = 0; i < endIndex; i++) {
```
rogatty: Would be nice to add a comment that this won't iterate over already replaced media items because they don't have `.article-media` class anymore.

front/scripts/main/views/ArticleView.ts
```diff
@@ -184,12 +188,16 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+			this.$(lazyImages[i]).replaceWith(this.createMediaComponent(lazyImages[i], model));
```
rogatty: `lazyImages` is an array of jQuery elements, right? No need to wrap it in `this.$()`

front/scripts/main/views/ArticleView.ts
```diff
@@ -184,12 +188,16 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
 	},
 
-	lazyLoadMedia: function (model: typeof App.ArticleModel): void {
+	replaceMediaPlaceholdersWithMediaComponents: function (model: typeof App.ArticleModel, endIndex: number): void {
 		var lazyImages = this.$('.article-media');
 
-		lazyImages.each((index: number, element: HTMLImageElement): void => {
-			this.$(element).replaceWith(this.createMediaComponent(element, model));
-		});
+		if(endIndex == -1 || endIndex > lazyImages.length){
```
rogatty: Missing spaces around brackets.

front/scripts/main/views/ArticleView.ts
```diff
@@ -184,12 +188,16 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		return component.$().attr('data-ref', ref);
 	},
 
-	lazyLoadMedia: function (model: typeof App.ArticleModel): void {
+	replaceMediaPlaceholdersWithMediaComponents: function (model: typeof App.ArticleModel, endIndex: number): void {
 		var lazyImages = this.$('.article-media');
```
rogatty: Can you rename it to `$lazyImages`? It would be obvious that this is jQuery object.

front/scripts/main/views/ArticleView.ts
katanka: That didn't work, looks like they aren't jQuery elements.

lizlux: User `$.fn.eq()` to get the jQuery element at a certain index.  https://api.jquery.com/eq/

front/scripts/main/views/ArticleView.ts
```diff
@@ -151,7 +152,17 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.handleInfoboxes();
 			this.handlePortableInfoboxes();
-			this.lazyLoadMedia(model.get('media'));
+
+			variation = Mercury.Utils.VariantTesting.getVariation('3066501061');
```
rogatty: https://github.com/Wikia/mercury/pull/939 introduces new method that has additional checks. Can you use it instead? Here is an example:
https://github.com/Wikia/mercury/pull/939/files#diff-ca00dc38f40fdedb6e32ced5881c292aR11

This PR should be merged today.

front/scripts/main/views/ArticleView.ts
```diff
@@ -132,7 +132,8 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 
 		var model = this.get('controller.model'),
 			article = model.get('article'),
-			isCuratedMainPage = model.get('isCuratedMainPage');
+			isCuratedMainPage = model.get('isCuratedMainPage'),
+			variation: any;
```
rogatty: If you keep this variable please be more verbose with the name. Variation of what?

server/views/_layouts/ember-main.hbs
```diff
@@ -97,10 +97,10 @@
 			<script src="/front/vendor/i18next/i18next.js"></script>
 			<script src="/front/vendor/vignette/dist/vignette.js"></script>
 			<script src="/front/vendor/numeral/numeral.js"></script>
-			<!-- @if WIKIA_ENVIRONMENT='prod' -->
+
 				<script src="/front/vendor/weppy/dist/weppy.js"></script>
 				<script src="/front/vendor/ember-performance-sender/dist/ember-performance-sender.js"></script>
```
rogatty: Fix indentation, please. Also no need for these empty lines.

front/scripts/main/views/ArticleView.ts
lizlux: https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#caching-jquery-objects

front/scripts/main/views/ArticleView.ts
```diff
@@ -151,7 +152,17 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.handleInfoboxes();
 			this.handlePortableInfoboxes();
-			this.lazyLoadMedia(model.get('media'));
+
+			variation = Mercury.Utils.VariantTesting.getVariation('3066501061');
+
+			if ( variation == 1 ) {
```
lizlux: Even though it's only a few lines of code, let's pull this logic out into a function called something like `handleMediaPlaceholderVariations`. Will keep this already too-big function from getting bigger.

front/scripts/main/views/ArticleView.ts
```diff
@@ -151,7 +151,15 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.handleInfoboxes();
 			this.handlePortableInfoboxes();
-			this.lazyLoadMedia(model.get('media'));
+
+			if ( Mercury.Utils.VariantTesting.getExperimentVariationNumber({prod: '0', dev: '3066501061'}) == 1 ) {
```
lizlux: Even though it's only a few lines of code, let's pull this logic out into a function called something like handleMediaPlaceholderVariations. Will keep this already too-big function from getting bigger.

lizlux:
Was part of the plan to include a variation where the first 1-4 images were loaded synchronously and the rest async? Looks like you're loading all async or all sync.

front/scripts/main/views/ArticleView.ts
```diff
@@ -364,6 +371,36 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 	},
 
 	/**
+	 * Handles the Optimizely variations for testing processing images async
+	 * Variations:
+	 * 	0 (default) => process all synchronously	
+	 *  1			=> process all async
```
lizlux: Spacing looks off here

katanka: huh it looks fine in sublime

front/scripts/main/views/ArticleView.ts
```diff
@@ -364,6 +371,36 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+	 * Variations:
+	 *	0 (default) => process all synchronously	
+	 *	1			=> process all async
+	 *	2			=> first 10 sync, rest async
+	 *	3			=> first 50 sync, rest async
+	 */
+	handleMediaPlaceholderVariations: function (): void {
+		var optimizelyVariation = Mercury.Utils.VariantTesting.getExperimentVariationNumber({prod: '0', dev: '3066501061'});
+
+		if ( optimizelyVariation == 1 ) {
```
lizlux: I realize this isn't explicitly stated in our guidelines (it probably should be) but we follow JSLint's eqeqeq rule ([configured here](https://github.com/Wikia/guidelines/blob/master/JavaScript/.jshintrc#L11)) to enforce triple equals for strict type comparison.

front/scripts/main/views/ArticleView.ts
```diff
@@ -364,6 +371,36 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+			Ember.run.later(this, function() {
+		 		this.replaceMediaPlaceholdersWithMediaComponents(model.get('media'), -1); // Process the images async
```
lizlux: would be more DRY to assign `model.get('media')` to a local var.

rogatty:
Don't forget to look at comments from https://github.com/Wikia/mercury/pull/952

front/scripts/main/views/ArticleView.ts
```diff
@@ -151,7 +151,9 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.handleInfoboxes();
 			this.handlePortableInfoboxes();
-			this.lazyLoadMedia(model.get('media'));
+			// ====================================  <-- racing stripes to increase performance
```
lizlux: lol

front/scripts/main/views/ArticleView.ts
```diff
@@ -364,6 +371,36 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+	 */
+	handleMediaPlaceholderVariations: function (): void {
+		var optimizelyVariation = Mercury.Utils.VariantTesting.getExperimentVariationNumber({prod: '0', dev: '3066501061'});
+
+		if ( optimizelyVariation === 1 ) {
+			Ember.run.later(this, function() {
+		 		this.replaceMediaPlaceholdersWithMediaComponents(model.get('media'), -1); // Process the images async
+			}, 0);
+		} else if ( optimizelyVariation === 2 ) {
+			this.replaceMediaPlaceholdersWithMediaComponents(model.get('media'), 10);
```
lizlux: `model` is undefined.

PR #937, opened 2015-06-17
HG-761 adding first content perf tracking and sending weppy in dev env

lizlux:
One of the HG team performance goals is to decrease the time it takes to show the user content. In order to measure this, I set up this performance tracking call. 

The data sent with the request will look like `"data":{"mercury::firstContent":[[10193]]}`

I also removed the conditional to only load weppy code in a production environment. The tracking calls include `env: "dev"`, which is consistent with how MW sends dev calls, so I'm assuming it will be fine to send calls from dev environment to ease development. @wladekb can you confirm?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -14,7 +14,7 @@
     "headroom.js": "0.7.0",
     "fastclick": "1.0.6",
     "jquery.cookie": "1.4.1",
-    "weppy": "Wikia/weppy",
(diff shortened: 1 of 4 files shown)


wladekb:
@lizlux that's exactly how it should be

lizlux:
This PR is dependent on https://github.com/Wikia/weppy/pull/2

kenkouot:
LGTM

PR #938, opened 2015-06-17
Fix for INT-104: Move section zero edit button to new block

perjg:


front/templates/main/article.hbs
```diff
@@ -38,6 +33,11 @@
 					{{/each}}
 				{{/collapsible-menu}}
 			{{/if}}
+			{{#if view.editButtonsVisible}}
+				<span {{action 'edit' model.title '0'}} class='zerobuttonblock'>
```
inez: If you used <div> instead of <span> maybe you wouldn't have to style is as display: block - cause divs by default are block elements.

front/styles/main/module/_edit.scss
```diff
@@ -46,8 +46,17 @@ textarea {
...
 	&.zero {
-		position: absolute;
-		right: 1.125rem;
-		top: 3.5rem;
+		position: relative;
+		right: 0.5rem;
 	}
 }
+
+.zerobuttonblock {
```
inez: I think the naming convention for CSS class in Mercury might be different - isn't it separating each word with dash?

inez: Also "wrapper" might be better than "block"

perjg:
Looks like the display: block has to be there regardless. I think it is because it is in the ::after block.

I'll fix the naming according to your suggestion.

front/templates/main/article.hbs
```diff
@@ -38,6 +33,11 @@
 					{{/each}}
 				{{/collapsible-menu}}
 			{{/if}}
+			{{#if view.editButtonsVisible}}
+				<span {{action 'edit' model.title '0'}} class='zero-button-wrapper'>
```
Grunny: Should still be `model.cleanTitle`.

front/styles/main/module/_edit.scss
```diff
@@ -46,8 +46,17 @@ textarea {
...
 }
+
+.zero-button-wrapper {
+	padding: 0.1rem;
+}
+
+.zero-button-wrapper::after {
+	content: '';
+	display: block;
+	clear: both;
```
Grunny: Properties should be alphabetised per [coding conventions](https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#propertysortorder).

front/styles/main/module/_edit.scss
```diff
@@ -46,8 +46,17 @@ textarea {
 	height: 2.2rem;
 	width: 2.2rem;
 	&.zero {
-		position: absolute;
-		right: 1.125rem;
-		top: 3.5rem;
+		position: relative;
+		right: 0.5rem;
```
Grunny: Not sure if we really need it brought in from the right at all now that it's in the content area, and removing this would make it consistent with the position of the other section edit buttons, though it doesn't make a lot of visual difference.

perjg: This wasn't really specified in the markup, I just thought it looked better slightly to the left.

inez:
LGTM, but I don't speak CSS... Hm, @kvas-damian maybe you have some opinion?

kvas-damian:
@inez This CSS can't be code reviewed without going deeper and analysing current styles. Unfortunately I have to focus on our project this week. Please find someone else.

PR #941, opened 2015-06-18
Soc 515 | Registration - username and password validation errors

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-515
https://wikia-inc.atlassian.net/browse/SOC-516

### DIFF
--- a/front/locales/en/auth-front.json
+++ b/front/locales/en/auth-front.json
@@ -2,6 +2,11 @@
   "errors" : {
     "email_already_exists": "Email is already registered on Wikia",
     "email_invalid": "Email is not valid",
-    "registration_error": "We can not complete your registration at this time."
+    "registration_error": "We can not complete your registration at this time.",
+    "username_already_exists": "Username is taken",
+    "username_unavailable": "TODO antispoof blocked",
+    "username_ip_address": "TODO username_ip_address",
+    "username_exceeds_max_length": "TODO username_exceeds_max_length 50 chars",
+    "username_illegal_characters": "Username includes invalid characters"
   }
 }
--- a/front/locales/en/auth-front.json
+++ b/front/locales/en/auth-front.json
@@ -7,6 +7,8 @@
     "username_unavailable": "TODO antispoof blocked",
     "username_ip_address": "TODO username_ip_address",
     "username_exceeds_max_length": "TODO username_exceeds_max_length 50 chars",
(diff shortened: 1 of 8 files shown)


front/locales/en/auth-front.json
```diff
@@ -2,6 +2,13 @@
   "errors" : {
     "email_already_exists": "Email is already registered on Wikia",
     "email_invalid": "Email is not valid",
-    "registration_error": "We can not complete your registration at this time."
+    "registration_error": "We can not complete your registration at this time.",
```
bkoval: I think in all the i18n files we use dashes instead of underscores for separating the words :)

server/views/signup.hbs
```diff
@@ -5,12 +5,12 @@
 			<label class="floating-label" for="signupEmail">{{i18n "auth:signup.email-label"}}</label>
 		</div>
 		<div class="input-container">
-			<input type="text" name="username" id="signupUsername" required />
+			<input type="text" name="username" id="signupUsername" maxlength="50" required />
```
bkoval: Do you think putting these values into some kind of config would make sense? In this case I don't think we need since it's just used in one / two places but it might increase maintainability of this in the future :)

front/locales/en/auth-front.json
```diff
@@ -2,6 +2,13 @@
...
+    "username_already_exists": "Someone already has this username. Try a different one!",
+    "username_unavailable": "Oops, your username can only contain letters and numbers from a single alphabet and must include at least one letter. Please try another name.",
+    "username_ip_address": "Oops, your username can only contain letters and numbers from a single alphabet and must include at least one letter. Please try another name.",
```
bkoval: Should this string be the same as the one above it? :)

front/locales/en/auth-front.json
kvas-damian: Those are error codes returned by Helios. I decided to keep its original names and break our guidelines

front/locales/en/auth-front.json
kvas-damian: According to messages provided in ticket, yes

bkoval:
LGTM :+1:

jsutterfield:
LGTMx2

kvas-damian:
Will open new PR because SOC-684 was just merged to dev.

PR #942, opened 2015-06-18
SOC-878 Adding active state to birthdate field

bkoval:
@kvas-damian @rafalkalinski https://wikia-inc.atlassian.net/browse/SOC-878

This ticket was filed to fix the state on a placeholding input type=date, but I added an active state on the new date field, so the PR is targeted at SOC-480.

### DIFF
--- a/front/scripts/auth/common/BirthdateInput.ts
+++ b/front/scripts/auth/common/BirthdateInput.ts
@@ -38,7 +38,7 @@ class BirthdateInput {
 			inputContainer = <HTMLElement> this.wrapper.parentElement,
 			target: HTMLElement;
 
-		inputContainer.addEventListener('focus', (() => {
+		inputContainer.addEventListener('focus', ((): void => {
 			target = <HTMLElement> event.target;
 			if (target === this.realInput) {
 				this.realInput.type = 'hidden';
@@ -48,7 +48,7 @@ class BirthdateInput {
 			this.wrapper.classList.add('focused');
 		}).bind(this), true);
 
-		inputContainer.addEventListener('blur', (() => {
+		inputContainer.addEventListener('blur', ((): void => {
 			this.wrapper.classList.remove('focused');
 		}).bind(this), true);
 	}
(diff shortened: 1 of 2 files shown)


front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -34,13 +34,23 @@ class BirthdateInput {
...
+		var firstInput: HTMLInputElement = <HTMLInputElement>this.fakeInputs[0],
+			inputContainer = <HTMLElement> this.wrapper.parentElement,
+			target: HTMLElement;
 
-		this.realInput.addEventListener('focus', (() => {
-			this.realInput.type = 'hidden';
-			this.wrapper.classList.remove('hide');
-			firstInput.focus();
-		}).bind(this));
+		inputContainer.addEventListener('focus', (() => {
```
rogatty: Missing return type :P

front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -34,13 +34,23 @@ class BirthdateInput {
...
+			target = <HTMLElement> event.target;
+			if (target === this.realInput) {
+				this.realInput.type = 'hidden';
+				this.wrapper.classList.remove('hide');
+				firstInput.focus();
+			}
+			this.wrapper.classList.add('focused');
+		}).bind(this), true);
+
+		inputContainer.addEventListener('blur', (() => {
```
rogatty: Missing return type.

rafalkalinski:
except missing return types, lgtm :)

kvas-damian:
I see a lot is happening here, can't we just update styles in: https://github.com/Wikia/mercury/blob/dev/front/styles/auth/module/_forms.scss#L48 to reflect styleguide animations?
Or is that because we're replacing input with 3 different?

bkoval:
The second thing. Input[type=date] will be replaced with three number inputs, that's why we need to apply these styles to their wrapper which acts like a one big input consisting of three smaller ones.

kvas-damian:
LGTM

PR #943, opened 2015-06-18
DAT-2897 handle article references

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2897

@rogatty

### DIFF
--- a/front/scripts/main/views/ApplicationView.ts
+++ b/front/scripts/main/views/ApplicationView.ts
@@ -102,12 +102,27 @@ App.ApplicationView = Em.View.extend({
 	 * Determine if we have to apply special logic to the click handler for MediaWiki / UGC content
 	 */
 	shouldHandleClick: function (target: EventTarget): boolean {
-		var $target = $(target);
+		var $target: JQuery = $(target),
+			isReference: boolean = this.targetIsReference(target);
 
 		return (
 			$target.closest('.mw-content').length &&
 				// ignore polldaddy content
-			!$target.closest('.PDS_Poll').length
+			!$target.closest('.PDS_Poll').length &&
+				// don't need special logic for article references
+			!isReference
+		);
+	},
+
+	/**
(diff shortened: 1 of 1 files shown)


front/scripts/main/views/ApplicationView.ts
```diff
@@ -102,12 +102,27 @@ App.ApplicationView = Em.View.extend({
...
+	},
+
+	/**
+	 * Determine if the clicked target is an reference/in references list (in text or at the bottom of article)
+	 */
+	targetIsReference: function (target: EventTarget): boolean {
+		var $target: JQuery = $(target);
+
+		return !!(
+			$target.closest('.references').length ||
```
rogatty: `closest` seems a little bit risky. What if user creates another element with class like that? This word isn't too unique. Would `parent` work for all cases?

front/scripts/main/views/ApplicationView.ts
```diff
@@ -102,12 +102,27 @@ App.ApplicationView = Em.View.extend({
...
-		var $target = $(target);
+		var $target: JQuery = $(target),
+			isReference: boolean = this.targetIsReference(target);
 
 		return (
 			$target.closest('.mw-content').length &&
 				// ignore polldaddy content
-			!$target.closest('.PDS_Poll').length
+			!$target.closest('.PDS_Poll').length &&
+				// don't need special logic for article references
```
rogatty: Fix indentation, please. Here and above.

front/scripts/main/views/ApplicationView.ts
dianafa: it would not. the references at the bottom of the page are nested more deeply, and using `parents` is even worse than closest.
However I will change it for the reference in text, thanks.

rogatty: This might come back to us as a P3 one day but seems like there is no simple solution available.

rogatty:
LGTM

PR #944, opened 2015-06-18
SOC-909 Redirect to old signup on /signup if config flag not set

bkoval:
Since we can see /signup on prod now and we should handle it somehow, let's redirect to the old /wiki/Special:UserSignup for now.

For development purposes set "enableNewSignup = true" in your localSettings to be able to see the new signup page.


lizlux:
LGTM!

bkoval:
Let's treat it as a beginning of a discussion though, as this make it harder to test new signup on sandboxes.

lizlux:
Hmmm good point. In that case, one workaround could be to set and get `M.prop('environment')` like the main ember app does, and check that it's not 'prod'.

kvas-damian:
'prod' check is must have if we want to merge it. On the other hand we don't link to this page anywhere so I don't expect users coming here. Is it worth preventing this page from being displayed for users that dig into our code?

lizlux:
IMO it's bad UX for someone coming to that URL directly expecting to be able to register. That said, it is definitely an edge case, and I don't think too much time should be spent on a solution.

rafalkalinski:
From my perspective: this code is not a real solution, because even if it fixes the problem, it could cause another one. We will talk about it more for sure, but I think that in this particular case (edge case; page not linked from anywhere) UX is not our first priority. That's why I'm closing this PR.

frankfarmer:
Is it difficult to simply configure the enableNewSignup flag on sandboxes?

lizlux:
This is my suggestion https://github.com/Wikia/mercury/pull/969

kvas-damian:
It's not difficult. It basically has no sense to protect page that is practically not available for users (not linked anywhere). Additionally we're going to release working signup in 1-2 weeks.

PR #946, opened 2015-06-19
CONCF-608 encode category and section name before sending to API

bognix:
ping @rogatty @vforge


rogatty:
`encodeURIComponent` can throw exceptions. We should check if this can happen here and if we need to wrap it in `try {} catch()`.

bognix:
String which is throwing exceptions: \uD800

I added section with name "\uD800" - exception wasn't thrown. It was nicely encoded.
I added category with name "\uD800" (actually with "%5CD800 because I can't add not encoded category) - exception wasn't thrown, worked nicely
I added article with name "\uD800" to category with name "\uD800" - exception wasn't thrown, worked nicely.

I can add exception because better save than sorry.
The question is how we should handle it?

vforge:
LGTM

rogatty:
LGTM

PR #953, opened 2015-06-19
HG-764: Add "cookie" to "vary" header

inez:
This is required by Fastly's varnishes, specifically it tells them that particular object should be cached separately (varied) based on cookie presence - however fastly has additonal logic to whitelist which cookies matter (for instance wikicities_session does matter, but some ad related cookie does not).

More info:
https://wikia-inc.atlassian.net/browse/HG-764
https://wikia-inc.atlassian.net/browse/OPS-6571
https://wikia-inc.atlassian.net/browse/HG-736


inez:
Assigning to Kenislaw

kenkouot:
L ^_^ G ^_^ T ^_^ M ^_^

PR #955, opened 2015-06-21
Portable infobox component

dianafa:
This PR doesn’t change anything but that infoboxes are now the Ember components. Their body, however, is still just HTML inserted into article with no difference for their inside components like groups, fields, images, titles etc. However, since now the infobox logic is placed in one place and managed internally, it’s much easier to improve infoboxes in next iterations, when they will be returned from API as an object like the media items are handled or in other, more proper form.

@hakubo @RafLeszczynski


front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
+/// <reference path="../app.ts" />
+/// <reference path="../mixins/ArticleContentMixin.ts" />
+/// <reference path="../mixins/ViewportMixin.ts" />
+'use strict';
+
+App.PortableInfoboxComponent = Em.Component.extend(App.ArticleContentMixin, App.ViewportMixin, {
+	classNames: ['portable-infobox'],
+	classNameBindings: ['collapsed'],
+	expandButtonClass: 'portable-infobox-expand-button',
+	layoutName: 'components/portable-infobox',
```
rogatty: It's the default path, no need to put it here explicitly.

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
...
+	height: null,
+	infoboxHTML: '',
+	collapsed: false,
+
+	/**
+	 * @desc determines if this infobox is
+	 * a short one or a long one (needs collapsing)
+	 * @return boolean true if infobox is long
+	 */
+	isLongInfobox: Em.computed({
```
rogatty: `Em.computed('collapsedHeight', 'height', {`

http://emberjs.com/api/classes/Ember.ComputedProperty.html
>In the following example we declare a computed property (by calling .property() on the fullName function) and setup the property dependencies (depending on firstName and lastName). The fullName function will be called once (regardless of how many times it is accessed) as long as its dependencies have not changed. Once firstName or lastName are updated any future calls (or anything bound) to fullName will incorporate the new values.

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
...
+		get(): boolean {
+			var collapsedHeight = this.get('collapsedHeight'),
+				height = this.get('height');
+
+			return height > collapsedHeight;
```
rogatty: You can use http://emberjs.com/api/classes/Ember.computed.html#method_gt instead.

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
...
+	}),
+
+	/**
+	 * @desc return height which should have the collapsed infobox,
+	 * basing on the viewport width.
+	 * It's taken from 9/16 proportions of screen (width * 16 / 9 + 100px).
+	 * We want to always show the image AND some other infobox informations to
+	 * indicate that this is infobox, not only an ordinary image.
+	 */
+	collapsedHeight: Em.computed('viewportDimensions.width', function() {
```
rogatty: `function (): number`

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
...
+		var deviceWidth = this.get('viewportDimensions.width');
+
+		return Math.floor(deviceWidth * 16 / 9) + 100;
+	}),
+
+	handleCollapsing: function(): void {
```
rogatty: Use spaces between `function` and `()` - everywhere ;)

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
...
+		var collapsedHeight = this.get('collapsedHeight'),
+			$this = this.$();
```
rogatty: What's the point of creating variables if they're used only once and are not so long (no need to shorten the line width)?

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,94 @@
...
+	 * Changes 'collapsed' property.
+	 */
+	onInfoboxClick: function(): void {
+		var body: HTMLElement,
+			button: HTMLElement,
+			collapsedHeight: number,
+			expandButtonClass: string,
+			scrollTo: (top?: boolean) => void,
+			collapsed = this.get('collapsed'),
+			$target = $(event.target),
```
rogatty: `event` will be undefined.

front/scripts/main/components/PortableInfoboxComponent.ts
dianafa: We came out it has to be here due to fact, that this component is created artificially.

front/scripts/main/components/PortableInfoboxComponent.ts
dianafa: nice! :) thank you

front/scripts/main/components/PortableInfoboxComponent.ts
dianafa: it's more clear and visible. At some cases it's no necessary, but at some help with readability.

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,92 @@
...
+			collapsed = this.get('collapsed'),
+			$target = $(event.target);
+
+		if (!$target.is('a') && !collapsed) {
+			body = window.document.body;
+			expandButtonClass = this.get('expandButtonClass');
+			button = this.$('.' + expandButtonClass)[0];
+			scrollTo = body.scrollIntoViewIfNeeded || body.scrollIntoView;
+
+			this.$().height(this.get('collapsedHeight'));
```
hakubo: this looks very similar to:https://github.com/Wikia/mercury/pull/955/files#diff-20f218083878a332aead8f28926ef4b8R43
how about handleCollapsing would cover collapse/expand and could be called from here?

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,92 @@
+/// <reference path="../app.ts" />
+/// <reference path="../mixins/ArticleContentMixin.ts" />
+/// <reference path="../mixins/ViewportMixin.ts" />
+'use strict';
+
+App.PortableInfoboxComponent = Em.Component.extend(App.ArticleContentMixin, App.ViewportMixin, {
+	classNames: ['portable-infobox'],
```
hakubo: don't it have portable-infobox-component class by default?

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,92 @@
...
+			collapsedHeight: number,
+			expandButtonClass: string,
+			scrollTo: (top?: boolean) => void,
+			collapsed = this.get('collapsed'),
+			$target = $(event.target);
+
+		if (!$target.is('a') && !collapsed) {
+			body = window.document.body;
+			expandButtonClass = this.get('expandButtonClass');
+			button = this.$('.' + expandButtonClass)[0];
```
hakubo: why do you need button here?

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,92 @@
...
+	},
+
+	/**
+	 * @desc In case of long infobox, setups click
+	 * handling function to this infobox component.
+	 */
+	didInsertElement: function () {
+		if (this.get('isLongInfobox')) {
+			this.handleCollapsing();
+			this.$().click((event: JQueryEventObject) => {
```
hakubo: this.$().click(this.onInfoboxClick) you don't need another function on stack

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (!$target.is('a') && !collapsed) {
+			body = window.document.body;
+			expandButtonClass = this.get('expandButtonClass');
+			button = this.$('.' + expandButtonClass)[0];
+			scrollTo = body.scrollIntoViewIfNeeded || body.scrollIntoView;
+
+			this.$().height(this.get('collapsedHeight'));
+			scrollTo.apply(button);
+		} else {
+			this.$().height('auto');
```
hakubo: If I click on A tag
should infobox be expanded?

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		this.toggleProperty('collapsed');
```
hakubo: is this class used anywhere?

front/scripts/main/components/PortableInfoboxComponent.ts
dianafa: no, probably because of 'artificial' creating and inserting

front/scripts/main/components/PortableInfoboxComponent.ts
dianafa: we want to scroll to it, when is kept in variable, code is more clean

front/scripts/main/components/PortableInfoboxComponent.ts
dianafa: yes, it adds: 
  overflow: hidden;
  position: relative;
when inbox is collapsed.

front/scripts/main/components/PortableInfoboxComponent.ts
```diff
@@ -0,0 +1,93 @@
...
+	 * @desc return height which should have the collapsed infobox,
+	 * basing on the viewport width.
+	 * It's taken from 9/16 proportions of screen (width * 16 / 9 + 100px).
+	 * We want to always show the image AND some other infobox informations to
+	 * indicate that this is infobox, not only an ordinary image.
+	 */
+	collapsedHeight: Em.computed('viewportDimensions.width', function (): number {
+		var deviceWidth = this.get('viewportDimensions.width');
+
+		return Math.floor(deviceWidth * 16 / 9) + 100;
```
RafLeszczynski: It's a small improvement but consider moving the `100` and `16 / 9` to component properties so they are not hardcoded inside a method

PR #963, opened 2015-06-23
HG-768 adding query string param for bucky sampling

lizlux:
This is a convenience for QA and troubleshooting. 

I'm not entirely sure it's in the right place though, b/c that function is called `prepareArticleData`. I'm open to suggestions on where to put it.


rogatty:
Yeah, we need to clean up this file. I hope that we'll do at least a part of it when working on routable Curated Main Page as it will involve further decoupling from articles.

For now it LGTM.

PR #970, opened 2015-06-24
SOC-517 | Registration - birthdate validation

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-517


front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -67,6 +67,12 @@ class BirthdateInput {
 	 * Set the value for the input that will ultimately be saved upon form submission
 	 */
 	private setRealValue(): void {
-		this.realInput.value = this.year.value + '-' + this.month.value + '-' + this.day.value;
+		this.realInput.value = this.padleftWithZeros(this.year.value, this.year.maxLength) + '-' +
+			this.padleftWithZeros(this.month.value, this.month.maxLength) + '-' +
```
bkoval: padLeft? :)

front/scripts/auth/common/BirthdateInput.ts
```diff
@@ -67,6 +67,12 @@ class BirthdateInput {
...
+			this.padleftWithZeros(this.day.value, this.day.maxLength);
+	}
+
+	private padleftWithZeros(input: string, length: number): string {
```
bkoval: Could you add a doc to this method? Seems like a code is not self-explanatory in an obvious way, so I think this would be beneficial :)

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -29,6 +29,8 @@ class SignupForm {
 		Array.prototype.forEach.call( errorNodes, (node: HTMLElement): void => {
 			if (node.tagName === 'INPUT') {
 				node.classList.remove('error');
+			} else if (node.classList.contains('input')) {
+				node.classList.remove('error');
 			} else {
 				node.parentNode.removeChild( node );
```
bkoval: Is typescript fine with not informing it about the type of the parent? :)

front/scripts/auth/common/BirthdateInput.ts
rafalkalinski: Shouldn't it be "padLeftWithZoros"?
Or even "padLeft" where additional parameter would be the parameter for the .join method

front/scripts/auth/signup/SignupForm.ts
kvas-damian: ```parentNode``` is ```Node``` type. What's the problem?

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -48,9 +50,14 @@ class SignupForm {
 
 	private displayFieldValidationError(err: HeliosError): void {
 		var errorNode: HTMLElement = this.createValidationErrorHTMLNode(err.description),
-			input: HTMLFormElement = <HTMLFormElement> this.form.elements[err.additional.field];
+			input: HTMLFormElement = <HTMLFormElement> this.form.elements[err.additional.field],
+			specialFieldContainer: HTMLElement;
 		input.parentNode.appendChild(errorNode);
-		input.classList.add('error');
+		if (specialFieldContainer = <HTMLElement> (<HTMLElement> input.parentNode).querySelector('.input')) {
```
rafalkalinski: Is it totally okay here to base on a fact that only birthdate parent has "input" class?
If yest, then please add some comment in the code, because will be not so obvious for someone in a month from now :)

rafalkalinski:
one minor comment, but LGTM

front/scripts/auth/signup/SignupForm.ts
kvas-damian: It is ok. We're imitating input here, so we need to add error class there.
I'll add some info in code

bkoval:
SOC-517 has been merged to dev as part of SOC-879 here: https://github.com/Wikia/mercury/pull/981

PR #976, opened 2015-06-24
SOC-913 | Retire wikicitiesUserID

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-913
+ small cleanup in routes.ts

### DIFF
--- a/server/facets/operations/prepareArticleData.ts
+++ b/server/facets/operations/prepareArticleData.ts
@@ -47,7 +47,7 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 		result.weppyConfig.samplingRate = result.queryParams.buckySampling / 100;
 	}
 
-	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
+	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : undefined;
 }
 
 export = prepareArticleData;
--- a/server/facets/operations/prepareArticleData.ts
+++ b/server/facets/operations/prepareArticleData.ts
@@ -47,7 +47,7 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 		result.weppyConfig.samplingRate = result.queryParams.buckySampling / 100;
 	}
 
-	result.userId = request.state.wikicitiesUserID ? request.state.wikicitiesUserID : 0;
+	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
 }
 
 export = prepareArticleData;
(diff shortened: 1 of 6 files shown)


server/facets/operations/prepareArticleData.ts
```diff
@@ -47,7 +47,7 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 		result.weppyConfig.samplingRate = result.queryParams.buckySampling / 100;
 	}
 
-	result.userId = request.state.wikicitiesUserID ? request.state.wikicitiesUserID : 0;
+	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
```
bkoval: I'm not sure but if not authenticated, userId should be =0, right? This would let us simplify the code to have just +	result.userId = request.auth.credentials.userId;

bkoval:
I'll rebuild Jenkins since the error it outputs is strange.

server/routes.ts
```diff
@@ -144,18 +148,22 @@ articlePagePaths = [
...
-authenticatedRoutes = [];
-
 unauthenticatedRoutes = unauthenticatedRoutes.map((route) => {
 	return Hoek.applyToDefaults(unauthenticatedRouteConfig, route);
 });
+authenticatedRoutes = authenticatedRoutes.map((route) => {
+	return Hoek.applyToDefaults(authenticatedRouteConfig, route);
+});
 
 routes = unauthenticatedRoutes.concat(authenticatedRoutes);
```
rafalkalinski: I can't get the sense of unauthenticatedRoutes and authenticatedRoutes arrays here, since the one action for them is   concatenation them both.

rafalkalinski:
LGTM

server/routes.ts
kvas-damian: unauthenticated routes don't need to verify if ```access_token``` is correct

server/facets/operations/prepareArticleData.ts
```diff
@@ -47,7 +47,7 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 		result.weppyConfig.samplingRate = result.queryParams.buckySampling / 100;
 	}
 
-	result.userId = request.state.wikicitiesUserID ? request.state.wikicitiesUserID : 0;
+	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : undefined;
```
rogatty: This looks strange. Wouldn't be better to:
- set it to null OR
- not set it at all?

kvas-damian: Fixed in: 4ee07cf

PR #977, opened 2015-06-24
SOC-898 New auth flow in wiki language

bkoval:
@rafalkalinski @kvas-damian
https://wikia-inc.atlassian.net/browse/SOC-898

I cannot link the new signup url yet since we don't want this page to be live but we can QA the login path (signup path works in the same way).


front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -9,7 +9,9 @@ App.LoginIconComponent = Em.Component.extend({
 
 	click: function (): void {
 		if (Mercury.wiki.enableNewAuth) {
-			window.location.href = '/join?redirect=' + encodeURIComponent(window.location.href);
+			window.location.href = '/join?redirect=' +
+			encodeURIComponent(window.location.href) +
+			'&uselang=' + Mercury.wiki.language.content || 'en';
```
kvas-damian: can't we just skip ```uselang``` when there is no language?

server/lib/AuthUtils.ts
```diff
@@ -29,6 +29,14 @@ export function getLoginUrlFromRedirect(redirect: string): string {
 	return url.format(forgotPasswordUrlObj);
 }
 
+export function getNewLoginUrl(request: Hapi.Request): string {
```
kvas-damian: I don't like ```new``` in method names. How would we name this method when newer login comes?
Can't it be just ```getLoginUrl```?

rafalkalinski: good point!

front/scripts/main/components/LoginIconComponent.ts
rafalkalinski: and it would be good if the uselang value will be uri-encoded :)

kvas-damian:
LGTM

kvas-damian:
@bkoval Please resolve conflicts.

BTW I don't see where you pass language to Helios

kvas-damian:
LGTM

bkoval:
Closing to file a new one, trageted at a different branch.

PR #986, opened 2015-06-25
SOC-563 Geo-targeting capabilities for New Auth

bkoval:
@kvas-damian @jsutterfield This gives us possibility of reading geo-cookie set by Varnish (and reading any cookie we want in the future) as requested in https://wikia-inc.atlassian.net/browse/SOC-563


front/scripts/auth/common/Cookie.ts
```diff
@@ -0,0 +1,19 @@
...
+			if (cookieStart != -1) {
+				cookieStart = cookieStart + cookieName.length + 1;
+				cookieEnd = cookie.indexOf(';', cookieStart);
+				if (cookieEnd == -1) {
+					cookieEnd = cookie.length;
+				}
+				return decodeURIComponent(cookie.substring(cookieStart, cookieEnd));
+			}
+		}
+		return '';
```
kvas-damian: Empty string when cookie doesn't exist?

front/scripts/auth/common/Cookie.ts
```diff
@@ -0,0 +1,19 @@
+class Cookie {
```
kvas-damian: Is it possible to write some unit tests for this class?

front/scripts/auth/common/Geo.ts
```diff
@@ -0,0 +1,22 @@
+interface GeoCookie {
+	city?: string;
+	country?: string;
+	continent?: string;
+}
+
+class Geo {
+	cookieName: string = 'Geo';
+
+	country: string;
```
kvas-damian: What do you think about adding getter for this property?

kvas-damian:
For consideration. Is it possible to convert ```Geo``` to utility that could be shared in Ember and Auth pages?

front/scripts/auth/common/Geo.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	constructor () {
+		var geoCookie: string = Cookie.get(this.cookieName),
+			parsedGeoCookie: GeoCookie;
+		try {
```
kvas-damian: ```geoCookie``` can be null. Don't we need an if statement?

bkoval: parsing null will not throw but this is cover by an if statement below :)

front/scripts/auth/common/Geo.ts
```diff
@@ -0,0 +1,32 @@
+interface GeoCookie {
+	city?: string;
```
jsutterfield: Do we care about the city part of the cookie? It's not used for anything (that I know of), probably b/c the value is always "FIXME".

front/scripts/auth/common/Geo.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	country: string;
+
+	constructor () {
+		var geoCookie: string = Cookie.get(this.cookieName),
+			parsedGeoCookie: GeoCookie;
+		try {
+			parsedGeoCookie = JSON.parse(geoCookie);
+		} catch (e) {
+			parsedGeoCookie = null;
```
jsutterfield: Can we just set `parsedGeoCookie` to `{}` here and eliminate lines 21-24?

jsutterfield: Sorry, I forgot how to read for a second. Thankfully I regained that ability and saw your explanation above. Disregard :)

front/scripts/auth/common/Cookie.ts
bkoval: Done!

front/scripts/auth/common/Geo.ts
bkoval: Done!

front/scripts/auth/common/Geo.ts
bkoval: I was thinking about that, but since it always gets set I think it's beneficial to inform there will be a field called "city" in javascript object after parsing, even if it's always set tu "FIXME" (mostly in case of calling Object.keys or iterating over keys in some other way). What do you think?

test/specs/front/scripts/auth/common/Geo.js
```diff
@@ -0,0 +1,14 @@
...
+	ok(typeof window.Geo === 'function');
+});
+
+QUnit.test('Geo class exposes country getter', function () {
+	ok(typeof new window.Geo().getCountry === 'function');
+});
+
+QUnit.test('Country getter returns string with a country name if Geo cookie is set', function () {
+	document.cookie = 'Geo={%22city%22:%22FIXME%22%2C%22country%22:%22PL%22%2C%22continent%22:%22EU%22}';
+	equal(new window.Geo().getCountry(), 'PL');
```
kvas-damian: Probably we should mock ```Cookie``` class.

lizlux:
Let's add a `getContinent` method too while we're at it.

front/scripts/auth/common/Geo.ts
jsutterfield: Sure that makes sense. I guess it doesn't hurt to leave it in :+1:

kenkouot:
So this is getting into code duplication in the same repo. For the Ember app, we load a different set of dependencies and hence have a cookie parsing jQuery plugin. Hence we didn't have to implement something like this to simply parse a cookie and cache it's results.

I do not think we should have two client side methods for parsing this cookie. I think it should also be safe to assume a standard set of dependencies across Wikia's new front ends. Doing so will prevent us from reinventing the wheel (or cookie parsing) every time.

As it stands, I much prefer this implementation: https://github.com/Wikia/mercury/pull/988/files
The prerequisite to this implementation is requiring jQuery, jQuery Cookie and the Mercury set of scripts.

In the near future, I do plan to abstract all the Mercury (non Ember) scripts out as it's own utility library and then we can encapsulate this a bit better.

bkoval:
@kenkouot @kvas-damian 
I agree with the code duplication problem, I'm also not super happy about that. However:

1) We do not make any calls to jQuery in the auth code so far
2) This code is unit-tested and allows for easy refactor
3) This pull request happened first, before #988 (a bit of luck) and it had business reasons to be working solely on Auth pages, that's why it doesn't have any external dependencies
4) We're actually looking into dropping mercury.js from auth pages, since soon they won't be served only to mercury users and browser cache reusability won't be possible at all
5) "Use jQuery everywhere" seems sensible, but currently it is getting cocnatenated with different vendor files in main app so it's not reusable in Auth Pages from browser / varnish cache.
6) We're planning to release the signup feature early this week, so any refactoring (pretty straightforward since it's just two super simple classes) / project restructuring will happen after we actually have to have it working.

kvas-damian:
LGTM

jsutterfield:
LGTM

PR #988, opened 2015-06-25
HG-777 adding geo cookie reading

lizlux:
We'll need geolocation data for general international features, so this adds it to `M.prop`. 

Note - the cookie isn't set on local dev environments so it will either need to be set manually or `M.prop('geo')` will be `undefined`. 

@mari-mcmurtrie @inez


bkoval:
Hey @lizlux We need the same functionality for the Auth pages and it was implemented yesterday (now in CR). It looks pretty similar, except it doesn't have any dependencies on the Mercury's Front-End code (jQuery, Mercury). We could think of a way to make it reusable for the main app as well and this way it will be easier to cover it with unit tests which I'm planning to do today :)

front/scripts/main/app.ts
```diff
@@ -156,3 +157,20 @@ App.initializer({
...
+ */
+App.initializer({
+	name: 'geo',
+	after: 'setupTracking',
+	initialize(container: any, application: typeof App): void {
+		var geoCookie = $.cookie('Geo');
+		if (geoCookie) {
+			M.prop('geo', JSON.parse(geoCookie));
+		} else {
+			Ember.debug('Geo cookie is not set');
```
kenkouot: Does it need this check? It can just set the cookie and the it will either set it or be undefined. Then invocations of it can decide how to handle failure to grab it.

lizlux: If `geoCookie` is undefined, `JSON.parse` will throw an exception. That said, I could change line 168 to: `var geoCookie = $.cookie('Geo') || null;` and `JSON.parse` can accept the `null`. 

The reason I like using `Ember.debug` is b/c on our local machines the cookie won't be there and I don't want engineers to be confused as to why their geo-related code doesn't work.

kenkouot:
LGTM

PR #994, opened 2015-06-26
HG-772 - fixed table bug

katanka:
Now handles links in tables correctly.

### DIFF
--- a/front/scripts/main/views/ArticleView.ts
+++ b/front/scripts/main/views/ArticleView.ts
@@ -299,7 +299,7 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 	},
 
 	handleTables: function (): void {
-		var $tables = this.$('table:not([class*=infobox], .dirbox)').not('table table'),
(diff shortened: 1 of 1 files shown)


front/scripts/main/views/ArticleView.ts
```diff
@@ -319,7 +319,7 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 	},
 
 	handleTables: function (): void {
-		var $tables = this.$('table:not([class*=infobox], .dirbox)').not('table table'),
+		var $tables = $('table:not([class*=infobox], .dirbox)').not('table table'),
```
rogatty: Aren't all tables inside an article? `this.$()` refers to the ArticleView top element and allows to traverse inside it.

katanka: Yes, this is weird. Using `this.$()` causes the bug, `$()` does not. I also tried 
`$('.article-content table:not etc')`, but that causes the bug as well.

kenkouot:
Merge conflict!

PR #998, opened 2015-06-29
Make use of buildUrl when constructing API calls

Grunny:
Make use of buildUrl when constructing API calls now that it is available so we
support all environments including running it locally.

/cc @rogatty


rogatty:
LGTM

PR #999, opened 2015-06-29
Hg 753 - Adding ability to load first article async

lizlux:
This article does a few things:
* Adds the ability to toggle between loading article content in the page source and asynchronously
* Implements loading all article content in the page source when `?_escaped_fragment_=` is present in the URL
* Includes the appropriate meta tag in the page source when article content is loaded asynchronously

Note that this PR assumes we'll be enabling async article loading on a per-wiki basis. Once https://wikia-inc.atlassian.net/browse/HG-755 is figured out we can modify that scheme.


kenkouot:
lgtm outside of conflict

rogatty:
@lizlux is this the expected behaviour? It looks odd. Previously we were overlaying spinner over the content and now we render it into application.hbs `{{outlet}}`.

>After transitioning into a loading substate, the corresponding template for that substate, if present, will be rendered into the main outlet of the parent route, e.g. foo.bar.loading's template would render into foo.bar's outlet. (This isn't particular to loading routes; all routes behave this way by default.)

from http://guides.emberjs.com/v1.12.0/routing/loading-and-error-substates/

lizlux:
@rogatty I was hoping to fix the fact that before the first article loads, if we're loading it asynchronously, there's no loading spinner. That will have to wait for a different ticket though, as this ticket was not specifically for that. You're implementation is fine for now. Thanks!

kenkouot:
This looks good but please don't merge yet as it will interfere with some tests we have planned.

lizlux:
Checked with @kenkouot, it's fine to merge while tests are in progress.

PR #1000, opened 2015-06-29
HG-782 - added state check back

katanka:
There are so many commits because I had to test on sandbox.


inez:
@katanka I suggest you squash those test/debug commits into one before merging.
Also - why you had to test on sandbox and not just devbox?

katanka:
There was some issue with domains, to use the edit functionality I needed to be on a *.wikia.com. I tried spoofing it, it didn't work.

PR #1002, opened 2015-06-30
Added new route for Wikia Japan homepage with placeholder content

perjg:
Initial commit for work on creating a new Wikia Japan homepage served via the mercury server app.

Added the following:
* Hapi routes and templates for the homepage with placeholder content
* Gulp tasks and changes to compile the homepage

Base imports for scss files are loaded directly from mercury (e.g. foundation grid).

Client side scripting set up with TypeScript, but can be changed to ES6 via babeljs ES5 transpiler if desired.

See INT-121 and INT-115 for details.


inez:
@perjg It's not mergeable at this moment.
@perjg It looks good to me, but given the scope of change let's have someone from mercury team review it as well. @kenkouot can you take a look? (reassigning to you in JIRA :))

perjg:
@inez Thanks. I'll take care of merging with dev once the code review is done.

inez:
@perjg It actually should be mergeable to dev before it is could reviewed - so reviewer is reviewing most up to date version that is most likely to be merged. Otherwise it is slightly missing a point - is reviewer is reviewing something that is known to be changed soon (to make it mergeable).

perjg:
@inez Got it. This change adds routes to the root controller, and easily ends up in a state that is not mergable, but the changes needed to merge are usually trivial. I'll have a look at it later today.

gulp/tasks/scripts-homepage.js
```diff
@@ -0,0 +1,43 @@
...
+			if (gutil.env.testing && environment.isProduction) {
+				console.error('Build contains some typescript errors/warnings');
+				process.exit(1);
+			}
+		})
+		.pipe(concat(folder + '.js'))
+		.pipe(gulpif(environment.isProduction, uglify()))
+		.pipe(gulpif(!environment.isProduction, sourcemaps.write('.' , {
+			includeContent: true,
+			sourceRoot: '/ja/scripts/',
```
rogatty: How about moving it to the config? Can (and should?) it live in the `/front/` path?

Routing is pretty messy right now, see https://github.com/Wikia/wikia-vcl/blob/master/wikia.com/control-skin.vcl#L66. In my opinion we should try to keep it as simple as possible.

server/facets/homepage/homepageView.ts
```diff
@@ -0,0 +1,84 @@
+/// <reference path='../../../typings/hapi/hapi.d.ts' />
+
+import caching = require('../../lib/Caching');
+import localSettings = require('../../../config/localSettings');
+import url = require('url');
+
+module homepageView {
```
kenkouot: Module names are PascalCased

server/facets/homepage/index.ts
```diff
@@ -0,0 +1,25 @@
...
+		lang = i18n.lng();
+
+	context = deepExtend(
+		homepageView.getDefaultContext(request),
+		{
+			title: 'Wikia Japan',
+			i18nContext: i18n
+		});
+
+ 	console.log(context);
```
kenkouot: dirty log

server/routes.ts
```diff
@@ -131,6 +131,20 @@ unauthenticatedRoutes = [
...
 		path: '/editorPreview',
 		handler: require('./facets/editorPreview')
+	},
+	/**
+	 * Homepage Routes
+	 * @description The following routes should be related to home page
+	 */
+	{
+		method: 'GET',
+		path: '/ja',
```
kenkouot: will there be other routes like this in the future? or just JA for now

perjg: There will be one more ('/ja/{path*}'), to serve assets.

gulp/tasks/scripts-homepage.js
perjg: I'd prefer keeping it in the /homepage path, as the homepage is an independent webapp, and dependencies and code may diverge from /front

kenkouot: In that case, why don't we just put it in an independent repo?

rogatty: @kenkouot great question!
@perjg why is it in this repo? It looks like a fully separate app. Does it share anything at all with the mercury code?

perjg: @kenkouot Well, the conclusion when I discussed this with you a few weeks back (when I was still in SF) was that the JA homepage should be added inside the Mercury server app, but not neccesarily use ember.js or the same css structure as the main mercury app.

Some of the arguments (as I remember them):
* Allow re-use of mercury components and assets when desired
* Ease future integration with the new auth-related functionality (which is being added in a similar way, as separate hapi routes)
* Leverage existing deployment and hosting of mercury
* Use MediaWiki api wrappers in the mercury server app

I can separate everything out into a completely separate app too, as my reasons for integrating it into the mercury server app were as a result of what we discussed. If I misunderstood, or did the integration in an undesirable way, I can look into that too.

server/routes.ts
hakubo: I don't get that
why would you want to serve assets from ('/ja/{path*}')?
shouldn't they be loaded from something like '/public/{path*}'?

hakubo:
I think we should think how can we decouple this code from Mercury server (same goes for Login service)
as now we are not able to build and deploy them separately and we might brake one accidentally by deploying some fixes to another service.

Unless those homepage pages are to be part of Mercury service in such case it be best if they were using full stack (Ember and the same asset pipeline).

I am fine with this code being in the same repo
but we SHOULD be able to build and serve it independently
I think we should go towards what has been proposed here: 
https://one.wikia-inc.com/wiki/User_blog:Damian_Jozwiak/Auth_Pages_and_Mercury_Dependencies

gulp/tasks/scripts-homepage.js
perjg: @kenkouot Just for clarification: I am ok with either approach (separate app, or keeping it in mercury but in  a separate route like now).

server/routes.ts
perjg: Were you thinking of serving all assets from /public?
Currently, front assets are served from the front subdirectory, so to make this change I guess it makes sense to move assets for all the different "modules" into one place. Separate PR?

server/facets/homepage/homepageView.ts
```diff
@@ -0,0 +1,84 @@
...
+			{
+				layout: 'homepage'
+			}
+		);
+
+		caching.disableCache(response);
+		return response;
+	}
+
+	export function getRedirectUrl (request: Hapi.Request): string {
```
kvas-damian: This code looks like Copy&Pasted from auth pages. I don't think it's best solution for this problem. Any changes in login code will end up with changes in two places, which is not ideal.

I don't know what is your usecase, but I'd suggest using iframe or link to authpages instead of creating next instance of the same service.

perjg: I used the auth controller as a base for the implementation.
Probably, I can streamline the code a bit more, but essentially it will be different from the auth pages, so I don't think there will be a need to keep the two classes synchronized. If the need arises, I will make sure to refactor the two to depend on a common ancestor class.

kvas-damian: What kind of differences do you expect on auth pages? Aren't we targeting to have consistent UI for the same functionality across entire service?

perjg:
Rewritten to be independent of mercury app. Closing PR and deleting branch.

PR #1004, opened 2015-06-30
SOC-965 Rename login/signup to signin/register

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-965 @kvas-damian @mklucsarits


server/facets/auth/register.ts
```diff
@@ -4,11 +4,13 @@
 
 import BirthdateInput = require('./BirthdateInput');
 import dateUtils = require('../../lib/DateUtils');
+import authUtils = require('../../lib/AuthUtils');
```
kvas-damian: Why there are new requires? I don't see any usages.

server/lib/AuthUtils.ts
```diff
@@ -38,3 +38,9 @@ export function getCacheBusterUrl(redirect: string): string {
 
 	return url.format(cacheBustedUrlObj);
 }
+
+export function getRedirectUrlWithQueryString(route: string, request: Hapi.Request) {
+	var redirectUrl = request.url;
+	redirectUrl.pathname = route;
+	return url.format(redirectUrl);
```
kvas-damian: Isn't it more readable to use ```redirectUrl.format()```?

server/facets/auth/register.ts
bkoval: authUtils are used in new handlers :)

server/facets/auth/register.ts
```diff
@@ -6,9 +6,10 @@ import BirthdateInput = require('./BirthdateInput');
 import dateUtils = require('../../lib/DateUtils');
 import localSettings = require('../../../config/localSettings');
 import authView = require('./authView');
-var deepExtend = require('deep-extend');
+var deepExtend = require('deep-extend'),
+	url = require('url');
```
kvas-damian: And where are we using this one? :P

kvas-damian:
LGTM

front/locales/de/auth.json
```diff
@@ -1,13 +1,13 @@
 {
   "common":{
     "server-error":"Es tut uns leid, bei der Verarbeitung deiner Anfrage ist ein Fehler aufgetreten. Bitte versuche es später noch einmal.",
-    "login-callout":"Hast du bereits ein Benutzerkonto?",
+    "signin-callout":"Hast du bereits ein Benutzerkonto?",
```
mklucsarits: Why change these?

mklucsarits:
I don't understand the necessity and don't personally approve of renaming "login" and "signup" to "signin" and "register" in the codebase. What is the reason for doing so? Is there a reason beyond the creation of the /signin and /register paths? Renaming "signup" to "register" is alright I guess, but pragmatically speaking, if someone is looking for log-in code in our codebase, no one is going to think to search for "signin" first. This will only lead to confusion. I also don't believe changing our log-in and registration paths necessitates changing the naming of variables, etc. in the codebase. It's important that we name classes and variables that make sense to us as developers, regardless of whether they match the words we show to our end-users.

bkoval:
Main reason is consistency - since we're renaming the routes to reflect our product mission in a more elegant way and we want to be consistent in language everywhere in the UI, renaming of variables that correspond to route handlers, templates and context objects should make sense in this case.

I also think that if someone is going to be grepping the code, it's better to get all the results once the phrase "signin" is typed into a search form (especially that it corresponds to the route in the browser's address bar and to the product language), than get half / not full portion of results where a bit of them is referenced under signup-something and the others use login-something pattern.

Since routes, templates, i18n strings and context objects are tightly connected with user-facing views, I believe consistency is the biggest value that we should align with.

mklucsarits:
But the naming was already consistent. And because we're keeping the /login and /signup paths, the new naming will be inconsistent with those. So we're not going to be able to have consistent naming between routes and variables/classes anyway, in either case.

front/locales/de/auth.json
```diff
@@ -1,13 +1,13 @@
...
     "login-link-text":"Anmelden",
```
mklucsarits: "login" is still used here. If you're renaming everything, be sure to grep and rename it all.

bkoval:
We're keeping /login and /signup just to make them redirect to /signin and /register.

We can continue this discussion for a long time, but I believe variable naming (as long as it's consistent) is not worth discussing too much :)

mklucsarits:
OK, I still don't think it's a good idea, but I won't hold up the merge.

PR #1009, opened 2015-07-01
SOC-898 New Registration in wiki language

bkoval:
#977 for more info.


bkoval:
Let's wait with merging it until we merge the change that enables new signup :)

PR #1011, opened 2015-07-01
DAT-2952 Infobox image cropping

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-2952

This PR should be merged after the https://github.com/Wikia/app/pull/7646 as it contains a part of if (branch DAT-2899 is the DAT-2952 parent: https://github.com/Wikia/mercury/compare/DAT-2899...DAT-2952). After merging DAT-2899 to dev, the diff will be smaller.

@Warkot @RafLeszczynski


front/scripts/main/components/ImageMediaComponent.ts
```diff
@@ -45,34 +51,82 @@ App.ImageMediaComponent = App.MediaComponent.extend(App.ArticleContentMixin, {
...
+	 * @desc return the params for getThumbURL for infobox image.
+	 * In case of very high or very wide images, crop them properly.
+	 * @return {
+	 *   mode: string crop mode
+	 *   height: number height of image
+	 *   width: number width of image
+	 * }
+	 */
+	infoboxImageParams: Em.computed({
+		get(): thumbnailerParams {
```
rogatty: This is the first place where we have getter but no setter for a computed property. Are you sure that it's allowed and won't cause any issues? What happens when you try to set this property?

I'm aware that setters don't always make sense but we don't want to throw exceptions or sth.

front/scripts/main/components/ImageMediaComponent.ts
```diff
@@ -45,34 +51,82 @@ App.ImageMediaComponent = App.MediaComponent.extend(App.ArticleContentMixin, {
...
 	/**
+	 * @desc return the params for getThumbURL for infobox image.
+	 * In case of very high or very wide images, crop them properly.
+	 * @return {
+	 *   mode: string crop mode
+	 *   height: number height of image
+	 *   width: number width of image
+	 * }
+	 */
+	infoboxImageParams: Em.computed({
```
rogatty: This looks like a perfect candidate for unit testing.

front/scripts/main/components/ImageMediaComponent.ts
dianafa: https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/computed.js#L430 imo it will not cause any problems

dianafa:
test added

SebastianMarzjan:
:+1:

rogatty:
LGTM

PR #1012, opened 2015-07-01
HG-721 - WIP - lazy load all js

katanka:


bower.json
```diff
@@ -23,6 +23,7 @@
   "devDependencies": {
     "ember-qunit": "0.3.0",
     "sinonjs": "1.12.2",
-    "sinon-qunit": "https://github.com/cjohansen/sinon-qunit.git#2.0.0"
+    "sinon-qunit": "https://github.com/cjohansen/sinon-qunit.git#2.0.0",
+    "system.js": "~0.18.3"
```
hakubo: Ember CLI uses loader.js
I think we should consider using the same toolset for easier switch

hakubo: Ember CLI uses loader.js
I think we should consider using the same toolset for easier switch

kenkouot: I think in the final implementation we'll use loader.js. For now we're just testing this and it was easy + well documented enough for @katanka to just get started. Loader's repo had no documentation.

PR #1017, opened 2015-07-02
CONCF-591 | add load more button for categories

bognix:
https://wikia-inc.atlassian.net/browse/CONCF-591


bognix:
ping @rogatty @vforge

front/scripts/main/components/CuratedContentComponent.ts
```diff
@@ -21,6 +21,15 @@ App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.T
 			} else {
 				this.trackClick('modular-main-page', 'curated-content-item-other');
 			}
+		},
+		loadMore: function(): void {
+			this.showLoader();
+			App.CuratedContentModel.fetchItemsForSection(this.get('title'), 'category', this.get('offset'))
```
rogatty: Why is `'category'` hardcoded?

front/scripts/main/components/CuratedContentComponent.ts
```diff
@@ -21,6 +21,15 @@ App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.T
...
+				.then((data: any) => {
+					this.set('offset', data.offset);
+					this.set('model', this.get('model').pushObjects(data.items));
+					this.hideLoader();
+				});
```
rogatty: We should handle a reject somehow.

front/scripts/main/components/CuratedContentComponent.ts
bognix: Two things:
1. I don't have the information about type of element. I would need to pass it to this component like I did with title and offset
2. With current implementation only categories API is supporting offset so category type is the only one which would have "load more" button and the only when item has category type this function is going to be executed

front/templates/main/components/curated-content.hbs
```diff
@@ -6,5 +6,10 @@
 
 	{{! the empty newline above lets the last inline item to be justified to the right if it's even }}
 	</nav>
+	{{#if offset}}
+		<button class="load-more-items" {{action 'loadMore'}}>
+			{{i18n 'app.load-more'}}
```
rogatty: We were supposed to create a new string.

IMO it's fine this way - I don't see how a context change might change the meaning of this simple string.

bognix: It is a new string. We didn't have it

front/scripts/main/components/CuratedContentComponent.ts
rogatty: Ok, can you add the second thing as a comment? Thanks.

front/locales/en/main.json
```diff
@@ -77,6 +77,8 @@
 		"user-menu-profile": "Profile",
 		"user-menu-log-out": "Log out",
 		"curated-content-error-category-not-found": "Category not found",
-		"curated-content-error-section-not-found": "Section not found"
+		"curated-content-error-section-not-found": "Section not found",
+		"categories-load-more": "Load more",
+		"something-went-wrong": "Oops! something went wrong."
```
rogatty: Can't we be more specific? How about "Sorry, we can't load more items"?

front/locales/en/main.json
```diff
@@ -77,6 +77,8 @@
 		"user-menu-profile": "Profile",
 		"user-menu-log-out": "Log out",
 		"curated-content-error-category-not-found": "Category not found",
-		"curated-content-error-section-not-found": "Section not found"
+		"curated-content-error-section-not-found": "Section not found",
+		"categories-load-more": "Load more",
```
rogatty: `curated-content-load-more-items`?

front/scripts/main/components/CuratedContentComponent.ts
```diff
@@ -21,6 +21,20 @@ App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.T
 			} else {
 				this.trackClick('modular-main-page', 'curated-content-item-other');
 			}
+		},
+		loadMore: function(): void {
+			this.showLoader();
+			App.CuratedContentModel.fetchItemsForSection(this.get('title'), 'category', this.get('offset'))
+				.then((data: any) => {
+					this.set('offset', data.offset);
+					this.set('model', this.get('model').pushObjects(data.items));
```
rogatty: You don't have to set the model, pushing is enough.

front/scripts/main/components/CuratedContentComponent.ts
```diff
@@ -21,6 +21,20 @@ App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.T
...
+				})
+				.catch((reason) => {
```
rogatty: Can we use `Em.Logger.error` to log it to the console? Would be easier to debug.

front/locales/en/main.json
vforge: Also `curated-content-error-load-more-items` ?

front/scripts/main/models/CuratedContentModel.ts
```diff
@@ -19,14 +14,17 @@ interface CuratedContentItem {
...
 		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
 			var url = App.get('apiBase');
 			// If this is first PV we have model for curated content already so we don't need to issue another request
 			// When resolving promise we need to set Mercury.curatedContent to undefined
 			// because this data gets outdated on following PVs
 			if (Mercury.curatedContent && Mercury.curatedContent.items) {
-				resolve(App.CuratedContentModel.sanitizeItems(Mercury.curatedContent.items));
+				resolve({
+					items: App.CuratedContentModel.sanitizeItems(Mercury.curatedContent.items),
+					offset: Mercury.curatedContent.offset
```
rogatty: I think we shouldn't use `Mercury` like that. See https://github.com/Wikia/mercury/pull/586

server/lib/MediaWiki.ts
```diff
@@ -133,7 +133,9 @@ export class ArticleRequest extends BaseRequest {
 			abstract: 0,
 			width: thumbSize.width,
 			height: thumbSize.height,
-			category: categoryName
+			category: categoryName,
+			offset: offset,
+			limit: 24
```
vforge: How about saving it as some kind of const, so we won't need to search for this number later in order to change it.

front/scripts/main/components/CuratedContentComponent.ts
rogatty: `front/scripts/main/components/CuratedContentComponent.ts(34,13): 7006 Parameter 'reason' implicitly has an 'any' type.`

front/scripts/main/components/CuratedContentComponent.ts
```diff
@@ -21,6 +21,23 @@ App.CuratedContentComponent = Em.Component.extend(App.LoadingSpinnerMixin, App.T
...
+			// "category" type is hardcoded because only Categories API supports returning and accepting offset.
+			// And only when offset is returned and set this action can be triggered.
+			App.CuratedContentModel.fetchItemsForSection(this.get('title'), 'category', this.get('offset'))
+				.then((data: any) => {
+					this.set('offset', data.offset);
+					this.set('model', this.get('model').pushObjects(data.items));
+				})
+				.catch((reason) => {
+					this.controllerFor('application').addAlert('error', i18n.t('app.something-went-wrong'));
+					Ember.debug(reason);
```
rogatty: Why not `Em.Logger.debug`? We're using it everywhere else.

server/lib/MediaWiki.ts
bognix: Well this number is used only here and this is the place where actual API call happens.
I thought it would increase readability to have this number here not inside some variable given that you can see what parameter has this value (limit). What is more if someone wants to debug what is happening and from where this number of categories comes from this is the place were this person will be at the end.

rogatty:
LGTM

vforge:
LGTM

server/lib/MediaWiki.ts
vforge: I'm good with that. Proceed. :+1:

Warkot:
LGTM

Warkot:
LGTM

server/routes.ts
```diff
@@ -103,12 +103,12 @@ unauthenticatedRoutes = [
 	},
 	{
 		method: 'GET',
-		path: localSettings.apiBase + '/curatedContent/{sectionName}',
+		path: localSettings.apiBase + '/main/section/{sectionName}',
 		handler: require('./facets/api/curatedContent').get
```
rogatty: Let's change it to `mainSection`.

server/routes.ts
```diff
@@ -103,12 +103,12 @@ unauthenticatedRoutes = [
...
 	},
 	{
 		method: 'GET',
-		path: localSettings.apiBase + '/category/{categoryName}',
+		path: localSettings.apiBase + '/main/category/{categoryName}',
 		handler: require('./facets/api/category').get
```
rogatty: Let's change it to `mainCategory`.

server/routes.ts
bognix: why? In code everywhere we use sectionName. Do you want to change it there also?
In my opinion it is readable as it is.

server/routes.ts
bognix: well it is in fact categoryName not mainPageCategory.
This category name is used across site, not only on main pages.

rogatty: I meant file name, not the parameter from the line above. The same for previous comment.

bognix: aaa, now it makes sense :+1:

rogatty:
LGTM, let's merge it and test on CONCF-655 branch.

tomasznapieralski:
LGTM

Warkot:
:+1:

PR #1018, opened 2015-07-02
Fix small typos on dev

dianafa:
While running server there was several messages in the console.

@rogatty @bkoval


rogatty:
LGTM and thanks, it was bothering me too. We should discourage people from merging code that causes warnings.

SebastianMarzjan:
Great idea!
Is there an idea you might have already as to how we can automate that so it happens with no additional effort?

dianafa:
The thing we can do now is to encourage people to install pre commit hooks. Unfortunatelly, the issue will probably remain unresolved, as above warning messages were visible in console on each server start from about one week, and nobody took a look on them.

In my opinion we should try to introduce to our stack this tool: https://www.npmjs.com/package/js-beautify which really helps with silly typos. I'll check this out.
However, some of this issues, like typings, would be harder to repair automatically. It's an interesting issue! Thanks.

PR #1026, opened 2015-07-06
DAT-2969 - fix issue with transparent color in linear gradient on iOS

RafLeszczynski:


SebastianMarzjan:
The change itself looks good, though I'm not thrilled about the semantics of $color-blue-gray-light as a var name

dianafa:
will `fade-out` work as well here?

LGTM

PR #1035, opened 2015-07-07
SOC-1005 Trigger number keyboard when b-date input focused

bkoval:
@mklucsarits @lizlux https://wikia-inc.atlassian.net/browse/SOC-1005

### DIFF
--- a/server/views/register.hbs
+++ b/server/views/register.hbs
@@ -16,7 +16,7 @@
 		<div class="input-container">
 			<div class="birthdate-container input hide">
 				{{#birthdateInputs}}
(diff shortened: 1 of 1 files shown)


lizlux:
I remember there was a reason I didn't do this in the first place, or at least I went back and forth on number vs text field. It's possible whatever reason the text input was used has gone away since then. Just make sure the auto-tab still works well and that the number up/down arrows don't show up.

bkoval:
We'll move it through normal QA with full regression to make sure it works properly :)

mklucsarits:
LGTM

PR #1038, opened 2015-07-07
INT-130: Adding country parameter in context for performance data.

mari-mcmurtrie:
We want to measure performance of traffic from Japan. Adding country info read by users geo cookie enables filtering per geo for performance data in grafana.


lizlux:
LGTM

mari-mcmurtrie:
I fixed indentation.

PR #1042, opened 2015-07-08
MAIN-4895 adding handling for url hash

lizlux:
https://wikia-inc.atlassian.net/browse/MAIN-4895

Quick fix for p3 - handling hash links inside articles. Ideally we'd add an automated test for this. (https://wikia-inc.atlassian.net/browse/HG-799)

@rogatty @hakubo


lizlux:
There's an issue when navigating from a different page to the page with the anchor on it. Closing this and going to see if a different solution would be better.

lizlux:
Actually, the issue I was seeing came from the title of the article showing up with escaped spaces as opposed to underscores, when accessing it via search results. There's no issues with this actual implementation, though the search results escaping issue should be fixed.

rogatty:
Unfortunately it breaks links to other pages with anchor in the URL.

For example I'm on the page `/wiki/Article1` and I have a link to `/wiki/Article2#Anchor`. After clicking the link:
- window scrolls to element `#Anchor` if there is one in the Article1 OR
- nothing happens if there is no such element

This is the logic that isn't run because of the change in this PR:
https://github.com/Wikia/mercury/blob/MAIN-4895/front/scripts/mercury/utils/articleLink.ts#L76-L89

lizlux:
Okay, looks like someone will need to put some more time into this fix.

PR #1044, opened 2015-07-09
mvp implementation of utilizing article sections api

kenkouot:
Really simple implementation of utilizing @lizlux's new section article work.


lizlux:
Please add a per-wiki and possibly per-article toggle so we can do a limited release of this feature.

typings/mercury/mercury-server.d.ts
```diff
@@ -14,6 +14,7 @@ interface ArticleRequestParams {
 	title?: string;
 	redirect?: any;
 	headers?: any;
+	sections?: string|number[];
```
lizlux: This should always be a string. The API no longer accepts arrays.

lizlux:
When given a sections parameter, the API returns an actual array of sections, not a string. It looks like the Mercury side is still expecting a string. Would you rather it returned a string?

server/lib/MediaWiki.ts
```diff
@@ -89,16 +89,23 @@ export class ArticleRequest extends BaseRequest {
 	 * @param redirect
 	 * @return {Promise<any>}
 	 */
-	article (title: string, redirect: string): Promise<any> {
+	article (title: string, redirect: string, sections?: string|number[]): Promise<any> {
```
lizlux: same here, should just be a string

server/lib/MediaWiki.ts
```diff
@@ -89,16 +89,23 @@ export class ArticleRequest extends BaseRequest {
...
 		var urlParams: any = {
 				controller: 'MercuryApi',
 				method: 'getArticle',
-				title: title
+				title: title,
```
lizlux: trailing comma and extra white space.

server/facets/operations/prepareArticleData.ts
```diff
@@ -49,19 +51,11 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
...
 	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
-	result.asyncArticle = shouldAsyncArticle(result);
-}
 
-/**
- * (HG-753) This allows for loading article content asynchronously while providing a version of the page with
- * article content that search engines can still crawl.
- * @see https://developers.google.com/webmasters/ajax-crawling/docs/specification
- */
-function shouldAsyncArticle(result: any): boolean {
```
rogatty: Can you also update `prepareMainPageData.ts` and `processCuratedContentData.ts`? Thanks.

We will have to refactor it anyway (there are no articles there) but let's not break what we have currently.

rogatty: Or maybe you have some idea how to make `M.prop('asyncArticle', {{asyncArticle}});` work without adding `asyncArticle` to `result`? If we don't add it then we have syntax error - `M.prop('asyncArticle', )`

lizlux:
It would be better if we didn't have to replace all the article content, but could append the remaining sections. I'm still seeing a jump back to the top of the page once the whole article has loaded. It shouldn't be too hard to implement on either the MW or HG side. What do you think @kenkouot?

lizlux:
@kenkouot I added the `3+` syntax here https://github.com/Wikia/app/pull/7876.

server/facets/operations/prepareMainPageData.ts
```diff
@@ -52,21 +54,14 @@ function prepareMainPageData (request: Hapi.Request, result: any): void {
 	}
 
 	result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
-	result.asyncArticle = shouldAsyncArticle(result);
 
-	delete result.adsContext;
-}
+	result.asyncArticle = (
```
kenkouot: @rogatty does this suffice for main pages?

rogatty: Yup, thanks!

PR #1047, opened 2015-07-09
SOC-490 Facebook Login

bkoval:
This PR covers https://wikia-inc.atlassian.net/browse/SOC-490
(logging in through Facebook for accounts already connected in the past)

@lizlux @mklucsarits @rogatty


config/localSettings.base.ts
```diff
@@ -104,7 +104,10 @@ var localSettings: LocalSettings = {
 	// CDN prefix with no tailing slash
 	cdnBaseUrl: '//mercury.nocookie.net',
 	// array of wiki dbnames to load first article async instead of in page source
-	asyncArticle: []
+	asyncArticle: [],
+	facebook: {
+		appId: 112328095453510
```
bkoval: @lizlux We expose this var in javascript anyway, the question is if we should obfuscate it more, e.g. with passing it from chef?

front/scripts/auth/facebook/FacebookSDK.ts
```diff
@@ -0,0 +1,28 @@
+interface Window {
+	FB?: any;
+	fbAsyncInit?: Function;
+}
+
+class FacebookSDK {
+	constructor (onLoad: Function = Function.prototype) {
```
bkoval: This is just a modified code taken from the SDK website, a bit typescripted and with callback.

config/localSettings.base.ts
lizlux: Looks like the main thing we care about protecting is the app secret. The ID is fine to leave in a public repo https://developers.facebook.com/docs/apps/register#app-id.

front/scripts/auth/facebook/FacebookSDK.ts
```diff
@@ -0,0 +1,28 @@
...
+		var js: HTMLScriptElement,
+			fjs: HTMLScriptElement = window.document.getElementsByTagName('script')[0];
```
mklucsarits: Can this variable be renamed to firstJS, or something else more descriptive? It's not immediately apparent that "f" means "first". My initial thought was "f" meant "Facebook".

front/scripts/auth/facebook/FacebookSDK.ts
```diff
@@ -0,0 +1,28 @@
...
+		js = window.document.createElement('script');
+		js.id = 'facebook-jssdk';
+		js.src = "//connect.facebook.net/en_US/sdk.js";
+		fjs.parentNode.insertBefore(js, fjs);
+
+		window.fbAsyncInit = function(): void {
+			window.FB.init({
+				appId: M.prop('facebookAppId'),
+				cookie: true,
+				version: 'v2.2'
```
mklucsarits: Should this version string be an "immutable" (meaning a config value, for us)? How often does this version number change?

front/scripts/auth/facebook/FacebookLogin.ts
```diff
@@ -0,0 +1,92 @@
+interface FacebookResponse {
+	status: string;
+	authResponse: FacebookAuthResponse;
+}
+
+interface FacebookAuthResponse {
```
mklucsarits: Can this be renamed to, perhaps, FacebookAuthData? Technically there's only one response object. Any objects within it can't also be a response. It's a little confusing to have two response interfaces when only one of them is the complete, actual request response.

lizlux:
@bkoval what's the best way to run this code? On my local instance, facebook showed an error: `Given URL is not allowed by the Application configuration: One or more of the given URLs is not allowed by the App's settings. It must match the Website URL or Canvas URL, or the domain must be a subdomain of one of the App's domains.` I tried it with different variations of `localhost` and `www.wikia.local`, etc. Also having trouble getting my devbox running mercury but that's a separate issue.

bkoval:
By default localSettings use production appId so it should be triggered from .wikia.com domain (e.g. on sandbox). To run this from wikia-dev, you should override appId to Wikia Dev appId in your localSettings.ts. Currently there's no way to run it locally as (I believe) Facebook SDK needs to match the domain whitelist of the Facebook App it connects to.

front/scripts/auth/facebook/FacebookSDK.ts
rogatty: Put it in a comment.

rogatty:
LGTM

front/scripts/auth/facebook/FacebookLogin.ts
```diff
@@ -0,0 +1,92 @@
...
+	private onUnsuccessfulLogin(response: FacebookResponse): void {
+		this.activateButton();
+	}
+
+	private getHeliosInfoFromFBToken(facebookAuthResponse: FacebookAuthResponse): void {
+		var facebookTokenXhr = new XMLHttpRequest(),
+			data = <HeliosFacebookToken> {
+				fb_access_token: facebookAuthResponse.accessToken
+			},
+			url = this.loginButton.getAttribute('data-helios-facebook-uri');
```
lizlux: I'm trying to decide if this line is problematic - getting the URL for a POST request from the DOM. For instance, if the helios `authResponse` contains sensitive data, and the user changes to the url and posts to a different site. @bkoval, what do you think?

lizlux:
Just one question but other than that, LGTM.

front/scripts/auth/facebook/FacebookSDK.ts
bkoval: I can see keeping this string here is not too clean, but do you mean moving it to localSettings or rather creating a *const* at the top of the file here?

front/scripts/auth/facebook/FacebookLogin.ts
bkoval: Yeah, I was thinking what is the easiest way of passing it, but we already use the DOM passing, e.g. here: https://github.com/Wikia/mercury/blob/dev/server/views/register.hbs#L2 (fallback to no-js registration) so it seemed more natural. We have to pass the URL either way and even if it's available only in JS, it's still prone to modification.

@lizlux Would you rather pass the URL through M.prop?

lizlux: I'm sure this is fine. I pinged @Grunny to get his opinion but this isn't a blocker. LGTM.

front/scripts/auth/facebook/FacebookSDK.ts
mklucsarits: If you want to make this change, ```const``` does make sense if we're only using it in this file. But if end up using it in multiple places, we should move it to a place where the value can be shared, whether it's in localSettings or whatever.

mklucsarits:
LGTM

front/scripts/auth/facebook/FacebookLogin.ts
```diff
@@ -0,0 +1,92 @@
...
+	private deactivateButton(): void {
+		this.loginButton.classList.add('on');
+		this.loginButton.classList.add('disabled');
+	}
+
+	private onSuccessfulLogin(response: FacebookResponse): void {
+		this.getHeliosInfoFromFBToken(response.authResponse);
+	}
+
+	private onUnsuccessfulLogin(response: FacebookResponse): void {
```
rafalkalinski: maybe for better distinction it should be called like "onFailedLogin"

server/views/_layouts/auth.hbs
```diff
@@ -42,6 +42,7 @@
 		<!-- endbuild -->
 		<script>
 			M.prop('tracking', {{{json trackingConfig}}});
+			M.prop('facebookAppId', {{facebookAppId}});
```
kvas-damian: Didn't we agree to avoid using Mercury code in auth pages?

kvas-damian: Instead of creating empty variable on every auth page we could put it only on pages that requires this value.
Maybe we could use similar patter to ```pageParams``` in:  https://github.com/Wikia/mercury/blob/SOC-1062/server/views/_layouts/auth.hbs#L36

lizlux: @kenkouot is going to abstract `Mercury.Utils` including `M.prop` into a separate repo that can be brought in via bower, so it's safe to use this in all apps inside the mercury repo and at Wikia.

PR #1049, opened 2015-07-09
SOC-898 New Registration in wiki language

bkoval:
#977 for details


front/scripts/main/components/LoginIconComponent.ts
```diff
@@ -27,5 +29,13 @@ App.LoginIconComponent = Em.Component.extend({
 		});
 
 		window.location.href = href;
+	},
+
+	getUselangParam: function (): string {
+		var lang: string = Mercury.wiki.language.content;
```
rogatty: Are you sure there is always `Mercury.wiki.language`? If not then this can throw exceptions.

PR #1051, opened 2015-07-09
SOC-1017 Enable /both/ as default trackingMethod setting for New Auth

bkoval:
@rogatty @jsutterfield https://wikia-inc.atlassian.net/browse/SOC-1017

### DIFF
--- a/front/scripts/auth/tracking.ts
+++ b/front/scripts/auth/tracking.ts
@@ -6,6 +6,20 @@
 		return (document.body.className.indexOf(pageType) !== -1);
 	}
 
+	function setupTracking(): void {
+		//Auth pages live on www.wikia.com and don't have access to WikiVariables
+		//hence there's a need to provide this data inline
+		window.Mercury.wiki = {
+			id: 80433,
+			dbName: 'wikiaglobal',
+			language: {
+				user: 'en'
+			}
+		};
+
+		setTrackingDimensions();
+	}
+
 	function setTrackingDimensions (): void {
 		var dimensions: (string|Function)[] = [];
 		// Skin
@@ -128,7 +142,7 @@
 	}
 
 	function init (): void {
-		setTrackingDimensions();
+		setupTracking();
 
 		if (checkPageType('join-page')) {
 			setTrackingForJoinPage();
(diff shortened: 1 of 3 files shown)


rogatty:
LGTM

jsutterfield:
LGTM

front/scripts/auth/tracking.ts
```diff
@@ -6,6 +6,20 @@
 		return (document.body.className.indexOf(pageType) !== -1);
 	}
 
+	function setupTracking(): void {
```
bkoval: Unfortunately, Internal Tracker is breaking since there's no WikiVariables call there. However, all the auth pages live on www.wikia.com domain, so we could work around this and just use the inline data for this particular Wikia.

This sounds like a hack and I'm more than glad to hear any other suggestions, however we need this working ASAP so if the suggested way needs more time my preferred path would be creating a ticket and tackling it in the future.

@rogatty @lizlux @jsutterfield @mklucsarits Any ideas?

rogatty: http://www.wikia.com/wikia.php?controller=MercuryApi&method=getWikiVariables seems to be working. Why not use it in Hapi the same way as the Mercury?

bkoval: That would introduce a request to Mediawiki (which we do not use on auth pages at all) on every pageview and we want to avoid that. However, if that's the only solution (and it seems so) we could tackle this in a seperate ticket.

jsutterfield: Hmm, I can't think of anything else that would be better at this time. It does feel a little hacky, but it's straightforward, and I think the comment does a good job of explaining clearly what it's doing and why. 

+1 for pushing this out and if someone thinks of a better approach tackling that in a separate ticket.

mklucsarits: If the requirement for internal tracking is just a wiki ID, it doesn't make sense to send even one request for wiki variables, for auth pages. I think storing it statically on the front end is preferred. For other pages that need a specific wiki ID for internal tracking, yes we'll have to think of a better approach.

front/scripts/auth/tracking.ts
```diff
@@ -6,6 +6,20 @@
...
+		//Auth pages live on www.wikia.com and don't have access to WikiVariables
+		//hence there's a need to provide this data inline
+		window.Mercury.wiki = {
```
lizlux: Please use `M.provide()`

PR #1056, opened 2015-07-09
INT-117: Photo/Edit icons style change for Japanese Wikia

mari-mcmurtrie:
When the section title takes up the whole width of the screen on a mobile device, the edit and photo icons are displaced and moved to a second line. We want these icons aligned in 1 line with section header to avoid displacement. When we show the icons, we shorten section header to make room for icons and display them vertical-top alignment.

### DIFF
--- a/front/scripts/main/views/ArticleView.ts
+++ b/front/scripts/main/views/ArticleView.ts
@@ -212,9 +212,7 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 
 		this.$(':header[section]').each((i: Number, item: any): void => {
 			var $sectionHeader = this.$(item);
-
-			$sectionHeader.prepend(this.$(iconsWrapper));
-			$sectionHeader.addClass('short-header');
+			$sectionHeader.prepend(iconsWrapper).addClass('short-header');
 		});
 		this.setupButtonsListeners();
 	},
@@ -237,14 +235,14 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 				});
 			})
 			.on('change', '.upload-photo', (event: JQueryEventObject): void => {
-				var $photoUpload = $(event.target).parent(),
+				var $uploadPhotoContainer = $(event.target).parent(),
 				    sectionIndex: number = parseInt($(event.target).closest(':header[section]').attr('section'), 10);
-				this.onPhotoIconChange($photoUpload, sectionIndex);
+				this.onPhotoIconChange($uploadPhotoContainer, sectionIndex);
 			});
 	},
 
-	onPhotoIconChange: function(photoUpload: JQuery, sectionNumber: number): void {
-		var photoData = (<HTMLInputElement>photoUpload.find('.file-input')[0]).files[0];
+	onPhotoIconChange: function(uploadPhotoContainer: JQuery, sectionNumber: number): void {
+		var photoData = (<HTMLInputElement>uploadPhotoContainer.find('.file-input')[0]).files[0];
 		this.get('controller').send('addPhoto', this.get('controller.model.title'), sectionNumber, photoData);
 	},
 
--- a/front/scripts/main/views/ArticleView.ts
+++ b/front/scripts/main/views/ArticleView.ts
@@ -194,6 +194,7 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		// TODO: There should be a helper for generating this HTML
 		var pencil = '<div class="edit-section"><svg class="icon pencil" role="img"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#pencil"></use></svg></div>',
 			photo = '<div class="upload-photo"><svg class="icon camera" role="img"><use xlink:href="#camera"></use></svg><input class="file-input" type="file" accept="image/*" capture="camera"/></div>',
+			iconsWrapper = '<div class="icon-wrapper">' + pencil + photo + '</div>',
 			$photoZero = this.$('.upload-photo');
 
 		$photoZero
@@ -210,12 +211,10 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			});
 
 		this.$(':header[section]').each((i: Number, item: any): void => {
-			var $sectionHeader = this.$(item),
-				$pencil = this.$(pencil),
-				$photo = this.$(photo);
+			var $sectionHeader = this.$(item);
 
+			$sectionHeader.prepend(this.$(iconsWrapper));
 			$sectionHeader.addClass('short-header');
-			$sectionHeader.after($pencil, $photo);
 		});
 		this.setupButtonsListeners();
 	},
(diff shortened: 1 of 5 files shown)


front/scripts/main/views/ArticleView.ts
```diff
@@ -212,9 +212,11 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		this.$(':header[section]').each((i: Number, item: any): void => {
 			var $sectionHeader = this.$(item),
 				$pencil = this.$(pencil),
-				$photo = this.$(photo);
+				$photo = this.$(photo),
+				sectionHeaderNewWidth = $sectionHeader.width() - 85;
```
inez: Where is that 85 coming from?

mari-mcmurtrie: That is about the size of 2 icons.

mari-mcmurtrie: I found a way to retrieve this number programmatically so let me re-write this part.

front/scripts/main/views/ArticleView.ts
```diff
@@ -222,12 +224,12 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 	setupButtonsListeners: function () : void {
 		this.$('.article-content')
 			.on('click', '.pencil', (event: JQueryEventObject): void => {
-				var $sectionHeader = $(event.target).parent();
+				var $sectionHeader = $(event.target).parent().prev('h2');
```
inez: Section headers are not always h2 tags (it could be h3 and h4, and maybe even h5, as well I believe).
I have doubts that relying on prev() and even prev().prev(...) few lines below is a solid approach - it's very sensitive to changes in the HTML so it carries maintenance cost. Would you be able to come with some other solution? Maybe you can take advantage of CSS classes.

front/scripts/main/views/ArticleView.ts
```diff
@@ -237,13 +239,14 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 				});
 			})
 			.on('change', '.upload-photo', (event: JQueryEventObject): void => {
-			var $sectionHeader = $(event.target).parent().parent();
-				this.onPhotoIconChange($sectionHeader, $sectionHeader.attr('section'));
+				var $photoUpload = $(event.target).parent(),
```
inez: "$(event.target).parent()" called twice

front/scripts/main/views/ArticleView.ts
```diff
@@ -212,9 +212,11 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 		this.$(':header[section]').each((i: Number, item: any): void => {
 			var $sectionHeader = this.$(item),
 				$pencil = this.$(pencil),
-				$photo = this.$(photo);
+				$photo = this.$(photo),
+				sectionHeaderNewWidth = $sectionHeader.width() - 85;
 
-			$sectionHeader.append($photo, $pencil);
+			$sectionHeader.css({ display: 'inline-block', width: sectionHeaderNewWidth });
```
inez: You should avoid modifying CSS properties on the fly - please try to limit it to toggling certain CSS class if you really need to have different styling for section header depending of button (edit section, add photo) presence.

inez: I'm also pretty surprised that you had to modify display property - seems like a big and significant change - make sure to test it very well.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Combined with your comment about CSS below, I removed dynamic css change, and instead, add a class to apply different styling.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I agree this is a bit fragile. I found a better and more robust way to get section header.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I changed in next commit.

mari-mcmurtrie: I improved to remove duplicate call here. Thanks.

front/scripts/main/views/ArticleView.ts
```diff
@@ -222,12 +223,11 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 	setupButtonsListeners: function () : void {
 		this.$('.article-content')
 			.on('click', '.pencil', (event: JQueryEventObject): void => {
-				var $sectionHeader = $(event.target).parent();
+				var $sectionHeader = $(event.target).parent().prevAll(':header[section]:first');
```
inez: Please consider using $.closest - I think it would do exactly what you want in more efficient manner.

front/scripts/main/views/ArticleView.ts
inez: Btw. that variable name $photoUpload is not very intuitive - it does not indicate what element it is referencing.

front/styles/main/module/edit/_edit-common.scss
```diff
@@ -29,3 +36,8 @@
 		}
 	}
 }
+
+.short-header {
+	display: inline-block;
+	width: calc(100% - #{$contribution-icons-space});
```
inez: The approach with either margin, padding or float didn't work out?

inez: If the header is really multiline this would trim (horizontally) all lines of that header - as opposed to trimming only first one or two lines - just so the icons can fit. Is it desired?

mari-mcmurtrie: Float does not work since text wraps around it. I tried yesterday and could not work with float approach.
If I add margin/padding, then I cannot place the icons right next to header, correct? Since there is margin there.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Ok, so I and Inez agreed that $closest does not work. I keep the code as-is.

inez:
@mari-mcmurtrie This is how I would consider handling displaying buttons (edit/upload) associated with sections: http://codepen.io/anon/pen/yNjXBe

inez:
I think this looks great now. I have only very small minor points that I'm going to add(comment) now.

front/scripts/main/views/ArticleView.ts
```diff
@@ -210,11 +211,10 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			});
 
 		this.$(':header[section]').each((i: Number, item: any): void => {
-			var $sectionHeader = this.$(item),
-				$pencil = this.$(pencil),
-				$photo = this.$(photo);
+			var $sectionHeader = this.$(item);
 
-			$sectionHeader.append($photo, $pencil);
+			$sectionHeader.prepend(this.$(iconsWrapper));
```
inez: I think $.prepend take HTML as well, so there is no need to call this.$.
jQuery can be chained, we often write like this:
```
$sectionHeader
    .prepend(iconsWrapper)
    .addClass('short-header');
```

front/scripts/main/views/ArticleView.ts
inez: This one about var name is still not addressed.

front/styles/main/module/edit/_edit-common.scss
```diff
@@ -1,3 +1,10 @@
+$contribution-icons-space : 5.2rem;
```
inez: Now - that this variable is used only once - I would just hardcoded it instead (it just doesn't seem generic at all). Specifically I would hardcode and put comment where it comes from, e.g.:
```
padding-right: 5.2rem; /* twice the width of contribution icon + extra margin */
```
or something like that.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I missed it! I address in next commit.

inez:
LGTM. Feel free to merge it if tested.

PR #1059, opened 2015-07-10
INT-140: Edit/Photo icons positioning fix for section header.

mari-mcmurtrie:
Edit and photo icon positions were optimized for h2 section header only. Fixed styling so that those icons are positioned correctly in any size of headers.


inez:
LGTM

PR #1064, opened 2015-07-13
HG-803 - un-normalized file name for lightbox handling

katanka:


rogatty:
When you open http://mediawiki119.127.0.0.1.xip.io:8000/wiki/MercuryGallery there is a section "Images as Files". Try clicking on the first image captioned "caption 1".

It doesn't work with your fix because in `MediaModel#getRefsForLightboxByTitle` we do this check: `galleryItem.title === title` where `galleryItem` is an item from the media model that comes from MW API and `title` is the one that we pass from `openLightboxForMedia`.

Titles in model are without underscores so in this example `Midnight Tango (2011) - Theatrical Trailer for Midnight Tango`. Query params are sanitized (spaces are replaced with underscores) in https://github.com/Wikia/mercury/blob/dev/front/scripts/main/components/LightboxMediaComponent.ts#L81 to be consistent with the rest of platform (article titles always have underscores and files are articles in MW).

After you removed `M.String.normalize`, the condition mentioned above won't pass because `'Midnight Tango (2011) - Theatrical Trailer for Midnight Tango' !== 'Midnight_Tango_(2011)_-_Theatrical_Trailer_for_Midnight_Tango'`.

lizlux:
I set up this test page: http://muppet.wikia.local:8000/wiki/Muppet_caps_(New_Era) (adjust the URL to your DNS settings) to try out this code change, and I'm seeing an issue. When a title in the media array has spaces instead of underscores, and the image is not part of a gallery array, the lightbox fails when reloading the page with '?file=xyz' in the url. Here's the article API for that page http://muppet.liz.wikia-dev.com/wikia.php?controller=MercuryApi&method=getArticle&title=Muppet_caps_(New_Era). Notice some images come through with underscores, and some don't, but the value of the `file` param in the URL appears to always include underscores. 

We may want to normalize both the title from the URL and the title in the media array that we're checking it against in order to be sure we're comparing apples to apples.

lizlux:
Haha, just noticed that @rogatty wrote basically the same comment already. D'oh!

lizlux:
I'm wondering if the best solution is to make sure the article API always returns image title with underscores, instead of the mis-match. @katanka, what do you think?

front/scripts/main/models/MediaModel.ts
```diff
@@ -50,13 +50,13 @@ App.MediaModel = Em.Object.extend({
 					return (<ArticleMedia[]>mediaItem).some(findInGallery, {
 						mediaIndex: mediaIndex
 					});
-				} else if ((<ArticleMedia>mediaItem).title === title) {
+				} else if (M.String.normalizeToUnderscore((<ArticleMedia>mediaItem).title) === M.String.normalizeToUnderscore(title)) {
```
lizlux: This line is more than 120 chars https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#white-space-guidelines

lizlux:
Just a couple comments but other than that, and the failing test, this LGTM.

PR #1065, opened 2015-07-13
HG-691 setting option to get locale folders names as all lower case

lizlux:
Fixing issue with country codes that have dashes in them not working well in the auth pages server-side translation code. Setting the `lowerCaseLng` option. 

Here's where that option is referenced: https://github.com/i18next/i18next/blob/master/i18next.js#L1851-L1853

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -68,7 +68,8 @@ plugins = [
 				useCookie: false,
 				detectLngFromHeaders: false,
 				detectLngFromQueryString: true,
-				detectLngQS: 'uselang'
+				detectLngQS: 'uselang',
+				lowerCaseLng: true
 			}
 		}
 	}


bkoval:
Wow, this is just awesome! One liner and it changes everything :) I've rebuilt Jenkins as it seems it failed because of non-mercury reason there.

lizlux:
Yeah, lots of work, one line of code!

It was odd when i18n worked fine on my Mac running prod mode, but not on my devbox or sandbox in any mode. I think the underlying issue is that Unix machines are case-sensitive when it comes to file names, but OS X machines aren't. Did a quick search and found this to back that up:  http://smartwebdeveloper.com/apache/htaccess-problems

katanka:
LGTM

PR #1066, opened 2015-07-13
Discuss

garthwebb:


front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,8 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionForumPostController = Em.Controller.extend({
+	init: function () {
+		this.controllerFor('application').set('noAds', true);
```
rogatty: Wow, this is hacky :D It will add ?noads=true to the URL.
/cc @nandy-andy

front/scripts/main/router.ts
```diff
@@ -70,6 +70,12 @@ App.Router.map(function () {
 		path: articlePath + 'addPhoto/:title'
 	});
 
+	this.resource('discussion', {path: 'd'}, function () {
```
rogatty: Please don't use `resource` as it's deprecated, see http://guides.emberjs.com/v1.13.0/routing/defining-your-routes/#toc_resetting-nested-route-namespace

front/scripts/main/views/DiscussionIndexView.ts
```diff
@@ -0,0 +1,5 @@
+/// <reference path='../app.ts' />
+
+App.UserProfileView = Em.View.extend({
```
rogatty: Please don't use views as they're deprecated, see http://emberjs.com/deprecations/v1.x/#toc_ember-view

You should use components instead.

front/templates/main/application.hbs
```diff
@@ -34,8 +34,11 @@
 			toggleVisibility='toggleUserMenu'
 		}}
 	{{/if}}
-	{{ad-slot name='INVISIBLE_HIGH_IMPACT' noAds=noAds}}
-	{{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
+	{{#unless noAds}}
```
rogatty: Currently we have to render ad slots and ads code will handle hiding them. This will be refactored in https://wikia-inc.atlassian.net/browse/ADEN-2189 but for now this condition shouldn't exist.

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Wikia Web Application",
   "main": "server.js",
   "scripts": {
-    "dev": "rm -rf www && gulp",
+    "dev": "rm -rf www && gulp --nosync",
+    "local": "rm -rf www && gulp",
```
rogatty: +1 for this change, can you open a separate PR to dev?

server/views/_partials/open-graph-tags.hbs
```diff
@@ -1,15 +1,15 @@
-<meta property='og:type' content='{{#if isMainPage}}website{{else}}article{{/if}}' />
+<meta property="og:type" content="{{openGraph.type}}" />
```
rogatty: How about a separate PR to dev?

front/scripts/main/controllers/DiscussionForumPostController.ts
nandy-andy: What's the `adsContext.opts.pageType` on those pages? You can see it in the browser console by executing: `Mercury.Modules.Ads.getInstance()`

front/templates/main/application.hbs
nandy-andy: I agree. We've just started talking about rebuilding/refactoring ads in Mercury. I'd like not to introduce any new changes without letting us know ;)

front/scripts/main/controllers/DiscussionForumPostController.ts
mklucsarits: @rogatty That didn't add the query parameter for me. You can see at http://glee.mattk.wikia-dev.com/d/1833581173276673026/2525402718156097295 (if/when Mercury is running there). I'm open to suggestions on how to control display of ads from other controllers; this was the simplest solution I found.

front/templates/main/application.hbs
mklucsarits: If there is a better way to do this, that's fine. What I needed was a way for ads to not display at all. Even if we use ```display: none``` on the ad containers, we have to make sure there are no margins around them that will mess with the page layout. In other words, if you plan to render the ad slots, but hide them programmatically, it should look to the user as if they were never rendered. This is not currently the case.

package.json
garthwebb: @rogatty, moved this PR here: https://github.com/Wikia/mercury/pull/1197

front/scripts/main/views/DiscussionIndexView.ts
garthwebb: I don't think we even used this.  Removing the file.

garthwebb:
@rogatty I think you have the last open questions in this PR.  The OpenGraph stuff has been copied to the PR https://github.com/Wikia/mercury/pull/1187.  If that one looks good can you give a thumbs up/down on this PR as well?

front/templates/main/components/post-detail.hbs
```diff
@@ -0,0 +1,25 @@
...
+			<img src="{{author.avatarPath}}">
+		</a>
+	</figure>
+	<a href="{{author.profileUrl}}">{{author.name}}</a>
+	<span class="timestamp">
+		&bull; {{timeAgo timestamp}}
+	</span>
+</div>
+<div class="discussion-content">
+	{{{content}}}
```
rogatty: Does it come escaped and safe from the API?

rogatty:
@garthwebb from what I see there is still an issue how to handle ads. Please sync with @nandy-andy on that.

front/templates/main/components/post-detail.hbs
garthwebb: I'll check with the team building the API but it will have to; this content will need to contain simple HTML.

front/scripts/main/controllers/DiscussionForumPostController.ts
mklucsarits: @nandy-andy To address your comment further, the issue I was dealing with is purely CSS styling. As long as the Ember ad slot component is allowed to render -- even if an ad is not displayed -- the margin and spacing at the top of the page is wrong. There is whitespace where an ad would be, whitespace that should not be there. This is not really an issue of, what is the correct way to hide ads. It's, if ads are hidden, the page should display as if there is no space for the ad to show, strictly in terms of margins and whitespace.

mklucsarits:
@rogatty @nandy-andy I've removed the changes to hide ads from this branch, so that it can be merged.

server/routes.ts
```diff
@@ -196,13 +196,6 @@ authenticatedRoutes = [
 		handler: function (request: Hapi.Request, reply: any): Hapi.Response {
 			return reply.redirect(authUtils.getRedirectUrlWithQueryString('signin', request));
 		}
-	},
-	{
-		method: 'GET',
-		path: '/signup',
```
rogatty: Is this intended as a part of this PR?
/cc @bkoval

rogatty:
LGTM

config/localSettings.d.ts
```diff
@@ -73,6 +63,18 @@ interface LocalSettings {
 	weppy: WeppyConfig;
 	workerCount: number;
 	workerDisconnectTimeout: number;
-	cdnBaseUrl: string;
-	asyncArticle: string[];
+	authCookieDomain?: string;
```
lizlux: This interface is sorted alphabetically until here.

front/scripts/main/components/PostReplyComponent.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.PostReplyComponent = Em.Component.extend({
+	classNames: ['post-reply'],
+
+	author: null,
+
+	init: function (): void {
+//		App.UserModel.find({userId: this.get('authorId')}).then((result: any): any => {
```
bkoval: It's a development mode, right? In other words, we should leave this commented line + the hardcoded userId for now?

front/templates/main/components/post-detail.hbs
```diff
@@ -0,0 +1,25 @@
+<div class="post-details">
```
lizlux: you shouldn't need this wrapping div since the component will wrap it in the same `div.post-details`

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -0,0 +1,15 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.PostDetailComponent = Em.Component.extend({
+	classNames: ['post-detail'],
+
+	author: null,
```
lizlux: It's nice to list out the properties that are passed in, i.e. `authorId: null,`, for clarity.

config/localSettings.d.ts
mklucsarits: I thought we had agreed at one time to put optional parameters last, sorted alphabetically on their own. Although I don't see it here https://github.com/Wikia/guidelines/tree/master/TypeScript

front/scripts/main/router.ts
```diff
@@ -70,6 +70,12 @@ App.Router.map(function () {
 		path: articlePath + 'addPhoto/:title'
 	});
 
+	this.route('discussion', {path: '/d'}, function () {
```
bkoval: How about resources? I noticed we don't use them in Mercury but it's nicely explained here -> http://stackoverflow.com/questions/14973556/what-is-the-difference-between-a-route-and-resource-in-new-router-api

rogatty: @bkoval deprecated, see http://guides.emberjs.com/v1.13.0/routing/defining-your-routes/#toc_resetting-nested-route-namespace

server/facets/api/discuss.ts
```diff
@@ -0,0 +1,46 @@
...
+		url: string,
+		article: Article.ArticleRequestHelper;
+
+	article = new Article.ArticleRequestHelper({
+		wikiDomain: wikiDomain
+	});
+
+	article.getWikiVariables((error: any, wikiVariables: any): void => {
+		if (error) {
+			// ???
```
bkoval: Maybe a ToDo would be nicer here? :)

front/scripts/main/components/PostReplyComponent.ts
mklucsarits: This was left in by accident. It doesn't matter though, because first of all this won't affect production, but more importantly, these init functions that fetch from UserModel are completely going away with new changes that I have not yet committed.

server/routes.ts
bkoval: Yeah, this is interesting. I don't think we need to remove the /signup redirect unless we need this path for something else or we want to remove it from varnish :)

front/scripts/main/router.ts
garthwebb: AFAIK resources are going away:
http://discuss.emberjs.com/t/this-resource-deprecation-and-loading-error-route-handling/7878/4

server/routes.ts
mklucsarits: Not sure how that happened. I will add it back!

config/localSettings.d.ts
lizlux: Ah okay. Feel free to make a PR for that. I sorta remember discussing this.

front/scripts/main/components/PostReplyComponent.ts
lizlux: Oh good. I was going to comment that there's probably a better way to set the `author` property, i.e. computed property or passing it into the component.

front/styles/main/component/_lightbox.scss
```diff
@@ -108,7 +108,7 @@
 	}
 
 	&::after {
-		background: linear-gradient(top, transparent 10%, black 100%);
+		background: linear-gradient(to bottom, transparent 10%, black 100%);
```
lizlux: :+1:

front/styles/main/component/_site-head.scss
```diff
@@ -12,6 +12,13 @@
 	text-align: center;
 	top: 0;
 	z-index: $z-site-head;
+
+	.avatar {
+		float: right;
+		position: relative;
```
lizlux: can't see this b/c login is broken on my local dev but could margins work instead of `position:relative`? Positioning is more heavy-handed if all you need is some spacing.

front/templates/main/components/post-reply.hbs
```diff
@@ -0,0 +1,22 @@
...
+		</figure>
+		<a href="{{author.profileUrl}}">
+			{{author.name}}
+		</a>
+		<span class="timestamp">
+			&bull; {{timeAgo timestamp}}
+		</span>
+	</div>
+	<div class="discussion-content">
+		{{content}}
```
lizlux: might run into escaping issues here if this `content` contains HTML.

lizlux:
This is a WIP that has minimal user-facing changes, so I don't see any blockers for merging to dev. I would comment that, going forward, it's easier to code review PRs that are per-ticket as opposed to per-feature. As it stands, there's a lot of WIP code that's being merged to dev and it will be hard to make sure it's all fixed in upcoming PRs.

front/styles/main/component/_site-head.scss
mklucsarits: Possibly. But this is was just moved to be nested here, from further down in the file. I don't want to mess with these styles right now.

front/templates/main/components/post-detail.hbs
mklucsarits: This is wrapping a small portion of the component markup, which contains some "details". Suggestions for a better class name are welcome. I realize that this whole markup will also be wrapped in `div.post-detail`.

server/facets/api/discuss.ts
mklucsarits: Ah, this whole file is more code that is already outdated and is already changed, awaiting commit. Discussion code is changing fast; that's why we just want to push this PR to dev and can continue updating the code.

front/templates/main/components/post-detail.hbs
lizlux: Ah I see, missed the `s`

garthwebb:
@lizlux this happened to be done as a feature because there was a lot of monkeying around at first with routes and faking out Hg to do what we wanted before it really supported it.  We wanted a place to work safely and that became this largish PR.  We're trying to get this merged down quickly because we want to start working in ticket branches.

front/scripts/main/components/PostReplyComponent.ts
bkoval: Let's commit the changes then, so the dev branch is as clear as possible :)

front/scripts/main/router.ts
bkoval: :+1:

server/facets/api/discuss.ts
bkoval: I'm not a fan of pushing this kind of things to dev, especially that they don't require a lot of work to look nicer and since it is an open repository we can have people from outside watching code on our default (theoretically, production-ready) branch :) But I get it, if it's going to be swapped quickly, then I'm fine with it :santa:

PR #1072, opened 2015-07-14
HG-809 - revison assets in all .hbs files

katanka:


gulp/tasks/build-views.js
```diff
@@ -39,7 +39,7 @@ gulp.task('build-views', ['scripts-front', 'copy-ts-source', 'vendor', 'build-ve
 
 		// TODO: Leave this in for now to run the normal template based assets pipeline for the duration
 		// of the test
-		gulpif('**/_layouts/auth.hbs', piper(
+		gulpif('**/_layouts/*.hbs', piper(
```
lizlux: This change looks fine, especially for a hotfix.

Just noticing some inconsistencies where, above, the glob is `'**/_layouts/**.hbs'`, whereas here, the glob is `'**/_layouts/*.hbs'`. Looks like there's no subdirectories in the `_layouts` folder, but for code clarity it would be nice if they matched. 

Also, the same conditional is being used on line 53: `gulpif('**/_layouts/*.hbs' ...` so theoretically they could be combined. 

In general, though, this hotfix LGTM.

kenkouot:
LGTM

PR #1075, opened 2015-07-15
HG-810 | run build-combined when ts file changes

bognix:
/cc @kenkouot @lizlux

### DIFF
--- a/gulp/tasks/watch.js
+++ b/gulp/tasks/watch.js
@@ -31,7 +31,7 @@ gulp.task('watch', ['build', 'build-views'], function () {
 	gulp.watch(path.join(
 		paths.scripts.front.src,
 		paths.scripts.front.files
-	), ['tslint', 'scripts-front', 'copy-ts-source', 'build-combined']).on('change', function (event) {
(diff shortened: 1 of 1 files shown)


rogatty:
LGTM

gulp/tasks/watch.js
```diff
@@ -31,7 +31,7 @@ gulp.task('watch', ['build', 'build-views'], function () {
 	gulp.watch(path.join(
 		paths.scripts.front.src,
 		paths.scripts.front.files
-	), ['tslint', 'scripts-front', 'copy-ts-source']).on('change', function (event) {
+	), ['tslint', 'scripts-front', 'copy-ts-source', 'build-combined']).on('change', function (event) {
```
lizlux: I agree that `build-combined` should be run every time a ts file is changed. We can remove `scripts-front` from this list though b/c `build-combined` includes `scripts-front`. 

@kenkouot does that match your vision for how this ought to work?

lizlux:
Merging this for now and @katanka will do the cleanup with HG-721

PR #1079, opened 2015-07-16
HG-721 - clean up gulp tasks

katanka:


hakubo:
This PR has more merges inside that actual changes introduced... :)
also we are considering moving to broccoli does it make sense to update those tasks?

rogatty:
This is just adding comments and renaming of a single task, so not really "updating those tasks" ;)
Would be nice to have it in one or two commits though.

LGTM anyway.

gulp/tasks/default.js
```diff
@@ -1,3 +1,8 @@
+/*
+ * default
+ * The default build task
```
lizlux: Might be helpful to add a note that the `default` task is what gets run when just calling `gulp` by itself.

gulp/tasks/docs-front.js
```diff
@@ -1,3 +1,8 @@
+/*
+ * docs-front
+ * Creates documentation for front scripts
+ */
```
lizlux: I didn't realize we had this! @katanka, do you happen to know where the docs can be found? Might be helpful to put that info here.

lizlux:
LGTM

PR #1100, opened 2015-07-23
INT-124: Wikia-In-Your-Language for Mercury

mari-mcmurtrie:
As a mobile user, I want to know if there is a wikia in my own language when i am on a wikia. We show a notification if a user is on wikia which has the same language wikia as the user's browser setting. If a user close the alert without going to the alternative wikia, then we record the action and now show the same alert for 1 day.


mari-mcmurtrie:
@Grunny, @inez  will you code review this? Thanks!

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,23 @@ App.AlertNotificationComponent = Em.Component.extend({
...
 	alert: null,
 
-	notificationExpiry: 10000,
 	timeout: null,
 
 	didInsertElement: function (): void {
-		this.set('timeout', Em.run.later(this, (): void => {
-			this.dismissNotification();
-		}, this.get('notificationExpiry')));
+		var expiry = this.get('alert').notificationExpiry;
```
inez: Do it like this instead of having to rely on typeof.
```
var expiry = this.get('alert').notificationExpiry || 1000;
```

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,23 @@ App.AlertNotificationComponent = Em.Component.extend({
...
+			expiry = 10000; //default
+		}
+
+		if (expiry > 0) {
+			this.set('timeout', Em.run.later(this, (): void => {
+				this.dismissNotification();
+			}, expiry));
+		}
+
+		if (typeof this.get('alert').callbacks.onInsertElement === 'function') {
```
inez: What is this for?

front/scripts/main/views/ArticleView.ts
```diff
@@ -435,5 +437,49 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+
+	handleWikiaInYourLang: function(): void {
+		if (this.shouldShowWikiaInYourLang()) {
+			App.WikiaInYourLangModel.check().then(function(model: typeof App.WikiaInYourLangModel): void {
+				if (model.exists) {
+					var appController = this.get('controller').get('controllers.application');
+					appController.addAlert('', model.message, 60000, {
+						onInsertElement: function(alert: any): void {
+							alert.on('click', 'a:not(.close)', (event: any) => {
+								M.track({
```
inez: This is way way too nested.

front/scripts/main/views/ArticleView.ts
```diff
@@ -435,5 +437,49 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+									action: M.trackActions.click,
+									category: 'wikiaInYourLangAlert',
+									label: 'link'
+								});
+							});
+						},
+						dismissAlert: function(): void {
```
inez: Please use consistent names for callback - I suggest this one would be called onDismissAlert

front/scripts/main/views/ArticleView.ts
```diff
@@ -435,5 +437,49 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+							var key = 'wikiaInYourLang.alertDismissed',
```
inez: No need to define those variables

front/templates/main/components/alert-notification.hbs
```diff
@@ -1,2 +1,2 @@
-{{alert.message}}
+{{{alert.message}}}
```
inez: Per our discussion with @Grunny. Please do not by default display unescaped HTML - it might be dangerous in the future when we start using more of alerts.

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,41 @@
...
+	exists: null,
+	message: null,
+	nativeDomain: null
+});
+
+App.WikiaInYourLangModel.reopenClass({
+	check: function(): Em.RSVP.Promise {
+		var browserLang = navigator.language || navigator.browserLanguage;
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			Em.$.ajax({
```
inez: Well, you could use getJson instead of ajax so you wouldn't have to specify explicitly dataType? couldn't you?

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,41 @@
...
+				url: M.buildUrl({ path: '/wikia.php' }),
+				data: {
+					controller: 'WikiaInYourLangController',
+					method: 'getNativeWikiaInfo',
+					format: 'json',
+					targetLanguage: browserLang
+				},
+				dataType: 'json',
+				success: (resp: any): void => {
+					var modelInstance = App.WikiaInYourLangModel.create({
```
inez: I would go with:
```
var modelInstance = App.WikiaInYourLangModel.create({
    exists: resp.success,
    nativeDomain: resp.nativeDomain || null,
    message: resp.message || null
});
```
Up to you

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,41 @@
...
+'use strict';
+
+App.WikiaInYourLangModel = Em.Object.extend({
+	exists: null,
+	message: null,
+	nativeDomain: null
+});
+
+App.WikiaInYourLangModel.reopenClass({
+	check: function(): Em.RSVP.Promise {
```
inez: From check method I would expect boolean resolution/return, but this actually resolved to a model object - does not seem right to me.

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: Not all alert has "onInsertElement" in callbacks. That is why I am checking if it is defined and it is a function.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -40,6 +40,9 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, App.TrackClickMix
 			M.VariantTesting.activate();
 			this.controller && this.controller.hideLoader();
 
+			// Clear notification alerts for the new route
+			this.controller.get('alertNotifications').clear();
```
inez: As we talked before - shouldn't notification componant itself react to some transition event? That way you have less of coupling.

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: I see what you mean. I should change function name.

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: So if resp.success is false, then the response still has "nativeDomain" but missing "message".  The response looks like:
{"success":false,"nativeDomain":"ja.visualeditor.wikia.com","error":"A native wikia not found."}

If wikia in the language does not exists, then we do not care about nativeDomain and message at all. So I would rather not initialized those data fields at all. Thanks.

front/scripts/main/routes/ApplicationRoute.ts
mari-mcmurtrie: I and Inez discussed about this and came to conclusion that the code is ok as is.

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: Thanks, changed.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Nested part is factored out.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Changed as suggested.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Changed.

front/templates/main/components/alert-notification.hbs
mari-mcmurtrie: Thanks, changed as we discussed.

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: I changed this to getJson.

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: Ok, so I and Inez discussed about this and agreed that if no such wiki exists, then modelInstance = null is good. I made that change.

mari-mcmurtrie:
@Grunny, will you please code review this? Thanks!

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,20 @@ App.AlertNotificationComponent = Em.Component.extend({
...
 	alert: null,
 
-	notificationExpiry: 10000,
 	timeout: null,
 
 	didInsertElement: function (): void {
-		this.set('timeout', Em.run.later(this, (): void => {
-			this.dismissNotification();
-		}, this.get('notificationExpiry')));
+		var expiry = this.get('alert').notificationExpiry || 10000; //Default to 10 seconds.
```
hakubo: could this be a normal property
then you could access that by
`this.defaultExpiry;`

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,20 @@ App.AlertNotificationComponent = Em.Component.extend({
...
+
+		if (expiry > 0) {
+			this.set('timeout', Em.run.later(this, (): void => {
```
hakubo: this.set('timeout', Em.run.later(this, this.dismissNotification, expiry));

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -28,6 +35,9 @@ App.AlertNotificationComponent = Em.Component.extend({
 	actions: {
 		close: function (): void {
 			this.dismissNotification();
+			if (typeof this.get('alert').callbacks.onCloseAlert === 'function') {
```
hakubo: Ember has something called
Em.K especially for this usecase
so you could alwyas call onCloseAlert not caring if user specified a function or not

front/scripts/main/mixins/AlertNotificationsMixin.ts
```diff
@@ -4,10 +4,13 @@
 App.AlertNotificationsMixin = Em.Mixin.create({
 	alertNotifications: Ember.A(),
 
-	addAlert: function (alertType: string, message: string): void {
+	addAlert: function (alertType: string, message: string, notificationExpiry: number, unsafe: boolean = false, callbacks: any = {}): void {
```
hakubo: 5 parameters - please consider using object as param

front/scripts/main/mixins/AlertNotificationsMixin.ts
```diff
@@ -4,10 +4,13 @@
...
 		this.get('alertNotifications').pushObject({
 			type: alertType,
-			message: message
+			message: message,
```
hakubo: in ES6 you can use object shorthands:

```js
var obj = {
  message,
  unsafe,
  callback
}
```
and so on

front/scripts/main/mixins/AlertNotificationsMixin.ts
inez: @mari-mcmurtrie just a reminder - per our discussion - consider having parameter "type" instead of "alertType" - in the same way you have "message" and not "alertMessages".

front/scripts/main/views/ArticleView.ts
```diff
@@ -260,5 +262,59 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+					label: err || 'error'
+				});
+			});
+		}
+	},
+
+	createAlert: function(model: typeof App.WikiaInYourLangModel): void {
+		var appController = this.get('controller').get('controllers.application');
+		appController.addAlert('', model.message, 60000, true, {
+			onInsertElement: function(alert: any): void {
```
inez: Well - nothing really to be addressed but just wanted to comment on this :) It would be awesome if actually addAlert returned some sort of alert object with perhaps $element as it's property so you could subscribe to some events without having to wait till the alert is really inserted into the page. But yeah, as I said - it's not really about this PR.

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,63 @@
...
+					); //write to cache
+					resolve(modelInstance);
+				}
+			).fail(function(err: any): void {
+				reject(err);
+			});
+		});
+	},
+
+	checkCacheForLoad: function(browserLang: string): typeof App.WikiaInYourLangModel {
```
inez: @mari-mcmurtrie We talked about it when you had method called just "check" - it sort of implies it's gonna return boolean, not really a model object (or null). Also, I would say that "cache" is internal detail of the implementation and information about it shouldn't really be exposed outside (as it is now through the name of the method) unless there is a reason for it.

front/scripts/main/views/ArticleView.ts
```diff
@@ -260,5 +262,59 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
...
+					label: 'close'
+				});
+			}
+		});
+	},
+
+	shouldShowWikiaInYourLang: function(): boolean {
+		var value = window.localStorage.getItem('wikiaInYourLang.alertDismissed'),
+		    now = new Date().getTime(),
+			hasNotCloseWikiaInYourLangAlert = !value || (now - value > 86400000), //1 day 86400000
```
Grunny: Whitespace inconsistency. This variable name is also pretty long, can you shorten it (i.e. something like `active` or `notDismissed`)?

front/scripts/main/mixins/AlertNotificationsMixin.ts
Grunny: `notificationExpiry` can probably just be `expiry` as well.

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,63 @@
...
+		    valueJson = JSON.parse(window.localStorage.getItem(key)),
+		    now = new Date().getTime();
+		if (!valueJson || now - valueJson.timestamp > 2592000000) { //we cache 30 days (2592000000)
+			return null;
+		}
+		return valueJson.model;
+	},
+
+	getCacheKey: function(lang: string): string {
+		return lang + '-' + window.location.host;
```
Grunny: The cache key should be something like `lang + '-WikiaInYourLang'`. It's local storage so it's per domain anyway, and a key of `ja-starwars.wikia.com` isn't specific to WikiaInYourLang at all.

front/scripts/main/components/AlertNotificationComponent.ts
rogatty: If `alert.callbacks` is undefined this will fail.

```
onInsertElement = this.get('alert.callbacks.onInsertElement');
if (typeof onInsertElement === 'function') {
  onInsertElement(this.$());
}
```

front/scripts/main/models/WikiaInYourLangModel.ts
rogatty: Agreed with the first part, `getFromCache` would be more descriptive.
As for the second comment - getting from cache is exactly what this method does so I don't see why not put in in the name. And it's used internally only.

front/scripts/main/views/ArticleView.ts
```diff
@@ -128,6 +129,7 @@ App.ArticleView = Em.View.extend(App.AdsMixin, App.LanguagesMixin, App.ViewportM
 			this.loadTableOfContentsData();
 			this.injectAds();
 			this.setupAdsContext(model.get('adsContext'));
+			this.handleWikiaInYourLang();
```
rogatty: Why is it in `ArticleView`? It doesn't have anything to do with articles, right? How about moving all this logic to `ApplicationView`?

rogatty: Or even better - create a `WikiaInYourLangComponent` and use it in `application.hbs`.

test/specs/front/scripts/main/components/AlertNotificationsComponent.js
```diff
@@ -1,8 +1,8 @@
 moduleForComponent('alert-notifications', 'AlertNotificationsComponent');
 
 test('Dismissing alert', function () {
-	var alertOne = { type: 'success', message: 'Success message' },
-		alertTwo = { type: 'error', message: 'Error message' },
+	var alertOne = { type: 'success', message: 'Success message', callbacks: {} },
+		alertTwo = { type: 'error', message: 'Error message', callbacks: {} },
```
rogatty: Why changing it? What will happen if there is no `callbacks` passed?

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: callbacks has default set.

front/scripts/main/mixins/AlertNotificationsMixin.ts
mari-mcmurtrie: Do I have to use it just because I can use shorthand?

test/specs/front/scripts/main/components/AlertNotificationsComponent.js
mari-mcmurtrie: Test fails.

rogatty: Why does it fail? Will all alerts without callbacks stop working?

front/scripts/main/components/AlertNotificationComponent.ts
rogatty: Still, IMO this version looks more elegant.

mari-mcmurtrie: "This version" is the code snippet you wrote in your comment? If so, then I totally agree.

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: ok I got that I suck naming. I fix it as suggested. About the caching, it is internal use as Igor said. In other words, cache implementation is well hidden from outside of WikiaInYourLangModel, so callers who want to get the model has no knowledge of if the model it gets is from cache or not. That is my intention and I do not see any change necessary.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Grunny told me to put in ArticleView so let me talk to him about it first.

test/specs/front/scripts/main/components/AlertNotificationsComponent.js
mari-mcmurtrie: In this test, alertOne and alertTwo are instantiated not via addAlert() function, but rather directly. That is why I had to add callbacks={}.
In code, alerts are instantiated via addAlert(), which has default callbacks set.

front/scripts/main/views/ArticleView.ts
inez: Btw. I just noticed it, having method createAlert in ArticleView which is actually very specific to Wikia in your language feature (and required WikiaInYourLangModel) seems pretty odd. ArticleView is definitely not the place where I would search for that feature, and method createAlert on ArticleView is definitely not the one from which I would expect handling it.

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: Thanks but this code is the same as the one above Igor left note. I follow his suggested pattern for consistency.

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: I will change to have default set when alert is created, then as you suggested below, I am gonna make parameters for addAlert to be an object.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Thanks for the note, Inez!

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: I renamed as suggested.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: Fixed white space. As for variable name, I tried to be descriptive but I guess it is too long. Changed as suggested.

front/scripts/main/mixins/AlertNotificationsMixin.ts
mari-mcmurtrie: I changed this as object. Variables are renamed as suggested.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: createAlert() is just a wrapper for appController.addAlert(). I just factored out logic from handleWikiaInYourLang(). So there is nothing significant relation between createAlert() function an ArticleView. Anyways, you disagree that code lives in ArticleView to start with, I consider to factor out to somewhere more appropriate.

front/scripts/main/mixins/AlertNotificationsMixin.ts
```diff
@@ -1,13 +1,30 @@
...
+}
+
 App.AlertNotificationsMixin = Em.Mixin.create({
 	alertNotifications: Ember.A(),
 
-	addAlert: function (alertType: string, message: string): void {
+	addAlert: function(alertData: AlertData): void {
+		var type = alertData.type,
+		    message = alertData.message,
+		    expiry = alertData.expiry ? alertData.expiry : 10000,
```
Grunny: I'd shorten these down to just `expiry = alertData.expiry || 10000,` rather than the ternary operator.

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,19 @@ App.AlertNotificationComponent = Em.Component.extend({
...
-		this.set('timeout', Em.run.later(this, (): void => {
-			this.dismissNotification();
-		}, this.get('notificationExpiry')));
+		var expiry = this.get('alert').expiry;
+
+		if (expiry > 0) {
+			this.set('timeout', Em.run.later(this, this.dismissNotification, expiry));
+		}
+
+		var onInsertElement = this.get('alert.callbacks.onInsertElement');
```
Grunny: One var per function, so this should go up with `expiry`.

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -28,6 +34,10 @@ App.AlertNotificationComponent = Em.Component.extend({
 	actions: {
 		close: function (): void {
 			this.dismissNotification();
+			var onCloseAlert = this.get('alert.callbacks.onCloseAlert');
```
Grunny: Should be moved to the top of the function block.

front/scripts/main/views/ArticleView.ts
Grunny: I just asked if it was specific to articles or just all page views, so per our discussion, I think ApplicationView is the right way to go.

@rogatty WikiaInYourLang just makes use of alert notifications for now so I'm not sure a separate component is needed. If this is redone later with a new design and needs its own component rather than making use of alert notifications, that might make sense at that time. What do you think?

inez:
@mari-mcmurtrie I just noticed that this branch can't be merged at the moment because of conflicts - please make sure to solve them soon - so people are reviewing code that is actually to be merged. Thanks.

front/scripts/main/views/ArticleView.ts
mari-mcmurtrie: I moved the logic into ApplicationView. Thanks

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: Changed.

front/scripts/main/components/AlertNotificationComponent.ts
mari-mcmurtrie: thanks, changed.

front/scripts/main/mixins/AlertNotificationsMixin.ts
mari-mcmurtrie: I forgot about that! Thanks

mari-mcmurtrie:
@inez I merged dev so this should be good. @Grunny please code review this again. Thank you!

front/scripts/main/components/AlertNotificationsComponent.ts
```diff
@@ -4,6 +4,8 @@
 interface AlertNotification {
 	type: string;
 	message: string;
+	notificationExpiry: number;
```
inez: Wouldn't calling it "expiry" be sufficient enough? (I would actually call it duration :))

front/scripts/main/controllers/AddPhotoController.ts
```diff
@@ -12,7 +12,7 @@ App.AddPhotoController = Em.Controller.extend({
 	handleAddContentSuccess: function(data: any): void {
 		var title = this.get('model.title');
 		this.transitionToRoute('article', title).then((): void => {
-			this.get('controllers.application').addAlert('success', i18n.t('app.add-photo-success'));
+			this.get('controllers.application').addAlert({ type: 'success', message: i18n.t('app.add-photo-success') });
```
inez: Reads way better - doesn't it?
```
this.get('controllers.application').addAlert({
    type: 'success',
    message: i18n.t('app.add-photo-success')
});
```

front/scripts/main/views/ApplicationView.ts
```diff
@@ -147,5 +150,67 @@ App.ApplicationView = Em.View.extend({
...
+			}
+		};
+		appController.addAlert(alertData);
+	},
+
+	shouldShowWikiaInYourLang: function(): boolean {
+		var value = window.localStorage.getItem('wikiaInYourLang.alertDismissed'),
+		    now = new Date().getTime(),
+		    notDismissed = !value || (now - value > 86400000), //1 day 86400000
+		    isJpOnNonJpWikia = this.get('isJapaneseBrowser') && !this.get('isJapaneseWikia');
```
inez: Jp or Ja? ;)

front/scripts/main/views/ApplicationView.ts
```diff
@@ -147,5 +150,67 @@ App.ApplicationView = Em.View.extend({
...
+		console.log('qqq notDiimissed', notDismissed); console.log('qqq isJPOnNonJP', isJpOnNonJpWikia);
```
inez: console.log - well I wouldn't use it on production :)

front/scripts/main/views/ApplicationView.ts
```diff
@@ -147,5 +150,67 @@ App.ApplicationView = Em.View.extend({
...
+					label: err || 'error'
+				});
+			});
+		}
+	},
+
+	createAlert: function(model: typeof App.WikiaInYourLangModel): void {
+		var appController = this.get('controller'),
+		    alertData = {
+			type: '',
```
inez: Make empty a default. Per our conversation.

front/scripts/main/components/AlertNotificationsComponent.ts
inez: I have a problem understanding what's the point of having AlertNotification and AlertData - can't review further without knowing it :)

front/scripts/main/views/ApplicationView.ts
```diff
@@ -147,5 +150,67 @@ App.ApplicationView = Em.View.extend({
...
+					alert.on('click', 'a:not(.close)', (event: any) => {
+						M.track({
+							action: M.trackActions.click,
+							category: 'wikiaInYourLangAlert',
+							label: 'link'
+						});
+					});
+				},
+				onCloseAlert: function(): void {
+					window.localStorage.setItem('wikiaInYourLang.alertDismissed', new Date().getTime().toString());
```
inez: "wikiaInYourLang.alertDismissed" - used twice - I suggest converting it to some sort of a constant.

mari-mcmurtrie: I added a function to return the string.

front/scripts/main/views/ApplicationView.ts
mari-mcmurtrie: done. thanks.

front/scripts/main/views/ApplicationView.ts
mari-mcmurtrie: It is "JP". 2-char locale :)

front/scripts/main/controllers/AddPhotoController.ts
mari-mcmurtrie: Agreed. Also, "type" is now optional so I switched an order - message is 1st, then type.

front/scripts/main/components/AlertNotificationsComponent.ts
mari-mcmurtrie: This was actually a good catch. AlertNotification and AlertData are duplicate. I consolidated.

mari-mcmurtrie:
@inez @Grunny when you have time, please review. thank you so much!

front/scripts/main/views/ApplicationView.ts
inez: I was pretty sure JA is for language and JP is for region. And we are here depending on language settings not on geo location (nor region).

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,63 @@
...
+				}
+			).fail(function(err: any): void {
+				reject(err);
+			});
+		});
+	},
+
+	getFromCache: function(browserLang: string): typeof App.WikiaInYourLangModel {
+		var key = App.WikiaInYourLangModel.getCacheKey(browserLang),
+		    valueJson = JSON.parse(window.localStorage.getItem(key)),
```
inez: Why this variable name have postfix "Json"?

front/scripts/main/views/ApplicationView.ts
mari-mcmurtrie: Hmmm yeah you are right. 'JA' is language code. I was using a country code ('JP'). In addition, we use 'JA' as our Japanese wiki domain (ja.destiny.wikia.com) so we should use 'JA'.  thanks for checking.

front/scripts/main/models/WikiaInYourLangModel.ts
mari-mcmurtrie: I changed to "value" from "valueJson". I tried to be descriptive :(

mari-mcmurtrie:
@inez please review again. thank you!

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,63 @@
...
+
+'use strict';
+
+App.WikiaInYourLangModel = Em.Object.extend({
+	message: null,
+	nativeDomain: null
+});
+
+App.WikiaInYourLangModel.reopenClass({
+	load: function(): Em.RSVP.Promise {
```
Grunny: Since you're doing it throughout, just a note that our [convention](https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#functions) is for there to be a space between the function keyword and the left parenthesis in anonymous function.

front/scripts/main/views/ApplicationView.ts
```diff
@@ -147,5 +150,69 @@ App.ApplicationView = Em.View.extend({
 				controller.send('handleLink', target);
 			}
 		}
+	},
+
+	handleWikiaInYourLang: function(): void {
+		if (this.shouldShowWikiaInYourLang()) {
+			App.WikiaInYourLangModel.load()
+			.then(function(model: typeof App.WikiaInYourLangModel): void {
```
Grunny: You can use the [arrow function syntax](https://github.com/Wikia/guidelines/blob/master/TypeScript/README.md#lambda-syntaxarrow-functions) here to capture the context, rather than the `.bind` below.

front/scripts/main/models/WikiaInYourLangModel.ts
rogatty: Even better, you can use `load(): Em.RSVP.Promise {`.

We started to use it in multiple places (for example in [CuratedContentEditorModel](https://github.com/Wikia/mercury/blob/dev/front/scripts/main/models/CuratedContentEditorModel.ts#L23)) and will probably convert all code to use it, one day.

mari-mcmurtrie: I changed as Igor suggested.

front/scripts/main/views/ApplicationView.ts
mari-mcmurtrie: thanks, I changed.

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,63 @@
...
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			Em.$.getJSON(
+				M.buildUrl({ path: '/wikia.php' }),
+				{
+					controller: 'WikiaInYourLangController',
+					method: 'getNativeWikiaInfo',
+					format: 'json',
+					targetLanguage: browserLang
+				},
+				function (resp: any): void {
```
Grunny: Since you're using `.fail` below, I'd use `.done` here for consistency, which would also de-indent this, i.e.
```typescript
Em.$.getJSON(
    M.buildUrl({ path: '/wikia.php' }),
    {
        controller: 'WikiaInYourLangController',
        method: 'getNativeWikiaInfo',
        format: 'json',
        targetLanguage: browserLang
    }
).done((resp: any): void => {
    var modelInstance: any = null;
    if (resp.success) {
        modelInstance = App.WikiaInYourLangModel.create({
            nativeDomain: resp.nativeDomain,
            message: resp.message
        });
    }
    window.localStorage.setItem(
        App.WikiaInYourLangModel.getCacheKey(browserLang),
        JSON.stringify({ model: modelInstance, timestamp: new Date().getTime() })
    ); //write to cache
    resolve(modelInstance);
}).fail((err: any): void => {
    reject(err);
});
```

front/scripts/main/models/WikiaInYourLangModel.ts
```diff
@@ -0,0 +1,63 @@
...
+	nativeDomain: null
+});
+
+App.WikiaInYourLangModel.reopenClass({
+	load(): Em.RSVP.Promise {
+		var browserLang = navigator.language || navigator.browserLanguage,
+		    model = App.WikiaInYourLangModel.getFromCache(browserLang); //read from cache
+
+		if (model) {
+			return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
```
Grunny: Just a thought, but since you're returning a promise anyway, I think it might be slightly cleaner to remove this one and just have something like this at the start of one below:
```typescript
if (model) {
	resolve(model);
	return;
}

...
```

Grunny:
In general, I think this looks OK now. @inez @rogatty Any more thoughts on this?

rogatty:
Back to the https://github.com/Wikia/mercury/pull/1100#discussion_r35722532.

I still think that we should have the logic outside of the `ApplicationView`. Components don't need renderable templates to make sense. Wikia-In-Your-Language is a separate feature and it seems wrong to have whole application coupled with it. Especially `createAlert` and `getAlertKey` functions in `ApplicationView` don't look good :see_no_evil:.

Can we just put `{{wikia-in-your-language}}` in `application.hbs` and move all new logic to `WikiaInYourLangComponent`? There is only one thing to change: `appController.addAlert(alertData);` -> `this.addAlert(alertData)`, just remember to extend component with `App.AlertNotificationsMixin`.

front/scripts/main/models/WikiaInYourLangModel.ts
rogatty: Agreed, that's what we do in `ArticleModel`.

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,19 @@ App.AlertNotificationComponent = Em.Component.extend({
...
 	alert: null,
 
-	notificationExpiry: 10000,
 	timeout: null,
 
 	didInsertElement: function (): void {
-		this.set('timeout', Em.run.later(this, (): void => {
-			this.dismissNotification();
-		}, this.get('notificationExpiry')));
+		var expiry = this.get('alert').expiry,
```
rogatty: `this.get('alert.expirty')` is safer as it won't fail if `alert` is null.

front/scripts/main/components/AlertNotificationComponent.ts
```diff
@@ -8,13 +8,19 @@ App.AlertNotificationComponent = Em.Component.extend({
...
+		    onInsertElement = this.get('alert.callbacks.onInsertElement');
+
+		if (expiry > 0) {
+			this.set('timeout', Em.run.later(this, this.dismissNotification, expiry));
```
rogatty: Will this work if there are more than 1 alerts with `expiry > 0`?

Grunny: Yep, each alert is a separate `AlertNotificationComponent`, and so a separate timeout is set.

mari-mcmurtrie:
@Grunny @rogatty please code review. @rogatty, I now have better understanding of what "component" is. Thanks!

Grunny:
LGTM. Looks much cleaner now. :)

front/scripts/main/views/ApplicationView.ts
```diff
@@ -18,7 +20,7 @@ interface EventTarget {
 	tagName: string;
 }
 
-App.ApplicationView = Em.View.extend({
+App.ApplicationView = Em.View.extend(App.LanguagesMixin, {
```
rogatty: Is it still needed?

front/scripts/main/components/WikiaInYourLangComponent.ts
```diff
@@ -0,0 +1,76 @@
...
+						label: 'close'
+					});
+				}
+			}
+		};
+		this.addAlert(alertData);
+	},
+
+	shouldShowWikiaInYourLang: function (): boolean {
+		var value = window.localStorage.getItem(this.getAlertKey()),
```
rogatty: You should use component property instead, so it would be `this.get('alertKey')`.

front/scripts/main/components/WikiaInYourLangComponent.ts
```diff
@@ -0,0 +1,76 @@
+/// <reference path="../app.ts" />
+/// <reference path="../mixins/AlertNotificationsMixin.ts" />
+/// <reference path="../mixins/LanguagesMixin.ts" />
+
+
+'use strict';
+
+App.WikiaInYourLangComponent = Em.Component.extend(App.AlertNotificationsMixin, App.LanguagesMixin, {
+	classNames: ['wikia-in-your-lang'],
```
rogatty: No need to add class if you don't want to render it. Also you can use `tagName: ''` to not render wrapper element at all. I found no official documentation for it unfortunately :(

mari-mcmurtrie: @rogatty I have a question.
When I first added {{wikia-in-your-lang}} in application.hbs, it cannot find WikiaInYourLangComponent.ts. That is why I added classNames: ['wikia-in-your-lang'] there. After I added the line, it started working.

But maybe my perception is wrong because now I remove classNames: ['wikia-in-your-lang'] and still works.  I just cannot find why application.hbs can find WikiaInYourLangComponent.ts by this {{wikia-in-your-lang}} reference. Who maps WikiaIYourLangComponent.ts to {{wikia-in-your-lang}}?

front/scripts/main/views/ApplicationView.ts
mari-mcmurtrie: good catch. removed.

mari-mcmurtrie:
@rogatty please CR. thank you!

rogatty:
Unit test is failing:
```
WikiaInYourLangComponent shouldShowWikiaInYourLang FAILED
	should show if never shown + isJaOnNonJaWikia
	Expected: true
	Actual: false
```

front/scripts/main/components/WikiaInYourLangComponent.ts
```diff
@@ -0,0 +1,72 @@
...
+		}
+	},
+
+	createAlert: function (model: typeof App.WikiaInYourLangModel): void {
+		var alertData = {
+			message: model.message,
+			expiry: 60000,
+			unsafe: true,
+			callbacks: {
+				onInsertElement: (alert: any): void => {
```
rogatty: Does it work like that? The standard way to write it would be `onInsertElement(alert: any): void => {`.

The same below.

rogatty:
LGTM apart from the failing unit test.

PR #1104, opened 2015-07-23
SOC-1089 Renaming zh-tw to zh in mercury repository

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-1089

@kenkouot @lizlux @Warkot @kvas-damian Please, let me know if this change won't break anything for your teams.


kvas-damian:
@Grunny Don't we need to update crowdin scripts to match this change?

Warkot:
LGTM, but be aware that Crowdin outputs the files in the `lang-variant` format, so they have to be altered manually after each project build→download→update.

kenkouot:
Don't think this will affect anything on HG core end atm so LGTM

Grunny:
@kvas-damian @Warkot we can probably just setup aliasing for mercury for `zh-TW` -> `zh` in the crowdin tools, so you don't need to modify it, as we do for MediaWiki and iOS, or you can do this in the crowdin.conf files using something like:
```
export.zh-TW = "${export.base}/zh/"
```

kvas-damian:
thanks @Grunny it works. Update for config here: https://github.com/Wikia/mercury/pull/1136

PR #1118, opened 2015-07-27
INT-121: Hapi app for Japan homepage with base gulp tasks and TS/SCSS

perjg:
This is an alternative implementation of INT-121 (As a Japanese user, I want to access a localized homepage.)

The original pull request is here:
https://github.com/Wikia/mercury/pull/1002

It adds routes to serve the Japan homepage from the main mercury app.

There were several comments in the PR suggesting that it would be better to serve the homepage as a separate webapp, but with the code kept inside the mercury repo.

This PR contains the following:

* Independent Japanese homepage webapp contained within the homepage directory
* bower packages, npm packages separate from main app
* Hapi used for backend app (plain JS for now)
* Typescript used for client app
* Separate gulp tasks to build everything

Missing:
* Build for deployment (currently only builds and runs dev version)
* Bunyan logging


hakubo:
Do you think you could make 'homepage' inherit some of the gulp task instead of C/P'ing them?

perjg:
@hakubo I cleaned up the gulp tasks a bit. Now I use the environment.js script in utils directly from the mercury directory rather than copying it.

As for inheriting gulp tasks, do you mean to include them directly from the mercury directory? This might be possible to do for a few of the tasks, but wouldn't it be better to refactor the shared tasks first, otherwise it would be easy to accidentally break the homepage build by changing tasks in mercury.

hakubo:
Yes, I meant including them directly from some shared folder in mercury repo. 
Yes this need a bit of refactor, but I think it'll be much less to refactor now as opposed to when shared folder and ja folder diverge and we'll have hard time thinking how to make it work together back again.

I'm a small fun of copypaste as well for quick and dirty solutions, but I worry that if we leave it as is now  we will have to pay for this soon.

homepage/.jshintrc
```diff
@@ -0,0 +1,84 @@
+{
```
bkoval: Do we really want to use a different jshint setup only for this directory? Why not stick to the global jshint?

homepage/front/scripts/main/main.ts
```diff
@@ -0,0 +1,4 @@
+/// <reference path="../../../../typings/jquery/jquery.d.ts" />
+/// <reference path="../../../../typings/slick/slick.d.ts" />
+
+'use strict';
```
bkoval: This looks like an empty file, is it intentional to add it?

homepage/server/index.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Wikia (Japan) Homepage
+ *
+ * @author Per Johan Groland <pgroland@wikia-inc.com>
+ */
+
+var Hapi = require('hapi');
```
bkoval: I think there's a couple of coding conventions that could be addressed in this file: https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md

homepage/front/scripts/main/main.ts
perjg: Yes, it is a placeholder file, it will get more content from the next PR.
I wanted it in there to make sure that TS compilation tasks are working.

perjg:
@hakubo I added the gulp-bower task just as a convenience. It is just a personal preference, I prefer to have everything controlled from the gulp tasks, and it works the same way as 'bower install'

perjg:
@hakubo Regarding refactoring the gulp tasks to share more between the main mercury tasks and the homepage tasks. Would it be ok to handle this as a separate PR?

I have some more adjustments to do for another PR to build on this one (INT-134), and once I have all the features I need on the new homepage set up, I should have a better idea of the gulp tasks needed and which ones can be shared.

homepage/server/index.js
perjg: Fixed. I was missing the proper git hooks from the guidelines.

perjg:
Unless there are further things to fix, I will merge this PR on Monday.

I am leaving gulp task refactoring for a separate task.

.jshintrc
```diff
@@ -56,27 +54,20 @@
 	"browser": true,
 	"devel": false,
 	"globals": {
-		"App": false,
-		"Em": false,
-		"Ember": false,
```
rogatty: Please don't remove our globals :sob:

rogatty:
@perjg is this supposed to be exclusively for Japanase homepage or are there plans for using it for all languages? I just wonder if `homepage` directory name is the most appropriate one.

Anyway, this is no blocker as we can rename it later.

homepage/gulp/paths.js
```diff
@@ -0,0 +1,76 @@
...
+    baseServer = '/server',
+    baseFront = '/front';
+
+module.exports = {
+    base: basePath,
+    baseFull: path.resolve(basePath),
+    baseFullServer: path.resolve(baseServer),
+    baseFullFront: path.resolve(baseFront),
+    vendor: {
+        homepage: {
```
hakubo: since this is in homepage folder do we need to have another layer here?

homepage/gulp/tasks/scripts.js
```diff
@@ -0,0 +1,48 @@
...
+ * scripts
+ * Compiles client TypeScript files
+ */
+
+var gulp = require('gulp'),
+	uglify = require('gulp-uglify'),
+	gulpif = require('gulp-if'),
+	folders = require('gulp-folders'),
+	ts = require('gulp-typescript'),
+	sourcemaps = require('gulp-sourcemaps'),
```
hakubo: This is buggy and got removed
Now because this is a copy, you'd have to do it here as well
can we consider storing configuration in homepage folder
but leave tasks itself in parent folder so they are shared?

rogatty: This was buggy and got removed from the main app probably because we've started to load resources asynchronously. It may not be a problem here.

.jshintrc
perjg: Sorry about that, reverting to the latest .jshintrc on dev

homepage/gulp/paths.js
perjg: My plan is to try and refactor some of the gulp tasks so that they can be shared between mercury and the homepage later. For that, I wanted to keep all the config consistently at an extra layer.

homepage/gulp/tasks/scripts.js
perjg: It is working ok for me. I'l remove it later if it turns out to be a problem.

PR #1125, opened 2015-07-27
SOC-1062 | Registration - terms of use in JP

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1062


front/scripts/auth/login/SubmitValidator.ts
```diff
@@ -30,10 +27,15 @@ class SubmitValidator {
...
-			input: HTMLInputElement;
-		for (i = 0; i < this.inputFields.length; i++) {
-			input = <HTMLInputElement> this.inputFields[i];
-			if (!input.value) {
+			input: HTMLInputElement,
+			inputFields: NodeList = this.form.querySelectorAll('input[type=text], ' +
+			'input[type=password], input[type=number], input[type=email], input[type=checkbox]:required');
+
+
+		for (i = 0; i < inputFields.length; i++) {
```
bkoval: Maybe array.some would be nicer here? :)

front/scripts/auth/login/SubmitValidator.ts
```diff
@@ -30,10 +27,15 @@ class SubmitValidator {
...
+			input = <HTMLInputElement> inputFields[i];
+
+			if (!input.value || (input.type == 'checkbox' && !input.checked)) {
```
bkoval: Why not triple '==='? :)

bkoval:
Two comments, except that LGTM.

front/scripts/auth/login/SubmitValidator.ts
jsutterfield: Ha, I was just going to comment and ask if there was a compelling reason to do the `Array.slice` approach since I find it harder to read. 

The `for` loop which returns on a certain condition is such a common idiom I find it easier to understand automatically, I have to slow down and think more when it comes to the `Array.slice` approach. But that could just be b/c I don't write enough JS :)

Since we already changed it once, I think this is fine. Just different approaches.

jsutterfield:
LGTM

PR #1128, opened 2015-07-28
rebuild templates on watch task

kenkouot:
This rebuilds templates properly. The browser sync task still fires too early sometime. Any ideas on that? @rogatty @hakubo


rogatty:
@kenkouot browser is updated too early because `server.restart` callback is run immedietaly instead of waiting for server to restart. I've tried to find a quick fix but failed. Ember CLI is coming anyway! :speak_no_evil: 

LGTM

PR #1134, opened 2015-07-29
SOC-1035 New JA translation for login and registration

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-1035


server/facets/auth/register.ts
```diff
@@ -36,7 +36,7 @@ export function get (request: Hapi.Request, reply: any): Hapi.Response {
 			headerText: 'auth:join.sign-up-with-email',
 			heliosRegistrationURL: localSettings.helios.host + '/users',
 			title: 'auth:join.sign-up-with-email',
-			termsOfUseLink: 'http://www.wikia.com/Terms_of_Use',
+			termsOfUseLink: '<a href="' + i18n.t('auth:register.terms-of-use-link-url') + '" target="_blank">' + i18n.t('auth:register.terms-of-use-link-title') + '</a>',
```
bkoval: How about moving it to AuthUtils with a method like "getTermsOfUseMarkup" or sth? :)

front/locales/en/auth.json
```diff
@@ -30,8 +30,9 @@
 		"footer-callout-link": "Log in",
 		"register-action": "Register",
 		"page-title": "Register with Email",
-		"terms-of-use-text": "By creating an account, you agree to",
-		"terms-of-use-link": "Wikia's Terms of use",
+		"terms-of-use-text": "By creating an account, you agree to __termsOfUseLink__",
+		"terms-of-use-link-title": "Wikia's Terms of use",
+		"terms-of-use-link-url": "http://www.wikia.com/Terms_of_Use",
```
kvas-damian: I don't like storing URLs in i18n files. What do you think about moving it to ```...Settings.ts``` file?

server/facets/auth/register.ts
kvas-damian: Can't it be done in hbs template?

kvas-damian:
LGTM

front/locales/en/auth.json
```diff
@@ -30,7 +30,7 @@
 		"footer-callout-link": "Log in",
 		"register-action": "Register",
 		"page-title": "Register with Email",
-		"terms-of-use-text": "By creating an account, you agree to",
+		"terms-of-use-text": "By creating an account, you agree to __termsOfUseLink__",
```
jsutterfield: Is `__termsOfUseLink__` a variable which gets interpolated later? None of the other languages seem to have it. Or is it something else? The leading and trailing double underbars make me think it's some kind of special symbol.

rafalkalinski: It is a variable. It needs to be in the other languages, but as far I don't know them, I'm not sure it the variable should be placed at the end of the sentence - it is up to translators.

PR #1135, opened 2015-07-29
CONCF-744 add arrows for sorting

bognix:
https://wikia-inc.atlassian.net/browse/CONCF-744

/cc @rogatty @hakubo


front/scripts/main/mixins/CuratedContentEditorSortableItemsMixin.ts
```diff
@@ -0,0 +1,30 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.CuratedContentEditorSortableItemsMixin = Em.Mixin.create({
+	sortableItems: Em.computed('model.items', function (): any {
```
hakubo: hmm can't model.items already implement whatever Em.A implements?

front/scripts/main/mixins/CuratedContentEditorSortableItemsMixin.ts
```diff
@@ -0,0 +1,30 @@
...
+		return Em.A().pushObjects(this.get('model.items'));
+	}),
+
+	actions: {
+		moveUp(item: CuratedContentEditorItemModel): void {
```
hakubo: maybe we could abstract a bit of logic here and craete some function like
moveBy
`moveBy(1)` `moveBy(-1)`

front/scripts/main/mixins/CuratedContentEditorSortableItemsMixin.ts
rogatty: @hakubo: we've discussed it and yes, it can be done in `App.CuratedContentEditorModel#sanitize`. This would require full regression testing though, to see if we don't break anything by such change. We've decided to keep it simple but if you disagree maybe we should reconsider the other solution.

hakubo: 'In automated tests we trust' and if we don't we should dedicate more time to it.

Generally I am against of "let's leave it like this for NOW" as now is hard to define.
There are cases where it make sense, but here it looks like we should not add another layer on top of model, not to increase complexity of the sysem.

rogatty: Functionality that `Em.A` provides is needed here and only here, it doesn't give us anything in the model. You propose to add another layer in the model and this adds it only where it's actually used.

Also, there are no tests for this tool yet ;)

bognix: I agree with Igor. We should try to keep the model as clean as it is possible so we don't have to sanitize it too much before saving and when referencing to this model in other places in code we can easily predict what elements are inside it. Adding there more "Ember friendly objects" adds unnecessary level of abstraction in all remaining use cases.

front/scripts/main/components/CuratedContentEditorSectionComponent.ts
```diff
@@ -30,6 +32,7 @@ App.CuratedContentEditorSectionComponent = Em.Component.extend(App.CuratedConten
 
 		done(): void {
 			if (this.get('notEmptyItems')) {
+				this.set('model.items', this.get('sortableItems'));
```
rogatty: Shouldn't we get rid of `Em.A` here? `$.extend` would do the job, right?

front/scripts/main/mixins/CuratedContentEditorSortableItemsMixin.ts
```diff
@@ -0,0 +1,23 @@
...
+		moveBy(offset: number, item: CuratedContentEditorItemModel): void {
+			var items: Em.NativeArray = this.get('sortableItems'),
+				currentItemIndex = items.indexOf(item);
+
+			if (offset > 0 && currentItemIndex > 0) {
+				items.removeAt(currentItemIndex);
+				items.insertAt(currentItemIndex - 1, item);
+			} else if (offset < 0 && currentItemIndex < items.length - 1) {
+				items.removeAt(currentItemIndex);
+				items.insertAt(currentItemIndex + 1, item);
```
rogatty: Where did `offset` go?

front/templates/main/components/curated-content-editor-row.hbs
```diff
@@ -1,8 +1,18 @@
...
+	<div class="item-description">
+		<div class="title">{{model.label}}</div>
+		{{#if model.items}}
+			{{!-- @TODO CONCF-956 add translations to items --}}
+			<div class="subtitle">{{model.items.length}} items</div>
+		{{/if}}
+	</div>
+</div>
+<div class="navigation">
+	<div {{action 'moveBy' 1}}>
```
rogatty: Moving up is using `index - 1`, right?

front/scripts/main/mixins/CuratedContentEditorSortableItemsMixin.ts
```diff
@@ -0,0 +1,23 @@
...
+	sortableItems: Em.computed('model.items', function (): any {
+		return Em.A().pushObjects(this.get('model.items'));
+	}),
+
+	actions: {
+		moveBy(offset: number, item: CuratedContentEditorItemModel): void {
+			var items: Em.NativeArray = this.get('sortableItems'),
+				currentItemIndex = items.indexOf(item);
+
+			if (offset > 0 && currentItemIndex > 0) {
```
hakubo: `items.removeAt(currentItemIndex);` - could be called before if
or how about
```js
var newItemIndex = currentItemIndex;

if (offset > 0 && currentItemIndex > 0) {
  newItemIndex = newItemIndex - offset;
} else if (offset < 0 && currentItemIndex < items.length - 1) {
  newItemIndex = newItemIndex + offset;
}

items.removeAt(currentItemIndex);
items.insertAt(newItemIndex, item);
```

rogatty: Anyway, we should always use `+` operator. Subtracting is already covered by the number being negative.

bognix: @hakubo it won't work.

Imagine you have an item with index 0. You press arrow up. offset is -1, currentItemIndex is 0, newItemIndex is -1. InsertAt uses http://emberjs.com/api/classes/Ember.NativeArray.html#method_replace and passing negative index doesn't work well for us. It adds this item as the second last but without it's properties.

hakubo:
lgtm

rogatty:
LGTM

tomasznapieralski:
LGTM

PR #1138, opened 2015-07-29
Disable source maps

lizlux:
This implementation is buggy and tends to break the `debugger` statement functionality. My hope is that we're moving to ES6 so we'll no longer need TS source maps. (We may want ES6 source maps but we can handle that later.)


kenkouot:
LGTM

rogatty:
@lizlux thanks! We actually wanted to do the same thing but didn't find time.

PR #1140, opened 2015-07-29
decouple article route from application in ember and hapi

kenkouot:


server/facets/showApplication.ts
```diff
@@ -0,0 +1,39 @@
...
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host),
+		wikiVariables = new MW.WikiRequest({wikiDomain: wikiDomain}).getWikiVariables();
+
+	wikiVariables.then((response) => {
+		var result: any = {};
+
+		// TODO: These transforms could be better abstracted, as such, this is a lot like prepareArticleData
+		result.wiki = response.data;
+		result.server = Utils.createServerData(wikiDomain);
+		result.asyncArticle = Utils.shouldAsyncArticle(localSettings, request.headers.host);
```
rogatty: IMO there should be no article references when we're not loading articles.

rogatty:
@kenkouot when working on this please keep in mind that `ApplicationView` is now deprecated and should be a component.

server/facets/showApplication.ts
```diff
@@ -0,0 +1,39 @@
...
+		result.queryParams = Utils.parseQueryParams(request.query, []);
+		console.log(Tracking);
```
rogatty: This is still work in progress, right? ;)

server/views/_layouts/ember-main.hbs
```diff
@@ -81,7 +81,16 @@
         -->
 		<div id="ember-container">
 			<div id="article-preload">
```
lizlux: probably want to change this id since the page might not include an article.

server/views/_layouts/ember-main.hbs
```diff
@@ -58,8 +63,9 @@
 				M.prop('mainPageData', {{{json mainPageData}}});
 			{{/if}}
 			{{#if error}}
-			M.provide('article.error', {{{error}}});
+				M.provide('article.error', {{{error}}});
```
lizlux: Is this line still pertaining specifically to articles? If not, maybe it should it be namespaced differently. If so, maybe it should go inside the `{{#if article}}` block.

lizlux:
Don't need to import `localSettings` in `Article.ts` anymore.

server/lib/Utils.ts
```diff
@@ -211,3 +213,23 @@ export function parseQueryParams (obj: any, allowedKeys: string[]): any {
...
 }
+
+/**
+ * Create server data
+ *
+ * @returns ServerData
+ */
+export function createServerData(wikiDomain: string = ''): ServerData {
+	// if no environment, pass dev
+	var env = localSettings.environment || 4;
```
rogatty: `Environment.Dev` equals to 4 and looks better, right?

front/scripts/main/router.ts
```diff
@@ -77,6 +77,8 @@ App.Router.map(function () {
 			path: '/*url'
 		});
 	}
+
+	this.route('raw');
```
rogatty: Is this for testing purposes only? Do you plan to merge it?

lizlux:
@kenkouot, just want to make sure you've considered @mklucsarits's comment on the ticket: "As part of this ticket, can the `.wiki-container` section tag be moved out of `application.hbs`? It doesn't make sense semantically for non-wiki pages, and its styling adds unwanted spacing."  

Looks like the container is still in there, adding `2rem` of padding at the top. That padding is overridden in the `editor-preview` pages, so maybe it's something we can move to the article wrapper template?

server/facets/showApplication.ts
```diff
@@ -0,0 +1,36 @@
...
+import localSettings = require('../../config/localSettings');
+
+function showApplication(request: Hapi.Request, reply: Hapi.Response): void {
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host),
+		wikiVariables = new MW.WikiRequest({wikiDomain: wikiDomain}).getWikiVariables();
+
+	wikiVariables.then((response) => {
+		var result: any = {};
+
+		// TODO: These transforms could be better abstracted, as such, this is a lot like prepareArticleData
```
lizlux: Can we create tickets to track these TODO's and reference the here?

server/routes.ts
```diff
@@ -226,9 +226,16 @@ articlePagePaths.forEach((path) => {
 	});
 });
 
+authenticatedRoutes.push({
```
lizlux: Maybe we should include this route in a comment as an example of how to create a "raw" page as opposed to setting up the actual route.

lizlux: Or set up the discussions route but put it behind a flag.

front/templates/main/components/main-page.hbs
```diff
@@ -1,21 +1,23 @@
...
-{{#if curatedContent}}
-	{{curated-content model=curatedContent openCuratedContentItem='openCuratedContentItem'}}
-{{/if}}
-{{#if trendingArticles}}
-	{{trending-articles model=trendingArticles}}
-{{/if}}
-{{#if trendingVideos}}
-	{{trending-videos model=trendingVideos openLightbox='openLightbox'}}
-{{/if}}
+<div class='wiki-container'>
```
mklucsarits: Shouldn't these HTML attributes use double quotes?

mklucsarits:
LGTM :ship:

PR #1145, opened 2015-07-30
HG-816 checking if window is loaded before binding event

lizlux:
See https://github.com/Wikia/mercury/pull/1137 for a description of the problem. 

This will fix the spikiness of the graphs as well as the apparent decrease in performance. After loading scripts asynchronously, it became possible for the code that binds the onload event to run after the event already fired. This will check if onload has fired before binding the event, and will send page performance stats if it has.

### DIFF
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -91,7 +91,12 @@ App.initializer({
 			return;
 		}
 
-		$(window).load(() => M.sendPagePerformance());
+		// Send page performance stats after window is loaded
+		if (document.readyState === 'complete') {
+			M.sendPagePerformance()
+		} else {
+			$(window).load(() => M.sendPagePerformance());
+		}
 
 		EmPerfSender.initialize({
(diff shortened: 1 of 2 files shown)


front/scripts/main/app.ts
```diff
@@ -91,7 +91,12 @@ App.initializer({
 			return;
 		}
 
-		$(window).load(() => M.sendPagePerformance());
+		// Send page performance stats after window is loaded
+		if (document.readyState === 'complete') {
+			M.sendPagePerformance()
+		} else {
+			$(window).load(() => M.sendPagePerformance());
```
rogatty: Do I understand correctly that this piece of code can be run after `window.load` event already happened? If so it would be nice to have it in a comment.

lizlux: Added a comment

rogatty: Thanks!

rogatty:
LGTM

PR #1149, opened 2015-07-31
CONCF-745 Add modal dialog with help text to Curated Content Tool

rogatty:
https://wikia-inc.atlassian.net/browse/CONCF-745

Work in progress, needs styling.


front/locales/en/main.json
```diff
@@ -87,6 +87,9 @@
 		"curated-content-error-category-not-found": "Category not found.",
 		"curated-content-error-section-not-found": "Section not found.", 
 		"curated-content-load-more-items": "Load more",
-		"curated-content-error-load-more-items": "Sorry, we couldn't load more items."
+		"curated-content-error-load-more-items": "Sorry, we couldn't load more items.",
+		"curated-content-editor-featured-help": "This is easy!",
+		"curated-content-editor-curated-help": "This is easy!",
+		"curated-content-editor-optional-help": "This is easy!"
```
tomasznapieralski: `"This is easy!"` is temporary? If not, maybe it should be under one key instead of multiplying value?

rogatty: I added it as a placeholder, I don't think we should upload it to crowdin.
@Wikia/x-wing?

bognix: Agreed. We should feed it with content from mocks

Warkot: But if we merge this, someone else will probably upload it.

front/scripts/main/components/CuratedContentEditorBlockComponent.ts
```diff
@@ -16,6 +16,10 @@ App.CuratedContentEditorBlockComponent = Em.Component.extend(App.CuratedContentE
 			this.sendAction('editItem', item, this.get('block'));
 		},
 
+		showHelp(): void {
+			this.sendAction('showHelp', i18n.t(this.get('helpText')));
```
bognix: which is better passing this.get('helpText') to component and inside it calling i18n helper or doing it this way?
In https://github.com/Wikia/mercury/pull/1150/files I used component helper, we should unify approach one way or the other

front/locales/en/main.json
rogatty: @bognix: we don't have this text on mocks. Would it be better to remove translation strings for now and just put some placeholders to not pollute crowdin?

front/scripts/main/components/CuratedContentEditorComponent.ts
```diff
@@ -13,8 +20,11 @@ App.CuratedContentEditorComponent = Em.Component.extend({
 			this.sendAction('editBlockItem', item, block);
 		},
 
-		addSection(): void {
-			this.sendAction('addSection');
+		showHelp(helpText: string): void {
+			this.setProperties({
```
bognix: why not doing this "spinner mixin way" so have two methods: show and hide inside mixin and you would call those methods instead of setting props on component. This way you wouldn't need to pass showHelp action inside components

front/scripts/main/components/CuratedContentEditorBlockComponent.ts
rogatty: If we move the i18n logic to `ModalDialogComponent` then it will require that all strings passed to it are translation keys, not cool. Now I think that we should use block form of components and yield provided content in modal instead of passing it this way. See http://guides.emberjs.com/v1.13.0/components/wrapping-content-in-a-component/ for reference.

front/locales/en/main.json
vforge: `<placeholder />` ;)

bognix:
LGTM

tomasznapieralski:
LGTM

PR #1150, opened 2015-07-31
CONCF-741 Add photo upload to the Curated Content Tool

bognix:
https://wikia-inc.atlassian.net/browse/CONCF-741

/cc @inez @mari-mcmurtrie @hakubo @rogatty 

related PR: https://github.com/Wikia/app/pull/7978


front/scripts/main/components/CuratedContentEditorItemComponent.ts
```diff
@@ -89,6 +89,30 @@ App.CuratedContentEditorItemComponent = Em.Component.extend(App.CuratedContentEd
 			if (confirm('Are you sure about removing this item?')) {
 				this.sendAction('deleteItem');
 			}
+		},
+
+		fileUpload(files: any[]): void {
+			this.showLoader();
+			App.AddPhotoModel.load('main', null, files[0])
```
inez: Perhaps null would be better than 'main'.

front/scripts/main/components/FileInputComponent.ts
```diff
@@ -0,0 +1,8 @@
+App.FileInputComponent = Ember.Component.extend(Ember.Evented, {
+	change: function(e: Event) {
```
tomasznapieralski: return type

front/styles/main/component/_curated-content-editor.scss
```diff
@@ -172,6 +172,19 @@ $row-height: 5rem;
...
+.curated-content-editor-item {
+	.file-upload-input {
+		display: none;
+	}
+
+	.file-upload-label {
+		position: absolute;
+		top: 14px;
+		width: 100vw;
+		height: 90vw;
```
tomasznapieralski: alphabetical order

front/scripts/main/components/CuratedContentEditorItemComponent.ts
```diff
@@ -89,6 +89,32 @@ App.CuratedContentEditorItemComponent = Em.Component.extend(App.CuratedContentEd
...
+		},
+
+		fileUpload(files: any[]): void {
+			this.showLoader();
+			App.AddPhotoModel.load(null, null, files[0])
+				.then(function (photoModel: typeof App.AddPhotoModel) {
+					return App.AddPhotoModel.upload(photoModel);
+				})
+				.then((data: any) => {
+					if (data && data.url) {
```
hakubo: What should happen when this is false?

front/scripts/main/components/CuratedContentEditorItemComponent.ts
```diff
@@ -89,6 +89,32 @@ App.CuratedContentEditorItemComponent = Em.Component.extend(App.CuratedContentEd
...
+						this.setProperties({
+							'model.image_url': this.generateThumbUrl(data.url),
+							'imageErrorMessage': null,
+							// article_id comes from MW because in MW files are like any other articles
+							// so there is no such thing as image_id from MW perspective.
+							'model.image_id': data.article_id
```
rogatty: I would change the order, so `model.image_url` and `model.image_id` are next to each other.

front/scripts/main/components/CuratedContentEditorItemComponent.ts
rogatty: IMO this should be checked in `AddPhotoModel` and reject if false.

front/scripts/main/components/CuratedContentEditorItemComponent.ts
```diff
@@ -165,9 +191,8 @@ App.CuratedContentEditorItemComponent = Em.Component.extend(App.CuratedContentEd
 				}
 			})
 			.catch((err: any): void => {
-				//@TODO CONCF-956 add translations
 				Em.Logger.error(err);
-				this.set('imageErrorMessage', 'Oops! An API Error occured.');
+				this.set('imageErrorMessage', 'app.curated-content-image-upload-error');
```
rogatty: This isn't about uploading but about getting default image. This message seems unappropriate here.

bognix: mistake. It shouldn't be changed at all.

front/scripts/main/models/AddPhotoModel.ts
```diff
@@ -101,7 +101,7 @@ App.AddPhotoModel.reopenClass(App.EditMixin, {
 			this.temporaryUpload(model.photoData)
 				.then((addmediatemporary: any): void => {
 					if (addmediatemporary.tempName === undefined) { //we already have the file. No need to upload.
-						return resolve({'title': addmediatemporary.title});
+						return resolve(addmediatemporary);
```
rogatty: This function has `void` return type, so why we return here?

front/templates/main/components/file-input.hbs
```diff
@@ -0,0 +1,2 @@
+<label for="fileUpload" class="file-upload-label"></label>
+<input id="fileUpload" class="file-upload-input" type="file" accept="image/*"/>
```
rogatty: This id seems too general. How about passing it as a component attribute?

bognix: I was thinking about it. On the other hand maybe we should "parametrize" it when it's needs to be parametrized?

front/scripts/main/components/CuratedContentEditorItemComponent.ts
bognix: @rogatty - I agree, but on the other hand I don't know the logic for addPhoto functionality. What if they don't need url?

rogatty: `AddPhoto` doesn't seem to use `url` for anything. Still, it's the same API and it should always return consistent data, right? Does API not returning the image URL mean that something went wrong? If yes, then we should reject the promise. Everyone using `App.AddPhotoModel.upload` can handle rejection in `.catch()` anyway they want, too.

@inez what do you think?

bognix: Regarding handling rejection - yes it can be handled inside catch but not some specific rejection but all of them. I meant handling "missing URL"  situation like a special case. URL was returned by temporary upload but not by permanent upload. Apparently it wasn't needed.

front/scripts/main/components/FileInputComponent.ts
```diff
@@ -0,0 +1,8 @@
+App.FileInputComponent = Ember.Component.extend(Ember.Evented, {
+	change(e: Event): void {
```
hakubo: e - xception?
e - viction?
e - commerce?
e - rror?
event! :)

front/scripts/main/components/CuratedContentEditorItemComponent.ts
```diff
@@ -89,6 +89,35 @@ App.CuratedContentEditorItemComponent = Em.Component.extend(App.CuratedContentEd
 			if (confirm('Are you sure about removing this item?')) {
 				this.sendAction('deleteItem');
 			}
+		},
+
+		fileUpload(files: any[]): void {
+			this.showLoader();
+			App.AddPhotoModel.load(files[0])
+				.then(function (photoModel: typeof App.AddPhotoModel) {
+					return App.AddPhotoModel.upload(photoModel);
```
rogatty: Would `.then(App.AddPhotoModel.upload)` work?

rogatty:
LGTM

vforge:
LGTM

hakubo:
LGTM

PR #1157, opened 2015-08-03
HG-810 - only build old way in prod

katanka:
This should fix the error in `build-views` when a server template is changed.


lizlux:
:+1:

gulp/tasks/build-views.js
```diff
@@ -44,15 +44,17 @@ gulp.task('build-views', ['scripts-front', 'copy-ts-source', 'vendor', 'build-ve
 
 		// TODO: Leave this in for now to run the normal template based assets pipeline for the duration
```
lizlux: I would be good to update/clarify this comment since it doesn't refer to which test and as far as I know our tests are over.

kenkouot:
LGTM

PR #1159, opened 2015-08-04
Revert "decouple article route from application in ember and hapi"

kenkouot:
Reverts Wikia/mercury#1140


kenkouot:
cdnBaseUrl bug

PR #1164, opened 2015-08-04
Hg 829 Adding comments and removing unused gulp task

lizlux:
I started this as part of https://wikia-inc.atlassian.net/browse/HG-829 but now it's just adding comments and code cleanup. 

@katanka @kenkouot


katanka:
LGTM

gulp/tasks/build-views.js
```diff
@@ -57,8 +61,9 @@ gulp.task('build-views', ['scripts-front', 'copy-ts-source', 'vendor', 'build-ve
 		)),
 
 		gulpif('**/_layouts/*.hbs', gulp.dest('www/server/views/_layouts')),
+
+		// for ember-main.hbs, find the file paths in the template source and replace them for prod CDN
 		gulpif(environment.isProduction, piper(
```
rogatty: How does it know to do this only for `ember-main.hbs`?

lizlux: https://github.com/Wikia/mercury/blob/dev/gulp/options/prod.js#L53

This file could use some more cleanup for sure, but this PR is a start.

rogatty: Oh, right, thanks! That's the info that we've missed when quickly investigating with @bkoval.

PR #1165, opened 2015-08-04
INT-149: Refactor article contribution buttons into separate component

mari-mcmurtrie:
so that it is more modular. ArticleContributionComponent.ts and article-contrinution.hbs are refactored files. ArticleContributionComponent is referred from ArticleWrapperComponent for section 0, and from ArticleContentComponent for section 1+. This is done that way because section 0 is outside of ArticleContentComponent. Once buttons are tapped, the actions are propagated into ArticleController essentially, then the rest of flow/logic is not changed.

@lizlux please code review. Thank you!


front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -62,6 +73,32 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, {
...
+	setupContributionButtons: function (): void {
+		var headers = this.get('headers'),
+			$sectionHeader: any = null,
+			contributionComponent: any = null;
+
+		headers.forEach((header: ArticleSectionHeader): void => {
+			contributionComponent = this.createArticleContributionComponent(header.section);
+			$sectionHeader = this.$(header.element);
+			$sectionHeader.prepend(contributionComponent).addClass('short-header');
+			contributionComponent.wrap('<div class="icon-wrapper"></div>');
```
lizlux: Does it work if you put this line before the previous one? That way you're manipulating the DOM only once per iteration instead of twice.

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -62,6 +73,32 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, {
...
+
+	createArticleContributionComponent: function(section: number): JQuery {
+		var article = this.get('content'),
+			title = this.get('cleanTitle'),
+			component = this.createChildView(App.ArticleContributionComponent.create({
+				article: article,
+				section: section,
+				title: title,
+				edit: 'edit',
+				addPhoto: 'addPhoto'
```
lizlux: Why do you need these two properties?

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -76,12 +113,17 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, {
 						element: elem,
 						level: elem.tagName,
 						name: elem.textContent,
-						id: elem.id
+						id: elem.id,
+						section: $(elem).attr('section')
```
lizlux: If you wanted to reduce the overhead of creating the jQuery object you could do `elem.getAttribute('section')`.

lizlux:
The alignment looks off on the buttons. 
screen shot 2015-08-04 at 1 25 50 pm

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+			this.setupContributionButtons();
+		});
+	}).on('init'),
+
+	setupContributionButtons: function (): void {
+		var $photoIcon = this.$('.upload-photo'),
+		    $editIcon = this.$('.edit-section');
+
+		if (this.section === 0) { //need different html for different styling for section 0
+			$photoIcon.addClass('zero');
```
lizlux: It looks like adding this extra class is no longer necessary since all it does is override the height with the same value https://github.com/Wikia/mercury/blob/dev/front/styles/main/module/edit/_edit.scss#L23

lizlux:
Left a few comments, but overall this looks way better than it used to. :+1:

front/styles/main/module/edit/_edit-common.scss
```diff
@@ -47,3 +47,7 @@
 	padding-right: 5.2rem; /* edit/photo icons width (2.2 * 2) + margin (0.8) in between */
 	position: relative;
 }
+
+input.file-input {
+	display: none;
```
rogatty: This seems too general, especially for `_edit-common.scss` file. Please make it more scoped like in https://github.com/Wikia/mercury/blob/dev/front/styles/main/component/_curated-content-editor.scss#L190.

rogatty: If you believe that every file input should have this style then you can create stylesheet file for `FileInputComponent` and put it there.

front/templates/main/components/article-contribution.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="edit-section">
+	<span {{action 'edit' model.title}}>
+		{{svg 'pencil' role='img' class='icon pencil'}}
+	</span>
+</div>
+<div class="upload-photo" {{action 'select'}}>
+	{{svg 'camera' role='img' class='icon camera'}}
+</div>
+<input class="file-input" type="file" accept="image/*" {{action 'addPhoto' on='change'}}/>
```
rogatty: Can you use `FileInputComponent` instead? Here you have example:
https://github.com/Wikia/mercury/blob/dev/front/templates/main/components/curated-content-editor-item.hbs#L15
https://github.com/Wikia/mercury/blob/dev/front/scripts/main/components/CuratedContentEditorItemComponent.ts#L94

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,60 @@
+/// <reference path='../../../../typings/ember/ember.d.ts' />
+/// <reference path='../app.ts' />
+
+App.ArticleContributionComponent = Em.Component.extend({
+	tagName: 'div',
```
rogatty: This is the default, no need to add it.

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+	classNames: ['contribution-container'],
+	templateName: 'components/article-contribution',
+	article: null,
+	section: null,
+	title: null,
+
+	articleContent: Em.computed('article', function (): any {
+		return this.get('article');
```
rogatty: What is `articleContent` for? Why can't you just use `article`?

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+
+		if (this.section === 0) { //need different html for different styling for section 0
+			$photoIcon.addClass('zero');
+			$editIcon.addClass('zero');
+		}
+	},
+
+	actions: {
+		edit: function (title: string): void {
+			//App.VisibilityStateManager.reset(); check if I do not need it anymore
```
rogatty: If it has to be done it shouldn't be done here but on `EditRoute`.

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+			M.track({
+				action: M.trackActions.click,
+				category: 'sectioneditor',
+				label: 'edit',
+				value: this.section
```
rogatty: `this.get('section')`

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,60 @@
...
+			});
+			this.sendAction('edit', this.title, this.section);
```
rogatty: You pass `title` to this action so maybe you want to use it?

front/scripts/main/components/ArticleContentComponent.ts
mari-mcmurtrie: Ah I first thought to do that pure javaScript then I was like, "maybe in CR someone will point out I should use jQuery" so I changed. Funny.

lizlux: Ha, yeah, it really doesn't make a big difference, just noticed you're not using it for any other `elem` properties.

front/scripts/main/components/ArticleContentComponent.ts
mari-mcmurtrie: This maps 'action' to 'actionName'. sendAction() 's first parameter is 'action', which then execute a function with its 'actionName'. This is the mapping for action: actionName passed into component when creating component. (see: https://github.com/emberjs/ember.js/blob/v1.13.6/packages/ember-views/lib/views/component.js#L212)

front/scripts/main/components/ArticleContentComponent.ts
mari-mcmurtrie: I cannot use wrap before component is attached to DOM :(

front/templates/main/components/article-contribution.hbs
mari-mcmurtrie: @rogatty file-input.hbs, input tag has id="fileUpload" so I do not want to use this since I need multiple of file inputs in a page. How do you think?

front/scripts/main/components/ArticleContributionComponent.ts
mari-mcmurtrie: I removed. I think this is left-over when I was trying out something.

front/scripts/main/components/ArticleContributionComponent.ts
mari-mcmurtrie: Turned out I dont have to pass it. I already have title as property.

mari-mcmurtrie:
@lizlux @rogatty please CR again when you have time. Thank you.

front/templates/main/components/article-contribution.hbs
rogatty: Id is hardcoded because we didn't need it to be dynamic. You can always modify `FileInputComponent` to your needs. Id has to be there to open file dialog on label click.

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,41 @@
+/// <reference path='../../../../typings/ember/ember.d.ts' />
+/// <reference path='../app.ts' />
+
+App.ArticleContributionComponent = Em.Component.extend({
+	classNames: ['contribution-container'],
+	templateName: 'components/article-contribution',
+	article: null,
```
rogatty: It's not used, right?

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,41 @@
...
+	section: null,
+	title: null,
+
+	// articleContent: Em.computed('article', function (): any {
+	// 	return this.get('article');
+	// }),
```
rogatty: Leftover?

front/templates/main/components/article-wrapper.hbs
```diff
@@ -27,15 +27,13 @@
 			{{/if}}
 			{{#if contributionFeatureEnabled}}
 				<div class="contribution-container">
```
lizlux: This div is redundant, since there's a div inside it with the same class.

front/scripts/main/components/ArticleContentComponent.ts
lizlux: Okay thanks for checking.

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -58,10 +69,36 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, {
...
+			$sectionHeader.prepend(contributionComponent).addClass('short-header');
+			contributionComponent.wrap('<div class="icon-wrapper"></div>');
+		});
+	},
+
+	createArticleContributionComponent: function(section: number): JQuery {
+		var article = this.get('content'),
+			title = this.get('cleanTitle'),
+			component = this.createChildView(App.ArticleContributionComponent.create({
+				article: article,
```
vforge: This (and next 3 lines) can be simplified:
```
component = this.createChildView(App.ArticleContributionComponent.create({
article,
section,
title,
...
```

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -58,10 +69,36 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, {
...
 	 * @param {string} content HTML containing whole article
 	 */
-	hackIntoEmberRendering(content: string) {
+	hackIntoEmberRendering(content: string): void {
 		this.$().html(content);
 	},
 
+	setupContributionButtons: function (): void {
+		var headers = this.get('headers'),
+			$sectionHeader: any = null,
```
vforge: This and next line can have `JQuery` type instead of `any` - we'll know instantly what those are for :)

Also: As you're prepending variable name with `$` for `sectionHeader`, can you please add it also to `contributionComponent` name?

front/scripts/main/components/ArticleContentComponent.ts
mari-mcmurtrie: I know but it does not apply last 2 parameters that is why I did not do it for sake of consistency.

front/scripts/main/components/ArticleContentComponent.ts
mari-mcmurtrie: Thanks, fixed.

mari-mcmurtrie:
@rogatty I tried to use FileInputComponent and changed code to use it (See: 7551d21ebcc67234d8ca78aa248bbda7e836654b), but it did not work. What was not working is that, "Change" event was not detected for icons section 1 and later. For section0, it worked. I think the reason why it did not work is because my code is creating button component dynamically (createChildView()), and the button component has fileInputComponent, which is created dynamically. Creating views dynamically screws up something that even though I add "id" to the input element, it never detect "change" event. 
I did not want to spend too much time on it so I reverted the change and go back to my original implementation, which works very solid.

mari-mcmurtrie:
@lizlux @rogatty please check this again. Thank you so much!

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -76,12 +112,17 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, {
...
+						id: elem.id,
+						section: elem.getAttribute('section')
 					};
 				}
 			}).toArray();
 
-		this.sendAction('updateHeaders', headers);
+		this.set('headers', headers);
+
+		if (this.get('contributionFeatureEnabled')) {
```
lizlux: An optimization for this might be to set up an observer for `headers` so that when it changes, we set up the contribution buttons. As it stands, this is a little spaghetti code-ish. I don't think it's a blocker but would be good to keep in mind.

front/styles/main/module/edit/_edit-common.scss
```diff
@@ -47,3 +48,12 @@
...
 }
+
+.top-contribution-container {
+	height: 2.4rem;
+
+	.contribution-container {
+		float: right;
+		width: 5.2rem;
+	}
+}
```
lizlux: should be a new line here.

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,39 @@
+/// <reference path='../../../../typings/ember/ember.d.ts' />
+/// <reference path='../app.ts' />
+
+App.ArticleContributionComponent = Em.Component.extend({
+	classNames: ['contribution-container'],
+	templateName: 'components/article-contribution',
+	section: null,
+	title: null,
+	fileInputId: Em.computed('section', function(): string {
```
lizlux: looks like this isn't being used anymore.

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -0,0 +1,39 @@
...
+		},
+
+		select: function (): void {
+			M.track({
+				action: M.trackActions.click,
+				category: 'sectioneditor',
+				label: 'addPhoto',
+				value: this.get('section')
+			});
+			this.$('.file-upload-input').click();
```
lizlux: I just ran a test by calling this from my browser console: `$('.file-upload-input').first().click()`. It doesn't actually activate the file input. What does activate it is tapping on it when you tap on the icon, which is overlaid on top of the file input. Here's a stackoverflow about it: http://stackoverflow.com/questions/793014/jquery-trigger-file-input

front/templates/main/components/file-input.hbs
```diff
@@ -1,2 +1,2 @@
 <label for="fileUpload" class="file-upload-label"></label>
```
bognix: please fix label too when changing input's id. Label's "for" should be the same as ID of an input.

front/templates/main/components/article-contribution.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="edit-section">
+	<span {{action 'edit'}}>
+		{{svg 'pencil' role='img' class='icon pencil'}}
+	</span>
+</div>
+<div class="upload-photo">
+	<span {{action 'select'}}>
+		{{svg 'camera' role='img' class='icon camera'}}
+	</span>
```
bognix: why didn't you use file input helper? If you would like to use it you need to set up the same "ID" attr for input as "FOR" attr for label. It should look like this:
```
<div class="upload-photo">
		{{file-input fileUpload='add photo' id="<your id>" reset=resetFileInput}}
</div>
```
and as label's background you can use camera's svg

mari-mcmurtrie: I tried and did not work (my comment is above) but, I did not know that label's "for" attribute has to have the same value as input's "id". Let me try that again.

front/scripts/main/components/ArticleContributionComponent.ts
mari-mcmurtrie: Yeah this is left over from using FileInputComponent. Forgot to remove it.

front/templates/main/components/article-contribution.hbs
mari-mcmurtrie: Ok, I tried and still not work. I will not use file input helper.

front/templates/main/components/file-input.hbs
mari-mcmurtrie: thanks, changed.

front/scripts/main/components/ArticleContributionComponent.ts
mari-mcmurtrie: I removed the line invoking click since it does not do anything.

mari-mcmurtrie:
@lizlux will you please check this again when you have time? Thanks!

lizlux:
Would be nice to have the file input working, but it can be done at a later date so this can get merged. 

LGTM.

tomasznapieralski:
@mari-mcmurtrie i revered your changes, because table of content on articles was not present.
https://github.com/Wikia/mercury/pull/1209

PR #1167, opened 2015-08-05
Lazy load jquery.cropper when entering to curated content editor

hakubo:


front/scripts/main/app.ts
```diff
@@ -22,6 +22,10 @@ declare var EmPerfSender: any;
 declare var optimizely: any;
 declare var FastClick: any;
 
+$.ajaxSetup({
```
rogatty: Can we put it in an initializer? I don't like adding stuff here like that, it will become a mess.

Warkot:
:+1:

front/scripts/main/routes/CuratedContentEditorIndexRoute.ts
```diff
@@ -3,10 +3,66 @@
...
+}
+
 App.CuratedContentEditorIndexRoute = Em.Route.extend({
 	renderTemplate(): void {
 		this.render('curated-content-editor', {
 			model: this.modelFor('curatedContentEditor')
 		});
 	},
+
+	loadingCropper: false,
```
rogatty: This is false at the beginning and then set to true, period. Does it mean it will be loading forever? ;)

hakubo: do you think setting it back to false, will help us here?

rogatty: No, but maybe we should use another name? For example `cropperLoadingInitialized` or something. Let's not muddle people's heads up with unclear naming.

hakubo: K, thanks! that make sense

rogatty:
`front/scripts/main/routes/CuratedContentEditorIndexRoute.ts(30,14): 2339 Property 'define' does not exist on type 'Window'.`

@hakubo so how do you like TS? ;)

rogatty:
LGTM as a temporary solution.

hakubo:
@tomasznapieralski when you're done with component please move this initialisation to wherever makes the most sense

rogatty:
This is now a part of https://github.com/Wikia/mercury/compare/CONCF-743

PR #1170, opened 2015-08-05
HG-667 - disable logging for ember performance sender

katanka:


lizlux:
Note that this will disable all logging, including logging on dev. IMO, this is okay since I don't see those logs as being very useful. Just pointing it out b/c the ticket specified production.

rogatty:
Please disable it only for production, you can just use `M.prop('environment')` for that.

Logs aren't very useful in MOST situations but surely can be helpful when trying to improve performance.

lizlux:
^ :+1:

lizlux:
LGTM!

PR #1171, opened 2015-08-06
INT-139: Company info page for Japan homepage

perjg:


inez:
googlemaps.png - "28 minutes drive home" :) well! it's so inaccurate!

perjg:
@inez haha... good catch. In my case it is pretty accurate, and it also means that google knows where I live!

perjg:
Will merge on Friday if there are no protests.

PR #1175, opened 2015-08-06
Add assert with type to all jQuery AJAX calls

vforge:
Add assert for all jQuery AJAX call (as described in `jquery.d.ts`) do webStorm's TSLINT is not showing errors.

/cc @rogatty @kenkouot @hakubo


rogatty:
LGTM

PR #1176, opened 2015-08-06
HG-838 Don't display deprecations when running unit tests

rogatty:
https://wikia-inc.atlassian.net/browse/HG-838

/cc @hakubo @lizlux @mari-mcmurtrie

### DIFF
--- a/test/helpers/startEmberApp.js
+++ b/test/helpers/startEmberApp.js
@@ -32,3 +32,6 @@ App.initializer({
 setResolver(Ember.DefaultResolver.create({
 	namespace: App
 }));
+
+// Don't clutter test log with deprecations
+Ember.deprecate = function(){};
--- a/test/helpers/startEmberApp.js
+++ b/test/helpers/startEmberApp.js
@@ -29,9 +29,9 @@ App.initializer({
 	}
 });
 
-setResolver(Ember.DefaultResolver.create({
+setResolver(Em.DefaultResolver.create({
 	namespace: App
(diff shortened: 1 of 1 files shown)


test/helpers/startEmberApp.js
```diff
@@ -32,3 +32,6 @@ App.initializer({
 setResolver(Ember.DefaultResolver.create({
 	namespace: App
 }));
+
+// Don't clutter test log with deprecations
+Ember.deprecate = function(){};
```
hakubo: Em.K :)

hakubo:
LGTM

lizlux:
LGTM. Could probably use a more descriptive comment about what `Em.deprecate` and `Em.K` are (I had to look both of them up).

rogatty:
@lizlux well... Should we add explanatory comments for all Ember mechanics or only for some subset? ;) IMO that's how it goes when you work with a framework - you need to check documentation for classes or functions you haven't seen before. I don't think that we should copy and paste docs into comments, only write about things that are not obvious either from the context or Ember documentation.

lizlux:
You have a good point, and I'm fine with leaving it as is, but this could be more descriptive and more helpful to someone looking at it quickly: "Set deprecation warning method to Ember's version of noop to declutter test logs".

rogatty:
@lizlux: clarified in the comment, thanks for suggestion.

lizlux:
:+1:

PR #1185, opened 2015-08-07
Initialize featured and optional sections with correct data

bognix:
/cc @rogatty @vforge

### DIFF
--- a/front/scripts/main/models/CuratedContentEditorModel.ts
+++ b/front/scripts/main/models/CuratedContentEditorModel.ts
@@ -94,8 +94,9 @@ App.CuratedContentEditorModel.reopenClass({
 	 */
 	sanitize(rawData: any): CuratedContentEditorModel {
 		/**
-		 * Label has to be initialized with string value because it can't be null.
-		 * If it's null it will fail rendering main page.
+		 * Label inside "optional" has to be initialized with empty string value.
+		 * Code inside CuratedContentController:getSections (MW) decides based on this label
+		 * if it's optional or not. If it's null it will fail rendering main page.
 		 */
 		var featured: any = {
 				items: <any>[],
--- a/front/scripts/main/models/CuratedContentEditorModel.ts
+++ b/front/scripts/main/models/CuratedContentEditorModel.ts
@@ -93,14 +93,20 @@ App.CuratedContentEditorModel.reopenClass({
 	 * @returns {Object}
 	 */
 	sanitize(rawData: any): CuratedContentEditorModel {
+		/**
+		 * Label has to be initialized with string value because it can't be null.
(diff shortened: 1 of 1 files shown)


front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -93,14 +93,20 @@ App.CuratedContentEditorModel.reopenClass({
...
+				items: <any>[],
+				featured: 'true',
 			},
 			curated = {
 				items: <any>[]
 			},
 			optional = {
-				items: <any>[]
+				items: <any>[],
+				label: ''
```
vforge: Why do we need this? Optional section shouldn't have any label as far as I know.

rogatty: In multiple places we use empty string to identify optional section, for example in https://github.com/Wikia/app/blob/dev/extensions/wikia/CuratedContent/CuratedContentController.class.php#L347

```
// there also might be some categories without SECTION, lets find them as well (optional section)
$optionalSections = $this->getSectionItems( $content, '' );
```

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -93,14 +93,20 @@ App.CuratedContentEditorModel.reopenClass({
 	 * @returns {Object}
 	 */
 	sanitize(rawData: any): CuratedContentEditorModel {
+		/**
+		 * Label has to be initialized with string value because it can't be null.
```
rogatty: Seems too general, this is true only for "optional" block.

rogatty:
```
front/scripts/main/models/CuratedContentEditorModel.ts(115,6): 2322 Type 'CuratedContentEditorRawSectionInterface' is not assignable to type '{ items: any; featured: string; }'.
  Property 'featured' is optional in type 'CuratedContentEditorRawSectionInterface' but required in type '{ items: any; featured: string; }'.
```

front/scripts/main/models/CuratedContentEditorModel.ts
bognix: https://github.com/Wikia/app/blob/dev/extensions/wikia/CuratedContent/CuratedContentController.class.php#L357 and this particular line was the problematic one.

front/scripts/main/models/CuratedContentEditorModel.ts
bognix: true. At the beginning I thought it applies also to featured. I'll fix the comment.

vforge:
LGTM

hakubo:
LGTM

rogatty:
LGTM

PR #1188, opened 2015-08-07
HG-819 - make blockquote styles conform to style guide

katanka:
There was blockquote styling in mercury that was conflicting with the style guide, causing blockquotes to be borked. This removes the custom stylings.


lizlux:
LGTM!

PR #1189, opened 2015-08-09
HG-825 Fix itemController deprecation

rogatty:
https://wikia-inc.atlassian.net/browse/HG-825

It's a simple fix. There is no need to normalize category title as it's coming from the API without underscores already.

/cc @lizlux @hakubo


hakubo:
LGTM

lizlux:
LGTM

PR #1146, opened 2015-07-30
HG-788 - replace inline SVG sprite with requested asset

katanka:
Instead of inlining the icons SVG, we will now request it asynchronously to increase performance on mobile devices.

PR #1192, opened 2015-08-10
Revert "HG-788 - replace inline SVG sprite with requested asset"

bkoval:
Reverts Wikia/mercury#1146


bkoval:
Tested on sandbox, seems like this revert doesn't break anything else.

rogatty:
LGTM

bkoval:
This change is causing auth pages to display error message on the page as it cannot find common.svg:
https://wikia-inc.atlassian.net/browse/SOC-1154

PR #1198, opened 2015-08-11
Mobile tweaks for nav with search and login window for Japan homepage

perjg:


perjg:
Will merge on Friday if there are no protests.

PR #1199, opened 2015-08-11
Changing default port to work better on devboxes

bkoval:
Devboxes are looking for mercury on port 7000. Let's add it as a default so when we start mercury on devbox it doesn't need any additional localSettings to be able to work out of the box.

@rogatty @bognix @rafalkalinski

PR #1201, opened 2015-08-11
Revert "Changing default port to work better on devboxes"

hakubo:
Reverts Wikia/mercury#1199

This brakes sandbox environment.
We don't have time to fix this right now, has to be done more holistically


rogatty:
LGTM

rogatty:
@lizlux Mercury stopped working on sandboxes after this change because chef doesn't override default port there. We don't know what would happen on production.

If we want to make this change we have to check what has to be done in chef-repo.

lizlux:
Got it. Thanks!

PR #1203, opened 2015-08-11
CONCF-967 Allow title duplicates and disallow label duplicates in Curated Content Tool

vforge:
https://wikia-inc.atlassian.net/browse/CONCF-967

/cc @rogatty @bognix @hakubo


front/scripts/main/components/CuratedContentEditorComponent.ts
```diff
@@ -59,14 +60,19 @@ App.CuratedContentEditorComponent = Em.Component.extend(
 			.finally((): void => this.hideLoader());
 	},
 
-	processValidationError(reason: string) {
-		if (reason === 'itemsMissing') {
+	processValidationError(type: string, reason: string) {
+		if (type == 'featured') {
```
rogatty: `===`

rogatty:
One small thing and LGTM

tomasznapieralski:
LGTM

hakubo:
LGTM

Warkot:
LGTM

bognix:
LGTM

Warkot:
I'm not sure if all the checks I added are needed — desktop tool allows to save duplicated labels between section items in curated content block and items in optional block. Maybe we should also adjust the desktop tool? @Wikia/x-wing

rogatty:
@Warkot do you mean that for example label "123" can be both in the named section and the one without name (optional)? It's not what I see on `app@CONCF-967`

front/scripts/main/components/CuratedContentEditorComponent.ts
```diff
@@ -59,14 +60,19 @@ App.CuratedContentEditorComponent = Em.Component.extend(
...
+	processValidationError(type: string, reason: string) {
+		if (type === 'featured') {
 			//@TODO CONCF-956 add translations
-			this.addAlert('alert', 'Please fix errors inside Explore the Wiki section.');
+			this.addAlert('alert', 'Please fix errors inside Featured section.');
 		} else {
-			// if other items occur that means user somehow bypassed validation of one or more items earlier
-			//@TODO CONCF-956 add translations
-			this.addAlert('alert', 'Please fix errors inside items');
+			if (reason === 'itemsMissing') {
```
rogatty: This can be `else if`

front/scripts/main/components/CuratedContentEditorComponent.ts
```diff
@@ -59,14 +60,19 @@ App.CuratedContentEditorComponent = Em.Component.extend(
...
+				//@TODO CONCF-956 add translations
+				this.addAlert('alert', 'Please fix errors inside Explore the Wiki section.');
```
rogatty: CONCF-956 is already in progress. Should we do something here?\
/cc @vforge

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,15 +162,51 @@ App.CuratedContentEditorModel.reopenClass({
...
-		if (Array.isArray(parentSection.items)) {
-			return parentSection.items.map((childItem: CuratedContentEditorItemModel): string => {
-				return childItem.label !== childLabel ? childItem.label : null
-			}).filter(String);
+	getAlreadyUsedLabels(sectionOrBlock: CuratedContentEditorItemModel, excludedLabel: string = null): string[] {
+		return this.getLabels(sectionOrBlock, excludedLabel, {isLabelAlreadyExcluded: false}).filter(this.isString);
+	},
+
+	getAlreadyUsedNonFeaturedItemsLabels(block: CuratedContentEditorModel, excludedLabel: string = null): string[] {
+		var nonFeaturedSectionsItemLabels: string[] = [],
```
rogatty: It makes sense to have a verbose function name but in the function scope I'd use simpler ones, like `labels`.

front/scripts/main/components/CuratedContentEditorSectionComponent.ts
```diff
@@ -68,7 +68,8 @@ App.CuratedContentEditorSectionComponent = Em.Component.extend(
 					this.sendAction('done', this.get('model'));
 				} else {
 					if (data.error) {
-						data.error.forEach((error: any) => this.processValidationError(error.reason));
+						data.error.forEach((error: CuratedContentValidationResponseErrorInterface)
```
hakubo: how can you iterate over an error :)
we should standardise things like this :(

rogatty:
Please rename the PR title as it's not up to date.

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,15 +162,51 @@ App.CuratedContentEditorModel.reopenClass({
...
+
+	getLabels(
+		section: CuratedContentEditorItemModel,
+		excludedLabel: string = null,
+		duplicationInfo: {isLabelAlreadyExcluded: boolean;} = {isLabelAlreadyExcluded: false}
+	): string[] {
+		if (Array.isArray(section.items)) {
+			return section.items.map((sectionItem: CuratedContentEditorItemModel): string => {
+				if (
+					!this.isString(excludedLabel) ||
```
hakubo: this looks rather complex - maybe a separate function would do for it?

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,15 +162,51 @@ App.CuratedContentEditorModel.reopenClass({
...
+			);
+		});
+
+		return nonFeaturedSectionsItemLabels.concat(this.getLabels(block.optional, excludedLabel, duplicationInfo));
+	},
+
+	getLabels(
+		section: CuratedContentEditorItemModel,
+		excludedLabel: string = null,
+		duplicationInfo: {isLabelAlreadyExcluded: boolean;} = {isLabelAlreadyExcluded: false}
```
rogatty: What's the point of using object with a single property? Isn't bool variable enough?

front/scripts/main/components/CuratedContentEditorSectionComponent.ts
vforge: Yes, but we're re-using existing API that return `error` as an array of errors :-1: And I want to refactor as little code as possible.

front/scripts/main/components/CuratedContentEditorComponent.ts
vforge: Yup, this should be changed to proper `i18n.t` call.

front/scripts/main/models/CuratedContentEditorModel.ts
Warkot: Looks weird, I know, but object variables are passed to functions by reference, others by value. I want to hold the changes to this variable between the function calls.
I could also declare a primitive Boolean variable in the scope above and change it inside the function without passing it. Do you this this would be a better solution?
```
var flag = false;
function enableFlag () {
  flag = true;
}
```

front/scripts/main/components/CuratedContentEditorComponent.ts
bognix: And please check this comment: https://wikia-inc.atlassian.net/browse/CONCF-956?focusedCommentId=188744&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-188744

vforge: Yes and no - we would need to iterate through all sent data and then we'll get label than has error - BUT the label can be shared between regular/featured/etc.

front/scripts/main/components/CuratedContentEditorComponent.ts
```diff
@@ -59,14 +60,19 @@ App.CuratedContentEditorComponent = Em.Component.extend(
 			.finally((): void => this.hideLoader());
 	},
 
-	processValidationError(reason: string) {
-		if (reason === 'itemsMissing') {
+	processValidationError(type: string, reason: string) {
+		if (type === 'featured') {
 			//@TODO CONCF-956 add translations
-			this.addAlert('alert', 'Please fix errors inside Explore the Wiki section.');
+			this.addAlert('alert', 'Please fix errors inside Featured section.');
```
bognix: please make it as i18n string since CONCF-956 is already in progress

rogatty:
LGTM

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,13 +162,45 @@ App.CuratedContentEditorModel.reopenClass({
 		return item;
 	},
 
-	getAlreadyUsedLabels(parentSection: CuratedContentEditorItemModel, childLabel: string = null): string[] {
-		if (Array.isArray(parentSection.items)) {
-			return parentSection.items.map((childItem: CuratedContentEditorItemModel): string => {
-				return childItem.label !== childLabel ? childItem.label : null
-			}).filter(String);
+	getAlreadyUsedLabels(sectionOrBlock: CuratedContentEditorItemModel, excludedLabel: string = null): string[] {
```
Warkot: After the cleanup, this looks like a redundant wrapper. Opinion, @rogatty?

rogatty: Agreed, fixed.

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,13 +162,41 @@ App.CuratedContentEditorModel.reopenClass({
...
-	getAlreadyUsedLabels(parentSection: CuratedContentEditorItemModel, childLabel: string = null): string[] {
-		if (Array.isArray(parentSection.items)) {
-			return parentSection.items.map((childItem: CuratedContentEditorItemModel): string => {
-				return childItem.label !== childLabel ? childItem.label : null
-			}).filter(String);
+	getAlreadyUsedNonFeaturedItemsLabels(block: CuratedContentEditorModel, excludedLabel: string = null): string[] {
+		var labels: string[] = [];
+
+		// Labels of section items
+		block.curated.items.forEach((section: CuratedContentEditorItemModel): void => {
```
bognix: consider renaming block to something like root or model since it's not block but model of all CuratedContent

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,13 +162,41 @@ App.CuratedContentEditorModel.reopenClass({
...
+	getAlreadyUsedLabels(
+		sectionOrBlock: CuratedContentEditorItemModel,
+		excludedLabel: string = null
+	): string[] {
+		var labels: string[] = [];
+
+		if (Array.isArray(sectionOrBlock.items)) {
+			labels = sectionOrBlock.items.map((item: CuratedContentEditorItemModel): string => {
+				var itemLabel: string = Em.get(item, 'label') || '',
+					itemLabelLowerCase: string = itemLabel.toLowerCase(),
```
bognix: no need to lower casing. Resources are case sensitive so iTem and itEm are not the same

bognix:
Two small things. Except that - LGTM

rogatty:
LGTM

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,13 +162,35 @@ App.CuratedContentEditorModel.reopenClass({
...
+		excludedLabel: string = null
+	): string[] {
+		var labels: string[] = [];
+
+		if (Array.isArray(sectionOrBlock.items)) {
+			labels = sectionOrBlock.items.map((item: CuratedContentEditorItemModel): string => {
+				var itemLabel: string = Em.get(item, 'label');
+
+				return (excludedLabel === null || itemLabel !== excludedLabel) ? itemLabel : null;
+			}).filter((item: any): boolean => typeof item === 'string');
```
vforge: you can use `.filter(String)` instead

front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -156,13 +162,35 @@ App.CuratedContentEditorModel.reopenClass({
 		return item;
 	},
 
-	getAlreadyUsedLabels(parentSection: CuratedContentEditorItemModel, childLabel: string = null): string[] {
-		if (Array.isArray(parentSection.items)) {
-			return parentSection.items.map((childItem: CuratedContentEditorItemModel): string => {
-				return childItem.label !== childLabel ? childItem.label : null
-			}).filter(String);
+	getAlreadyUsedNonFeaturedItemsLabels(modelRoot: CuratedContentEditorModel, excludedLabel: string = null): string[] {
```
hakubo:
```ts
getAlreadyUsedNonFeaturedItemsLabels(modelRoot: CuratedContentEditorModel, excludedLabel: string = null): string[] {

  //flatten the array
  return [].concat.apply([], modelRoot.curated.items.map((section: CuratedContentEditorItemModel) => 
	  // Labels of section items
      this.getAlreadyUsedLabels(section, excludedLabel)
  )
  .concat(
	  // Labels of optional block items
	  this.getAlreadyUsedLabels(modelRoot.optional, excludedLabel)
  ));
}
```

front/scripts/main/models/CuratedContentEditorModel.ts
hakubo: neat

hakubo:
one minor thing, LGTM

front/scripts/main/models/CuratedContentEditorModel.ts
Warkot: Not really.
```
[1, 2, 3, 'a', 'b', 'c', [], [1], ['a'], {}, {foo: 'bar'}, null, undefined].filter(String);
[1, 2, 3, "a", "b", "c", Array[1], Array[1], Object, Object, null, undefined]
```
It filtered out just the empty array.

hakubo: not neat...

front/scripts/main/models/CuratedContentEditorModel.ts
rogatty: It would work but I have two questions:
1. Does it improve performance much? I guess not.
2. Is it easier to read? In my opinion not.

front/scripts/main/models/CuratedContentEditorModel.ts
Warkot: It also filtered out the empty string. :open_mouth: 
```
[0, 1, 2, 3, '', 'a', 'b', 'c', [], [1], ['a'], {}, {foo: 'bar'}, null, undefined].filter(String);
[0, 1, 2, 3, "a", "b", "c", Array[1], Array[1], Object, Object, null, undefined]
```

vforge: :(

front/scripts/main/models/CuratedContentEditorModel.ts
hakubo: My small affection toward functional programming might show :)

I tend to try to avoid situation where my code mutates some 'global' state
in this case 'label'

I'll leave it to authors discretion to review this. I am not strongly standing for such style (yet :))

front/scripts/main/models/CuratedContentEditorModel.ts
Warkot: Yup, only labels which are strings or the `null` object for the current object's label. But when using `.filter(String)` the `null` stays in the array.

front/scripts/main/models/CuratedContentEditorModel.ts
Warkot: It looks nice and clean and I like it, but I'm not sure about the `apply()`, it might just look like this:
```
return [].concat(modelRoot.curated.items.map(…
```

bognix:
LGTM

hakubo:
LGTM

vforge:
LGTM

PR #1208, opened 2015-08-12
DAT-3121 Related pages images fix

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3121

@rogatty


rogatty:
LGTM

hakubo:
LGTM

dianafa:
@rogatty @hakubo Can you please take a look on this one more time?

Strange issue - the main change happens here:
https://github.com/Wikia/mercury/blob/DAT-3121/front/templates/main/components/content-recommendations.hbs#L8

front/scripts/main/components/ImageMediaComponent.ts
```diff
@@ -54,7 +54,7 @@ App.ImageMediaComponent = App.MediaComponent.extend(App.ArticleContentMixin, App
 	 * @desc return the thumbURL for media.
 	 * If media is an icon, use the limited width.
 	 */
-	url: Em.computed('media', 'computedHeight', 'articleContent.width', 'imageSrc', {
+	url: Em.computed({
```
rogatty: Why?

front/templates/main/components/content-recommendations.hbs
```diff
@@ -5,7 +5,7 @@
 			{{#each pages as |page|}}
 				<li>
 					{{#link-to 'article' page.title data-tracking-category='related'}}
-						{{image-media url=page.imgUrl caption=page.title height=200 limitHeight=true}}
+						{{image-media limitHeight=true height=200 url=page.imgUrl caption=page.title}}
```
rogatty: Does the order matter?

dianafa: YES :D

vforge: Care to explain?

front/scripts/main/components/ImageMediaComponent.ts
dianafa: Because we don't really want to recompute the url each time the media, computedHeight, articleContent.width or imageSrc change, do we? It was like this some time ago, and I changed to it to be consistent with computed property syntax. So nothing new here.
Basically, what happens here when one time ImageMediaComponents are artificially created by replacing in article and when they are normal ember components like in content recommendations needs to be refactored someday, as well as set() here, which always calls the Thumbnailer.mode.topCrop where the default crop mode AFAIK is thumbnailDown - are we sure the set() is called only for content-recommendations? grząski grunt z tymi obrazkami :D

front/templates/main/components/content-recommendations.hbs
dianafa: Not yet. I need to investigate it. I mean it's not normal :D Maybe I'll write to Ember team.

Reason of this change is that the `url` depends of `height` (to compute the `computedHeight` which is used in `url#set`). `url` depends of `limitHeight` as well which is used by `getThumbUrl` to know how to generate the url. While debugging I recognised, that if we pass the `height` and `limitHeight` AFTER we call the `url` set() in moment of function execution this values are undefined! If we switch the order, they are already has been defined and everything works good. It's surprising and shouldn't be like this. Maybe it's an Ember regression.

front/scripts/main/components/ImageMediaComponent.ts
```diff
@@ -17,9 +17,9 @@ App.ImageMediaComponent = App.MediaComponent.extend(App.ArticleContentMixin, App
 		'emptyGif'
 	),
 
-	hasCaption: Em.computed('media.caption', 'isIcon', function (): boolean {
+	caption: Em.computed('media.caption', 'isIcon', function (): string {
```
hakubo: string|boolean ?

hakubo:
@dianafa are you planning to do the investigation before merging this branch?
It'd be good to at least know what have to be done after merging this - what is 'grząski grunt'
Care to summarise it somewhere so we could then take and implement upon? or implement needed improvements with this PR?

dianafa:
In my opinion we should proceed with this PR and fix the issue. I've created a ticket for refactoring this logic: https://wikia-inc.atlassian.net/browse/CONCF-1080

vforge:
LGTM :)

PR #1216, opened 2015-08-13
SOC-884 Post detail page updates

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-884
This pull request is co-dependent on https://github.com/Wikia/mercury/pull/1218. They should be merged at the same time.

@lizlux @garthwebb


front/scripts/main/router.ts
```diff
@@ -70,7 +70,7 @@ App.Router.map(function () {
 		path: articlePath + 'addPhoto/:title'
 	});
 
-	this.route('discussion', {path: '/d'}, function () {
+	this.route('discussion', {path: '/d/:thing'}, function () {
```
lizlux: What is `thing`?

front/styles/main/component/_site-head.scss
```diff
@@ -65,3 +73,51 @@
...
 	}
 }
+
+hr.theme {
+	border-width: 0;
+	height: 10px;
+	margin: 0;
+}
+
+.books-vertical {
```
rogatty: You can use a loop to generate these styles, see https://github.com/Wikia/mercury/blob/74eb4c8a197c438bab104cf589cf8838e2a1503c/front/styles/main/component/_smart-banner.scss#L130 for example.

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionForumPostController = Em.Controller.extend({
+	init: function () {
+		// Enables vertical-colored theme bar in site-head component
+		this.controllerFor('application').set('themeBar', true);
```
rogatty: I don't see it being set to `false` anywhere. How about setting it to `true` in the route in [activate hook](http://emberjs.com/api/classes/Ember.Route.html#method_activate) and to `false` in [deactivate hook](http://emberjs.com/api/classes/Ember.Route.html#method_deactivate)?

front/scripts/main/components/SiteHeadComponent.ts
```diff
@@ -5,6 +5,7 @@
 
 App.SiteHeadComponent = Em.Component.extend(App.TrackClickMixin, {
 	classNames: ['site-head'],
+	classNameBindings: ['themeBar'],
```
rogatty: Please add a default for `themeBar` property.

front/scripts/main/controllers/DiscussionForumPostController.ts
garthwebb: @mklucsarits care to comment?  This must be another merge artifact :/

front/scripts/main/router.ts
garthwebb: @mklucsarits ...  and another merge artifact :/  Should this still be here?

mklucsarits: It's the type of page for discussion, which would be 'f' for forum, 'p' for post, or something else. Should it be called 'type' instead?

lizlux: Yeah I think "type" is better than "thing". Wishing there was something better than "type", but that sounds good for now.

garthwebb:
LGTM

garthwebb:
@rogatty I believe @mklucsarits addressed all your feedback so I'm going to merge this.  If you feel there is any additional feedback to give, please feel free to leave it here or open a separate ticket.

PR #1218, opened 2015-08-13
SOC-959 OpenGraph attributes for discussion posts

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-959

This pull request is co-dependent on https://github.com/Wikia/mercury/pull/1216. They should be merged at the same time.

Related wikia/app change: https://github.com/Wikia/app/pull/8106

@garthwebb @kenkouot


server/facets/api/discussion.ts
```diff
@@ -0,0 +1,82 @@
...
+							content.substr(0, 172) + '...' : content;
+						openGraphData.image = wikiVars.image;
+						resolve(openGraphData);
+					})
+					.catch((): void => {
+						resolve(openGraphData);
+					});
+			});
+		}
+	}
```
rogatty: Function return type should be consistent so for example `} else { return null; }`

server/facets/api/discussion.ts
```diff
@@ -0,0 +1,82 @@
...
+
+function getOpenGraphAttributes (path: string, pathParams: any, wikiVars: any): any {
+	var apiUrl: string,
+		openGraphData: any = {},
+		regexMatch: any,
+		promise: Promise<any>;
+
+	// Post
+	if (pathParams.thing === 'p' && pathParams.id) {
+		regexMatch = pathParams.id.match(/^[\w-]*?(\d+)$/);
```
rogatty: Can you put it in a variable and add a comment explaining what it's supposed to match?

server/facets/api/discussion.ts
```diff
@@ -0,0 +1,82 @@
...
+	});
+}
+
+function getOpenGraphAttributes (path: string, pathParams: any, wikiVars: any): any {
+	var apiUrl: string,
+		openGraphData: any = {},
+		regexMatch: any,
+		promise: Promise<any>;
+
+	// Post
```
kenkouot: this comment is not very useful

server/routes.ts
```diff
@@ -230,8 +230,8 @@ articlePagePaths.forEach((path) => {
 unauthenticatedRoutes.push({
 	// Discussion forums
 	method: 'GET',
-	path: '/d/{subpaths*}',
-	handler: require('./facets/showApplication')
+	path: '/d/{thing}/{id}/{action?}',
```
kenkouot: `thing`

server/facets/api/discussion.ts
```diff
@@ -0,0 +1,82 @@
...
+
+		if (result.wiki.language) {
+			var userDir = result.wiki.language.userDir;
+			result.isRtl = (userDir === 'rtl');
+		}
+
+		// TODO: I'm not fond of there being several patterns for transforming the object here.
+		Tracking.handleResponse(result, request);
+
+		openGraph = getOpenGraphAttributes(request.path, request.params, result.wiki);
```
kenkouot: Why couldn't we have just put this openGraph logic in the generic `showApplication` controller?

mklucsarits: Good question. I actually spent several hours trying to fit this OpenGraph logic into the showApplication handler. The main problem I dealt with was that the request to get discussion API data for the OG attributes had to be wrapped in a promise. The OG attributes for article pages on the other hand were readily available and didn't require a promise. I tried a few different approaches to combine the OG attribute gathering logic into one function for both discussions and for articles, but it was difficult when one case required an async call and one didn't. In addition to the logic necessary to parse the route path -- to see which discussion API endpoint needed to be hit for the OG attributes -- the code seemed messy any way I tried to arrange it. The easiest option still looked to be giving discussion routes their own handler, despite copying some code from showApplication.

I'm sure there's a way to make everything more generalized but I'm just not sure of the best way. HapiJS also supports pre-handlers and I considered moving showApplication stuff into a pre-handler, for wiki variables and other basic things needed on every page, rather than moving the complicated logic for OG tags for discussions into showApplication.

Again there's probably a better way, but I spent the better part of a day refactoring it in different ways, but couldn't come up with an elegant solution. So I just resorted to the simplest solution. I'm open to suggestions, but make sure those suggestions take into consideration the async-versus-sync logic for getting OG data on different routes.

rogatty: In my opinion it makes sense to create a promise even if the data isn't async just to keep things consistent. We do this already, for example [ArticleModel](https://github.com/Wikia/mercury/blob/dev/front/scripts/main/models/ArticleModel.ts#L74) is synchronously filled with data from DOM if available and then resolved.

mklucsarits: @rogatty That was one of the ways I considered doing it. I wasn't sure if that was considered bad practice or not. That would make certainly make it consistent and simpler. I'll refactor it that way unless anyone has a better suggestion.

server/facets/api/discussion.ts
garthwebb: This is a pretty minor quibble, but I would suggest either being more rigid, explicitly adding the '-' that is supposed to immediately precede the ID, e.g.:
```
regexMatch = pathParams.id.match(/^([\w-]+-)?(\d+)$/);
```
or less rigid and match only the part that you care about, e.g.:
```
regexMatch = pathParams.id.match(/(\d+)$/);
```

server/routes.ts
garthwebb: @kenkouot you mean ```{thing}``` and ```{id}``` should be combined into just ```{thing}``` per the email thread?  Agreed if that's what you meant, request for more detail if not ;)

server/facets/api/discussion.ts
garthwebb: Btw, I am not worried about the performance of this regex **at all** but this site has some cool debugging output for regexes: https://regex101.com/

server/facets/api/discussion.ts
rogatty: @mklucsarits see http://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html, section `Advanced mistake #1: not knowing about Promise.resolve()` (no anchors there...).

You can use `Em.RSVP.Promise.resolve().then(...)` instead of creating new instance manually, it's a nice shortcut.

mklucsarits:
Hey @kenkouot and @rogatty can you have another look at the route handler portion of this?

server/facets/showApplication.ts
```diff
@@ -1,36 +1,48 @@
...
-		result.weppyConfig = localSettings.weppy;
-		result.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
-
-		if (result.wiki.language) {
-			var userDir = result.wiki.language.userDir;
-			result.isRtl = (userDir === 'rtl');
+		context.wiki = response.data;
+
+		if (context.wiki.language) {
+			var userDir = context.wiki.language.userDir;
```
rogatty: `var` should be at the top of the scope

server/facets/showApplication.ts
```diff
@@ -1,36 +1,48 @@
...
-		Tracking.handleResponse(result, request);
+		return OpenGraph.getAttributes(request.path, request.params, context.wiki);
+	}).then((openGraphData: any): void => {
+		// Add OpenGraph attributes to context
+		context.openGraph = openGraphData;
 
-		reply.view('application', result);
+		outputResponse(request, reply, context);
+	}).catch((): void => {
+		// In case of any unforeseeable errors, attempt to output with the context we have so far
```
rogatty: Can we log the error that comes as `catch` argument? Might be helpful to debug.

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,58 @@
+import localSettings = require('../../config/localSettings');
+import MW = require('./MediaWiki');
+
+interface OpenGraphAttributes {
+	description?: string;
+	image?: string;
```
rogatty: Optional parameters go to the bottom, right?

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,58 @@
...
+		openGraphData: any = {},
+		regexMatch: any,
+		defaultPromise: Promise<any>;
+
+	defaultPromise = new Promise((resolve: Function, reject: Function): void => {
+		resolve(openGraphData);
+	});
+
+	// Discussions
+	if (path.split('/')[1] === 'd') {
```
rogatty: I'd leave `if` here but extract the rest below into another function as it's related exclusively to a single route.

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,58 @@
...
+						})
+						.catch((): void => {
+							resolve(openGraphData);
+						});
+				});
+			}
+		}
+	}
+
+	return defaultPromise;
```
rogatty: No need for a variable, just `return Promise.resolve(openGraphData);`

server/lib/OpenGraph.ts
mklucsarits: Not anymore. I think the latest consensus is that it was causing more confusion than it was helping, so we're alphabetizing everything and just leaving optional parameters where they are, in that order.

lizlux: Yeah I think there was discussion of it but it never made it into the guidelines.

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,63 @@
...
+	if (request.params.thing === 'p' && request.params.id) {
+		// Get post ID, which might be prepended with slug text
+		regexMatch = request.params.id.match(/(\d+)$/);
+		if (regexMatch !== null) {
+			apiUrl = 'http://' + localSettings.servicesDomain + '/'
+				+ localSettings.discuss.baseAPIPath + '/' + wikiVars.id
+				+ '/threads/' + regexMatch[1];
+
+			openGraphData.type = 'article';
+			openGraphData.url = wikiVars.basePath + request.path;
```
rogatty: These two seem like defaults. Would it make sense to add them to the `openGraphData` outside of this function?

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,63 @@
...
+				// Fetch discussion post data from the API to complete the OG data
+				MW.fetch(apiUrl)
+					.then((response: any): void => {
+						var content = response._embedded.firstPost[0].rawContent;
+						openGraphData.title = response.title ? response.title :
+							i18n.t('discussion:open-graph.generic-title', {siteName: wikiVars.siteName});
+						// Keep description to 175 characters or less
+						openGraphData.description = content.length > 175 ?
+							content.substr(0, 172) + '...' : content;
+						openGraphData.image = wikiVars.image;
```
rogatty: This seems like a default, `wikiVars` is available already in `showApplication.ts`

server/facets/operations/prepareArticleData.ts
```diff
@@ -46,7 +46,8 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 	result.queryParams = Utils.parseQueryParams(request.query, allowedQueryParams);
 	result.openGraph = {
 		type: 'article',
-		title: title
+		title: title,
```
hakubo: you can use object shorthand:
```
result.openGraph = {
  type: 'article',
  title,
  url: result.canonicalUrl
};
```

server/facets/operations/prepareArticleData.ts
```diff
@@ -46,7 +46,8 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
...
+		url: result.canonicalUrl
```
hakubo: should we send this data twice?
as 
result.openGraph.url
and 
result.canonicalUrl?

can't openGraph use result.canonicalUrl?

server/facets/operations/processCuratedContentData.ts
```diff
@@ -53,7 +53,8 @@ function prepareData (request: Hapi.Request, result: any): void {
 	result.queryParams = Utils.parseQueryParams(request.query, ['noexternals', 'buckysampling']);
 	result.openGraph = {
 		type: 'website',
-		title: result.wiki.siteName
+		title: result.wiki.siteName,
+		url: result.canonicalUrl
```
hakubo: same here, do we have to pass this data twice?

server/routes.ts
```diff
@@ -246,7 +246,7 @@ articlePagePaths.forEach((path) => {
 unauthenticatedRoutes.push({
 	// Discussion forums
 	method: 'GET',
-	path: '/d/{subpaths*}',
+	path: '/d/{thing}/{id}/{action?}',
```
hakubo: What is this thing? Can it have a meaningful name?
What is the action? Can be that handled by HTTP method?

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,63 @@
...
+			return new Promise((resolve: Function, reject: Function): void => {
+				// Fetch discussion post data from the API to complete the OG data
+				MW.fetch(apiUrl)
+					.then((response: any): void => {
+						var content = response._embedded.firstPost[0].rawContent;
+						openGraphData.title = response.title ? response.title :
+							i18n.t('discussion:open-graph.generic-title', {siteName: wikiVars.siteName});
+						// Keep description to 175 characters or less
+						openGraphData.description = content.length > 175 ?
+							content.substr(0, 172) + '...' : content;
```
hakubo: Should this be actually: U+2026 or &hellip;
also
does elipsis have the same meaning in all languages?
also
for RTL languages should that be in front?

server/lib/OpenGraph.ts
```diff
@@ -0,0 +1,63 @@
...
+				+ '/threads/' + regexMatch[1];
+
+			openGraphData.type = 'article';
+			openGraphData.url = wikiVars.basePath + request.path;
+
+			return new Promise((resolve: Function, reject: Function): void => {
+				// Fetch discussion post data from the API to complete the OG data
+				MW.fetch(apiUrl)
+					.then((response: any): void => {
+						var content = response._embedded.firstPost[0].rawContent;
```
hakubo: _embedded + rawContent does not sound like something that won't brake.
What are those?

server/facets/operations/prepareArticleData.ts
mklucsarits: Do you mean to use `result.canonicalUrl` in the open-graph-tags template? Not every page may set `result.canonicalUrl`, and not every page will use that URL -- for example, discussion posts in this same pull request.

server/lib/OpenGraph.ts
mklucsarits: These are only defaults for discussion posts.

server/lib/OpenGraph.ts
mklucsarits: Again this is only the default for discussion posts. I don't want to make any assumption of what image will be used for other pages.

server/routes.ts
mklucsarits: The `thing` parameter will be renamed. `action` is something like 'edit', which yes may depend on HTTP method. This route handling is subject to change before the discussion MVP is finished.

server/lib/OpenGraph.ts
mklucsarits: That's how the Services team structured the API response. If they change it, I expect them to notify us or else the discussions app will break in many places, not just here.

PR #1220, opened 2015-08-14
Hg 834 Adding the new top bar component to mercury discussion route

lizlux:
https://wikia-inc.atlassian.net/browse/HG-834

WIP


front/scripts/main/components/TopBarComponent.ts
```diff
@@ -0,0 +1,17 @@
...
+
+App.TopBarComponent = Em.Component.extend(App.TrackClickMixin, {
+	// todo: this adds a new top-bar on every transition. 
+	didInsertElement: function (): void {
+		// create style guide top-bar web component
+		new window.SG.Components.TopBar({
+			// send reference to imported document
+			importedDoc: document.getElementById('top-bar-import').import,
+			logoUrl: '/front/vendor/wikia-style-guide/dist/svg/wikia_logo.svg'
+		}).createElement();
```
kenkouot: I don't like that you have to explicitly `createElement` after an import. I know that's the only option using just the Web Components native APIs, but I like the Polymer approach where once you import a Polymer element, it's available for us (it is registered as part of the import).

kenkouot: In addition, this situation becomes even less clear once you have multiple places invoking a component. Do you just call createElement `n` # of times? I'll explore not having to do this in the next iteration.

PR #1222, opened 2015-08-14
Fix for horizontal scrollbar in mobile view for Japan Homepage

perjg:


perjg:
Will merge on Thursday unless there are objections.

PR #1224, opened 2015-08-14
SOC-904 Do not show header on Auth Pages in modal

bkoval:
@rafalkalinski @garthwebb 
https://wikia-inc.atlassian.net/browse/SOC-904
If a URL contains "modal=1" let's get rid of a corporate header and gray background and siplay the raw auth view.

### DIFF
--- a/server/facets/auth/authView.ts
+++ b/server/facets/auth/authView.ts
@@ -30,6 +30,7 @@ module authView {
 		headerText?: string;
 		bodyClasses?: string;
 		trackingConfig?: any;
+		isModal?: boolean;
 	}
 
 	export function view (template: string, context: AuthViewContext, request: Hapi.Request, reply: any): Hapi.Response {
@@ -82,18 +83,20 @@ module authView {
 	}
 
 	export function getDefaultContext (request: Hapi.Request): AuthViewContext {
-		var viewType: string = this.getViewType(request);
+		var viewType: string = this.getViewType(request),
+			isModal: boolean = request.query.modal === '1';
 		return {
 			title: null,
 			canonicalUrl: this.getCanonicalUrl(request),
 			exitTo: this.getRedirectUrl(request),
 			mainPage: "http://www.wikia.com",
+			isModal: isModal,
 			language: request.server.methods.i18n.getInstance().lng(),
 			trackingConfig: localSettings.tracking,
 			optimizelyScript: localSettings.optimizely.scriptPath +
 				(localSettings.environment === Utils.Environment.Prod ?
 					localSettings.optimizely.account : localSettings.optimizely.devAccount) + '.js',
-			standalonePage: (viewType === authView.VIEW_TYPE_DESKTOP),
+			standalonePage: (viewType === authView.VIEW_TYPE_DESKTOP && !isModal),
 			pageParams: {
 				viewType: viewType
 			}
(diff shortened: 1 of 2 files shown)


server/facets/auth/authView.ts
```diff
@@ -82,18 +83,20 @@ module authView {
 	}
 
 	export function getDefaultContext (request: Hapi.Request): AuthViewContext {
-		var viewType: string = this.getViewType(request);
+		var viewType: string = this.getViewType(request),
+			isModal: boolean = request.query.modal === '1';
```
rafalkalinski: I wonder if we need to check it the modal === '1'. Maybe if we need only one value here it would be more convenient just to check if there is 'modal' param in the query.

rafalkalinski: And I wonder if there is any sense to set it to true on mobile devices. I think we should also check the VIEW_TYPE here.

server/views/_layouts/auth.hbs
```diff
@@ -21,13 +21,15 @@
 				<a class="main-page" href="{{mainPage}}"><span>Wikia</span></a>
 			</nav>
 		{{else}}
-			{{#if exitTo }}
+			{{#unless isModal}}
+			{{#if exitTo}}
```
garthwebb: What do you think about simplifying this template logic?  I could see two options:
* Set a single variable you could test with ```#if```, e.g. ```{{#if shouldShowClose}}```
* In authView.ts, clear the value of ```exitTo``` whenever ```isModal``` is true.

bkoval: :+1:

server/facets/auth/authView.ts
```diff
@@ -82,18 +83,20 @@ module authView {
...
 		return {
 			title: null,
 			canonicalUrl: this.getCanonicalUrl(request),
 			exitTo: this.getRedirectUrl(request),
 			mainPage: "http://www.wikia.com",
+			isModal: isModal,
```
rafalkalinski: 'isModal' should go to pageParams.

server/facets/auth/authView.ts
```diff
@@ -82,18 +83,20 @@ module authView {
...
 			language: request.server.methods.i18n.getInstance().lng(),
 			trackingConfig: localSettings.tracking,
 			optimizelyScript: localSettings.optimizely.scriptPath +
 				(localSettings.environment === Utils.Environment.Prod ?
 					localSettings.optimizely.account : localSettings.optimizely.devAccount) + '.js',
-			standalonePage: (viewType === authView.VIEW_TYPE_DESKTOP),
+			standalonePage: (viewType === authView.VIEW_TYPE_DESKTOP && !isModal),
```
rafalkalinski: 'standalonePage' should probably also be in the pageParams

server/facets/auth/authView.ts
kvas-damian: @rafalkalinski I'm not sure if any value in ```modal``` should enable it. I can imagine someone typing: ```modal=0```.

rafalkalinski: @kvas-damian good point, lets leave it like it is

server/facets/auth/authView.ts
kvas-damian: @rafalkalinski This variable is used on every auth page and is not required to be exported to frontend JS. I don't see a reason now to to move it there.

server/facets/auth/authView.ts
bkoval: @rafalkalinski It's not needed in this story.

garthwebb:
LGTM

PR #1225, opened 2015-08-14
SOC-951 Allow users to paginate through post replies

garthwebb:
https://wikia-inc.atlassian.net/browse/SOC-951


front/styles/main/module/_discussion.scss
```diff
@@ -13,6 +13,15 @@
 		}
 	}
 
+	.see-more {
+		display: inline-block;
```
bkoval: Looks like this class lacks alphabet order of properties :)

front/styles/main/module/_discussion.scss
```diff
@@ -13,6 +13,15 @@
...
+		background-color: #E6E6E6;
```
bkoval: I would consider moving all colors to settings file, where they're easily configurable, what do you think?

front/templates/main/discussion/forum/post.hbs
```diff
@@ -6,13 +6,20 @@
 		upvoteCount=model.upvoteCount
 		replyCount=model.postCount
 	}}
-
+	{{#if showMore}}
+        <button {{action 'expand'}} style="width: 100%" class="see-more">
```
bkoval: Maybe this style attribute could be replaced with a rule in the stylesheet?

front/templates/main/discussion/forum/post.hbs
```diff
@@ -6,13 +6,20 @@
...
+			{{svg 'comments' viewBox='0 0 30 30' class='icon replies'}}
+            View Older Replies
+        </button>
+	{{/if}}
 	{{#each model.replies as |item index|}}
 		{{post-reply
 			authorId=item.creatorId
```
bkoval: Maybe we could just pass the whole "item" object to the component?

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,28 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionForumPostController = Em.Controller.extend({
+
+	init: function () {
+		this.controllerFor('application').set('noAds', true);
+	},
+
+	vertical: Em.computed(function (): string { return Mercury.wiki.vertical; }),
```
bkoval: I think the rule of thumb is to expand the function body in the new line, but I don't really know if we have guidelines for that :)

front/scripts/main/models/PostModel.ts
```diff
@@ -1,27 +1,79 @@
...
 	firstPost: null,
 	upvoteCount: 0,
-	postCount: 0
+	postCount: 0,
+	page: null,
+
+	loadNextPage() {
+		return new Em.RSVP.Promise((resolve: Function, reject: Function) => {
+			Em.$.ajax({
+				url: 'https://services.wikia-dev.com' +
```
bkoval: Why not merge strings that happen one after another? E.g. 'https://services.wikia-dev.com/discussion/' and '?responseGroup=full&sortDirection=descending'.

Also, I think this URL is so complicated that it deserves it's own builder method :)

front/scripts/main/models/PostModel.ts
```diff
@@ -1,27 +1,79 @@
...
+					 '?responseGroup=full&sortDirection=descending&limit=' + postInstance.replyLimit,
+
 				dataType: 'json',
 				success: (data: any) => {
+					var replies = data._embedded['doc:posts'],
+						pivotId = replies[0].id,
+						i;
+
+					// @todo dev testing remove
+					for (i=0; i<replies.length; i++) {
```
bkoval: So is this whole loop a dev testing code? If so, I believe we could just leave it on a feature branch and not merge to dev :)

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,28 @@
...
+	showMore: Em.computed(function (): boolean {
+			var model = this.get('model');
+
+			return model.replies.length < model.postCount;
+	}),
+
+	actions: {
+		expand: function() {
```
kenkouot: space after `function` keyword

front/scripts/main/controllers/DiscussionForumPostController.ts
garthwebb: @mklucsarits Could you check this out?  I know I didn't add this so I'm assuming this came in via a merge at some point.  Is this still what it should be?

front/scripts/main/models/PostModel.ts
garthwebb: Yup, debugging code.  Killed.

front/scripts/main/controllers/DiscussionForumPostController.ts
mklucsarits: Looks like lines 6-10 are old code that need to be removed.

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,28 @@
...
+			var model = this.get('model');
+
+			model.loadNextPage().then(function (){
+				if (model.replies.length >= model.postCount) {
+					this.set('showMore', false);
+				}
+			}.bind(this));
```
mklucsarits: If you use arrow notation for this function declaration (`(): void => {`) you don't need to bind `this` here.

front/styles/main/module/_discussion.scss
```diff
@@ -13,6 +13,15 @@
 		}
 	}
 
+	.see-more {
+		background-color: #E6E6E6;
+		border-width: 0;
+		color: #16ABC1;
```
mklucsarits: This should use the SCSS variable for the vertical color. See how I did it here https://github.com/Wikia/mercury/pull/1216/files#diff-adaefc77e753dfb73d772ff4a8a60ee7R83 or here https://github.com/Wikia/mercury/blob/dev/front/styles/main/module/_discussion.scss#L74

Look at Igor's comment there though; we should probably use loops for generating that stuff in Sass.

front/scripts/main/models/PostModel.ts
garthwebb: The host part actually needs to be in config as well and switch between dev/prod, but after talking to liz that turned out to be more complicated than I thought, so I decided to defer that work.

Where do you think a good place for the builder function would be?  I really want to get this same construction out of the three places it currently is.

I created a separate ticket for this btw: https://wikia-inc.atlassian.net/browse/SOC-1174.  I don't normally condone making new tickets to handle feedback on a PR but I think there is enough work here that it makes sense to split out and tackle on its own.

bkoval: If you want a configurable url switching between environments, add a property for it in LocalSettings.base and a define environment variables in chef, e.g. here https://github.com/Wikia/chef-repo/blob/master/cookbooks/_mercury/attributes/default.rb like it's done with Helios :santa: 

As for the builder, I can see there's a Mercury.Utils object and it already has things like this: https://github.com/Wikia/mercury/blob/dev/front/scripts/mercury/utils/articleLink.ts

Ping @rogatty @hakubo Is this the best place for discussions url builder?

front/styles/main/module/_discussion.scss
```diff
@@ -13,6 +13,20 @@
 		}
 	}
 
+	@for $i from 1 through length($verticals) {
```
mklucsarits: I think only the `color` property needed to be declared in this loop. You can leave the others like they were previously defined, because this will generate a lot of duplicate code that's not needed.

front/scripts/main/models/PostModel.ts
```diff
@@ -1,27 +1,73 @@
...
+				'?responseGroup=full' +
+				'&sortDirection=descending' +
+				'&limit=' + this.replyLimit +
+				'&pivot=' + this.pivotId +
+				'&page=' + (this.page+1),
+				dataType: 'json',
+				success: (data: any) => {
+					var newReplies = data._embedded['doc:posts'];
+
+					newReplies.reverse();
```
mklucsarits: Can you explain why you used `reverse` with sortDirection=descending instead of just using sortDirection=ascending?

garthwebb: Assume you have 50 replies where 1 is the first (oldest) reply and 50 is the last (most recent).  If I ask for 5 replies with no offset, sorted by time descending I get an array ordered like:
```
50
49
48
47
46
```
However, the UI requires the most recent items at the bottom.  So I do ```reverse``` to get:
```
46
47
48
49
50
```

front/styles/main/module/_discussion.scss
garthwebb: Whoops!

front/scripts/main/models/PostModel.ts
mklucsarits: Thanks, that makes sense.

front/scripts/main/models/PostModel.ts
```diff
@@ -1,27 +1,73 @@
...
 			Em.$.ajax({
 				url: `https://services.wikia-dev.com/discussion/${wikiId}/threads/${threadId}` +
-					 '?responseGroup=full&sortDirection=descending&limit=10',
+					 '?responseGroup=full&sortDirection=descending&limit=' + postInstance.replyLimit,
+
 				dataType: 'json',
 				success: (data: any) => {
+					var replies = data._embedded['doc:posts'],
+						pivotId = replies[0].id,
+						i;
```
mklucsarits: What's this `i` used for? If it will be used, it needs a type specified.

garthwebb: Nothing, nothing at all ...

front/scripts/main/models/PostModel.ts
rogatty: We have [buildUrl](https://github.com/Wikia/mercury/blob/3a965cbc8e8430ac3e854eac663486ac97f9404d/front/scripts/baseline/mercury/utils/buildUrl.ts) that @mklucsarits wrote but I'm not sure if it makes sense to expand it. Maybe a better option would be to have a similar `buildServiceUrl`?

We probably don't need it in the baseline so `Mercury.Utils` seems like a good fit for it.

mklucsarits:
LGTM

front/scripts/main/models/PostModel.ts
bkoval: :+1:

front/styles/main/module/_discussion.scss
```diff
@@ -13,6 +13,23 @@
 		}
 	}
 
+	.see-more {
+		background-color:#E6E6E6;
+		border-width:0;
+		display:inline-block;
+		font-size:10px;
```
bkoval: I guess all these properties lack spaces after colons :)

front/scripts/main/components/PostReplyComponent.ts
```diff
@@ -4,10 +4,11 @@
 App.PostReplyComponent = Em.Component.extend({
 	classNames: ['post-reply'],
 
+	item: null,
 	author: null,
 
 	init: function (): void {
-//		App.UserModel.find({userId: this.get('authorId')}).then((result: any): any => {
+//		App.UserModel.find({userId: this.get('item.authorId')}).then((result: any): any => {
```
bkoval: Commented code - still needed?

bkoval:
Last two minor things, except that LGTM :kimono:

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,23 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionForumPostController = Em.Controller.extend({
+
+	showMore: Em.computed(function (): boolean {
+			var model = this.get('model');
+
+			return model.replies.length < model.postCount;
```
rogatty: If `model.replies` is undefined it will fail. You can use `Em.get(model, 'replies.length')` to be safe.

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,23 @@
...
+
+	actions: {
+		expand: function () {
+			var model = this.get('model');
+
+			model.loadNextPage().then( (): void => {
+				if (model.replies.length >= model.postCount) {
+					this.set('showMore', false);
+				}
+			} );
```
rogatty: MW habits ;)

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,23 @@
...
+			return model.replies.length < model.postCount;
+	}),
+
+	actions: {
+		expand: function () {
+			var model = this.get('model');
+
+			model.loadNextPage().then( (): void => {
+				if (model.replies.length >= model.postCount) {
+					this.set('showMore', false);
```
rogatty: This shouldn't be needed as you already have `showMore` computed.

front/scripts/main/controllers/DiscussionForumPostController.ts
```diff
@@ -0,0 +1,23 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionForumPostController = Em.Controller.extend({
+
+	showMore: Em.computed(function (): boolean {
```
rogatty: Use `showMore: Em.computed('model', function (): boolean {` to recompute on every model change. Or you can be more specific and use `'model.replies', 'model.postCount'`

front/scripts/main/models/PostModel.ts
rogatty: Makes sense but can you add this explanation to a comment so everybody that reads the code can see it and not have to find this PR?

front/scripts/main/models/PostModel.ts
```diff
@@ -1,27 +1,72 @@
...
+	wikiId: null,
+	threadId: null,
+	pivotId: null,
+	replyLimit: 10,
 	replies: [],
 	firstPost: null,
 	upvoteCount: 0,
-	postCount: 0
+	postCount: 0,
+	page: null,
```
rogatty: Below you do `page: this.page+1` so maybe this should be 0 instead?

front/scripts/main/components/PostReplyComponent.ts
garthwebb: I think we still need to work out what's needed.  Certainly the line below needs to go; its a hard coded user ID for us to use until we get user information worked out.  This will get checked and covered by a task that ties the whole post+reply page together.

mklucsarits: This is irrelevant. We'll no longer be calling UserModel in these components to fetch user data, because the user data will be included in the discussion API response. I've already made those changes in https://github.com/Wikia/mercury/pull/1216 (which will cause a conflict with this change).

front/scripts/main/controllers/DiscussionForumPostController.ts
garthwebb: Haha :)

rogatty:
LGTM

bkoval:
LGTM :santa:

front/templates/main/discussion/forum/post.hbs
garthwebb: Its past the LGTM mark, but I've reconsidered this;  I think its better to be explicit about what a data a component needs.  Also, if the component is force internally to do things like {{item.creationDate.epochSecond}}, we can never reuse this component elsewhere without that code having to know how the first usage of this component structured its ```item``` data.

PR #1227, opened 2015-08-17
CONCF-956 Add translations for Curated Content Tool

bognix:
https://wikia-inc.atlassian.net/browse/CONCF-956


vforge:
LGTM

front/templates/main/components/curated-content-editor.hbs
```diff
@@ -2,43 +2,38 @@
...
 {{curated-content-editor-block
 	model=model.optional
 	block='optional'
-	title='Optional'
-	description='Drag each item into the order you prefer. Tap on the item to reveal configuration options.'
-	helpText='Lorem ipsum, this is a placeholder for the help text.'
-	addItemLabel='+ add category'
+	title=(i18n 'app.curated-content-editor-optional')
+	description=(i18n 'app.curated-content-editor-optional-description')
+	helpText=(i18n 'app.curated-content-editor-block-help-text')
```
rogatty: I thought that we will have 3 different texts for help, depending on the block. If not then it doesn't make sense to pass the same thing 3 times. We can just put it in `curated-content-editor-block.hbs` instead of `{{helpText}}`.

rogatty:
LGTM

PR #1230, opened 2015-08-17
SOC-601 | Set cookie on successful account registration

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-601

### DIFF
--- a/front/scripts/auth/common/Cookie.ts
+++ b/front/scripts/auth/common/Cookie.ts
@@ -1,12 +1,19 @@
+interface CookieAttributes {
+	expires?: string;
+	path?: string;
+	domain?: string;
+	secure?: boolean
+}
+
 class Cookie {
-	static get (cookieName: string): string {
+	static get (name: string): string {
 		var cookie = document.cookie,
 			cookieStart: number,
 			cookieEnd: number;
 		if (cookie.length) {
-			cookieStart = cookie.indexOf(cookieName + '=');
+			cookieStart = cookie.indexOf(name + '=');
 			if (cookieStart != -1) {
-				cookieStart = cookieStart + cookieName.length + 1;
+				cookieStart = cookieStart + name.length + 1;
 				cookieEnd = cookie.indexOf(';', cookieStart);
 				if (cookieEnd == -1) {
 					cookieEnd = cookie.length;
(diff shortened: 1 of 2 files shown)


front/scripts/auth/signup/SignupForm.ts
```diff
@@ -66,6 +65,8 @@ class SignupForm {
 			label: 'successful-registration'
 		});
 
+		document.cookie = 'registerSuccess=1; path=/; domain=' + this.getWikiaDomain();
```
garthwebb: Would Cookie.set be good to add to ```front/scripts/auth/common/Cookie.ts``` ?

garthwebb:
LGTM

front/scripts/auth/common/Cookie.ts
```diff
@@ -1,12 +1,19 @@
+interface CookieAttributes {
+	expires?: string;
+	path?: string;
+	domain?: string;
```
bkoval: AFAIK we are supposed to start introducing alphabetical order to interface properties.

bkoval:
LGTM :palm_tree:

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -49,13 +49,12 @@ class SignupForm {
 		};
 	}
 
-	private trackValidationErrors(errors: Array<string>): void {
```
rafalkalinski: Does it should be removed?

kvas-damian: It's already present in: https://github.com/Wikia/mercury/blob/SOC-601/front/scripts/auth/common/FormErrors.ts#L73

front/scripts/auth/signup/SignupForm.ts
```diff
@@ -66,6 +65,14 @@ class SignupForm {
 			label: 'successful-registration'
 		});
 
+		Cookie.set(
+			'registerSuccess',
+			'1',
+			{
+				domain: this.getWikiaDomain()
```
rafalkalinski: if there will not be a redirect parameter (leading to the same origin), then user is redirected to the www.wikia.com, and then the cookie will not be accessible there

rafalkalinski:
LGTM

PR #1236, opened 2015-08-18
Separate localSettings file for dev environment

kvas-damian:
It's time to separate ```localSettings``` for dev environment.

**Action required for devs:**
In order to use dev ```localSettings``` you need to update your ```localSettings.ts``` file to target new settings file.

Ping @rogatty @bkoval @rafalkalinski

### DIFF
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -92,7 +92,7 @@ var localSettings: LocalSettings = {
 		tv: '#00b7e0'
 	},
 	weppy: {
-		enabled: true,
+		enabled: process.env.WIKIA_ENVIRONMENT === 'prod',
 		host: 'http://speed.wikia.net/__rum',
 		samplingRate: 0.1,
 		aggregationInterval: 1000
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -19,7 +19,7 @@ var localSettings: LocalSettings = {
 	// Targeted environment [prod|preview|verify|dev|testing]
 	environment: Utils.getEnvironment(process.env.WIKIA_ENVIRONMENT),
 	helios: {
-		host: 'https://services.wikia.com/auth',
+		path: '/auth',
 		usernameMaxLength: 50,
 		passwordMaxLength: 50
 	},
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -11,7 +11,7 @@ var deepExtend: any = require('deep-extend');
 
 var localSettings: LocalSettings = {
 	apiBase: '/api/v1',
-	servicesDomain: process.env.WIKIA_ENVIRONMENT === 'prod' ? 'services.wikia.com' : 'services.wikia-dev.com',
+	servicesDomain: 'services.wikia.com',
 	// Default timeout for backend requests
 	// This timeout is the same as the MW app timeout
 	backendRequestTimeout: 300000,
@@ -19,7 +19,7 @@ var localSettings: LocalSettings = {
 	// Targeted environment [prod|preview|verify|dev|testing]
 	environment: Utils.getEnvironment(process.env.WIKIA_ENVIRONMENT),
 	helios: {
-		host: process.env.HELIOS_HOST,
+		host: 'https://services.wikia.com/auth',
 		usernameMaxLength: 50,
 		passwordMaxLength: 50
 	},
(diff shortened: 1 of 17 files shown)


bkoval:
LGTM :package: :santa: :jack_o_lantern:

server/facets/operations/prepareArticleData.ts
```diff
@@ -58,10 +58,13 @@ function prepareArticleData (request: Hapi.Request, result: any): void {
 	}
 
 	result.weppyConfig = localSettings.weppy;
+
 	if (typeof result.queryParams.buckysampling === 'number') {
 		result.weppyConfig.samplingRate = result.queryParams.buckysampling / 100;
 	}
 
+	result.localSettings = localSettings;
```
rogatty: Are we sure that there are no secrets in `localSettings`? There is something like `ironSecret` but I don't see it being used anywhere. Until this change we were explicitly saying what should be passed so this is a somewhat breaking change.

Also if we decide that this is fine then you could change things like `result.weppyConfig = localSettings.weppy;` few lines above.

kvas-damian: I don't know if secrets are stored in ```localSettings``` now, but I believe we'll add them there in future. On the other hand passing this variable to view file doesn't mean it's public.

I'm going to cleanup ```weppyConfig``` when we agree that it's fine ;)

config/localSettings.base.ts
```diff
@@ -19,7 +19,7 @@ var localSettings: LocalSettings = {
 	// Targeted environment [prod|preview|verify|dev|testing]
 	environment: Utils.getEnvironment(process.env.WIKIA_ENVIRONMENT),
 	helios: {
-		host: process.env.HELIOS_HOST,
+		host: 'https://services.wikia.com/auth',
```
hakubo: should this be 
``` 
host: servicesDomain + '/auth'
```
or
```
path: '/auth'
```
and be calculated in run time?

hakubo:
BTW, should we rename those files to 'config'
I don't like the 'local' name in them as they are being used on all different machines now.

server/views/_partials/open-graph-tags.hbs
```diff
@@ -10,6 +10,6 @@
 {{#if openGraph.image}}
 	<meta property="og:image" content="{{openGraph.image}}" />
 {{/if}}
-{{#if wiki.facebookAppId}}
```
hakubo: should we remove it from API response then?

config/localSettings.dev.ts
```diff
@@ -0,0 +1,30 @@
...
+var localSettings = baseLocalSettings.getSettings({
+	loggers: {
+		default: 'debug'
+	},
+	authCookieDomain: '.wikia-dev.com',
+	servicesDomain:'services.wikia-dev.com',
+	helios: {
+		host: 'https://services.wikia-dev.com/auth',
+	},
+	facebook: {
```
hakubo: We should think how can we share config easily between app and mercury

I think we'll have to rewrite some stuff in config repo to json
and create dependency in mercury on config repo

hakubo: this can be done as a separate PR though

hakubo:
Thanks, that make sense.

Ideally we'll move to consul to store configuration for mercury as well

server/views/_partials/open-graph-tags.hbs
kvas-damian: It is: https://github.com/Wikia/app/issues/8126

kvas-damian:
Having those settings in consul makes it almost impossible to start development by contributors that do not have access to our consul. There is the same issue with storing it in chef. 

While all those settings are already public (visible on frontend) I think we could created shared public repo for both app and mercury projects where we could store settings that are common. Then load them in both projects. I agree this should be handled by separate PR.

server/views/_partials/open-graph-tags.hbs
hakubo: ACK :)

kvas-damian:
According to: https://github.com/Wikia/mercury/pull/1236#issuecomment-132215122
I agree, but I'd like to avoid to many changes at once.

server/views/_partials/open-graph-tags.hbs
kvas-damian: FIN ;)

lizlux:
LGTM

rogatty:
LGTM

server/facets/operations/prepareMainPageData.ts
```diff
@@ -60,9 +60,11 @@ function prepareMainPageData (request: Hapi.Request, result: any): void {
 		}
 	}
 
-	result.weppyConfig = localSettings.weppy;
-	if (typeof result.queryParams.buckySampling === 'number') {
-		result.weppyConfig.samplingRate = result.queryParams.buckySampling / 100;
+	// clone object to avoid overriding real localSettings for futurue requests
+	result.localSettings = JSON.parse(JSON.stringify(localSettings));
```
rogatty: Is it better to use `JSON` or `deepExtend` maybe?

rogatty:
:+1:

PR #1242, opened 2015-08-19
Update README.md

kvas-damian:


README.md
```diff
@@ -9,15 +9,16 @@
 * Copy `config/localSettings.example.ts` to your own copy of `localSettings.ts` and set the `port` and `devboxDomain`.
   * The devboxDomain must have your devbox name (without the dev- prefix) in it.
   * The `wikiFallback` is useful but is not obligatory.
+  * For dev machines use `localSettings.dev` instead `localSettings.base`
```
rogatty: There is nothing said about `localSettings.base` before so this line doesn't make a lot of sense to me. How I understand it is that the default base file for dev is `localSettings.dev` and `localSettings.base` will be used on sandboxes / production machines only, regular developer shouldn't touch it.

Altough it would be nice to have an explanation what is `localSettings.dev` and what is `localSettings.base` and in what conditions each is being used.

PR #1243, opened 2015-08-19
Setting :7000 as a default port on dev env

bkoval:
@rogatty @hakubo @bognix @kvas-damian 

Let's make use of a new config architecture to revert your revert of this change :)

### DIFF
--- a/config/localSettings.dev.ts
+++ b/config/localSettings.dev.ts
@@ -20,6 +20,8 @@ var localSettings = baseLocalSettings.getSettings({
 	qualaroo: {
 		scriptUrl: '//s3.amazonaws.com/ki.js/52510/dlS.js'
 	},
+	//Default option on devboxes
+	port: 7000
 });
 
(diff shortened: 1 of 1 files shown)


config/localSettings.dev.ts
```diff
@@ -20,6 +20,8 @@ var localSettings = baseLocalSettings.getSettings({
 	qualaroo: {
 		scriptUrl: '//s3.amazonaws.com/ki.js/52510/dlS.js'
 	},
+	//Default option on devboxes
```
rogatty: All these options are default ones for dev environment ;)

bkoval: I know, I meant that this is the port that is used on devboxes for finding mercury service. Anyway, I removed a comment since it was not clear, but the port number itself seems clear enough :)

kvas-damian:
LGTM

hakubo:
LGTM

vforge:
LGTM

PR #1246, opened 2015-08-19
mvp implementation of importing web components

kenkouot:
This is an initial integration of the new style-guide top-bar. It's only showing up on the discussions landing page for now, as it's a WIP and so is discussions. We'll continue to iterated on this and replace hard-coded strings.


lizlux:
@garthwebb @mklucsarits could you give this PR a look?

bower.json
```diff
@@ -17,9 +17,11 @@
     "weppy": "Wikia/weppy#0.9.2",
     "ember-performance-sender": "1.1.1",
     "vignette": "wikia/vignette-js#2.1.1",
-    "wikia-style-guide": "Wikia/style-guide#1.0.2",
+    "wikia-style-guide": "Wikia/style-guide#1.0.3",
     "numeral": "~1.5.3",
-    "cropper": "~0.11.0"
+    "cropper": "~0.11.0",
+    "polymer": "Polymer/polymer#~1.1.0",
+    "webcomponentsjs": "~0.7.11"
```
mklucsarits: What is the relationship between polymer and webcomponentsjs? I thought webcomponentsjs was a polyfill that replaced polymer, making polymer not needed. Why are both needed?

gulp/tasks/build-vendor.js
```diff
@@ -13,7 +13,9 @@ var gulp = require('gulp'),
 
 gulp.task('build-vendor', ['vendor'], function () {
 	return piper(
-		gulp.src(['www/front/vendor/script.js/dist/script.js',
+		gulp.src([
+			'www/front/vendor/webcomponentsjs/webcomponents-lite.js',
+			'www/front/vendor/script.js/dist/script.js',
```
mklucsarits: This is a really generic name for a library. What does it do?

bower.json
garthwebb: Looks like polymer is based on webcomponents? (http://webcomponents.org/, scroll down to 'libraries')

gulp/tasks/build-vendor.js
lizlux: script loading. Not sure what it's being used for here. @kenkouot?

bower.json
lizlux: @garthwebb is correct, we need webcomponents.js for polymer.

PR #1254, opened 2015-08-20
INT-155: Add an option to turn off "upload a photo" from mobile

mari-mcmurtrie:
We want to add a configuration to turn on/off upload photo feature on Mercury. The config variable comes from backend (See: https://github.com/Wikia/app/pull/8168 and https://github.com/Wikia/config/pull/1249 for related backend change). Basically, if we disable upload photo, I add additional class to the element so that I will not display the photo icon.


front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -3,9 +3,14 @@
...
 App.ArticleContributionComponent = Em.Component.extend({
 	classNames: ['contribution-container'],
+	classNameBindings: ['enableUpload::no-photo'],
 	templateName: 'components/article-contribution',
 	section: null,
 	title: null,
+	uploadFeatureEnabled: null, //function
+	enableUpload: Em.computed('uploadFeatureEnabled', function() {
+		return this.get('uploadFeatureEnabled')();
+	}),
```
mari-mcmurtrie: So uploadFeatureEnabled() is function and enableUpload is boolean. I need a value as boolean to bind it to template. I am not sure if this is a right way to do. Any opinion?

mari-mcmurtrie:
@lizlux this is the front-end code of the other PL you reviewed (https://github.com/Wikia/app/pull/8168). I added 1 note since I am not 100% confident about my implementation there. Thanks for CR!

front/scripts/main/components/ArticleWrapperComponent.ts
```diff
@@ -70,6 +70,10 @@ App.ArticleWrapperComponent = Em.Component.extend(App.LanguagesMixin, App.TrackC
 		return !this.get('model.isMainPage') && this.get('isJapaneseWikia');
 	}),
 
+	uploadFeatureEnabled: function(): boolean {
```
rogatty: Wrap it into `Em.computed(` and it won't be a function anymore.

front/scripts/main/components/ArticleContributionComponent.ts
rogatty: After https://github.com/Wikia/mercury/pull/1254/files#r37614667 you won't need this property at all.

mari-mcmurtrie:
Hi @lizlux , Igor is on vacation and I changed code as he reviewed. Will you CR when you have a chance? Thanks!

lizlux:
LGTM

PR #1257, opened 2015-08-21
movnig polymer asset loading outside of main app for now

lizlux:
I noticed that polymer is loading a few different scripts, causing multiple requests. Since we're not going to release the top-bar component just yet, I moved asset loading to within the discussions pages to we don't pollute production. 

@kenkouot


kenkouot:
LGTM

PR #1259, opened 2015-08-21
INT-162: Featured commmunity panel for Japan Homepage. Content updates

perjg:


perjg:
Will merge on Tuesday unless there are any protests.

PR #1261, opened 2015-08-21
CONCF-1059 | Style Curated Content header

vforge:


front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,68 @@
...
+	align-items: center;
+	background: $color-blue-light;
+	color: $white;
+	display: flex;
+	height: $sub-head-height;
+	left: 0;
+	margin: 0 auto;
+	right: 0;
+	top: 0;
+	z-index: $z-7;
```
hakubo: 7?

front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,68 @@
...
+
+	&.fixed {
+		position: fixed;
+	}
+}
+
+/*a*/
```
hakubo: a?

front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,68 @@
...
+.sub-head--cancel {
+	align-items: center;
+	background: inherit;
+	cursor: pointer;
+	display: flex;
+	height: $sub-head-height;
+	justify-content: center;
+	order: 1;
```
hakubo: is it needed?
can we just lay our HTML in such a way that this would not be needed?

front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,68 @@
...
+	white-space: nowrap;
+}
+
+/*button*/
+.sub-head--done {
+	background: inherit;
+	border: 0;
+	color: inherit;
+	cursor: pointer;
+	font: inherit;
```
hakubo: isin't inherit a default?

front/styles/main/component/_sub-head.scss
rogatty: Use https://github.com/Wikia/mercury/blob/dev/front/styles/main/_wikia-variables.scss

front/styles/main/component/_sub-head.scss
vforge: Nope, as we're overriding button styles - that's what comment above this class means. Do you think It can be done differently?

front/styles/main/component/_sub-head.scss
vforge: We can, but with `order` we don't have to. Basically we're removing *need* to lay out HTML in some arbitrary manner that may be omitted in the future - by mistake or just by accident.

front/styles/main/component/_sub-head.scss
vforge: It's a comment that we're expecting `<a>` element here - and we're overriding specific styles.

front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,70 @@
...
+	background: inherit;
+	cursor: pointer;
+	display: flex;
+	height: $sub-head-height;
+	justify-content: center;
+	order: 1;
+	width: $sub-head-height;
+
+	svg {
+		fill: #fff;
```
bognix: $white?

front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,70 @@
...
+	top: 0;
+	z-index: $z-site-head;
+
+	&.fixed {
+		position: fixed;
+	}
+}
+
+/*a*/
+.sub-head--cancel {
```
Warkot: Why the `--` not `-`?

front/styles/main/component/_sub-head.scss
Warkot: We could fix it then.

front/styles/main/component/_sub-head.scss
Warkot: `$white`

front/styles/main/component/_sub-head.scss
```diff
@@ -0,0 +1,70 @@
...
+	background: inherit;
+	border: 0;
+	color: inherit;
+	cursor: pointer;
+	font: inherit;
+	height: inherit;
+	margin: 0;
+	order: 3;
+	padding: 0 13px;
+	text-transform: inherit;
```
bognix: it's inheriting text transform from where?

front/svg/symbols/main/back-arrow.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16px" height="16px" viewBox="0 0 16 16" version="1.1">
```
bognix: do we need this sizes if they are not used anywhere? what is more they are defined one more time here: https://github.com/Wikia/mercury/pull/1261/files#diff-aaa1311c6d796719f1872334d728c5baR33

front/styles/main/component/_sub-head.scss
vforge: from button - https://github.com/Wikia/mercury/pull/1261#discussion_r37626172

front/styles/main/component/_sub-head.scss
vforge: as it's part of the component  and we we're separating component name's words with a single dash we can use double dash or underscore here to make a difference - and I think double dash is better in this context

front/styles/main/component/_sub-head.scss
vforge: As we're supposed to separate content layer from presentation layer - why we should remove possibility to have free-layed HTML just to remove three attributes?

front/styles/main/component/_sub-head.scss
Warkot: Makes sense, thx for the info. :+1:

front/svg/symbols/main/back-arrow.svg
vforge: We don't need it - but those are part of the standard and are telling us how large is image - `viewBox` is doing something else. We need to define it's size once more as we don't use `svg` directly but via `use` directive which ignores `svg` size.

front/svg/symbols/main/back-arrow.svg
```diff
@@ -0,0 +1,5 @@
...
+  <g transform="translate(-503.000000, -56.000000)">
```
Warkot: I can see some room for optimization here — these values above could be directly applied to the coordinates in the `<path/>` below.

front/styles/main/component/_sub-head.scss
hakubo: ok, sounds good to me

front/styles/main/component/_sub-head.scss
hakubo: mkay
probably we can do it better, but we have to finally come up with a standard way of writing scss

hakubo:
LGTM

front/svg/symbols/main/back-arrow.svg
Warkot: I used https://github.com/svg/svgo recommended to me by @hakubo — now I recommend it too. :smiley:

Warkot:
LGTM

tomasznapieralski:
LGTM

PR #1268, opened 2015-08-24
SOC-1182 Use trackPageView for impressions of Auth Pages

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-1182

@kvas-damian @rafalkalinski


front/scripts/auth/tracking.ts
```diff
@@ -26,6 +26,8 @@
 		dimensions[4] = 'mercury';
 		// LoginStatus
 		dimensions[5] = 'anon';
+		//Page type
+		dimensions[8] = 'authPage';
```
bkoval: @kvas-damian @rafalkalinski We could use this dimension to show a type of a more specific auth page (e.g. 'join-page' on /join, 'signin-page' on /signin) but it feels like this should be more general flag, right? What do you think?

front/scripts/auth/common/AuthTracker.ts
```diff
@@ -16,9 +16,10 @@ class AuthTracker {
 	}
 
 	public trackPageView (pageType: string) {
-		if (pageType) {
-			this.track(pageType, M.trackActions.impression);
-		}
+		M.trackPageView({
+			a: null,
+			n: null
+		});
```
kvas-damian: It looks that by default we don't send info about current page to internal tracking system.

bkoval: @rogatty @hakubo How do you handle that in the main mercury app? Even if we add "a" and "n" parameters, the produced url looks like:
https://beacon.wikia-services.com/__track/view?ga_category=view&c=80433&x=wikiaglobal&lc=en&u=0&s=mercury&beacon=&cb=54508
and doesn't contain any information about a page (except referrer HTTP header, which might not be supported in all browsers)

front/scripts/auth/tracking.ts
```diff
@@ -26,16 +26,18 @@
...
 		// IsCorporatePage
 		dimensions[15] = 'No';
 		Mercury.Modules.Trackers.UniversalAnalytics.setDimensions(dimensions);
 	}
 
 	function setTrackingForSignInPage (): void {
-		var tracker = new AuthTracker('user-login-mobile');
+		var tracker = new AuthTracker('user-login-mobile', 'signin-page');
```
kvas-damian: urls should start from ```/```

front/scripts/mercury/utils/track.ts
```diff
@@ -163,7 +163,7 @@ module Mercury.Utils {
 	 *
 	 * trackPageView is called in ArticleView.onArticleChange
 	 */
-	export function trackPageView (adsContext: any) {
+	export function trackPageView (adsContext: any = context) {
```
kvas-damian: ```adsContext``` vs ```context```, few lines below we make a check which one to use. It's not clear for me what's going on here.

front/scripts/auth/tracking.ts
```diff
@@ -155,18 +162,52 @@
...
-		}
+		trackingSets = {
+			'join-page': setTrackingForJoinPage,
+			'signin-page': setTrackingForSignInPage,
+			'register-page': setTrackingForRegisterPage,
+			'fb-connect-page': setTrackingForFBConnectPage,
+			'register-fb-page': setTrackingForFBRegisterPage
+		};
+
+		Object.keys(trackingSets).some(function (pageType) {
```
kvas-damian: Can we avoid making DOM Queries for every defined page type?

kvas-damian:
LGTM

PR #1273, opened 2015-08-24
Hg 843

lizlux:
I'll admit I'm not 100% sure why this broke, but now instead of aborting the transition when there's a 404 we're transitioning to the `notFound` route instead.


kenkouot:
LGTM

PR #1274, opened 2015-08-24
Move ads out of application template

mklucsarits:
This change moves ad slots out of the main application template. The purpose of this change is to allow content on pages using ```application.hbs``` to easily present content that is flush at the top of the page with the site header, allowing them to instead place ads underneath or wherever they need to go, depending on each page. This was primarily motivated by development of Discussions, which has its own forum header that goes directly underneath the site header. A positive side effect of this change is that the margin above the ad slots no longer depends on the height of site header.

I've added the ad slots for articles into ```article.hbs```. I wasn't sure if ad slots are needed in any of the other main templates. If they are, please say so.

@nandy-andy @hakubo You'll both want to look at this one.

### DIFF
--- a/front/styles/main/module/_ads.scss
+++ b/front/styles/main/module/_ads.scss
@@ -14,7 +14,7 @@
 
 .mobile-top-leaderboard {
 	min-height: 50px;
-	margin-top: 60px;
+	margin-top: 10px;
 	position: relative;
 	z-index: $z-ads-top-leaderboard;
 }
--- a/front/templates/main/application.hbs
+++ b/front/templates/main/application.hbs
@@ -31,8 +31,6 @@
 		shouldBeVisible=userMenuVisible
 		toggleVisibility='toggleUserMenu'
 	}}
-	{{ad-slot name='INVISIBLE_HIGH_IMPACT' noAds=noAds}}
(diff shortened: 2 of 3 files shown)


nandy-andy:
The code changes looks good for me. Let me ran some selenium tests of ours on this branch to see if simple code changes don't cause any regression. I filled [ADEN-2365](https://wikia-inc.atlassian.net/browse/ADEN-2365) for it.

nandy-andy:
@mklucsarits sorry it tooks so long. Today @PiotrGackowski ran our tests against your branch and did a little bit of manual testing. All looks good on our side :+1:

mklucsarits:
Thank you @nandy-andy and @PiotrGackowski :+1:

tomasznapieralski:
unfortunately this breaks ad top leaderboard on curated content :/
http://qa-s2.wikia-prod:8080/view/Mercury-Release/job/Mercury-Release-TESTS/261/artifact/logs/log.html

mklucsarits:
@tomasznapieralski That just means I have to add the ad slots on those templates. I only re-added the ad slots on the article template, because I didn't know which other pages they were currently displayed on aside from that one.

tomasznapieralski:
@mklucsarits i know but we couldn't deploy broken code to production, that is why i reverted your changes. Please add it also to curated content and ping me, so i can test it before merge, ok?

PR #1282, opened 2015-08-26
Change mainPage route to the MW main page article route

kenkouot:
This PR changes the curated mobile main pages static route from `/` to the legacy MW style route `/wiki/{articleName}`. This is a temporary change intended to fix a P2 (MAIN-5225) until a longer term solution is available.


lizlux:
LGTM

PR #1288, opened 2015-08-27
HG-850 Implemented inifinite scroll for post list page

garthwebb:
https://wikia-inc.atlassian.net/browse/HG-850

### DIFF
--- a/front/scripts/main/components/ForumWrapperComponent.ts
+++ b/front/scripts/main/components/ForumWrapperComponent.ts
@@ -35,7 +35,7 @@ App.ForumWrapperComponent = Em.Component.extend({
 		});
 	}),
 
-	// we check if we are at the bottom of the page
+	// Check if scrolling should trigger fetching new posts
 	isScrolledToTrigger: function() {
 		var windowHeight = $(window).height(),
 			triggerDistance = 0.25*windowHeight,
--- a/front/scripts/main/components/ForumWrapperComponent.ts
+++ b/front/scripts/main/components/ForumWrapperComponent.ts
@@ -4,7 +4,7 @@
 App.ForumWrapperComponent = Em.Component.extend({
 	classNames: ['forum-wrapper'],
 
-	numPosts: 0,
+	postsDisplayed: 0,
 	totalPosts: 0,
 	pageNum: 0,
 	loadingMore: false,
@@ -12,7 +12,7 @@ App.ForumWrapperComponent = Em.Component.extend({
 	actions: {
 		goToPost: function (postId: number): void {
 			this.sendAction('goToPost', postId);
-		},
+		}
 	},
 
 	didScroll: function() {
@@ -26,19 +26,17 @@ App.ForumWrapperComponent = Em.Component.extend({
 	},
 
 	hasMore: function() {
-		return this.totalPosts > this.numPosts;
+		return this.totalPosts > this.postsDisplayed;
 	},
 
-	pageLoaded: Ember.observer('numPosts', function() {
-		this.setProperties({
-			loadingMore: false
-		});
+	pageLoaded: Ember.observer('postsDisplayed', function() {
+		this.set('loadingMore', false);
 	}),
 
 	// Check if scrolling should trigger fetching new posts
 	isScrolledToTrigger: function() {
 		var windowHeight = $(window).height(),
-			triggerDistance = 0.25*windowHeight,
+			triggerDistance = 0.25 * windowHeight,
 			distanceToViewportTop = $(document).height() - windowHeight,
 			viewPortTop = $(document).scrollTop();
 
--- a/front/scripts/main/components/ForumWrapperComponent.ts
+++ b/front/scripts/main/components/ForumWrapperComponent.ts
@@ -1,0 +1,55 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.ForumWrapperComponent = Em.Component.extend({
+	classNames: ['forum-wrapper'],
+
+	numPosts: 0,
+	totalPosts: 0,
+	pageNum: 0,
+	loadingMore: false,
+
+	actions: {
+		goToPost: function (postId: number): void {
+			this.sendAction('goToPost', postId);
+		},
+	},
+
+	didScroll: function() {
+		if (this.hasMore() && !this.loadingMore && this.isScrolledToTrigger()) {
+			this.setProperties({
+				pageNum: this.pageNum+1,
+				loadingMore: true
+			});
+			this.sendAction('loadPage', this.pageNum);
+		}
+	},
+
+	hasMore: function() {
+		return this.totalPosts > this.numPosts;
+	},
+
+	pageLoaded: Ember.observer('numPosts', function() {
+		this.setProperties({
+			loadingMore: false
+		});
+	}),
+
+	// we check if we are at the bottom of the page
+	isScrolledToTrigger: function() {
+		var windowHeight = $(window).height(),
+			triggerDistance = 0.25*windowHeight,
+			distanceToViewportTop = $(document).height() - windowHeight,
+			viewPortTop = $(document).scrollTop();
+
+		return distanceToViewportTop - viewPortTop < triggerDistance;
+	},
+
+	didInsertElement: function() {
+		$(window).on('scroll', this.didScroll.bind(this));
+	},
+
+	willDestroyElement: function() {
+		$(window).off('scroll', this.didScroll.bind(this));
+	}
+});
--- a/front/scripts/main/models/DiscussionForumModel.ts
+++ b/front/scripts/main/models/DiscussionForumModel.ts
@@ -1,15 +1,61 @@
 /// <reference path="../app.ts" />
 
 App.DiscussionForumModel = Em.Object.extend({
+	wikiId: null,
+	forumId: null,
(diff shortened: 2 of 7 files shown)


front/scripts/main/routes/DiscussionPostRoute.ts
```diff
@@ -27,6 +27,11 @@ App.DiscussionPostRoute = Em.Route.extend({
 			var model = this.modelFor('discussion.post');
 
 			model.loadNextPage();
+		},
+
+		didTransition(): boolean {
+			window.scrollTo(0, 0);
+			return true;
```
garthwebb: The scroll position from the post list was being retained when we went to the post detail page.  This puts scroll back to the top.

front/styles/main/module/_discussion.scss
```diff
@@ -99,6 +104,7 @@
 
 		.post-detail {
 			border-top: 1px solid #ccc;
+			cursor: pointer;
```
garthwebb: Make sure the post detail looks like its clickable when its on the post list page.

front/templates/main/discussion/forum.hbs
```diff
@@ -1,15 +1,9 @@
...
-			content=item._embedded.firstPost.0.rawContent
-			timestamp=item._embedded.firstPost.0.creationDate.epochSecond
-			upvoteCount=item.upvoteCount
-			replyCount=item.postCount
-		}}
-	{{/each}}
-</div>
+{{forum-wrapper
+	goToPost='goToPost'
+	loadPage='loadPage'
```
garthwebb: This is magic.  Without this the action bubbling doesn't seem to work.  @lizlux can back me.

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,55 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.ForumWrapperComponent = Em.Component.extend({
+	classNames: ['forum-wrapper'],
+
+	numPosts: 0,
```
mklucsarits: I think this variable could be better named. It could be confusing with ```totalPosts```. Can I suggest maybe ```postsShown``` or ```postsDisplayed```?

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,55 @@
...
+			this.sendAction('loadPage', this.pageNum);
+		}
+	},
+
+	hasMore: function() {
+		return this.totalPosts > this.numPosts;
+	},
+
+	pageLoaded: Ember.observer('numPosts', function() {
+		this.setProperties({
```
mklucsarits: No need for setProperties() if only setting one prop. Just use: ```this.set('loadingMore', false);```

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,55 @@
...
+			loadingMore: false
+		});
+	}),
+
+	// Check if scrolling should trigger fetching new posts
+	isScrolledToTrigger: function() {
+		var windowHeight = $(window).height(),
```
mklucsarits: In other places we've been calling jQuery through Ember. That is ```$``` is then ```this.$```. However, I see that not all devs have been doing this, so I'm not sure which is preferred or why.

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,55 @@
...
+			triggerDistance = 0.25*windowHeight,
```
mklucsarits: Spaces around operator.

front/scripts/main/models/DiscussionForumModel.ts
```diff
@@ -1,15 +1,61 @@
...
+		return new Em.RSVP.Promise((resolve: Function, reject: Function) => {
+			Em.$.ajax({
+				url: 'https://services.wikia.com/discussion/' + this.wikiId + '/forums/' + this.forumId +
+					 '?page=' + pageNum,
+				dataType: 'json',
+				success: (data: any) => {
+					var newPosts = data._embedded['doc:threads'],
+					    allPosts = this.posts.concat(newPosts);
+
+					this.setProperties({
```
mklucsarits: ```this.set``` can be used instead.

front/styles/main/module/_discussion.scss
```diff
@@ -65,6 +65,11 @@
 		}
 	}
 
+  	.load-more {
+	  width: 100%;
```
mklucsarits: A minor annoying thing, but these are not in alphabetical order. :)

front/styles/main/module/_discussion.scss
mklucsarits: :+1:

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,55 @@
...
+
+	numPosts: 0,
+	totalPosts: 0,
+	pageNum: 0,
+	loadingMore: false,
+
+	actions: {
+		goToPost: function (postId: number): void {
+			this.sendAction('goToPost', postId);
+		},
```
lizlux: extra comma

front/scripts/main/models/DiscussionForumModel.ts
```diff
@@ -1,15 +1,61 @@
 /// <reference path="../app.ts" />
 
 App.DiscussionForumModel = Em.Object.extend({
+	wikiId: null,
+	forumId: null,
+	name: null,
+	posts: null,
+	totalPosts: 0,
+
+	hasMore: function(): boolean {
```
lizlux: looks like this can be removed.

front/scripts/main/models/DiscussionForumModel.ts
```diff
@@ -1,15 +1,61 @@
...
 
 App.DiscussionForumModel.reopenClass({
 	find(wikiId: number, forumId: number) {
 		return new Em.RSVP.Promise((resolve: Function, reject: Function) => {
+			var forumInstance = App.DiscussionForumModel.create({
+				wikiId: wikiId,
+				forumId: forumId
+			});
+
 			Em.$.ajax({
```
lizlux: If you write it like this `Em.$.ajax(<JQueryAjaxSettings>{` WebStorm won't through typescript warnings.

front/templates/main/discussion/forum.hbs
lizlux: :+1:

lizlux:
Just a couple comments but this LGTM.

front/scripts/main/components/ForumWrapperComponent.ts
lizlux: `this.$` will call the jQuery function in the context of the component wrapper element. In this case, we're selecting the `window` object so we want to use just `$`.

front/scripts/main/models/DiscussionForumModel.ts
garthwebb: Hmm, my WebStorm isn't throwing those but I'll put it in anyway

front/styles/main/module/_discussion.scss
garthwebb: I never remember this one ...

mklucsarits:
LGTM

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,53 @@
...
+		if (this.hasMore() && !this.loadingMore && this.isScrolledToTrigger()) {
+			this.setProperties({
+				pageNum: this.pageNum+1,
+				loadingMore: true
+			});
+			this.sendAction('loadPage', this.pageNum);
+		}
+	},
+
+	hasMore: function() {
```
rogatty: This looks like a candidate for computed property.

front/scripts/main/components/ForumWrapperComponent.ts
```diff
@@ -0,0 +1,53 @@
...
+		}
+	},
+
+	didScroll: function() {
+		if (this.hasMore() && !this.loadingMore && this.isScrolledToTrigger()) {
+			this.setProperties({
+				pageNum: this.pageNum+1,
+				loadingMore: true
+			});
+			this.sendAction('loadPage', this.pageNum);
```
rogatty: If you're using `this.set` to set the property then please use `this.get` to get its value. Doing it like this might lead to some hard-to-debug race conditions.

front/scripts/main/components/ForumWrapperComponent.ts
rafalkalinski: @garthwebb 
Actually Igor had right here - it should be computed. Currently we have always true under hasMore property (cos it has a reference to a function), and since it is already on dev - we have neverending spinner spinning in the threads view on every branch that is synced with dev today.

garthwebb: Wait, why is it always true?  Its not a property, its a function, so it will always be evaluated when its called on L19 as a function.

The spinner is always running now (even if not on screen) unless there are no more items because https://github.com/Wikia/mercury/commit/27d7fee8e87a2f2981ab5d5976d88539b6054371

PR #1289, opened 2015-08-28
Hg 837 Dry run of updating discussions with latest version of top-bar

lizlux:
DON'T MERGE

We'll put all the finishing touches on this in https://wikia-inc.atlassian.net/browse/HG-846. This PR is to make sure the latest version of the style guide component can be easily inserted into the mercury app. 

Don't merge this yet b/c this PR is pointing to @mklucsarits's discussions work. Let's create a separate PR to dev once both these tickets are ready to go.


front/templates/main/discussion/index.hbs
```diff
@@ -1,17 +1 @@
-<link rel="import" href="/front/vendor/polymer/polymer.html">
```
mklucsarits: I think this template may be used eventually to display all forums for a wiki. I'm not sure it should be used as an outlet for other templates.

lizlux: I just updated it to say "discussions WIP"

mklucsarits:
LGTM

PR #1293, opened 2015-08-28
Xw 69 show prompt on back click

bognix:
/cc @vforge @hakubo

### DIFF
--- a/front/scripts/main/models/CuratedContentEditorModel.ts
+++ b/front/scripts/main/models/CuratedContentEditorModel.ts
@@ -41,7 +41,8 @@ type CuratedContentEditorModel = typeof App.CuratedContentEditorModel;
 App.CuratedContentEditorModel = Em.Object.extend({
 	featured: null,
 	curated: null,
-	optional: null
+	optional: null,
+	isDirty: false
 });
 
 App.CuratedContentEditorModel.reopenClass({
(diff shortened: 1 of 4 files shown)


front/scripts/main/models/CuratedContentEditorModel.ts
```diff
@@ -41,7 +41,8 @@ type CuratedContentEditorModel = typeof App.CuratedContentEditorModel;
 App.CuratedContentEditorModel = Em.Object.extend({
 	featured: null,
 	curated: null,
-	optional: null
+	optional: null,
+	isDirty: false
```
vforge: Is it really so good idea to set up a global-level flag inside model that we are instancing and re-use? I think that even global variable would be better, but I would really like another `Em.Object` for keeping this state. What @hakubo thinks of this idea?

bognix: on the other hand this flag represents state of this model and this state should be shared globally because model is shared globally

kenkouot: I would shy away from global state or things on the application and instead inject singletons into your classes. http://guides.emberjs.com/v1.10.0/understanding-ember/dependency-injection-and-service-lookup/

bognix: this looks great but this value won't be shared among different objects so I don't see the need to create a special service to replace one line of code so it can store one variable. In m opinion It would make sense to build this service if we have fancier logic which might be shared.

hakubo: it should be available on instance not on a 'class'
but this is good enough for now,

we should reevaluate Ember Data - that has 'isDirty' built in

We can refactor it later
I like that at least this is stored next to data it describes

hakubo:
LGTM

PR #1295, opened 2015-08-28
Hg 854 fixing polldaddy after transition

lizlux:
PollDaddy was working fine when going to the page directly, but navigating to the page via Ember transition wasn't working because the polldaddy script wasn't getting loaded. 

Since I was performing the same script loading functionality that ember-main was doing, I abstracted it into mercury modules.


front/scripts/baseline/mercury/utils/loadScript.ts
```diff
@@ -0,0 +1,20 @@
...
+/// <reference path="../../mercury.d.ts" />
+
+module Mercury.Utils {
+	function loadError (error: ErrorEvent): void {
+		var scriptTag: HTMLScriptElement = <HTMLScriptElement> error.target;
+		throw new URIError("The script " + scriptTag.src + " is not accessible.");
+	}
+
+	export function loadScript(src: string, fOnload: any) {
+		var script = document.createElement("script");
```
kenkouot: single quotes in js

PR #1298, opened 2015-08-31
INT-136: Use mercury signin/signup pages for Japan homepage login

perjg:
Switched login logic for JP homepage to use mercury /signin and /register pages.

Backend work to get helios cookie and query for token is still needed separately.


homepage/server/views/_layouts/default.html
```diff
@@ -13,5 +13,19 @@
 
 <script src="/vendor/jquery/dist/jquery.min.js"></script>
 <script src="/vendor/slick.js/slick/slick.min.js"></script>
+<script>
+  // Globals
+  function getGlobals () {
+    return {
+      {{#if loggedIn}}
```
hakubo: this will disallow us from caching this on varnish easily 
maybe a blocking script tag high in head would work as good?

homepage/server/facets/privacy.js
```diff
@@ -4,14 +4,13 @@
...
  */
 
-var util = require('../util');
+var deepExtend = require('deep-extend'),
+	util = require('../util');
 
 function privacy(request, reply) {
-	var data = {
+	var data = deepExtend({
 		title: 'ウィキア・ジャパン',
```
bkoval: Do you think it makes sense to use deepExtend even if you add only one property (title) to the GlobalData object? Maybe you could just store util.getGlobalData() in a variable and invoke myObject.title = <title> ?

perjg: You're right: There's no real point in using deepExtend in this case.

I changed all facets to use deepExtend regardless of what data is inherited for easy consistency.

homepage/server/views/_layouts/default.html
perjg: I didn't think of caching.

Is it enough to move the inline script into a .js file like so:

<script src="/globals.js"></script>
and adding globals.js as a route to the backend app that generates the correct current state instead of doing it via the template.

PR #1305, opened 2015-08-31
Move ad slots out of main application template

mklucsarits:
There was a previous pull request (https://github.com/Wikia/mercury/pull/1274) for these changes which neglected to include ad slots on the curated main page. That has been fixed here. Description from the previous PR:

"This change moves ad slots out of the main application template. The purpose of this change is to allow content on pages using application.hbs to easily present content that is flush at the top of the page with the site header, allowing them to instead place ads underneath or wherever they need to go, depending on each page. This was primarily motivated by development of Discussions, which has its own forum header that goes directly underneath the site header. A positive side effect of this change is that the margin above the ad slots no longer depends on the height of site header."

*The implication of this change is that any new page template on Mercury using ```application.hbs``` as a wrapper must include its own ad slot components.*


tomasznapieralski:
LGTM. All tests are passing and top leaderboard is displayed on both articles and curated main pages

PR #1307, opened 2015-09-01
SOC-1079 | VisitSource on authpages and mercury app

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1079


bkoval:
Please, resolve merge conflicts.

front/scripts/auth/common/VisitSource.ts
```diff
@@ -0,0 +1,15 @@
+class VisitSourceWrapper {
+	public static sessionVisitSource: any = (typeof VisitSource !== 'undefined') ?
```
bkoval: typeof VisitSource === 'function' would be safer here :)

front/scripts/auth/common/VisitSource.ts
```diff
@@ -0,0 +1,15 @@
...
+		new VisitSource('WikiaSessionSource', pageParams.cookieDomain) :
+		undefined;
+	public static lifetimeVisitSource: any = (typeof VisitSource !== 'undefined') ?
+		new VisitSource('WikiaLifetimeSource', pageParams.cookieDomain, false) :
+		undefined;
+
+	public static init() {
+		if (this.sessionVisitSource && this.sessionVisitSource) {
```
bkoval: probably you meant this.lifetimeVisitSource in the second argument of the logic sum :)

front/scripts/auth/main.ts
```diff
@@ -1,6 +1,7 @@
 /// <reference path="../../../typings/i18next/i18next.d.ts" />
 declare var translations: any;
 declare var language: string;
+declare var VisitSource: any;
```
bkoval: Isn't it a Function?

front/scripts/auth/main.ts
```diff
@@ -49,4 +50,8 @@ window.document.addEventListener('DOMContentLoaded', function ():void {
 	if (birthdateContainer) {
 		new BirthdateInput(birthdateContainer, formElement).init();
 	}
+
+	if (typeof VisitSource !== 'undefined') {
```
bkoval: Why not ===' function'? :)

front/scripts/main/app.ts
```diff
@@ -47,6 +48,16 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'visit-source',
+	initialize () {
+		if (typeof VisitSource !== 'undefined') {
```
bkoval: Same story here :)

front/scripts/auth/common/VisitSourceWrapper.ts
```diff
@@ -0,0 +1,15 @@
...
+		new VisitSource('WikiaSessionSource', pageParams.cookieDomain) :
+		undefined;
+	public static lifetimeVisitSource: any = (typeof VisitSource === 'function') ?
+		new VisitSource('WikiaLifetimeSource', pageParams.cookieDomain, false) :
+		undefined;
+
+	public static init() {
+		if (this.sessionVisitSource && this.lifetimeVisitSource) {
+			this.sessionVisitSource.checkAndStore();
+			this.lifetimeVisitSource.checkAndStore();
```
jsutterfield: Should these be:
```typescript
if (VisitSourceWrapper.sessionVisitSource && VisitSourceWrapper.lifetimeVisitSource) {
    VisitSourceWrapper.sessionVisitSource.checkAndStore();
    VisitSourceWrapper.sessionVisitSource.checkAndStore();
...
```
?

I don't know a lot about typescript, but it looks like we're accessing static variables as if they're instance variables. Or it could be some js voodoo I don't quite understand.

front/scripts/auth/common/VisitSourceWrapper.ts
```diff
@@ -0,0 +1,15 @@
+class VisitSourceWrapper {
+	public static sessionVisitSource: any = (typeof VisitSource === 'function') ?
+		new VisitSource('WikiaSessionSource', pageParams.cookieDomain) :
```
jsutterfield: What do you think about defining this and `WikiaLifetimeSource` as constants somewhere?

front/scripts/main/app.ts
```diff
@@ -47,6 +49,16 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'visit-source',
+	initialize () {
+		if (typeof VisitSource === 'function') {
```
jsutterfield: Are there times when we don't expect VisitSource to exist? ie, can we remove this check and just create/call the VisitSource objects below? If so, could we then remove the similar checks in the VisitSourceWrapper.ts?

kvas-damian: This check is required for unittests. Current unit testing platform allows us to use only compliled files. Unfortunately external libraries aren't loaded there so we need to use this kind of checks.

Fortunately this library has its own unittests.

front/scripts/auth/common/VisitSourceWrapper.ts
kvas-damian: Does it makes sense while it's used only once?

jsutterfield: Well, it's also used [here](https://github.com/Wikia/mercury/pull/1307/files#diff-9d7b84ee40c267f1c24af45132c86c20R55), right? It's a small thing, I'm going to move this over to QA.

jsutterfield:
LGTM - just need to resolve merge conflicts.

front/scripts/auth/common/VisitSourceWrapper.ts
kvas-damian: We're trying to make auth pages independent from mercury frontend JS. Having one constant for both will make it harder to drop mercury dependency at some point.

PR #1217, opened 2015-08-13
SOC-883 Share forum post to social networks

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-883

@lizlux @garthwebb

PR #1309, opened 2015-09-01
Registering discussion i18n namespace to make it visible to the app

bkoval:
@kvas-damian @garthwebb @mklucsarits @rafalkalinski It seems like the discussion namespace is not reachable by the app and we should fix it ASAP since it seems like open-graph needs a label form the translation file.


mklucsarits:
I spent some time dealing with this issue last week. This is fine, for server-side messages, but any front-end messages require more work. Additionally, @lizlux and I discussed this last week while I was tackling the same issue, and we decided it would make more sense to keep discussion messages in main.json, under a discussion section. See this open pull request: https://github.com/Wikia/mercury/pull/1217

PR #1311, opened 2015-09-01
HG-837 Top Bar MVP v3

kenkouot:


bower.json
```diff
@@ -17,11 +17,11 @@
     "weppy": "Wikia/weppy#0.9.2",
     "ember-performance-sender": "1.1.1",
     "vignette": "wikia/vignette-js#2.1.1",
-    "wikia-style-guide": "Wikia/style-guide#1.0.3",
+    "wikia-style-guide": "Wikia/style-guide#v1.3.0",
```
mklucsarits: Why the huge jump in version?

front/templates/main/application.hbs
```diff
@@ -1,3 +1,4 @@
+<link rel="import" href="/front/vendor/wikia&#45;style&#45;guide/dist/components/import&#45;built.html">
```
mklucsarits: No hyphens allowed?

front/templates/main/discussion/index.hbs
```diff
@@ -1,17 +1,2 @@
...
-	user-logged-in="false"
-	show-user-status="true"
-	user-avatar-src="http://img4.wikia.nocookie.net/__cb1372467243/common/avatars/thumb/e/eb/11096997.png/100px-11096997.png.jpg">
-</top-bar>
-
-<h1>Discussions</h1>
-
-<h2>Forums:</h2>
-<ul>
-{{#each _embedded.doc:forum as |item|}}
```
mklucsarits: Please leave this code in so that when we come back to it, we don't have to rewrite this loop to get all the forums from the model.

kenkouot: Whoops, bad auto merge. Will do.

bower.json
kenkouot: Answered offline

mklucsarits:
LGTM

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../main/mixins/HeadroomMixin.ts" />
+/// <reference path="../../baseline/mercury/utils/buildUrl.ts" />
+'use strict';
+
+App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
+	isUserLoggedIn: Em.computed('currentUser.isAuthenticated', function () {
+		// HTMLBars attribute binding only removes an attribute if it's value is set to null
+		if (this.get('currentUser.isAuthenticated') === false) {
```
rogatty: @kenkouot if `this.get('currentUser.isAuthenticated')` is undefined it will return true. Doesn't seem too robust.

kenkouot: It should never be `undefined` as `currentUser.isAuthenticated` is a result of `Em.computed.bool` on `M.prop('userId')`, which will always be at least 0.

kenkouot: TBH, seems like I should add a test more for `CurrentUser` property types than add in defensive programming checks here ^_^

rogatty: You assume that `currentUser` is never broken and always injected correctly. Shouldn't we at least try to protect ourselves from programming errors? Simple
```
return this.get('currentUser.isAuthenticated') === true ? true : null;
```
would do the trick.

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -0,0 +1,32 @@
...
+			return null;
```
hakubo: this should be false
as 'semantically' null is not a value
in other words null means nothing

false would be explicit here

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -0,0 +1,32 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../main/mixins/HeadroomMixin.ts" />
+/// <reference path="../../baseline/mercury/utils/buildUrl.ts" />
+'use strict';
+
+App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
+	isUserLoggedIn: Em.computed('currentUser.isAuthenticated', function () {
+		// HTMLBars attribute binding only removes an attribute if it's value is set to null
```
hakubo: could you shed some more info about that?
doesn't seem right and if so, I think more people would like to know more about that

PR #1316, opened 2015-09-02
DAT-3144 infobox builder

dianafa:
WIP!!!! DO NOT MERGE

https://wikia-inc.atlassian.net/browse/DAT-3144

This PR doesn't contain the visual part yet. Styles are temporary and will probably be changed.

please CR: @hakubo @Warkot @bognix


SebastianMarzjan:
Please address the TS issue highlighter by the check

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.InfoboxBuilderItemDataComponent = Em.Component.extend({
+	tagName: '',
```
bognix: do you need to initiate it with empty value?

dianafa: yes, othervise it will be wrapped in default ember `<div>` with id's and other stuff. Additional div's will break our styling.

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
```diff
@@ -0,0 +1,16 @@
...
+	attributeBindings: ['data-position'],
+	label: Em.computed('data', function(){
```
bognix: how about http://emberjs.com/api/classes/Ember.computed.html#method_alias?

front/scripts/main/components/InfoboxBuilderItemImageComponent.ts
```diff
@@ -0,0 +1,26 @@
...
+		return this.get('data.caption.defaultValue');
+	}),
+	position: Em.computed('infoboxBuilderData', function (){
+		return this.get('infoboxBuilderData.position');
+	}),
+	videoClasses: Em.computed(function (): string {
+		if (this.get('isVideo')) {
+			return 'video video-thumbnail small';
+		}
+		return '';
```
bognix: how about:
```
return this.get('isVideo') ? 'video video-thumbnail small' : ''

File: front/scripts/main/components/InfoboxBuilderWrapperComponent.ts [hunk 4]
```diff
@@ -0,0 +1,9 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.InfoboxBuilderWrapperComponent = Em.Component.extend({
+	tagName: '',
+
+	theme: '',
+	layout: ''
+});
```
bognix: do you need this component?

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,247 @@
...
+				} ,
+				defaultValue: 'path/to/image.jpg',
+			},
+			infoboxBuilderData: {
+				index: i,
+				position: this.get('_stateLength'),
+				component: this.createComponentName(itemType)
+			},
+			source: `image${i}`,
+			type: 'image'
```
bognix: have you considered using here itemType var instead of hardcoded 'image'?

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,247 @@
...
+			},
+			infoboxBuilderData: {
+				index: i,
+				position: this.get('_stateLength'),
+				component: this.createComponentName(itemType)
+			},
+			source: `title${i}`,
+			type: 'title'
+		});
+	},
```
bognix: There is a lot of duplicated code in addTitleItem, addImageItem and addDataItem. The only thing which differs them is data. Is it possible to create one more generic method?

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,247 @@
...
+	type: string;
+}
+
+interface SaveStateToTemplateResponse {
+	success: boolean;
+	errors: any[];
+	warnings: any[];
+}
+
+App.InfoboxBuilderModel = Em.Object.extend(App.ObjectUtilitiesMixin, {
```
bognix: you are not using method from this mixin.

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,247 @@
...
+
+		return JSON.stringify({data: plainState});
+	},
+
+	/**
+	 * increase index for given item type
+	 * @param {String} intexType
+	 * @returns {Number}
+	 */
+	increaseItemIndex(intexType: string): number {
```
hakubo: typo

front/styles/main/component/_infobox-builder.scss
```diff
@@ -0,0 +1,12 @@
+.infobox-builder {
+	margin: auto;
+	max-width: 700px;
+	padding: 50px;
+
+	.infobox-builder-buttons {
```
hakubo: does that have to be child?

dianafa: it's the draft, the whole styling will be changed

front/scripts/main/components/InfoboxBuilderWrapperComponent.ts
dianafa: yes, to define empty `tagName` (explained above)

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
hakubo: consider using
`tagName: false`
I think it a little bit better explains your intention

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,247 @@
...
+
+App.InfoboxBuilderModel = Em.Object.extend(App.ObjectUtilitiesMixin, {
+	_itemIndex: {
+		data: 0,
+		image: 0,
+		title: 0,
+	},
+	infoboxState: Em.A([]),
+	_stateLength: Em.computed('infoboxState', {
+		get(): number {
```
hakubo: get is default

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,247 @@
...
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			Em.$.ajax(<JQueryAjaxSettings>{
+				url: M.buildUrl({
+					path: '/wikia.php'
+				}),
+				data: {
+					controller: 'PortableInfoboxBuilderController',
+					method: 'publish',
+					title: this.get('title'),
+					data: this.prepareStateForSaving(this.get('infoboxState'))
```
hakubo: this.serialize?

front/scripts/main/routes/InfoboxBuilderRoute.ts
```diff
@@ -0,0 +1,230 @@
...
+			var ponto = window.Ponto;
+
+			ponto.setTarget(ponto.TARGET_IFRAME_PARENT, window.location.origin);
+
+			ponto.invoke(
+				'wikia.infoboxBuilder.ponto',
+				'isWikiaContext',
+				null,
+				function (data: any): void {
+					if (data && data.isWikiaContext && data.isLoggedIn) {
```
hakubo: what otherwise?

hakubo:
couple of remarks

nice usage of promises!

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
dianafa: I think it's not we meant to: 
https://www.dropbox.com/s/rt2wnknj5sgducr/Screenshot%202015-09-04%2012.45.44.png?dl=0

front/locales/en/main.json
```diff
@@ -136,6 +136,15 @@
 		"curated-content-editor-missing-items-error": "Sections can't be empty. Please add items to section.",
 		"curated-content-editor-error-inside-items-message": "Please fix errors inside items.",
 		"curated-content-editor-items-count": "__count__ items",
+		"infobox-builder-label-item-default-value": "Label",
```
bkoval: @kenkouot @hakubo @kvas-damian Do we have a standardized way of creating namespaces here? I feel like curated-content-editor and infobox-builder would be perfect candidates for namespaces here.

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
bkoval: Could you add a code comment explaining that is it intentional? :)

front/scripts/main/components/InfoboxBuilderItemTitleComponent.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.InfoboxBuilderItemTitleComponent = Em.Component.extend({
+	tagName: '',
+	attributeBindings: ['data-position'],
+	position: Em.computed('infoboxBuilderData', function(){
```
bkoval: Shouldn't there be spaces around the round brackets when there's a function involved? Also (just asking) should we consider indicating return types of computed properties?

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+App.InfoboxBuilderModel = Em.Object.extend({
+	_itemIndex: {
+		data: 0,
+		image: 0,
+		title: 0,
+	},
+	infoboxState: Em.A([]),
+	_stateLength: Em.computed('infoboxState', function (): number {
```
bkoval: Oh yeah, I got my answer ;)

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,250 @@
...
+						defaultValue: i18n.t('app.infobox-builder-alt-item-default-value'),
+					}
+				},
+				caption: {
+					source: `caption${i}`,
+					data: {
+						defaultValue: i18n.t('app.infobox-builder-caption-item-default-value'),
+					}
+				} ,
+				defaultValue: 'path/to/image.jpg',
```
bkoval: Shouldn't it default to an empty 1x1px image? Or it will be sanitized somewhere and won't trigger a 404-ed request?

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,250 @@
...
+			type: itemType
+		});
+	},
+
+	/**
+	 * add <title> item
+	 * addItem's methods can be refactored when figure out
+	 * stable version of infobox items and params
+	 */
+	addTitleItem() : void {
```
bkoval: Is the space after the colon aligned with guidelines?

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,250 @@
...
+				dataType: 'json',
+				method: 'POST',
+				success: (data: SaveStateToTemplateResponse): void => {
+					if (data && data.success) {
+						resolve(this.get('title'));
+					} else {
+						reject(data.errors);
+					}
+				},
+				error: (data: any): void => {
```
bkoval: Maybe a param here should be called "err"? Since you pass it to the reject method as you do in line 241 it seems like it's going to be an error object anyway :)

front/scripts/main/models/InfoboxBuilderModel.ts
dianafa: it is a draft, we wait for the designs with such an image. Thanks Bart :)

front/scripts/main/routes/InfoboxBuilderRoute.ts
```diff
@@ -0,0 +1,232 @@
...
+
+'use strict';
+
+interface InfoboxBuilderGetAssetsResponse {
+	css: string[];
+}
+
+App.InfoboxBuilderRoute = Em.Route.extend({
+	pontoLoadingInitialized: false,
+	pontoPath: '/front/vendor/ponto/web/src/ponto.js',
```
bkoval: Shouldn't this be loaded via nocookie? @rogatty

front/scripts/main/routes/InfoboxBuilderRoute.ts
```diff
@@ -0,0 +1,232 @@
...
+		});
+	},
+
+	/**
+	 * @desc shows error message for ponto communication
+	 * @param {string} message - error message
+	 */
+	showPontoError(message: any) {
+		if (window.console) {
+			window.console.error('Ponto Error', message);
```
bkoval: Is there any Ember-based logger that could be used here?

front/locales/en/main.json
rogatty: /cc @Warkot

Warkot: We don't have yet, but we probably should. We have over 40 CC Edit Tool strings. So maybe even a separate file would be a good way to handle it.

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
rogatty: `tagName: ''` is the only way to create tagless components in Ember, altough not officialy documented.

In this case we could instead remove `<div class="pi-item pi-data pi-item-spacing pi-border-color" data-position="{{position}}">` wrapper from the template and set those attributes here in JS but we want to keep templates consistent between mercury and app repos.

@dianafa @RafLeszczynski please add a comment explaining this.

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.InfoboxBuilderItemDataComponent = Em.Component.extend({
+	tagName: '',
+	attributeBindings: ['data-position'],
```
rogatty: It would make sense only if it wasn't a tagless component.
http://emberjs.com/api/classes/Ember.View.html#toc_html-attributes

front/scripts/main/components/InfoboxBuilderItemTitleComponent.ts
rogatty: Yes and yes.

front/scripts/main/components/InfoboxBuilderWrapperComponent.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.InfoboxBuilderWrapperComponent = Em.Component.extend({
+	tagName: '',
+
+	theme: '',
+	layout: ''
```
rogatty: This is a reserved property name, see http://emberjs.com/api/classes/Ember.View.html#property_layout

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+interface SaveStateToTemplateResponse {
+	success: boolean;
+	errors: any[];
+	warnings: any[];
+}
+
+App.InfoboxBuilderModel = Em.Object.extend({
+	_itemIndex: {
```
rogatty: Why the underscore?

front/scripts/main/router.ts
```diff
@@ -58,6 +58,10 @@ App.Router.map(function () {
 		});
 	});
 
+	this.route('infoboxBuilder', {
+		path: '/infoboxBuilder/:templateName'
```
rogatty: Consider using `/infobox-builder/` path.

front/scripts/main/routes/InfoboxBuilderRoute.ts
rogatty: @bkoval eventually yes, but it's fine for now as this route will be loaded rarely, by admins. This is copied straight from https://github.com/Wikia/mercury/blob/dev/front/scripts/main/routes/CuratedContentEditorRoute.ts#L77

We will tackle this problem in https://wikia-inc.atlassian.net/browse/XW-134 and try to create some kind of resource loader that will take care of that.

bkoval: Awesome!

front/scripts/main/router.ts
dianafa: why?

front/scripts/main/models/InfoboxBuilderModel.ts
dianafa: why?

front/scripts/main/models/InfoboxBuilderModel.ts
dianafa: to mark that this var is only for the internal use. Does this violate our guidelines?

front/scripts/main/router.ts
rogatty: Because it's the standard in web apps. Some reading about it:

http://stackoverflow.com/a/18450653/1050577
http://ux.stackexchange.com/a/9304

front/scripts/main/components/InfoboxBuilderWrapperComponent.ts
dianafa: true that! thank you!

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
dianafa: yes, we want to keep the templates between app and mercury repo the most consistent as possible. Thus, we decided to not change the tag names and bind the classes, but instead remove added by default ember div, and use our code.
We don't want to keep the default divs because of styling which can break because of this.

front/scripts/main/router.ts
kenkouot: +1 @rogatty Furthermore this is all over the place right now. Some query params are camelCased, others are lowercase. In the future URI proposal guidelines, it's hyphenated.

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
kenkouot: I understand the reason why you are doing this but I'm not in favor of this approach.`tagName: ''` and it's alternative, inheriting from `_MetamorphView` are undocumented private APIs, which we should not be using as there is no official support for relying on this functionality. This feels a lot like tribal knowledge that prevents us from hiring off-the-shelf Ember devs and is symptomatic of why it's so painful to work in the Wikia/app codebase today.

Also, I'm still unclear as to why most of these components have a empty tag name exactly. You need to keep the templates mirrored between Wikia/app and Wikia/mercury for what reasons exactly?

kenkouot: Here is a thread where a Em core team member confirms that this API is private: https://github.com/emberjs/ember.js/issues/9285#issuecomment-58362487

front/scripts/main/router.ts
RafLeszczynski: @rogatty, @kenkouot: +1 for the idea. Will will do it while continuing out work on Infobox Builder feature. Here is a link to Jira issue: https://wikia-inc.atlassian.net/browse/DAT-3233

front/scripts/main/components/InfoboxBuilderItemDataComponent.ts
RafLeszczynski: @kenkouot I totally agree with what you wrote above. Here is the reasons why we are doing it:
- while creating the infobox using infobox builder we want to show exactly how it will look like in the article on desktop (this is the product requirement)
- far that reason we want to use the same styles and same templates that are used to render infobox on desktop
- unfortunately we cannot convert mustache templates from oasis to Ember version of HandleBars (we've tired)
- because of that we decided to make a copy of all templates in mercury 
- we want to be as consistent as possible when it comes to how both ember and mustache templates look like so we decided not to extract the template wrapper and its classes to ember component  properties
- the default ember component wrapper was breaking our styles so we needed to remove it somehow
- in Angular there is a `replace` option but in Ember AFAIK the only way is `tagName: ''`

If you have any other idea how we could solve this issue, we are open for your feedback

RafLeszczynski:
Thank you all for you feedback. Infobox Builder is still WIP. We won't marge anything to dev until we implement all MVP features. For now I'm closing this PR. You can track progress on Infobox Builder project here: https://github.com/Wikia/mercury/compare/infobox-builder-feature-branch

PR #1326, opened 2015-09-04
DAT-3201 set single session "useskin=oasis" cookie for footer links

RafLeszczynski:


front/scripts/main/components/WikiaFooterComponent.ts
```diff
@@ -50,5 +50,17 @@ App.WikiaFooterComponent = Em.Component.extend(App.TrackClickMixin, {
 			href: '//www.wikia.com/Special:Contact',
 			text: 'footer-link-feedback'
 		}
-	]
+	],
+	actions: {
+		handleFooterLinkClick: function(text: string, href: string) {
+			var pattern = '?useskin=oasis';
```
SebastianMarzjan: I'm wondering if the ? is necessary - in case we have this param somewhere further in the query string

front/scripts/main/components/WikiaFooterComponent.ts
```diff
@@ -50,5 +50,17 @@ App.WikiaFooterComponent = Em.Component.extend(App.TrackClickMixin, {
 			href: '//www.wikia.com/Special:Contact',
 			text: 'footer-link-feedback'
 		}
-	]
+	],
+	actions: {
+		handleFooterLinkClick: function(text: string, href: string) {
+			var pattern = 'useskin=oasis';
```
hakubo: this might be monobook for example
so this might be 
useskin=monobook
can we handle that also?

front/scripts/main/components/WikiaFooterComponent.ts
```diff
@@ -50,5 +50,17 @@ App.WikiaFooterComponent = Em.Component.extend(App.TrackClickMixin, {
 			href: '//www.wikia.com/Special:Contact',
 			text: 'footer-link-feedback'
 		}
-	]
+	],
+	actions: {
+		handleFooterLinkClick: function(text: string, href: string) {
```
hakubo: what do you think about unit test for this function?
this way we won't introduce this P2 in the future

front/scripts/main/components/WikiaFooterComponent.ts
```diff
@@ -50,5 +50,17 @@ App.WikiaFooterComponent = Em.Component.extend(App.TrackClickMixin, {
...
+			var pattern = 'useskin=oasis';
+
+			// if link is overwritting skin to oasis, set cookie with skin overwrite for single session
```
hakubo: typo: ovrewriting

front/scripts/main/components/WikiaFooterComponent.ts
hakubo: nevermind, 
in the long run we don't want to support that skin anyway

hakubo:
LGTM,

+1 for test! :)

RafLeszczynski:
:-)

PR #1331, opened 2015-09-04
HG-848 Discussion header

mklucsarits:
https://wikia-inc.atlassian.net/browse/HG-848


front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -5,7 +5,55 @@
...
+		}
+	],
+
+	siteName: Em.computed(function (): string {
 		return Em.get(Mercury, 'wiki.siteName');
-	})
+	}),
+
+	init: function () {
+		this.set('sortMessageKey', this.getSortMessageKey(this.container.lookup('route:discussion.forum').get('sortBy')));
```
bkoval: Is there a way to simplify this line (split it)? I can see four methods invoked here :)

front/scripts/main/models/DiscussionForumModel.ts
```diff
@@ -24,19 +26,36 @@ App.DiscussionForumModel = Em.Object.extend({
 				error: (err: any) => reject(err)
 			});
 		});
+	},
+
+	getSortKey (sortBy: string): string {
+		switch (sortBy) {
```
bkoval: Can the content just be unsorted? (returning ' ') or should this method default to one of the two values?

front/scripts/main/models/DiscussionForumModel.ts
```diff
@@ -24,19 +26,36 @@ App.DiscussionForumModel = Em.Object.extend({
...
+	find (wikiId: number, forumId: number, sortBy: string) {
 		return new Em.RSVP.Promise((resolve: Function, reject: Function) => {
 			var forumInstance = App.DiscussionForumModel.create({
-				wikiId: wikiId,
-				forumId: forumId
-			});
+					wikiId: wikiId,
+					forumId: forumId
+				}),
+				requestData: any = {};
```
bkoval: Could this have an interface?

front/styles/main/module/_discussion.scss
```diff
@@ -1,30 +1,84 @@
+// TODO this height will be different or unused on desktop
+$site-header-height: 46px;
+$discussion-header-height: 40px;
+$discussion-header-line-height: 45px;
```
bkoval: Shouldn't we set the font-related sizes in rems to be adjustable when screen size changes?

front/styles/main/module/_discussion.scss
```diff
@@ -1,30 +1,84 @@
...
+$theme-bar-height: 10px;
+
 .discussion-header {
-	margin-top: 50px;
+	color: #fff;
```
bkoval: Sounds like a good candidate for a variable :)

front/styles/main/module/_discussion.scss
```diff
@@ -1,30 +1,84 @@
...
+		position: absolute;
+		right: 5px;
+		text-align: right;
+		top: 5px;
+		width: 110px;
+
+		li {
+			cursor: pointer;
+			font-size: $type-smallest;
+			line-height: 28px;
```
bkoval: How about rems here as well?

front/scripts/main/components/DiscussionHeaderComponent.ts
rogatty: Why not pass this property the standard way (as a component attribute)?

front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -5,7 +5,55 @@
...
+	 * Gets i18n message key for sort-by text
+	 * @param {string} sortBy
+	 */
+	getSortMessageKey: function (sortBy: string) {
+		var filtered = this.get('sortTypes').filter((obj: any): boolean => {
+				return obj.name === sortBy;
+			});
+
+			return filtered.length ? filtered[0].messageKey :
+				this.get('sortTypes')[0].messageKey;
```
rogatty: You're doing `this.get('sortTypes')` twice so maybe a new `var`?

front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -5,7 +5,55 @@
...
+	init: function () {
+		this.set('sortMessageKey', this.getSortMessageKey(this.container.lookup('route:discussion.forum').get('sortBy')));
+		this._super();
+	},
+
+	/**
+	 * Gets i18n message key for sort-by text
+	 * @param {string} sortBy
+	 */
+	getSortMessageKey: function (sortBy: string) {
```
rogatty: You can make `sortMessageKey` a computed property, I see no need for additional function.

front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -5,7 +5,55 @@
...
+	actions: {
+		setSortBy: function (sortBy: string): void {
+			// Send action up to route object
+			this.sendAction('setSortBy', sortBy);
+			// Update the translated sort-by text
+			this.set('sortMessageKey', this.getSortMessageKey(sortBy));
+		},
+
+		showSortSelector: function (): void {
+			this.$('.sort-select').show();
```
rogatty: Why jQuery? You can just have a property like `sortVisible` and use it in the template, for example `<ul class="sort-select {{unless sortVisible 'hidden'}}">`

front/scripts/main/components/DiscussionHeaderComponent.ts
bkoval: I was thinking about it. Getter doesn't neccesarily have to be invoked twice, it will only be triggered once if "filtered" array has elements so maybe this strategy will perform better?

front/scripts/main/models/DiscussionForumModel.ts
```diff
@@ -7,11 +7,13 @@ App.DiscussionForumModel = Em.Object.extend({
 	posts: null,
 	totalPosts: 0,
 
-	loadPage(pageNum: number) {
+	loadPage (pageNum: number) {
```
hakubo: there should be no space here
https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#functions

"There should be no space between the name of a function and the left parenthesis of its parameter list."

and I agree as that is how you'll be calling this function later on

this.loadPage(0);

front/scripts/main/routes/DiscussionForumRoute.ts
```diff
@@ -1,29 +1,28 @@
...
+
+App.DiscussionForumRoute = Em.Route.extend(App.FullPageMixin, {
+	forumId: null,
+	sortBy: null,
 
-App.DiscussionForumRoute = Em.Route.extend({
 	model (params: any) {
-		return App.DiscussionForumModel.find(Mercury.wiki.id, params.forumId);
+		this.set('forumId', params.forumId);
+		this.set('sortBy', params.sortBy);
```
rogatty: `this.setProperties`

front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -5,7 +5,55 @@
...
+		var filtered = this.get('sortTypes').filter((obj: any): boolean => {
+				return obj.name === sortBy;
+			});
+
+			return filtered.length ? filtered[0].messageKey :
+				this.get('sortTypes')[0].messageKey;
+	},
+
+	actions: {
+		setSortBy: function (sortBy: string): void {
```
hakubo: you can use object shorthand
setSortBy(sortBy) {
}

front/scripts/main/components/DiscussionHeaderComponent.ts
hakubo: agree!

front/scripts/main/mixins/FullPageMixin.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.FullPageMixin = Em.Mixin.create({
```
hakubo: This sounds like a hack :)
But we should solve it more holistically.

front/templates/main/discussion/forum.hbs
```diff
@@ -1,5 +1,7 @@
 {{top-bar}}
-{{discussion-header}}
+{{discussion-header
+	setSortBy='setSortBy'
+}}
```
rogatty: Related to https://github.com/Wikia/mercury/pull/1331/files#r38926423: `sortBy=sortBy`

front/scripts/main/mixins/FullPageMixin.ts
rogatty: Please add a comment that this is temporary and will be done properly in https://wikia-inc.atlassian.net/browse/XW-247

front/scripts/main/components/DiscussionHeaderComponent.ts
mklucsarits: @rogatty I spent hours trying to figure out a way to do this. You can't pass a property from a route to a component. Ember doesn't allow it. Ken's suggestion was to create a controller just for this property, to pass in to the component, but I didn't want to create a controller if Ember is moving away from controllers.

front/scripts/main/mixins/FullPageMixin.ts
mklucsarits: Well this just moves this logic out of multiple files that were setting this fullPage property, into a mixin. It was not intended as a hack but to make our code DRY.

front/scripts/main/models/DiscussionForumModel.ts
mklucsarits: It will always be sorted by the API. If sort_key is not specified in the API request, it will be sorted by creation_date. It may be better to have creation_date as the default value.

front/scripts/main/components/DiscussionHeaderComponent.ts
rogatty: Ember is creating controllers in the background anyway and before Routable Components this is the only way to pass data from route to component. This is exactly what happens when you use `model` in `front/templates/main/discussion/forum.hbs`. This is the model of dynamically created controller, passed there from route.

You can look at main pages for example how we implemented it:
https://github.com/Wikia/mercury/blob/dev/front/scripts/main/routes/MainPageRoute.ts#L16
https://github.com/Wikia/mercury/blob/dev/front/templates/main/main-page.hbs

front/scripts/main/mixins/FullPageMixin.ts
rogatty: It was hacky from day one, before this PR ;)

front/scripts/main/components/DiscussionHeaderComponent.ts
mklucsarits: Ah ok. It's not intuitive, but it makes sense. Thanks for pointing to that example. I'll change this code to work similarly.

front/styles/main/module/_discussion.scss
mklucsarits: In general, I agree. However, this particular line-height was set for the discussion header height, which has an exact px value decided by the UX team. There was not much consideration given to the current style guide for the discussion design. Instead we will have to update the style guide (when? who knows) to match these new design elements. You'll also find font-size attributes here with px values which unfortunately are not part of the style guide.

rogatty: /cc @vforge

front/scripts/main/controllers/DiscussionForumController.ts
```diff
@@ -0,0 +1,3 @@
+/// <reference path="../app.ts" />
+
+App.DiscussionForumController = Em.Controller.extend({});
```
kenkouot: Does this file need to exist?

front/scripts/main/mixins/FullPageMixin.ts
kenkouot: Agreed, the whole `fullPage` implementation is pretty dirty :D We need to deliberately design our application view hierarchy.

front/scripts/main/controllers/DiscussionForumController.ts
mklucsarits: Yes. If I call this.controllerFor('discussionForum') in DiscussionForumRoute without this file existing, stuff breaks.

rogatty: @mklucarits what do you mean by stuff breaks? It should work just fine without this file.

mklucsarits: There is an Ember error and the page doesn't load. There's just a blank page. I think the reason it works for MainPageRoute is because there is a MainPageController.

front/scripts/main/models/DiscussionForumModel.ts
mklucsarits: It could. Should it include all 7 optional parameters, even if we never use 6 of them here?

front/scripts/main/controllers/DiscussionForumController.ts
rogatty: Please add a `DiscussionRoute` with action `error` that will log all route errors in its child routes. Right now they're dropped and you can't see what's wrong. Take `MainPageRoute` as an example. This probably should be moved to `ApplicationRoute` at some point.

After I enabled logs I've seen `Assertion Failed: The controller named 'discussionForum' could not be found. Make sure that this route exists and has already been entered at least once. If you are accessing a controller not associated with a route, make sure the controller class is explicitly defined.`. This is caused by the fact that you set controller properties in `model` function when you actually should do it in the `setupController` function, see http://emberjs.com/api/classes/Ember.Route.html#method_setupController

The proper way to do it is:
```
setupController(controller, model, transition) {
	controller.set('sortBy', transition.params['discussion.forum'].sortBy);
}
```

Doing it like that stops the error from appearing altough there is now a deprecation warning:
`DEPRECATION: You attempted to access sortBy from (generated discussion.forum controller), but object proxying is deprecated. Please use model.sortBy instead.`

This is fixed by setting the property that you want to use on the controller like:
```
App.DiscussionForumController = Em.Controller.extend({
	sortBy: null
});
```

If I understand things correctly we'll be able to remove all those placeholder controllers (we have 7 of them for `CuratedContentEditor`!) after upgrading to Ember 2.0.

front/styles/main/module/_discussion.scss
vforge: @mklucsarits It's ok-ish for now, but we need to adjust it later when StyleGuide will be updated.

bkoval: Is typography and dimensions EXACTLY the same on desktop designs? I guess we'' aim for reusing as many styles as possible so if they're compliant we can live with px, but if there's a change between mobile and desktop then we should definitely stick to rem values.

I think we should raise this issue as soon as possible, since we're supposed to use style-guide, maybe there's a way of aligning the designs in order to use what's already there?

mklucsarits: Believe me, I have raised this issue with UX more than once since the discussion designs were first introduced. Being one of the devs at Wikia who first created the style guide, I am all for using things from the style guide, and making things re-usable so that they can be added to the style guide. The discussion header in this commit, and its styles, are for mobile only. These styles will not apply to screen widths greater than 320px. Obviously these styles will need to be re-organized quite a bit for desktop. At this point, because read-only discussions are past due, my priority is getting these styles done quickly for mobile and reorganizing the styles later on, while we work on desktop.

lizlux:
The `CuratedContentEditorRoute` and `AddPhotoRoute` files still have the `fullPage` hacks and should be able to start using the mixin.

front/scripts/main/models/DiscussionForumModel.ts
kenkouot: With the move to ES6 looming, I don't think this is a blocker for merging.

mklucsarits:
@lizlux I can update AddPhotoRoute and EditRoute to use the FullPageMixin, but the CuratedContentEditorRoute contains more complex logic for unsetting the fullPage property that isn't easily replaceable.

front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -5,7 +5,45 @@
...
+			messageKey: 'discussion.sort-by-latest'
+		},
+		{
+			name: 'trending',
+			messageKey: 'discussion.sort-by-trending'
+		}
+	],
+
+	siteName: Em.computed(function (): string {
 		return Em.get(Mercury, 'wiki.siteName');
```
kenkouot: Just use Mercury.wiki.siteName, this getter wrapper is of dubious usefulness.

kenkouot:
LGTM

lizlux:
@mklucsarits :+1:

front/scripts/main/routes/AddPhotoRoute.ts
```diff
@@ -17,17 +18,10 @@ App.AddPhotoRoute = Em.Route.extend({
...
 			return true;
 		},
-		willTransition: function(transition: EmberStates.Transition): boolean {
-			transition.then(() => {
-				this.controllerFor('application').set('fullPage', false);
-			});
-			return true;
-		},
 		didTransition: function(): boolean {
 			// AddPhotoRoute works in "fullPage mode" (unlike ArticleRoute) which means that it takes
```
lizlux: Please remove this comment too.

PR #1334, opened 2015-09-07
XW-230 Update Host header when there is a redirect

rogatty:
https://wikia-inc.atlassian.net/browse/XW-230

Trying to access www.starwars.wikia.com with Mercury was failing because Wreck was sending all requests with the original host, even when redirect location had a different one.

I'll update `package.json` when https://github.com/hapijs/wreck/compare/master...rogatty:before-redirect is merged. I still need to write tests and docs, no need to block P2 fix by that.

/cc @hakubo @Warkot @bognix

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "hoek": "2.12.0",
     "i18next": "1.7.10",
     "newrelic": "1.18.1",
-    "wreck": "6.1.0"
+    "wreck": "rogatty/wreck#before-redirect"
   },
   "devDependencies": {
     "browser-sync": "^2.5.3",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "hoek": "2.12.0",
     "i18next": "1.7.10",
     "newrelic": "1.18.1",
-    "wreck": "rogatty/wreck#before-redirect"
+    "wreck": "rogatty/wreck"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


bognix:
could you do something about: 
```
npm ERR! Command failed: Cloning into bare repository '/var/lib/jenkins/.npm/_git-remotes/https-github-com-rogatty-wreck-bbcf0992'...

File: package.json [hunk 1]
```diff
@@ -29,7 +29,7 @@
     "hoek": "2.12.0",
     "i18next": "1.7.10",
     "newrelic": "1.18.1",
-    "wreck": "6.1.0"
+    "wreck": "rogatty/wreck"
```
bognix: is this temporary solution?

rogatty: Yes, please read the PR description. I can't add a comment to JSON.

bognix: sorry, thanks! :) next time we should do the fork as Wikia so everybody in the company can do the fix easily :)

bognix:
LGTM as a temp solution. Please create a ticket for permanent fix and add it in TODO comment above  https://github.com/Wikia/mercury/pull/1334/files#diff-21236e6088aed77d5a421e8c69672a93R180

rogatty:
@bognix created XW-236 and added a comment.

hakubo:
LGTM

vforge:
LGTM

PR #1341, opened 2015-09-08
SOC-959 Use wordmark as default sharing image

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-959

This adds a Wikia logo image to use for sharing discussion posts, if the wiki does not have an image for the OpenGraph tag.

@kenkouot @garthwebb


server/lib/OpenGraph.ts
```diff
@@ -37,6 +38,9 @@ function getPromiseForDiscussionData (request: Hapi.Request, wikiVars: any): Pro
 
 			openGraphData.type = 'article';
 			openGraphData.url = wikiVars.basePath + request.path;
+			// Use Wikia logo as default image
+			openGraphData.image = 'http://' + Utils.getCachedWikiDomainName(localSettings, request.headers.host)
```
kenkouot: Using a template string here would be much cleaner: http://basarat.gitbooks.io/typescript/content/docs/template-strings.html

kenkouot:
Where will this image be getting served from, nocookie? or from the node server.

garthwebb:
Relaying the conversation on #mercury-core, we should probably have the generic image use ```localSettings.cdnBaseUrl``` rather than the wikia URL.  Also, from @kenkouot: 
> there should be a function [to generate the CDN URL]
> because on dev it’s `/front/`
> which gets replaced with the cdn url
> so we need a dynamic url builder that checks the env

mklucsarits:
@garthwebb See my latest change. I am using the CDN domain.

garthwebb:
@mklucsarits I put @kenkouot's quote in my comment because he's noting that we need more than the cdnUrl for this to work properly in prod and dev. For example, with the current change the URL on my dev box would be:

> http://glee.garth.wikia-dev.com/mercury-static/images/wikia-mark-128.png

which gets interpreted as an article.  This URL seems to do the right thing however:

> http://glee.garth.wikia-dev.com/front/images/wikia-mark-128.png

I feel like we should either make it so dev responds to `mercury-static` properly, or we should make sure the URL gets munged correctly to match the environment.

garthwebb:
LGTM!

server/lib/Utils.ts
```diff
@@ -232,3 +232,17 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
 	};
 }
 
+
+/**
+ * Gets the domain and path for a static asset
```
rogatty: Please add a comment that values here are strictly connected with https://github.com/Wikia/mercury/blob/dev/gulp/options/prod.js#L52-L56

server/lib/Utils.ts
```diff
@@ -232,3 +232,17 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+ *
+ * @param {LocalSettings} localSettings
+ * @param {string} wikiDomain
+ * @returns {string}
+ */
+export function getStaticAssetPath(localSettings: LocalSettings, wikiDomain: string): string {
+	var env = typeof localSettings.environment === 'number' ? localSettings.environment : Environment.Dev;
+	return (env === Environment.Prod || env === Environment.Sandbox)
+		? 'http:' + localSettings.cdnBaseUrl + '/mercury-static/'
+		: 'http://' + getCachedWikiDomainName(localSettings, wikiDomain) + '/front/';
```
rogatty: Do we really want to hardcode `http` here?

mklucsarits: What would you suggest instead? The OpenGraph image needs a fully qualified URI and can't be protocol-agnostic (it's not correct usage for any OpenGraph tag URL).

server/lib/Utils.ts
```diff
@@ -232,3 +232,18 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+/**
+ * Gets the domain and path for a static asset
+ *
+ * @param {LocalSettings} localSettings
+ * @param {string} wikiDomain
+ * @returns {string}
+ */
+export function getStaticAssetPath(localSettings: LocalSettings, wikiDomain: string): string {
+	var env = typeof localSettings.environment === 'number' ? localSettings.environment : Environment.Dev;
+	return (env === Environment.Prod || env === Environment.Sandbox)
```
kvas-damian: @mklucsarits What about ```preview```? Shouldn't preview follow the same rules as ```Sandbox``` and ```Prod```?

Maybe we could just use: ```env !== Environment.Dev```

mklucsarits: Yes, you're right. I'll do that instead.

server/lib/Utils.ts
rogatty: This function isn't strictly related to OpenGraph so maybe the protocol should be added somewhere else, later? I'm not sure what's the best way to handle it but `getStaticAssetPath` seems like something that should be more universal.

garthwebb:
LGTM

PR #1350, opened 2015-09-10
HG-839 - make sure image width is defined

katanka:
In the "Read More" section in articles, sometimes the image width was undefined when requesting it, which makes the request 404. This makes sure width is defined, and also makes them 16x9.


front/scripts/main/components/MediaComponent.ts
```diff
@@ -50,12 +50,24 @@ App.MediaComponent = Em.Component.extend(App.VisibleMixin, {
 			options.width = this.normalizeThumbWidth(options.width);
 		}
 
+		// Sometimes width is null, so we need to make sure it has a value.
+		// I chose 378 because that is the correct size on iPhone 6 Plus
+		options.width = options.width || 378;
+
+		// If there is no restriction on height, make the image square.
+		// Otherwise, make it 16x9
```
rogatty: Why would we want to restrict all images to only two possible aspect ratios? If it's needed only for content recommendations then this logic should be in `ContentRecommendationsComponent`.

See https://github.com/Wikia/mercury/blob/dev/front/scripts/main/mixins/CuratedContentThumbnailMixin.ts for reference. (We probably should extract common logic from there).

front/scripts/main/components/MediaComponent.ts
```diff
@@ -50,12 +50,24 @@ App.MediaComponent = Em.Component.extend(App.VisibleMixin, {
 			options.width = this.normalizeThumbWidth(options.width);
 		}
 
+		// Sometimes width is null, so we need to make sure it has a value.
+		// I chose 378 because that is the correct size on iPhone 6 Plus
+		options.width = options.width || 378;
```
rogatty: Please use `this.thumbSize` constant.

front/scripts/main/components/MediaComponent.ts
```diff
@@ -50,6 +50,9 @@ App.MediaComponent = Em.Component.extend(App.VisibleMixin, {
 			options.width = this.normalizeThumbWidth(options.width);
 		}
 
+		// Sometimes width is null, so we need to make sure it has a value.
```
lizlux: might make sense to check if `options.width` is defined before calling lines 49-51 so we can save on that logic.

katanka:
It looks like the vignette is only behaving weirdly on dev - this should work fine on production (it works on sandbox).
Related ticket: https://wikia-inc.atlassian.net/browse/PLATFORM-1444

front/scripts/main/components/MediaComponent.ts
kenkouot: Addressed on https://github.com/Wikia/mercury/pull/1350/files#diff-8a5706fd0efa824e3fe620034bbe6049R49

kenkouot:
LGTM

rogatty:
LGTM

PR #1351, opened 2015-09-10
INT-185 Internationalization for Wikia homepage link in header

mari-mcmurtrie:
In Mercury, there is "Wikia" logo in header section. The logo is linked to "www.wikia.com", no matter which wikia you are looking to, and which language you set in your account. We internationalize this so that:
1) If a user is not login, then we link to homepage of the same language of wikia the user currently is in.
2) If a user is login, then we link to homepage of the same language of the user's language.

The behavior is the same as existing Oasis header link to homepage. Ticket: https://wikia-inc.atlassian.net/browse/INT-185


mari-mcmurtrie:
@lizlux @kenkouot , will you please CR when you have a chance? Thank you!

front/locales/de/main.json
```diff
@@ -1,5 +1,6 @@
 {
   "app":{
+    "wikia-homepage": "http://de.wikia.com/Wikia",
```
hakubo: I think this file is meant to be translatable. For copies that translators can actually take and translate.
Translators won't necessarily know url structure to our service
In addition it is odd to have links in messages

I think this should be either
created in SPA
or send from server

rogatty: +1

lizlux: Do we have these values server-side already? If we do it's probably b/c they were entered into i18n files. 

The reason I suggested putting these values here is b/c it is actually internationalizing our content. Also, there's no great pattern for knowing which home page to link to based on wiki language, so it would be difficult to create these URLs programatically. 

If the suggestion is to set up a similar hash of hard-coded values either in the SPA or in MW then I can see that being a viable solution, though I still think this is a good place for them.

front/locales/fi/main.json
```diff
@@ -1,5 +1,6 @@
 {
   "app":{
+    "wikia-homepage": "http://www.wikia.com/Wikia",
```
lizlux: It's probably better to use only `http://www.wikia.com` and let VCL do the redirect to `/Wikia`, just in case the name for the home page article ever changes. (Same for all the other URLs.)

mari-mcmurtrie: I thought I would avoid redirect so I put /Wikia here, but I was not aware of the possibility you mentioned (hp article change). I will remove /Wikia.

front/locales/de/main.json
kenkouot: I agree with @lizlux and I was the one who suggested moving this to the template. These should just be internationalized with a fallback redirect to the EN version for `language.wikia.com` instances that don't work. I don't think a service that handles this is really appropriate.

Additionally, having i18n-ized versions of our communities actually be different subdomains and separate communities is an incredibly ugly part of our legacy architecture. Crystallizing that in a server seems like it just adds another layer of complexity, on top of an already poorly thought out implementation. Putting localized homepage URIs in i18n files is simple and costs very little, and limits the amount of *dealing with legacy* cost we have to keep paying.

hakubo: I have found WikiaLogoHelper::getCentralUrlForLang that returns exactly this
https://github.com/Wikia/app/blob/dev/extensions/wikia/WikiaLogo/WikiaLogoHelper.class.php#L24

You could add it to a response for wikivariables as Mercury does not 'care' about user language for now anyway (it always uses content lang), so that would be consistent across skins. We could then refactor it to take into consideration user language and would move sending this link to other resource. 

This way we have one logic that generates those links, so if anything changes/brakes it's consistent between skins. Examples of change:
- we are talking about moving to https
- we are talking about changing our domains
- someone moves a page

front/locales/fi/main.json
mari-mcmurtrie: I removed url path from all locale. I tested all of the URLs in file get redirected to correct home page (www.wikia.com -> 301 to /Wikia).

mari-mcmurtrie:
Fixed some URLs. @lizlux will you check it again? Thank you!

hakubo:
I have found WikiaLogoHelper::getCentralUrlForLang that returns exactly this
https://github.com/Wikia/app/blob/dev/extensions/wikia/WikiaLogo/WikiaLogoHelper.class.php#L24

You could add it to a response for wikivariables as Mercury does not 'care' about user language for now anyway (it always uses content lang), so that would be consistent across skins. We could then refactor it to take into consideration user language and would move sending this link to other resource.

This way we have one logic that generates those links, so if anything changes/brakes it's consistent between skins. Examples of change:
- we are talking about moving to https
- we are talking about changing our domains
- someone moves a page

mari-mcmurtrie:
@hakubo thanks! For login user's case, as you know I filed a bug (https://wikia-inc.atlassian.net/browse/HG-874 ). 
If I stick with current implementation using i18n file, then fixing the HG-874 bug automatically takes care of login user case for this feature, so no more works required. But if I change the homepage URL coming back from server as you suggested, then logic to generate a homepage URL depending on user's login status has to change (as you said, have to be refactored or something).

I have to make the feature work for login users. If I take your approach, then who will take care of backend refactor? Let me know.

mari-mcmurtrie:
Ok, so I created branch for app, please see: https://github.com/Wikia/app/pull/8452

mari-mcmurtrie:
@hakubo @lizlux I changed the way to retrieve locale homepage URL from backend. This code works with https://github.com/Wikia/app/pull/8452. Please CR when you have time. Thank you so much!

front/locales/en/main.json
```diff
@@ -144,4 +144,4 @@
 		"sort-by-latest": "Latest",
 		"sort-by-trending": "Trending"
 	}
-}
+}
```
lizlux: new line

lizlux:
:+1: LGTM

rogatty:
LGTM

front/scripts/main/components/SiteHeadComponent.ts
```diff
@@ -8,7 +8,7 @@ App.SiteHeadComponent = Em.Component.extend(App.TrackClickMixin, App.HeadroomMix
 	classNameBindings: ['themeBar'],
 	tagName: 'nav',
 	themeBar: false,
-
+	wikiaHomepage: Em.get(Mercury, 'wiki.homepage'),
```
rogatty: Please change it to `Em.getWithDefault(Mercury, 'wiki.homepage', 'http://www.wikia.com')` so we don't serve a broken link before the app part is released and its responses purged from all caches.

front/scripts/main/components/SiteHeadComponent.ts
```diff
@@ -8,7 +8,7 @@ App.SiteHeadComponent = Em.Component.extend(App.TrackClickMixin, App.HeadroomMix
 	classNameBindings: ['themeBar'],
 	tagName: 'nav',
 	themeBar: false,
-
+	wikiaHomepage: Em.get(Mercury, 'wiki.homepage', 'http://www.wikia.com'),
```
rogatty: `Em.getWithDefault` ;)

mari-mcmurtrie: Yeah my build failed :b

PR #1352, opened 2015-09-11
SOC-951 Add back controller so we can trigger the "show more" button

garthwebb:
When the model changes, we need to make sure the "show more" button disappears when appropriate.

https://wikia-inc.atlassian.net/browse/SOC-951


front/scripts/main/controllers/DiscussionPostController.ts
```diff
@@ -0,0 +1,29 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionPostController = Em.Controller.extend({
```
garthwebb: Yes, its a controller.  Talked w/ Ken and Liz and there isn't a good way to do these computed properties within the route apparently, or at least not until we've moved fully to Ember 2.x

front/scripts/main/controllers/DiscussionPostController.ts
```diff
@@ -0,0 +1,29 @@
...
+	}),
+
+	actions: {
+		expand: function () {
+			var model = this.get('model');
+
+			model.loadNextPage().then(function (): number {
+				var model = this.get('model');
+				this.set('loadedReplies', Em.get(model, 'replies.length'));
+			}.bind(this));
```
mklucsarits: If you want, you can avoid using .bind() here by using arrow notation for the function declaration:
```(): number => {```

File: front/scripts/main/controllers/DiscussionPostController.ts [hunk 3]
```diff
@@ -0,0 +1,29 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionPostController = Em.Controller.extend({
+	loadedReplies: null,
+
+	showMore: Em.computed('model', 'loadedReplies', function (): boolean {
```
rafalkalinski: 'showMore' method name doesn't tell much about what it really does. Maybe sth like 'isShowMore' or 'showMoreActive'.

front/scripts/main/controllers/DiscussionPostController.ts
```diff
@@ -0,0 +1,29 @@
...
+		}
+
+		return loadedReplies < model.postCount;
+	}),
+
+	actions: {
+		expand: function () {
+			var model = this.get('model');
+
+			model.loadNextPage().then(function (): number {
```
rafalkalinski: I can't see why there is ': number' return type

mklucsarits: Yes, it should be void.

front/templates/main/discussion/post.hbs
```diff
@@ -6,7 +6,7 @@
 		upvoteCount=model.upvoteCount
 		replyCount=model.postCount
 	}}
-	{{#if showMore}}
+	{{#if canShowMore}}
         <button {{action 'expand'}} class="see-more">
 			{{svg 'comments' viewBox='0 0 30 30' class='icon replies'}}
             View Older Replies
```
mklucsarits: Out of scope of this change maybe, but this will need to be an i18n string at some point.

mklucsarits:
LGTM

front/templates/main/discussion/post.hbs
garthwebb: Heh, I think its reasonable to fix this here, but we've just talked about having Armon do it, so I'll skip.

rafalkalinski:
LGTM

front/scripts/main/controllers/DiscussionPostController.ts
```diff
@@ -0,0 +1,29 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionPostController = Em.Controller.extend({
+	loadedReplies: null,
+
+	canShowMore: Em.computed('model', 'loadedReplies', function (): boolean {
+		var model = this.get('model'),
```
kvas-damian: Can you define type of variables?

front/scripts/main/controllers/DiscussionPostController.ts
```diff
@@ -0,0 +1,29 @@
...
+			loadedReplies = this.get('loadedReplies');
```
kvas-damian: I'm not sure if this variable name describes best what is stored in it.

garthwebb: @kvas-damian for subjective comments, e.g. "I think this could be better", don't forget to give a suggestion.  Otherwise the change becomes a guessing game:
> "Name 'A' could be better", 
> "How about 'B'?
> "Still not good"
> "How about C?"
> ...
Updating this to be what I think you're asking for...

front/scripts/main/controllers/DiscussionPostController.ts
garthwebb: Yes, you can.  It seems to be done inconsistently so I'm not sure if there is a standard to do this or not.  I've added it to this controller.

front/scripts/main/controllers/DiscussionPostController.ts
kvas-damian: Sorry, my fault.
I'm glad that you found what I was referring to.

front/scripts/main/controllers/DiscussionPostController.ts
kvas-damian: Fixed typings: https://github.com/Wikia/mercury/commit/27eb761a4b0ba5604088b1ac055ca194cbca9b6d

PR #1354, opened 2015-09-11
XW-204 Unwrap Wikia widgets

rogatty:
https://wikia-inc.atlassian.net/browse/XW-204

### DIFF
--- a/front/styles/main/module/_wikia-widget.scss
+++ b/front/styles/main/module/_wikia-widget.scss
@@ -1,0 +1,3 @@
+iframe[data-wikia-widget] {
+	width: 100%;
+}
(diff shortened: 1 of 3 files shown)


hakubo:
LGTM

front/styles/main/module/_wikia-widget.scss
```diff
@@ -0,0 +1,3 @@
+iframe[data-wikia-widget] {
+	width: 100%;
+}
```
bognix: is this needed if we handle all iframes somewhere else?

rogatty: We have
```
iframe {
	max-width: 100%;
}
```
in `_media.scss` but I don't want to put this rule there. The one above doesn't touch all iframes but only our widgets so I can't introduce any regression - I don't like introducing them ;)

bognix:
LGTM

tomasznapieralski:
LGTM

PR #1357, opened 2015-09-11
SOC-1279 Expose servicesDomain to ember and use for discussion api url

armonr:
/CC @kenkouot @mklucsarits @garthwebb 

https://wikia-inc.atlassian.net/browse/SOC-1279

### DIFF
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -11,7 +11,7 @@ var deepExtend: any = require('deep-extend');
 
 var localSettings: LocalSettings = {
 	apiBase: '/api/v1',
-	servicesDomain: 'services.wikia-dev.com',
+	servicesDomain: 'services.wikia.com',
 	// Default timeout for backend requests
 	// This timeout is the same as the MW app timeout
 	backendRequestTimeout: 300000,
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -11,7 +11,7 @@ var deepExtend: any = require('deep-extend');
 
 var localSettings: LocalSettings = {
 	apiBase: '/api/v1',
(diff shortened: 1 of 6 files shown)


config/localSettings.base.ts
```diff
@@ -11,7 +11,7 @@ var deepExtend: any = require('deep-extend');
 
 var localSettings: LocalSettings = {
 	apiBase: '/api/v1',
-	servicesDomain: 'services.wikia.com',
+	servicesDomain: 'services.wikia-dev.com',
```
mklucsarits: The dev URLs are already configured in localSettings.dev.ts. This one in base shouldn't be changed.

mklucsarits:
LGTM

front/scripts/main/models/DiscussionIndexModel.ts
```diff
@@ -7,7 +7,7 @@ App.DiscussionIndexModel.reopenClass({
 	find(wikiId: number) {
 		return new Em.RSVP.Promise((resolve: Function, reject: Function) => {
 			Em.$.ajax({
-				url: `https://services.wikia.com/discussion/${wikiId}/forums`,
+				url: 'https://' + M.prop('servicesDomain') + `/discussion/${wikiId}/forums`,
```
lizlux: Should the protocol be included in the `servicesDomain`? Would be cleaner from the perspective of someone using the var. Is there any use case for `http` here?

garthwebb: The `services.wikia.com` domain redirects to `HTTPS` if you try to access it via `HTTP`.  I'd say leave it for now and if it becomes necessary in the future for some reason we add `M.prop('servicesProto')`

garthwebb:
LGTM!

PR #1358, opened 2015-09-11
Hg 848 Incorporating Top Bar component into mercury discussion pages

lizlux:
This PR does more to incorporate the new `top-nav` component with the discussions pages. I'd like to CR and merge this so we can swarm on the rest of the tasks.  

Here's left to do:

* [x] Fix page structure so the top-nav is outside of the discussions div, and so it makes sense on mobile and desktop. 
* [ ] i18n for login/register text on desktop - https://wikia-inc.atlassian.net/browse/HG-878
* [x] passing user name into `user-status` component for `alt` attribute
* [x] fix padding for mobile screens and make top-bar padding consistent with page (15px) 
* [ ] Put the blue bar back on the post pages - https://wikia-inc.atlassian.net/browse/HG-879
* [x] Update the home page links to match @mari-mcmurtrie's work on the existing global nav 
* [ ] Possibly create a ticket for Polymer about a JS error I'm seeing when using `<content>` tags inside `dom-if` templates. Error is `Uncaught TypeError: Cannot set property '_insertionPoints' of undefined` but it doesn't seem to be affecting the application at all. - https://wikia-inc.atlassian.net/browse/HG-880


bower.json
```diff
@@ -17,7 +17,7 @@
     "weppy": "Wikia/weppy#0.9.2",
     "ember-performance-sender": "1.1.1",
     "vignette": "wikia/vignette-js#2.1.1",
-    "wikia-style-guide": "Wikia/style-guide#v1.4.0",
+    "wikia-style-guide": "Wikia/style-guide#tob-bar-logo",
```
lizlux: We can remove this change once https://github.com/Wikia/style-guide/pull/72 is merged and the version is bumped.

katanka: should this be `#top-bar-logo`?

front/styles/main/module/_discussion.scss
```diff
@@ -150,7 +149,7 @@
 }
 
 @for $i from 1 through length($verticals) {
-	.#{nth($verticals, $i)}-vertical .discussion a {
+	.#{nth($verticals, $i)}-vertical .discussion a, .#{nth($verticals, $i)}-vertical .discussion ::shadow a {
```
mklucsarits: Maybe you could take advantage of SCSS nesting here so this doesn't look so verbose.
```
.#{nth($verticals, $i)}-vertical {
  .discussion {
    a, ::shadow a { 
  ...
```

front/templates/main/discussion/index.hbs
```diff
@@ -1,4 +1,11 @@
-{{top-bar}}
```
mklucsarits: Is this file's changes intentional? It doesn't match with forum.hbs and post.hbs.

gulp/tasks/build-vendor.js
```diff
@@ -14,7 +14,7 @@ var gulp = require('gulp'),
 gulp.task('build-vendor', ['vendor'], function () {
 	return piper(
 		gulp.src([
-			'www/front/vendor/webcomponentsjs/webcomponents-lite.js',
+			'www/front/vendor/webcomponentsjs/webcomponents.js',
```
mklucsarits: + the whole kitchen sink

server/routes.ts
```diff
@@ -238,7 +240,7 @@ articlePagePaths.forEach((path) => {
 
 // For application routes that are not articles and require the Ember app, push a route object
 // that uses the `showApplication` route handler to get a basic Ember application instance
-unauthenticatedRoutes.push({
+authenticatedRoutes.push({
```
mklucsarits: We are only allowing logged-in users to view discussions?

front/templates/main/discussion/forum.hbs
```diff
@@ -1,14 +1,16 @@
...
-	sortBy=sortBy
-}}
-{{forum-wrapper
-	goToPost='goToPost'
-	loadPage='loadPage'
-	name=model.name
-	posts=model.posts
-	postsDisplayed=model.posts.length
-	totalPosts=model.totalPosts
+<div class="page-wrapper discussion">
```
mklucsarits: Can you explain why you merged the page-wrapper container with the content container ('discussion')? I think it's useful to keep a distinction between the two container elements. Additionally, there is now a ```discussion``` CSS class on both the page-wrapper and on a div inside the forum-wrapper component. It should only be used in one place, and I don't see why it needs to be on an element encapsulating the top-bar and discussion-header.

bower.json
lizlux: @katanka yes I did screw up the branch name, but at least it was consistent! :dancer:

katanka: haha

server/routes.ts
rogatty: @mklucsarits see https://github.com/Wikia/mercury/pull/1358/files#diff-5db3909f0cdac37d8bc699c50c366dcdR17
>routes where we want to know the user's auth status

server/views/_layouts/ember-main.hbs
```diff
@@ -7,6 +7,11 @@
 		<meta name="keywords" content="{{#if wiki.siteMessage}}{{wiki.siteMessage}},{{/if}}{{wiki.siteName}},{{wiki.dbName}},{{displayTitle}}">
 		<meta name="description" content="{{article.details.description}}">
 		<meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1, minimal-ui">
+		<script>
+			window.Polymer = window.Polymer || {};
```
rogatty: Is it possible to put it somewhere else? This template is too big already.

server/routes.ts
mklucsarits: Hmm I see. That was not the original intended purpose of the separation between authenticated and unauthenticated routes, so I hope whoever made that change took that into consideration.

lizlux: Thanks @rogatty. @mklucsarits it confused me too, which is why I added the comment. We may want to update the wording used.

front/scripts/main/components/SmartBannerComponent.ts
```diff
@@ -7,6 +7,10 @@ App.SmartBannerComponent = Em.Component.extend({
 	classNames: ['smart-banner'],
 	classNameBindings: ['noIcon'],
 
+	becameVisible() {
+		this.$().css('display', 'flex');
```
rogatty: This is hacky. If it's here to stay please add a comment why we're doing it and preferably number of a ticket to refactor. Also move it next to `didInsertElement`.

Probably could be avoided by removing use of `isVisible` from this component and using `classNameBindings` instead. See:
- https://github.com/Wikia/mercury/blob/96ec87a5b548b1e67c9175cc7710aa037d056ef5/front/templates/main/application.hbs#L11
- https://github.com/Wikia/mercury/blob/96ec87a5b548b1e67c9175cc7710aa037d056ef5/front/scripts/main/controllers/ApplicationController.ts#L160

lizlux: notifying @kenkouot

front/templates/main/discussion/index.hbs
```diff
@@ -1,4 +1,11 @@
...
+{{top-bar
+	logo-src="/front/vendor/wikia-style-guide/dist/svg/wikia_logo.svg"
+	logo-href="/"
+}}
+{{discussion-header
```
mklucsarits: This header should not be in index.hbs. It's only intended for the post list page (forum.hbs).

front/templates/main/discussion/index.hbs
lizlux: Changes are intentional at this point. Now I'm just removing the top-bar. This page doesn't load as far as I can tell so I can't test it.

gulp/tasks/build-vendor.js
lizlux: Yeah, we need this for shadow dom support in non-chrome.

server/routes.ts
lizlux: As far as I could see, this was the intent, since it's the only way that works/makes sense.

front/templates/main/discussion/forum.hbs
```diff
@@ -1,14 +1,15 @@
...
-	sortBy=sortBy
-}}
-{{forum-wrapper
-	goToPost='goToPost'
-	loadPage='loadPage'
-	name=model.name
-	posts=model.posts
-	postsDisplayed=model.posts.length
-	totalPosts=model.totalPosts
+<div class="discussion">
```
mklucsarits: I don't understand why this is here. There is already a <div class="discussion"> inside forum-wrapper.hbs. If you need another div to wrap these components, please don't use the ```discussion``` CSS class, because it will interfere with existing styles using the ```.discussion``` selector.

front/scripts/main/routes/DiscussionIndexRoute.ts
```diff
@@ -2,8 +2,14 @@
 /// <reference path="../mixins/FullPageMixin.ts"/>
```
mklucsarits: Please remove this reference here, too.

front/styles/main/module/_discussion.scss
```diff
@@ -1,8 +1,10 @@
+$horizontal-padding: 15px;
```
mklucsarits: Are these padding and width changes needed? Seems like they just support the ```<div class="discussion">``` element you added which also doesn't seem necessary. Can it all be removed and still work?

lizlux: The padding is no longer around the entire wrapper, it's in the modules themselves. This seems to work better with the existing page structure on mobile. We may want to change it for desktop. 

I removed the superfluous `div.discussion` instances too.

server/views/_layouts/ember-main.hbs
lizlux: Hm, not sure where to put it, though I can compress it a bit so it doesn't take up so much space on the page. Polymer settings need to be called before the html import, so it's awkward to add this to the async-loaded js. Not sure it belongs in baseline.js either b/c that's code that is, in theory, shared between all mercury apps (like auth), and they don't all necessarily use polymer.

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -4,6 +4,11 @@
 'use strict';
 
 App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
+	tagName: 'top-bar',
+	attributeBindings: ['logo-href', 'show-user-status'],
+	'logo-href': Em.getWithDefault(Mercury, 'wiki.homepage', 'http://www.wikia.com'),
+	'show-user-status': true,
```
rogatty: Why not `showUserStatus`?

front/scripts/main/routes/DiscussionForumRoute.ts
```diff
@@ -1,8 +1,7 @@
 /// <reference path="../app.ts" />
-/// <reference path="../mixins/FullPageMixin.ts"/>
 'use strict';
 
-App.DiscussionForumRoute = Em.Route.extend(App.FullPageMixin, {
```
rogatty: `App.UseNewNavMixin`?

front/styles/main/component/_discussion-header.scss
```diff
@@ -4,10 +4,11 @@ $discussion-header-color-text: #fff;
 .discussion-header {
 	color: $discussion-header-color-text;
 	margin-top: $global-nav-height;
+	max-width: inherit;
 	position: fixed;
 	top: 0;
 	width: 100%;
-	z-index: $z-site-head;
+	z-index: $z-site-head - 1; // drop-downs from top-bar component display over the discussion header
```
rogatty: https://github.com/Wikia/style-guide/blob/dev/src/scss/lib/_settings.scss#L701
> Never --ever-- implement z-index that doesn't use this scale, if you have a use-case that for some reason falls out of this scale, ask for guidance before implementing one-offs.

/cc @kenkouot

front/styles/main/module/_discussion.scss
```diff
@@ -30,6 +32,8 @@
 	.discussion-reply {
 		border-top: 1px solid #ccc;
 		padding-top: 12px;
+		padding-left: $horizontal-padding;
+		padding-right: $horizontal-padding;
```
rogatty: How about:
```
padding: {
  left: ...
  right: ...
  top: ...
}
```

server/views/_layouts/ember-main.hbs
```diff
@@ -14,6 +15,10 @@
 			<meta name="fragment" content="!">
 		{{/if}}
 
+		<!-- Web components and library import -->
+		<script> window.Polymer = window.Polymer || {}; window.Polymer.dom = 'shadow'; </script>
+		<link rel="import" href="/front/vendor/wikia-style-guide/dist/components/import-built.html" async>
```
rogatty: Would putting it in `_partials/web-components.hbs` make sense?

front/scripts/main/components/TopBarComponent.ts
lizlux: Learned something new... if you set the bound attribute to `logo-href`, Ember adds it as the element's attribute in the DOM. If you set it to `logoHref`, it actually becomes a property and not an attribute, so you can't see it in the DOM. The functionality is the same though either way. At the time I was assuming that since I didn't see the attribute in the DOM, it wasn't being applied.  

Turns out I don't need `show-user-status` after a change in the `top-bar` component but i've updated `logo-href` to `logoHref`.

front/styles/main/module/_discussion.scss
mklucsarits: OK, I guess I don't understand how the padding changes are in scope for this pull request. Was the padding affected by the top-bar changes?

kenkouot:
LGTM

server/views/_layouts/ember-main.hbs
lizlux: Sounds good. I'll look into this in a separate PR.

front/styles/main/module/_discussion.scss
lizlux: Resolved offline.

front/scripts/main/mixins/UseNewNavMixin.ts
```diff
@@ -0,0 +1,15 @@
...
+'use strict';
+
+/**
+ * As we roll out the new top-bar custom component, we'll be controlling it's use via the `useNewNav` property on the
+ * ApplicationController. Once we've rolled it out everywhere, we can remove this code.
+ */
+App.UseNewNavMixin = Em.Mixin.create({
+	activate: function (): void {
+		this.controllerFor('application').set('useNewNav', true);
+	},
```
rogatty: We probably should add `this._super();` here and in `deactivate`

lizlux: This would only make sense if we included the mixin as the second param of `Em.Route.extend`, which we don't typically do. IMO that would make it too hidden underneath all the route properties. Unfortunately, whichever route is using the mixin will have to call `this._super()` if it also has an `activate` or `deactivate` method.

rogatty: If someone adds `activate` to `DiscussionForumRoute` and wants to run also `activate` from `UseNewNavMixin` then (s)he should use `this._super();` in the route.

Altough if someone adds new `DiscussionRoute` which would be the parent for all discussion related routes and if it had `activate` then the only way to run it would be adding `this._super();` here. I'm thinking ahead, maybe it won't ever be the case ;)

lizlux: Ah i see what you're saying. That's an interesting idea.

rogatty: I've checked and what happens now is:
- `ApplicationRoute#activate` runs
- `UseNewNavMixin#activate` runs

After adding `activate () {}` to `ApplicationRoute` (which is already a parent of `DiscussionForumRoute`) and `DiscussionForumRoute`:
- `ApplicationRoute#activate` runs
- `DiscussionForumRoute#activate` runs

After adding `this._super()` to previously added `DiscussionForumRoute#activate`:
- `ApplicationRoute#activate` runs
- `DiscussionForumRoute#activate` runs
- `UseNewNavMixin#activate` runs

lizlux: Good call. https://github.com/Wikia/mercury/commit/b9fa06d241fec7f72e394406f8c1a6cae930871d

PR #1363, opened 2015-09-15
INT-136: Login state support for Japan homepage

perjg:
Read helios access_token cookie, and then call info to get token validity. Also use the return value from info to translate user id into user name.

Refactored facets to all use the auth_token cookie.

Missing: Requests are made every time. Should store user name into session cookie. Will do this in separate PR.


homepage/front/scripts/main/main.ts
```diff
@@ -58,7 +58,12 @@ $('.search-wikia').click(function(event) : void {
 });
 
 $('#loginIcon').click(function(event) : void {
-	window.location.href = getGlobals().loginUrl;
+	if ($(document).width() < 710) {
+		$('#userInfoToggle').toggle();
```
bkoval: Can you cache $('#userInfoToggle') in a var and use it inside a handler instead of querying dom on every click?

homepage/server/auth.js
```diff
@@ -0,0 +1,70 @@
...
+Auth.prototype.validateUser = function (username) {
+	var url = this.baseUrl + 'username/validation?' +
+		'username=' + username;
+
+	return requestWrapper(url);
+};
+
+Auth.prototype.getUserName = function (heliosInfoResponse) {
+	var url = this.servicesUrl + 'user-attribute/user/' +
+		heliosInfoResponse.user_id  + '/attr/username'; // jshint ignore:line
```
bkoval: Would jshint shout if you'd call heliosInfoResponse['user_id'] instead?

homepage/server/facets/globals.js
```diff
@@ -7,13 +7,10 @@
 var util = require('../util');
 
 function globals(request, reply) {
-	var loggedIn = util.getLoginState(),
-		loginUrl = util.getLoginUrl().replace(/&amp;/g, '&'),
+	var loginUrl = util.getLoginUrl().replace(/&amp;/g, '&'),
```
bkoval: Could it be the case to use decodeUriComponent?

homepage/server/util.js
```diff
@@ -25,14 +29,23 @@ exports.getUserLocale = function (/*request*/) {
...
-exports.getLoginState = function (/*request*/) {
-	// TODO: Parse access_token cookie from Helios
-	return false;
-};
-
-exports.getUserName = function (/*request*/) {
-	// TODO: Parse access_token cookie form Helios and get user information
-	return 'Test';
+exports.getLoginState = function (request) {
+	console.log('getLoginState');
```
bkoval: Forgotten development leftovers?

homepage/server/util.js
```diff
@@ -43,14 +56,26 @@ exports.getSignupUrl = function () {
...
+		});
+
+		reply.view(view, combinedData);
+	}
+
+	this.getLoginState(request).then(function (data) {
+		return auth.getUserName(data);
+	}).then(function (data) {
+		renderView(true, data.value);
+	}).catch(function (/*error*/) {
```
bkoval: Leftover "/*error*/" comment? I think "catch" is descriptive enough to ensure it's an error handler :)

homepage/server/auth.js
perjg: Tried that... then I just get "['user_id'] is better written in dot notation." instead :)

homepage/server/facets/globals.js
perjg: Actually, this is leftover code that isn't needed any more, I'll just remove the .replace part entirely.

homepage/server/util.js
perjg: It's a bit confusing, so I'll remove it.

The catch function has an error parameter, but if I just write 'function (error)' and don't use the parameter, jslint will flag it. This was just my workaround for showing that the function has a parameter that I don't use, not to show that you get here in an error situation (which as you say is obvious due to this being in the catch block.

homepage/front/scripts/main/main.ts
perjg: I don't know if we have a specific policy for this kind of situation, but since this is only called when the user clicks the icon, there shouldn't really be any performance impact by looking up the selector every time.

PR #1359, opened 2015-09-11
SOC-1119 Styles for post list on desktop

mklucsarits:
These changes add a CSS media query for desktop discussion styles, and override styles for the forum post list page.

https://wikia-inc.atlassian.net/browse/SOC-1119

@garthwebb @bkoval

PR #1367, opened 2015-09-15
SOC-1289 Header (hero unit) for desktop posts list page

bkoval:
@garthwebb @mklucsarits 

https://wikia-inc.atlassian.net/browse/SOC-1289

Please, do not merge until #1359 is merged.

This is still a work In progress so I expect it to do at least two rounds of CR. Also, I'm using style-guide fonts here so they might get changed if design review will lead to it.


front/locales/en/main.json
```diff
@@ -142,6 +142,8 @@
 	"discussion": {
 		"share-default-title": "Discussion on __siteName__",
 		"sort-by-latest": "Latest",
-		"sort-by-trending": "Trending"
+		"sort-by-trending": "Trending",
+		"discussions-title": "Discussions",
```
mklucsarits: I believe this title will need to come from the API response -- it will be the forum title, which doesn't need to be translated. Looks like it is the "name" attribute in the API response, which would come from DiscussionForumModel.

front/styles/main/component/_discussion-hero-unit.scss
```diff
@@ -0,0 +1,30 @@
+$discussion-hero-unit-text-color: #fff;
+
+.discussion-hero-unit {
+	color: $discussion-hero-unit-text-color;
+	display: none;
+	padding-bottom: 37px;
+	//height of the top-bar = 41 + 37px of internal padding
```
mklucsarits: I added a SCSS variable to _wikia-variables.scss called ```$global-nav-height```. This corresponds to the new global nav's height (;)), which you can use here in a calculation instead of hard-coding it.

front/locales/en/main.json
bkoval: Thanks, seems like I've misunderstood the concept :)

front/templates/main/components/post-detail.hbs
```diff
@@ -13,11 +13,11 @@
 	{{{content}}}
 </div>
 <div class="row">
-	<div class="small-5 columns upvote count">
+	<div class="small-4 columns upvote count">
```
bkoval: Oh, I missed that when submitting the pull request, I'll make sure to revert this change.

garthwebb:
Hi @bkoval, so a general note; it looks like you've created a DiscussionHeroUnit component which is different and outside the existing DiscussionHeader.  The content you add for this ticket should be within this header which should hopefully allow you to benefit from some of the common functionality required of all things in the header.

For example, the discussion title and post count you added should only be visible on the *desktop* post list page and should go away on the post detail page and shouldn't be visible anywhere in the *mobile* view.

bkoval:
@garthwebb I reverted the merge of the top-bar branch as it caused a lot of conflicts with the dev branch, a target of this pull request. 

As for the separate component, I decided to go this way for a variety fo reasons:

1) Mobile header is a headroom component, desktop one isn't. This means we'd have to watch for resize event in javascript (performance downgrade) and add / remove "headroom" class which we currently don't control in Ember so we'd need to introduce refactoring. Also, mobile header has a bit of styles that are hard to override (e.g. defined height)

2) Mobile and desktop components don't share much functionality except showing forum name, which is stored in the forum model anyway

3) Hero unit component doesn't necessarily need to be used only on desktop screens, could be used someday for mobile as well

4) It's fairly easy to refactor the code to use just one component in case we have any issues with having two, but this way, especially at the beginning of developing the feature for 2 platforms, feels like it's giving more clarity for developers (two separate, small templates instead of a big one).

Let's keep this discussion going, I think we should tackle these problems at the beginning of the desktop development so we have some rules to obey in the future, for the other components. Adding @mklucsarits @kvas-damian @rafalkalinski and @lizlux to the loop

front/locales/en/main.json
```diff
@@ -142,6 +142,8 @@
 	"discussion": {
 		"share-default-title": "Discussion on __siteName__",
 		"sort-by-latest": "Latest",
+		"sort-by-trending": "Trending",
+		"number-of-posts": "__numberOfPosts__ Posts"
```
garthwebb: Missing comma here

mklucsarits:
@bkoval I was initially in favor of using a unified component for both, but you make good points, and I think now that technically it does make sense for two components. However, I'm still concerned about the extra markup loaded for the hero header on mobile. For mobile users, we should make it a priority to load only what's needed, to increase load time for the user. Even if a hidden component is only a few bytes or kb, if we keep adding more and more, the extra size will be significant. So with that in mind, what do you think about lazy-loading the header component, depending on desktop versus mobile? We could use the ViewportMixin to observe screen width changes.

front/locales/en/main.json
bkoval: Aaargh, merge went wrong, thanks!

vforge:
@bkoval :+1: Those are two separate components, not one - as you discussed, both are used in slightly different context (device).

bognix:
@bkoval :+1:  @vforge :+1: 
What is more if they don't share the same logic (except Forum which is in ForumModel) there is no point to couple them at this stage

rogatty:
@bkoval 
>I think we should tackle these problems at the beginning of the desktop development so we have some rules to obey in the future, for the other components.

If it makes sense to separate these two components then fine, do it. But let's not make a rule like "every component needs two separate versions". I like "let's be reasonable and approach it case by case" better ;)

If Headroom is an issue then maybe we should get rid of it? I created a spike already, see https://wikia-inc.atlassian.net/browse/XW-367

front/styles/main/component/_discussion-hero-unit.scss
```diff
@@ -0,0 +1,31 @@
+$discussion-hero-unit-text-color: #fff;
```
Warkot: Please use `$white` variable.

Warkot:
Forcing anything to be coupled when it shouldn't be might lead to suboptimal code with suboptimal performance. As @rogatty said, common sense is what devs should use. :+1:

bkoval:
Thanks for the feedback guys! I totally agree with preserving common sense as our primary source of knowledge when making decisions like this one!

front/styles/main/component/_discussion-hero-unit.scss
```diff
@@ -0,0 +1,31 @@
+$discussion-hero-unit-text-color: $white;
```
mklucsarits: Please whenever possible use the Wikia-specific colors in the style guide, which is ```$color-white``` in this instance. The Foundation named colors were not intended to be kept forever, and we may remove them at some point.

And, related to this, if we are using a color which hasn't been defined in the style guide for Wikia (not by Foundation), bring it up with UX team, and if the UX team intends on using that color going forward, make a note that it should be added to the style guide. But in general, please do not use the Foundation-defined color variables. /cc @Warkot

Warkot: Valuable info, thanks for cc. :+1:

front/templates/main/discussion/forum.hbs
```diff
@@ -2,6 +2,10 @@
 	setSortBy='setSortBy'
 	sortBy=sortBy
 }}
+{{discussion-hero-unit
+	numberOfPosts=model.totalPosts
+	name=name
```
garthwebb: Model is provided by the route, but I don't see where name is coming from.  Maybe you want `model.name` here?

garthwebb:
LGTM

mklucsarits:
LGTM

PR #1370, opened 2015-09-16
SOC-1120 Change styling of discussion post sorting between desktop and mobile

mklucsarits:
https://wikia-inc.atlassian.net/browse/SOC-1120

These changes extract the discussion post list sorting into a component outside the discussion header. The goal is to extensively change the styling between mobile and desktop views (CSS) while using the same components underneath (JS). While the sorting markup has been moved to its own component template, much of the sort logic had to be moved outside the component to the route and controller. I was not able to completely compartmentalize the sorting while making it work the way we need to. I think this is alright because sorting is integral to the forum post list page. However, as you review this code, please think of ways this code could be better organized.

@bkoval @kvas-damian


kvas-damian:
Please fix integration tests

front/scripts/main/components/DiscussionSortComponent.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	didInsertElement: function (): void {
+		this.updateActive();
+		this._super();
+	},
+
+	updateActive: function (): void {
+		// Add the 'active' CSS class to the sort tab that's active,
+		// but right now this only applies to desktop styling.
+		var $discussionSort = $('.discussion-sort');
```
kvas-damian: Can we cache this selector in a variable and avoid querying DOM on every funtcion call?

front/scripts/main/components/DiscussionSortComponent.ts
```diff
@@ -0,0 +1,36 @@
...
+
+		$discussionSort.find('li').removeClass('active');
```
kvas-damian: I know there are only two elements in the list, but my dark side tells me that ```find('li.active').removeClass('active');``` should be little bit faster ;)

front/scripts/main/controllers/DiscussionForumController.ts
```diff
@@ -1,5 +1,40 @@
...
+			filtered = sortTypes.filter((obj: any): boolean => {
+				return obj.name === this.get('sortBy');
+			});
+
+			return filtered.length ? filtered[0].messageKey :
+				sortTypes[0].messageKey;
+	}),
+
+	actions: {
+		showSortSelector(): void {
```
kvas-damian: ```selector``` in this method name could be little confusing - can suggest DOM selector. What do you think about ```showSortModule```?

front/scripts/main/routes/DiscussionForumRoute.ts
```diff
@@ -12,7 +13,29 @@ App.DiscussionForumRoute = Em.Route.extend(App.UseNewNavMixin, {
 
 	setupController(controller: Em.Controller, model: Em.Object, transition: EmberStates.Transition) {
 		this._super(controller, model, transition);
-		controller.set('sortBy', transition.params['discussion.forum'].sortBy);
+		controller.set('sortBy', transition.params['discussion.forum'].sortBy || controller.get('sortTypes')[0].name);
+	},
+
+	viewportObserver: Em.observer('viewportDimensions.width', function (): void {
+		this.updateSortVisibility();
```
kvas-damian: Can that be handled in CSS using mediaqueries?

front/scripts/main/components/DiscussionSortComponent.ts
bkoval: I am wondering if we need to use it at all. Could we possibly use ember binding / action bubbling to communicate this change to discussion sort?

front/styles/main/component/_discussion-sort.scss
```diff
@@ -0,0 +1,56 @@
...
+	position: absolute;
+	right: 5px;
+	text-align: right;
+	top: 5px;
+	width: 110px;
+
+	li {
+		cursor: pointer;
+		font-size: $type-smallest;
+		line-height: 28px;
```
kvas-damian: Can we use styleguide value here? I think ```$line-height-smaller``` should be ok.

front/scripts/main/controllers/DiscussionForumController.ts
```diff
@@ -1,5 +1,40 @@
...
+		},
+		{
+			name: 'trending',
+			messageKey: 'discussion.sort-by-trending'
+		}
+	],
+
+	sortMessageKey: Em.computed('sortBy', function (): string {
+		var sortTypes = this.get('sortTypes'),
+			filtered = sortTypes.filter((obj: any): boolean => {
```
bkoval: How about making sortTypes an object with names as keys? Seems like it could speed up a process of finding the right sort type here.

front/scripts/main/routes/DiscussionForumRoute.ts
bkoval: Yeah, I'm also voting for using CSS everywhere we can (can we do it here @mklucsarits ?), but if there's a className switch involved, we need to use javascript.

front/scripts/main/routes/DiscussionForumRoute.ts
```diff
@@ -12,7 +13,29 @@ App.DiscussionForumRoute = Em.Route.extend(App.UseNewNavMixin, {
...
+	}),
+
+	updateSortVisibility: function (): void {
```
bkoval: Shouldn't this be implemented in the controller? It operates only on controller properties.

front/scripts/main/routes/DiscussionForumRoute.ts
```diff
@@ -25,7 +48,14 @@ App.DiscussionForumRoute = Em.Route.extend(App.UseNewNavMixin, {
 		},
 
 		setSortBy: function (sortBy: string): void {
-			this.controllerFor('discussionForum').set('sortBy', sortBy);
+			var controller = this.controllerFor('discussionForum');
```
bkoval: Same here, I think we can trigger a route transition from a controller, and feels like data manipulation (especially through controller properties) belongs more to the controller than route handler. What do you think?

front/styles/main/component/_discussion-sort.scss
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+@for $i from 1 through length($verticals) {
+	.#{nth($verticals, $i)}-vertical .discussion-sort {
+		color: #{nth($vertical-colors, $i)};
+	}
+
+	@media #{$desktop-range} {
```
kvas-damian: My gulp-sass can't resolve this variable.

kvas-damian: Sass isn't great at combining media queries, so it'll generate separate media query for every vertical. Please consider having separate loop inside ```@media```.

front/styles/main/component/_discussion-sort.scss
```diff
@@ -0,0 +1,56 @@
...
+		text-align: center;
+		width: 507px;
+
+		li {
+			border-bottom: 4px solid $border-color;
+			color: $border-color;
+			float: left;
+			font-size: $type-smaller;
+			height: 50px;
+			line-height: 50px;
```
kvas-damian: Please consider using styleguide values.

front/scripts/main/controllers/DiscussionForumController.ts
mklucsarits: Unfortunately I don't think it's possible to loop through them then in the template (although I could be wrong).

front/scripts/main/controllers/DiscussionForumController.ts
mklucsarits: Agreed. Even better, I'll call these "...SortComponent" since it refers directly to the DiscussionSortComponent.

front/scripts/main/routes/DiscussionForumRoute.ts
mklucsarits: Yes, I'll remove all this Viewport logic. It can be always be shown on desktop with a simple media query.

front/scripts/main/routes/DiscussionForumRoute.ts
mklucsarits: I'm trying to keep as much as possible in the Route class for now, because soon we're moving away from controllers in Ember, and I thought it would ease the transition if as much logic as possible is in the route instead.

front/styles/main/component/_discussion-sort.scss
mklucsarits: The reason I would rather not do that is that these line heights I'm using here are for vertically centering the text, and has nothing to do with line height in the context of a paragraph, which is what the style guide values are intended for. So I don't want to link this line height value with a value that is related to display of a font in the context of a paragraph (which could change while these values should not).

front/scripts/main/controllers/DiscussionForumController.ts
kvas-damian: :+1:

front/styles/main/component/_discussion-sort.scss
kvas-damian: Makes sense.

front/scripts/main/controllers/DiscussionForumController.ts
bkoval: :+1:

bkoval:
Lookin' good! :up:

front/styles/main/component/_discussion-sort.scss
```diff
@@ -0,0 +1,64 @@
...
+		margin: $global-nav-height auto 10px auto;
+		padding: 0;
+		position: static;
+		text-align: center;
+		width: 507px;
+		z-index: auto;
+
+		&.hide {
+			// These are needed to override the hide class
+			display: block !important;
```
kvas-damian: This is really not the code I was expecting here.
Please use different class name to hide elements on mobile + try to avoid ```important``` as it's really hard to override.

kvas-damian:
LGTM

front/scripts/main/components/DiscussionSortComponent.ts
```diff
@@ -0,0 +1,43 @@
...
+
+App.DiscussionSortComponent = Em.Component.extend({
+	classNames: ['discussion-sort', 'clearfix'],
+	classNameBindings: ['sortVisible::mobile-hidden'],
+	tagName: 'ul',
+
+	// Whether the component is currently visible
+	sortVisible: false,
+	// jQuery object for this component
+	$discussionSort: null,
```
rogatty: You can use `this.$()` which is the same thing.

front/scripts/main/routes/DiscussionForumRoute.ts
rogatty: @mklucsarits yup, thanks for that.

front/scripts/main/components/DiscussionSortComponent.ts
```diff
@@ -0,0 +1,43 @@
...
+	updateActive: function (): void {
+		// Add the 'active' CSS class to the sort tab that's active,
+		// but right now this only applies to desktop styling.
+		var $discussionSort: JQuery = this.get('$discussionSort');
+
+		if ($discussionSort === null) {
+			return;
+		}
+
+		$discussionSort.find('li.active').removeClass('active');
```
bkoval: Yeah, I've just spoken to @rogatty as he came to a conclusion that this might be implemented using Ember bindings instead of querying DOM like this.

I gotta agree with it - if it's possible, we should avoid using jQuery directly as much as possible. Using Ember bindings should make it easier for us to manipulate this code in the future.

mklucsarits: I agree using jQuery is not ideal. I'll try to get an Ember/CSS-only solution to this.

PR #1373, opened 2015-09-16
XW-239 Handle preview as one of possible envs in mercury

bognix:
Some requests on Mercury were hitting production instead of preview.
Root cause of this issue was WIKIA_ENVIRONMENT which was set to prod on staging-s3

Chef change: https://github.com/Wikia/chef-repo/pull/6558

### DIFF
--- a/gulp/utils/environment.js
+++ b/gulp/utils/environment.js
@@ -3,8 +3,6 @@ var gutil = require('gulp-util'),
 
 environment = (environment === 'production' ) ? 'prod' : environment;
 
-process.env.WIKIA_ENVIRONMENT = environment;
-
 module.exports = {
 	name: environment,
 	isProduction: environment === 'prod'
--- a/server/facets/api/article.ts
+++ b/server/facets/api/article.ts
@@ -30,7 +30,7 @@ function isRequestForRandomTitle (query: any): boolean {
  * @param result
  */
 export function get (request: Hapi.Request,  reply: any): void {
-	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host),
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers['x-original-host']),
 		params: ArticleRequestParams = {
 			wikiDomain: wikiDomain,
 			title: request.params.articleTitle,
--- a/server/facets/api/article.ts
+++ b/server/facets/api/article.ts
@@ -30,7 +30,7 @@ function isRequestForRandomTitle (query: any): boolean {
  * @param result
  */
 export function get (request: Hapi.Request,  reply: any): void {
-	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, Utils.getHostFromRequest(request)),
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request),
 		params: ArticleRequestParams = {
 			wikiDomain: wikiDomain,
 			title: request.params.articleTitle,
--- a/server/facets/api/article.ts
+++ b/server/facets/api/article.ts
@@ -30,7 +30,7 @@ function isRequestForRandomTitle (query: any): boolean {
  * @param result
  */
 export function get (request: Hapi.Request,  reply: any): void {
-	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers['x-original-host']),
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, Utils.getHostFromRequest(request)),
 		params: ArticleRequestParams = {
 			wikiDomain: wikiDomain,
 			title: request.params.articleTitle,
--- a/server/lib/Utils.ts
+++ b/server/lib/Utils.ts
@@ -235,3 +235,7 @@ export function getCDNBaseUrl(localSettings: LocalSettings): String {
 	return localSettings.environment !== Environment.Dev ? localSettings.cdnBaseUrl : ''
 }
 
+export function getHostFromRequest(request: Hapi.Request): String {
+	return request.headers['x-original-host'] || request.headers['host'];
+}
+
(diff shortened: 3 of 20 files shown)


vforge:
LGTM

gulp/utils/environment.js
```diff
@@ -3,8 +3,6 @@ var gutil = require('gulp-util'),
 
 environment = (environment === 'production' ) ? 'prod' : environment;
 
-process.env.WIKIA_ENVIRONMENT = environment;
```
rogatty: @hakubo it's your code but we believe it doesn't do anything good. Just letting you know that it's being removed.

rogatty:
LGTM

server/lib/Utils.ts
```diff
@@ -66,32 +67,16 @@ export function stripDevboxDomain (host: string): string {
...
-}
-
-/**
- * @desc Get fallback domain
- * @returns {string}
- */
-function getFallbackWiki (localSettings: LocalSettings): string {
-	return (localSettings.wikiFallback || 'community');
+function getDevDomainFromWikiName (localSettings: LocalSettings, wikiName: string): string {
+	return localSettings.devboxDomain ?
```
vforge: In what situations you could have `devboxDomain` empty? Is that even possible to have `wikiName + '.wikia-dev.com'` as working address?

server/lib/Utils.ts
```diff
@@ -119,36 +104,24 @@ export function getCachedWikiDomainName (localSettings: LocalSettings, host: str
...
-	if (passThroughEnv.hasOwnProperty(environment)) {
-		if (hostName) {
-			return hostName;
-		}
-		// This fallback is just for mediawikiDomain var in ServerData::createServerData
-		return passThroughEnv[environment].replace('%s', getFallbackWiki(localSettings));
+		match: RegExpMatchArray,
+		environment = localSettings.environment;
+
+	if (environment === Environment.Dev && hostName.indexOf('xip.io') > -1) {
```
vforge: you can use `endsWith` instead of `indexOf`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

rogatty: `http://xip.io.igor.wikia-dev.com`
;)

server/facets/api/article.ts
```diff
@@ -30,7 +30,7 @@ function isRequestForRandomTitle (query: any): boolean {
  * @param result
  */
 export function get (request: Hapi.Request,  reply: any): void {
-	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request.headers.host),
+	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, Utils.getHostFromRequest(request)),
```
rogatty: How about passing `request.headers` and calling `getHostFromRequest` inside `getCachedWikiDomainName`? Less code to write ;)

server/lib/Utils.ts
vforge: Precisely

server/lib/Utils.ts
rogatty: It made more sense in the full version:
```
// This is specific to the devbox setup		
// Curl requests will look like muppet.devname.wikia-dev.com so skip the "devname"		
// Web requests will look like muppet.123.123.123.123.xip.io so add the "devname"		
if (localSettings.devboxDomain && wikiSubDomain.indexOf(localSettings.devboxDomain) === -1) {
```

server/facets/api/article.ts
bognix: thought about it and wasn't sure if I should couple those two.
On the other hand I'm probably optimizing what didn't have to be optimized

server/lib/Utils.ts
bognix: why wikiName should be contained in localSettings.devboxDomain?
Right now this func should create url base on wikiName.
This is the only use case for this function

server/lib/Utils.ts
```diff
@@ -227,9 +199,15 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+		cdnBaseUrl: getCDNBaseUrl(localSettings)
 	};
 }
 
+export function getCDNBaseUrl(localSettings: LocalSettings): String {
+	return localSettings.environment !== Environment.Dev ? localSettings.cdnBaseUrl : ''
+}
+
+export function getHostFromRequest(request: Hapi.Request): string {
+	return request.headers['x-original-host'] || request.headers['host'];
```
rogatty: Please add JSDoc explaining why we're doing this and where is the logic that adds `x-original-host`.

server/lib/Utils.ts
```diff
@@ -227,9 +199,15 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+}
+
```
rogatty: Too much space! Can't look! :see_no_evil:

bognix: It's ok now :)

rogatty:
@bognix please remove all occurences of `wikiFallback`
https://github.com/Wikia/mercury/search?utf8=%E2%9C%93&q=wikiFallback

vforge:
LGTM

rogatty:
LGTM

server/lib/Utils.ts
```diff
@@ -227,9 +185,30 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+ * @param request
+ * @returns {string}
+ */
+export function getHostFromRequest(request: Hapi.Request): string {
+	return request.headers['x-original-host'] || request.headers['host'];
+}
+
+export function isXipHost(localSettings: LocalSettings, hostName: string): boolean {
+	return localSettings.environment === Environment.Dev &&
+		hostName.search(/(?:[\d]{1,3}\.){4}xip\.io$/) !== -1
```
rogatty: From my testing `request.headers.host` contains also port number which would cause this check to fail.

Warkot:
LGTM

PR #1375, opened 2015-09-16
HG-878 fixing locale gulp task and top-nav i18n for anons

lizlux:
The "Register or Login" text didn't match the mocks before, and now it does. Also, 

* Redirect params are added to anon cta links
* User-status component is broken out into it's own component now
* I fixed an issue where editing a locale file wouldn't restart the server, so the changes weren't reflected in the app. That should work more smoothly now. 

Note that click tracking will be handled with https://wikia-inc.atlassian.net/browse/HG-886.


front/locales/en/main.json
```diff
@@ -145,6 +145,8 @@
 		"sort-by-trending": "Trending"
 	},
 	"topBar": {
-		"anon-cta": "Sign In"
+		"register": "Register",
```
kenkouot: register and login are actually part of the user status component right?
also, I think it would be more clear if we labelled keys with their type hinted, eg `register-cta-label`

front/locales/en/main.json
```diff
@@ -145,6 +145,8 @@
...
+		"login": "Login",
+		"cta-separator": "or"
```
bkoval: What happens for the languages which require a different order of words than "Register or Login"? This could be a one translate'able sentence with 2 variables, containing preformatted login and reg links.

Also, do we want to lose /join step and link directly to form pages instead?

lizlux: @bkoval I'm totally open to suggestions as I don't think this is the best way to handle this. I think your suggestion is something I considered at one point, but wasn't sure the best way to implement. Is this an example of what you're suggesting? https://github.com/Wikia/mercury/blob/dev/server/views/_partials/auth/register-fb.hbs#L24. I'll try that.

lizlux: Also, to respond to the `/join` question, on mobile we'll link to `/join` but on desktop we'll link to the login and register pages. I'm working on that now. This PR is actually a WIP so I'll mark it as such.

bkoval: Fantastic!

btw Let me know if I can help with the i18n thing, we spent a bit of time figuring it out :)

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -8,6 +9,43 @@ App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
...
+	}),
+
+	joinHref: Em.computed('redirectString', function () {
+		if (Mercury.wiki.enableNewAuth) {
+			return '/join' + this.get('redirectString');
+		}
+		return '/Special:UserLogin';
+	}),
+
+	redirectString: Em.computed(window.location.href, function () {
```
rogatty: I don't think this is what you want to do. My guess is that Ember resolves value of `window.location.href` and then starts to observe property named like that. So for example if it's `http://wikia.com` then this will be recomputed only if `this.get('http://wikia.com')` changes. So never ;)

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -8,6 +9,43 @@ App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
 	attributeBindings: ['logoHref'],
 	logoHref: Em.getWithDefault(Mercury, 'wiki.homepage', 'http://www.wikia.com'),
 
+	loginHref: Em.computed('redirectString', function () {
```
rogatty: `: string` is missing in all those functions.

front/scripts/main/components/TopBarComponent.ts
```diff
@@ -33,4 +72,36 @@ App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
...
+	//		action: M.trackActions.click,
+	//		category: 'user-login-mobile',
+	//		label: label
+	//	});
+	//
+	//	window.location.href = href;
+	//},
+
+	getUselangParam: function (): string {
+		var lang: string = Mercury.wiki.language.content;
```
rogatty: If wikiVariables is changed then this will crash, that's why I use `Em.get` when going this deep into objects.

front/scripts/main/components/TopBarComponent.ts
bkoval: Maybe title from article controller would be a better choice?

rogatty: Then it would be tightly coupled with articles.

kenkouot: It's a component, I'd make sure that the dependent key was something explicitly passed into the component. Then it goes from coupling to declarative dependency.

lizlux: Trying `router.url`

gulp/tasks/watch.js
```diff
@@ -67,7 +67,7 @@ gulp.task('watch', ['build'], function () {
 		paths.base + '/front/scripts/*.js',
 		paths.base + '/front/styles/*.css',
 		paths.base + '/front/templates/*.js',
-		paths.base + '/front/locales/**/translate.json'
```
lizlux: This fixes the issue where server doesn't restart when editing i18n files.

front/scripts/main/components/UserStatusComponent.ts
```diff
@@ -0,0 +1,93 @@
...
+	}),
+
+	registerHref: Em.computed('redirectString', function (): string {
+		if (Mercury.wiki.enableNewAuth) {
+			return '/register' + this.get('redirectString');
+		}
+		return '/Special:UserSignup';
+	}),
+
+	redirectString: Em.computed('router.url', function (): string {
```
lizlux: this doesn't work. Will have to find a different solution

front/scripts/main/components/UserStatusComponent.ts
```diff
@@ -0,0 +1,93 @@
...
+		return '/Special:UserLogin';
+	}),
+
+	userLoggedIn: Em.computed('currentUser.isAuthenticated', function () {
+		// HTMLBars attribute binding only removes an attribute if it's value is set to null
+		if (this.get('currentUser.isAuthenticated') === false) {
+			return null;
+		}
+		return true;
+	}),
```
rogatty: IMO this should be:
```
if (this.get('currentUser.isAuthenticated') === true) {
    return true;
}
return null;
```

Let's be pessimistic here.

front/scripts/main/components/UserStatusComponent.ts
```diff
@@ -0,0 +1,93 @@
...
+
+	userAvatarSrc: Em.computed('currentUser.avatarPath', function (): string {
+		return this.get('currentUser.avatarPath');
```
rogatty: You can use `userAvatarSrc: Em.computed.oneWay('currentUser.avatarPath')`

front/scripts/main/components/UserStatusComponent.ts
```diff
@@ -0,0 +1,93 @@
...
+	}),
+
+	userName: Em.computed('currentUser.avatarPath', function (): string {
+		return this.get('currentUser.name');
```
rogatty: `userName: Em.computed.oneWay('currentUser.name')`

front/scripts/main/components/UserStatusComponent.ts
```diff
@@ -0,0 +1,108 @@
...
+/// <reference path="../../baseline/mercury/utils/buildUrl.ts" />
+/// <reference path="../../../../typings/i18next/i18next.d.ts" />
+'use strict';
+
+
+App.UserStatusComponent = Em.Component.extend(App.HeadroomMixin, {
+	tagName: 'user-status',
+	attributeBindings: ['userLoggedIn', 'userAvatarSrc', 'userName'],
+	classNames: ['user-status', 'needsclick'],
+	anonAvatarSrc: 'http://wikia.github.io/style-guide/assets/images/icons/icon_avatar.svg',
```
kenkouot: Can we switch to a internal asset

kenkouot:
LGTM

front/scripts/main/components/UserStatusComponent.ts
lizlux: done in 8e13756

PR #1376, opened 2015-09-16
HG-879 - add theme bar for discussion post pages - WIP

katanka:
On the post details page there was a theme-bar below the site-nav component. Now that we're using the top-bar component instead, the theme-bar will have to be re-inserted and styles fixed.

### DIFF
--- a/front/styles/main/_settings.scss
+++ b/front/styles/main/_settings.scss
@@ -13,3 +13,4 @@ $label-color-active: $color-blue-gray-dark;
 $label-color-inactive: $color-blue-gray;
 $label-color-valid: $color-blue-gray-dark;
 $link-color: $color-blue-light;
+$top-bar-height: 46px;
--- a/front/templates/main/components/discussion-header.hbs
+++ b/front/templates/main/components/discussion-header.hbs
@@ -1,9 +1,11 @@
-<div class="header clearfix">
-	<h1>{{siteName}}</h1>
-	<div {{action 'showSortComponent'}} class="sort">
-		<span>{{i18n sortMessageKey}}</span>
-		{{svg 'filter-dropdown-white' viewBox='0 0 7 4' class='icon dropdown'}}
+{{#if showContent}}
+	<div class="header clearfix">
+		<h1>{{siteName}}</h1>
+		<div {{action 'showSortComponent'}} class="sort">
+			<span>{{i18n sortMessageKey}}</span>
+			{{svg 'filter-dropdown-white' viewBox='0 0 7 4' class='icon dropdown'}}
+		</div>
+		<div {{action 'hideSortComponent'}} class="overlay"></div>
 	</div>
-	<div {{action 'hideSortComponent'}} class="overlay"></div>
-</div>
+{{/if}}
 <hr class="theme">
--- a/front/templates/main/discussion/post.hbs
+++ b/front/templates/main/discussion/post.hbs
@@ -1,3 +1,10 @@
+{{discussion-header
+	showContent=false
+	showSortComponent='showSortComponent'
+	hideSortComponent='hideSortComponent'
+	sortBy=sortBy
+	sortMessageKey=sortMessageKey
+}}
 <div class="discussion post">
 	{{post-detail
 		author=model.firstPost.createdBy
--- a/front/locales/en/main.json
+++ b/front/locales/en/main.json
@@ -150,6 +150,6 @@
   "userStatus": {
     "anon-cta": "__registerLink__ or __loginLink__",
     "register-text": "Register",
-    "login-text": "Login"
+    "login-text": "Sign In"
   }
 }
--- a/front/scripts/main/components/TopBarComponent.ts
+++ b/front/scripts/main/components/TopBarComponent.ts
@@ -4,7 +4,7 @@
 /// <reference path="../../../../typings/i18next/i18next.d.ts" />
 'use strict';
 
-App.TopBarComponent = Em.Component.extend(App.HeadroomMixin, {
+App.TopBarComponent = Em.Component.extend({
 	classNames: ['top-bar-component'],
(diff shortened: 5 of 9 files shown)


front/styles/main/component/_theme-bar.scss
```diff
@@ -0,0 +1,13 @@
+$theme-bar-height: 10px;
+
+.theme {
```
lizlux: From Andrew: The css re-implements some head-room functionality. This could possibly be improved upon.

kenkouot: I still don't get *why* it reimplements headroom functionality?

front/templates/main/application.hbs
```diff
@@ -24,11 +27,13 @@
 	{{/if}}
 
 	{{#if enableSharingHeader}}
-		{{share-feature
-			smartBannerVisible=smartBannerVisible
-			title=currentTitle
-			language=language
-		}}
+		{{#unless themeBar}}
```
kenkouot: A comment here explaining the intended results would be good.

front/styles/main/component/_theme-bar.scss
kenkouot: this class name should be `.theme-bar`

front/templates/main/application.hbs
```diff
@@ -14,6 +14,9 @@
 
 	{{#if useNewNav}}
 		{{top-bar}}
+		{{#if themeBar}}
+			{{theme-bar}}
```
rogatty: Seems strange. How about:
```
{{#if useThemeBar}}
  {{theme-bar}}
```
?

front/templates/main/components/theme-bar.hbs
```diff
@@ -0,0 +1 @@
+
```
rogatty: If it's gonna stay empty then there is no need for this file to exist.

front/scripts/main/components/ThemeBarComponent.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../main/mixins/HeadroomMixin.ts" />
+'use strict';
+
+App.ThemeBarComponent = Em.Component.extend(App.HeadroomMixin, {
+    tagName: 'hr',
+    classNameBindings: ['theme-bar'],
+    theme-bar: true
```
lizlux: `theme-bar` needs to be in quotes b/c you can't have `-` inside object keys.

lizlux: Actually, to set the class just use `className`. (Discussed offline)

front/styles/main/component/_site-head.scss
```diff
@@ -73,12 +73,6 @@
...
 
-hr.theme {
-	border-width: 0;
-	height: 10px;
-	margin: 0;
-}
-
 @for $i from 1 through length($verticals) {
 	.#{nth($verticals, $i)}-vertical {
 		hr.theme {
```
lizlux: Looks like these styles aren't being brought in

front/templates/main/application.hbs
```diff
@@ -24,11 +27,16 @@
...
 
 	{{#if enableSharingHeader}}
-		{{share-feature
-			smartBannerVisible=smartBannerVisible
-			title=currentTitle
-			language=language
-		}}
+		{{!--	Because the sharing header conflicts with the theme bar,
+				they should not both be active simultaneously. This should
+				only happen in the discussions feature. --}}
```
mklucsarits: There should be no either-or logic for the theme bar and the sharing header. On the discussions post detail page, the theme bar and the sharing header must both be able to show.

front/templates/main/components/discussion-header.hbs
```diff
@@ -10,4 +10,4 @@
 		{{svg 'filter-dropdown-white' viewBox='0 0 7 4' class='icon dropdown'}}
 	</div>
 </div>
-<hr class="theme">
+<hr class="theme-bar">
```
mklucsarits: Replace with component.

front/styles/main/_settings.scss
```diff
@@ -13,3 +13,4 @@ $label-color-active: $color-blue-gray-dark;
 $label-color-inactive: $color-blue-gray;
 $label-color-valid: $color-blue-gray-dark;
 $link-color: $color-blue-light;
+$top-bar-height: 46px;
```
mklucsarits: Why make this change? ```$global-nav-height``` is supposed to represent the height of the global nav, which is the top-bar component. If the height has changed, just change the value of ```$global-nav-height``` instead.

front/templates/main/discussion/post.hbs
```diff
@@ -1,3 +1,10 @@
+{{discussion-header
```
mklucsarits: There should be no need for adding the discussion-header on the post detail page, or this logic which shows or doesn't show the discussion-header content, if theme-bar is its own component. What happened to the theme-bar component approach?

front/styles/main/_settings.scss
lizlux: This is the height of the new top bar but not necessarily that of the old site-head component. I suppose I can confirm with the designer (Witcher) to make sure he understands that the two versions of the top nav will have different heights during the transition, if you think it will be an issue.

mklucsarits: ```$global-nav-height``` is meant only for the top-bar nav. It has no relation to the site-head component, and should be used as such. If we need to add a comment to the variable declaration to clarify that, then we should do that, but we shouldn't have duplicate variables for the same thing.

lizlux: Ah i see, okay. I'll update the variable to be `$top-bar-height` everywhere so others don't make the same mistake. Thanks for catching!

PR #1381, opened 2015-09-17
XW-236 Redirect to canonical host if needed

rogatty:
https://wikia-inc.atlassian.net/browse/XW-236

The main change is adding `redirectToCanonicalHostIfNeeded` method to utils and then using it after each `getWikiVariables` call. To do that I had to refactor a lot of code to get rid of callbacks and expose promises.

I hope that the server code is easier to understand now as there is no longer :spaghetti: of callbacks.


front/scripts/main/models/ArticleModel.ts
```diff
@@ -128,7 +128,7 @@ App.ArticleModel.reopenClass({
 		M.prop('articleContentPreloadedInDOM', false);
 
 		// On first page load the article content is available only in HTML
-		article.content = $('#preloadedContent').html();
+		article.content = $.trim($('#preloadedContent').html());
```
rogatty: I fixed the issue where user would get an empty article instead of message that it's empty when opening it directly. There was a white space in there.

server/facets/editorPreview.ts
```diff
@@ -15,45 +15,53 @@ function editorPreview (request: Hapi.Request, reply: Hapi.Response): void {
...
+			// TODO: why is this needed for the images to load?
+			result.tracking = localSettings.tracking;
+
+			reply.view('application', result);
+		})
+		.catch((error: any) => {
+			reply.view('application', {
+				error
+			}, {
+				layout: 'empty'
```
rogatty: @gabrys I couldn't find a way to test this route. Trying to send POST returned 403 every time I tried. I agreed with @adamkarminski that it's not essential to work on it right now as it's not used anyway. Feel free to reach out if I broke anything here and you need my help to fix it.

server/facets/api/article.ts
```diff
@@ -56,15 +59,12 @@ export function get (request: Hapi.Request,  reply: any): void {
...
-
-		Caching.disableCache(response);
-	});
+	article
+		.getArticle()
+		.then((result: any): void => {
+			handleArticleResponse(reply, result, allowCache);
+		})
+		.catch((result: any): void => {
+			handleArticleResponse(reply, result, allowCache);
```
bognix: why we are doing the same thing in catch? Could you add comment what is changing between try and catch (parameters values perhaps?)

front/scripts/main/models/ArticleModel.ts
vforge: :+1:

vforge:
more Ember :sparkles: magic

LGTM tho

server/lib/Utils.ts
```diff
@@ -233,3 +236,43 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+ * @param wikiVariables
+ * @throws RedirectedToCanonicalHost
+ */
+export function redirectToCanonicalHostIfNeeded(
+	localSettings: LocalSettings, request: Hapi.Request, reply: Hapi.Response, wikiVariables: any
+): void {
+	var requestedHost = getCachedWikiDomainName(localSettings, request.headers.host),
+		canonicalHost = Url.parse(wikiVariables.basePath).hostname,
+		isDev = localSettings.environment === Environment.Dev,
+		localPattern = /(?:[\d]{1,3}\.){4}xip\.io$/,
```
bognix: do you think checking for indexOf('xip.io') > -1 isn't enough?
I do the same check in my PR and I'm checking the indexOf.
We should unify how we do it :)

server/lib/Utils.ts
```diff
@@ -233,3 +236,43 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+	}
+}
+
+export class RedirectedToCanonicalHost {
+	constructor () {
+		Error.apply(this, arguments);
+	}
+}
+
+RedirectedToCanonicalHost.prototype = Object.create(Error.prototype);
```
bognix: nice idea :+1:

server/lib/Utils.ts
vforge: @bognix I've asked about your solution in your PR ;)

But I agree on common solution - let's use `endWith` - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

bognix: Created generic function here: https://github.com/Wikia/mercury/commit/50cdc93c7bbc0e16ab0316ae64f1d9fc8e9309f7

vforge: :+1:

Warkot:
Nice. :sunglasses: LGTM

server/lib/MediaWiki.ts
```diff
@@ -69,7 +69,7 @@ export class WikiRequest extends BaseRequest {
 	 *
 	 * @return {Promise<any>}
 	 */
-	getWikiVariables (): Promise<any> {
+	wikiVariables (): Promise<any> {
```
bognix: how about loadWikiVariables? name wikiVariables is not very verbose for me

test/specs/server/lib/Utils.js
```diff
@@ -232,3 +232,154 @@ test('parseQueryParams', function () {
...
+		{
+			environment: Environment.Prod,
+			hostName: 'muppet.127.0.0.1.xip.io',
+			expected: false
+		},
+		{
+			environment: Environment.Dev,
+			hostName: 'muppet.xip.io',
+			expected: false
+		}
```
bognix: how about one more test case with 
```
hostName: 'xip.io.wikia.com'
```
?

server/lib/MediaWiki.ts
rogatty: I just wanted to be consistent with the rest of file, see https://github.com/Wikia/mercury/blob/XW-236/server/lib/MediaWiki.ts#L85-L168

Do you think we should rename all of them?

test/specs/server/lib/Utils.js
```diff
@@ -232,3 +232,154 @@ test('parseQueryParams', function () {
...
+	];
+
+	testCases.forEach(function (testCase) {
+		equal(global.isXipHost({environment: testCase.environment}, testCase.hostName), testCase.expected);
+	});
+});
+
+test('redirectToCanonicalHostIfNeeded', function () {
```
bognix: :+1:

server/lib/MediaWiki.ts
bognix: didn't have the full context. If that's the case I'm fine with the name

server/facets/showArticle.ts
```diff
@@ -42,24 +44,43 @@ function showArticle (request: Hapi.Request, reply: Hapi.Response): void {
...
-					onArticleResponse(request, reply, error, result, allowCache);
-				});
-			}
-		});
+
+				return Promise.join(wikiVariables, article.getArticle());
+			})
+			.spread((wikiVariables: any, article: any): void => {
+				article = article.data;
+				onArticleResponse(request, reply, article.exception, {
```
hakubo: you might use object shorthand

server/facets/showMainPageCategory.ts
```diff
@@ -22,16 +23,24 @@ function showCategory (request: Hapi.Request, reply: Hapi.Response): void {
...
 
 	mainPage = new MainPage.MainPageRequestHelper(params);
-	mainPage.getWikiVariables((error: any, wikiVariables: any) => {
-		if (error) {
-			reply.redirect(localSettings.redirectUrlOnNoData);
-		} else {
+	mainPage
+		.getWikiVariables()
+		.then((data: any) => {
+			var wikiVariables = data.data;
```
hakubo: data.data :)
I see you did: `wikiVariables = wikiVariables.data;` above
already better

or how about returning data already from promise?
or is there anything that uses whole object?

server/lib/Utils.ts
```diff
@@ -233,3 +236,46 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
...
+): void {
+	var requestedHost = getCachedWikiDomainName(localSettings, request.headers.host),
+		canonicalHost = Url.parse(wikiVariables.basePath).hostname,
+		isLocal = isXipHost(localSettings, clearHost(request.headers.host)),
+		redirectLocation: string;
+
+	if (!isLocal && requestedHost !== canonicalHost) {
+		redirectLocation = wikiVariables.basePath + request.path;
+
+		if (Object.keys(request.query).length > 0) {
```
hakubo: I think you can pass whole object to stringify
```
querystring.stringify({ foo: 'bar', baz: ['qux', 'quux'], corge: '' })
// returns
'foo=bar&baz=qux&baz=quux&corge='
```

test/specs/server/lib/Utils.js
```diff
@@ -232,3 +232,159 @@ test('parseQueryParams', function () {
 	equal(typeof global.parseQueryParams(testCases[3], allowedKeys).allowed, 'string', 'Whitelisted parameter was not passed through');
 	equal(global.parseQueryParams(testCases[3], allowedKeys).allowed, '&lt;&#x2f;script&gt;', 'HTML in query not escaped properly');
 });
+
+test('isXipHost', function () {
```
hakubo: :+1:

server/facets/showMainPageCategory.ts
rogatty: Good idea, see https://github.com/Wikia/mercury/commit/3ba8bf49113ef552a539d5d2b3aacf210a71bd1f.

server/lib/Utils.ts
rogatty: That's what I do: `QueryString.stringify(request.query)`

This `if` is here so we know if `?` should be added to URL.

hakubo: ok, get it

PR #1387, opened 2015-09-18
SOC-1329 Adds a contributors component + an standalone avatar component

armonr:
This is the basic component needed for contributors/posters avatars.
A separate PR will be added to use this component in header and left rail.

/CC @bkoval @mklucsarits @garthwebb 

https://wikia-inc.atlassian.net/browse/SOC-1329

### DIFF
--- a/front/scripts/main/components/DiscussionAvatarComponent.ts
+++ b/front/scripts/main/components/DiscussionAvatarComponent.ts
@@ -1,0 +1,7 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionAvatarComponent = Em.Component.extend({
+	classNames: ['discussion-avatar']
+
+});
--- a/front/scripts/main/components/DiscussionAvatarComponent.ts
+++ b/front/scripts/main/components/DiscussionAvatarComponent.ts
@@ -1,7 +1,0 @@
-/// <reference path="../app.ts" />
-'use strict';
-
-App.DiscussionAvatarComponent = Em.Component.extend({
-	classNames: ['discussion-avatar']
-
-});
--- a/front/scripts/main/components/DiscussionContributorsComponent.ts
+++ b/front/scripts/main/components/DiscussionContributorsComponent.ts
@@ -14,8 +14,8 @@ App.DiscussionContributorsComponent = Em.Component.extend({
 	}),
 
 	additionalContributorCount: Em.computed('latestContributors',
-											'totalContributorCount',
-											function () {
+	                                        'totalContributorCount',
+	                                        function () {
 		return this.get('totalContributorCount') - this.get('latestContributors').length
 	})
 });
--- a/front/scripts/main/components/DiscussionContributorsComponent.ts
+++ b/front/scripts/main/components/DiscussionContributorsComponent.ts
@@ -4,6 +4,8 @@
 App.DiscussionContributorsComponent = Em.Component.extend({
 	classNames: ['discussion-contributors'],
 
+	contributors: [],
+
 	maxContributorDisplayCount: 7,
 
 	totalContributorCount: 0,
@@ -16,6 +18,6 @@ App.DiscussionContributorsComponent = Em.Component.extend({
 	additionalContributorCount: Em.computed('latestContributors',
 	                                        'totalContributorCount',
 	                                        function () {
-		return this.get('totalContributorCount') - this.get('latestContributors').length
+		return this.get('totalContributorCount') - this.get('latestContributors.length')
 	})
 });
--- a/front/scripts/main/components/DiscussionContributorsComponent.ts
+++ b/front/scripts/main/components/DiscussionContributorsComponent.ts
@@ -4,10 +4,18 @@
 App.DiscussionContributorsComponent = Em.Component.extend({
 	classNames: ['discussion-contributors'],
 
-	additionalContributorsCount: 0,
-	count: 7,
+	maxContributorDisplayCount: 7,
+
+	totalContributorCount: 0,
+
 	latestContributors: Em.computed('contributors', function () {
-		var limit = this.get('count');
-		return this.get('contributors').slice(0, limit);
+		var max = this.get('maxContributorDisplayCount');
+		return this.get('contributors').slice(0, max);
+	}),
+
+	additionalContributorCount: Em.computed('latestContributors',
+											'totalContributorCount',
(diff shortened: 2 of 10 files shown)


front/scripts/main/components/DiscussionContributorsComponent.ts
```diff
@@ -0,0 +1,21 @@
...
+
+	totalContributorCount: 0,
+
+	latestContributors: Em.computed('contributors', function () {
+		var max = this.get('maxContributorDisplayCount');
+		return this.get('contributors').slice(0, max);
+	}),
+
+	additionalContributorCount: Em.computed('latestContributors',
+											'totalContributorCount',
```
bkoval: Whitespacing seems to be a bit unusual here :)

front/styles/main/component/_discussion-contributors.scss
```diff
@@ -0,0 +1,33 @@
+$discussion-contributors-background-color: #666;
+$discussion-contributors-border-color: #ccc;
```
bkoval: I think we have style-guide values corresponding to grayscale colors and we probably should use these instead of hex values. /cc @mklucsarits and @vforge

front/templates/main/components/discussion-avatar.hbs
```diff
@@ -0,0 +1,8 @@
+<figure class="avatar">
```
bkoval: What is the purpose of the <figure> element? Maybe this could be simplified to just an anchor with an image inside? Ember component will create a div wrapping the whole component anyway :)

front/scripts/main/components/DiscussionAvatarComponent.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference path="../app.ts" />
```
mklucsarits: It might make more sense to have a general-purpose avatar component, instead of making it specific to discussions. I'm sure the UX team will be re-using circular user avatars in several places going forward. We also already have such a need for it in the new global nav component (top-bar) -- eventually we could merge an Ember avatar component and the webcomponent usage into one.

front/templates/main/components/discussion-avatar.hbs
armonr: @bkoval: This is merely how we render avatars in this project, and for DRYness sake, I pulled it into a component and will soon replace the repeated usages of this html block.
I kept the same DOM structure, but only made the user-name part configurable to fit both needs.

front/templates/main/components/discussion-avatar.hbs
```diff
@@ -0,0 +1,8 @@
...
+    <a href="{{authorUrl}}">
+        <img src="{{avatarUrl}}">
+    </a>
+</figure>
+{{#if displayAuthorName}}
```
mklucsarits: I think this avatar component should handle only the avatar image, and leave an optional corresponding user name to outside logic. If you do that, you can use 'figure' as the tagName property of the AvatarComponent. It would simply this component a lot.

front/scripts/main/components/DiscussionContributorsComponent.ts
armonr: Interesting. I used tabs, but IDEA is not doing what I expected.

front/styles/main/component/_discussion-contributors.scss
mklucsarits: Yep, there is a ```$border-color``` variable I recently added to the style guide for #CCC. #FFF can be replaced with ```$color-white```.

front/scripts/main/components/DiscussionAvatarComponent.ts
```diff
@@ -0,0 +1,7 @@
...
+'use strict';
+
+App.DiscussionAvatarComponent = Em.Component.extend({
+	classNames: ['discussion-avatar']
+
```
mklucsarits: Extra blank line can be removed.

front/scripts/main/components/DiscussionContributorsComponent.ts
```diff
@@ -0,0 +1,21 @@
...
+App.DiscussionContributorsComponent = Em.Component.extend({
+	classNames: ['discussion-contributors'],
+
+	maxContributorDisplayCount: 7,
+
+	totalContributorCount: 0,
+
+	latestContributors: Em.computed('contributors', function () {
+		var max = this.get('maxContributorDisplayCount');
+		return this.get('contributors').slice(0, max);
```
rogatty: This will fail if you use the component but don't pass `contributors` array. How about adding `contributors: []` as a component's default? Just to be safe.

front/scripts/main/components/DiscussionContributorsComponent.ts
```diff
@@ -0,0 +1,21 @@
...
+	}),
+
+	additionalContributorCount: Em.computed('latestContributors',
+	                                        'totalContributorCount',
+	                                        function () {
+		return this.get('totalContributorCount') - this.get('latestContributors').length
```
rogatty: You can use `this.get('latestContributors.length')` to be safe.

front/templates/main/components/user-avatar.hbs
```diff
@@ -0,0 +1,8 @@
+<figure class="avatar">
+    <a href="{{profileUrl}}">
+        <img src="{{avatarUrl}}">
+    </a>
```
rogatty: Please use tabs in .hbs files. If you're using JetBrains:
1. Settings > Editor > Code Style
2. Default Indent Options
3. Check ​"Use tab character"

vforge: We have `.editorconfig` file (https://github.com/Wikia/mercury/blob/dev/.editorconfig) - please configure your IDE to use it - see http://editorconfig.org/#download

armonr: Thanks. I'm using Intellij and `editorConfig` is enabled, so I am not sure why it did not take effect when the file was created.

mklucsarits:
LGTM

I still recommend moving non-avatar HTML (username text) out of the avatar component, but I won't block the PR for that opinion.

bkoval:
LGTM :japanese_ogre:

armonr:
Thanks @mklucsarits. I created [SOC-1335](https://wikia-inc.atlassian.net/browse/SOC-1335) to revisit this.

PR #1391, opened 2015-09-21
SOC-954 As a user, I want to upvote posts or replies

rafalkalinski:
https://wikia-inc.atlassian.net/browse/SOC-954

This covers upvoting content in the new discussions product. Downvoting will not be included. Upvoting can happen on either a post or a reply.


front/scripts/main/components/DiscussionUpvoteComponent.ts
```diff
@@ -0,0 +1,21 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionUpvoteComponent = Em.Component.extend({
+	classNames: ['small-4', 'columns', 'upvote', 'count'],
+
+	post: null,
+	classNameBindings: ['hasUpvoted'],
+	hasUpvoted: Em.computed('post._embedded.userData.@each.hasUpvoted', function(): boolean {
```
bkoval: Space after function :)

front/scripts/main/components/DiscussionUpvoteComponent.ts
```diff
@@ -0,0 +1,21 @@
...
+	classNameBindings: ['hasUpvoted'],
+	hasUpvoted: Em.computed('post._embedded.userData.@each.hasUpvoted', function (): boolean {
+		if (this.get('post')._embedded === undefined || this.get('post')._embedded.userData === undefined) {
+			return false
+		}
+		return this.get('post')._embedded.userData[0].hasUpvoted;
+	}),
+
+	actions: {
+		upvote(post: typeof App.DiscussionPostModel): void {
```
bkoval: How about a mixin for upvote action?

front/styles/main/base/_pointer.scss
```diff
@@ -2,4 +2,8 @@
 // adding cursor: pointer to that element makes it register click events in iOS
 [data-ember-action] {
 	cursor: pointer;
+
+	&.no-pointer {
```
bkoval: Please, add a comment explaining what happens here.

front/templates/main/components/forum-wrapper.hbs
```diff
@@ -1,13 +1,10 @@
 <div class="discussion forum">
 	{{#each posts as |item|}}
 		{{post-detail
-		action='goToPost'
 		goToPost='goToPost'
+		upvote='upvote'
+		post=item._embedded.firstPost.[0]
```
bkoval: :100:

front/scripts/main/components/DiscussionUpvoteComponent.ts
```diff
@@ -0,0 +1,21 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionUpvoteComponent = Em.Component.extend({
+	classNames: ['small-4', 'columns', 'upvote', 'count'],
+
+	post: null,
+	classNameBindings: ['hasUpvoted'],
+	hasUpvoted: Em.computed('post._embedded.userData.@each.hasUpvoted', function (): boolean {
+		if (this.get('post')._embedded === undefined || this.get('post')._embedded.userData === undefined) {
```
rogatty: `this.get('post._embedded')`
`this.get('post._embedded.userData')`

vforge: How about saving `this.get('post._embedded')` in variable?

front/styles/main/base/_pointer.scss
rogatty: Can you add explanation why we need it?

vforge: In general, I discourage anyone from this - `[data-ember-action] { cursor: pointer; }` is a fix for iOS, and not styling *per se* - turning it off will disable this fix.

server/views/_layouts/ember-main.hbs
```diff
@@ -44,7 +44,7 @@
 			M.prop('wikiaLanguage', '{{{wiki.language.user}}}');
 			M.prop('userId', {{{userId}}});
 			M.prop('cookieDomain', {{{json localSettings.authCookieDomain}}});
-			M.prop('servicesDomain', '{{{server.servicesDomain}}}' || 'services.wikia.com');
+			M.prop('servicesDomain', {{{json localSettings.servicesDomain}}});
```
rogatty: Should we remove `server.servicesDomain` from https://github.com/Wikia/mercury/blob/fd6a35519378e75a4ba7fae940acc1bf78b7d297/server/lib/Utils.ts#L186?

front/scripts/main/components/DiscussionUpvoteComponent.ts
bognix: this should be enough:
```
return this.get('post._embedded.userData[0].hasUpvoted')  || false
```
or:
```
return Em.isArray(this.get('post._embedded.userData')) ? this.get('post._embedded.userData')[0].hasUpvoted')  || false
```
not sure about array index

front/styles/main/base/_pointer.scss
rafalkalinski: removed

server/views/_layouts/ember-main.hbs
rafalkalinski: removed

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -1,19 +1,21 @@
...
 	classNames: ['post-detail'],
 
-	author: null,
 	postId: null,
-	authorUrl: Em.computed('author', function (): string {
+	authorUrl: Em.computed('post', function (): string {
 		return M.buildUrl({
 			namespace: 'User',
-			title: this.get('author.name')
+			title: this.get('post').createdBy.name
```
kvas-damian: Are you sure title doesn't require escaping? We already have similar issue here: https://wikia-inc.atlassian.net/browse/SOC-1337
Maybe we will need to update ```buildUrl```?

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+			hasUpvoted = Em.get(post._embedded.userData[0], 'hasUpvoted');
+			method = (hasUpvoted ? 'delete' : 'post');
+
+			// assuming the positive scenario, the change in the front-end is dome here
+			Em.set(post, 'upvoteCount', oldUpvoteCount + (hasUpvoted ? -1 : 1));
+			Em.set(post._embedded.userData[0], 'hasUpvoted', !hasUpvoted);
+
+			Em.$.ajax(<JQueryAjaxSettings>{
+				method: method,
+				url: 'https://' + M.prop('servicesDomain') +
```
kvas-damian: This is something we should definitely take care of.
we should about generating URLs like this in many places, that's why I suggest using some utils for that. 
BTW: Pathname is already defined in https://github.com/Wikia/mercury/blob/dev/config/localSettings.base.ts#L27

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+				'/discussion/' + post.siteId + '/votes/post/' + post.id,
+				dataType: 'json',
+				xhrFields: {
+					withCredentials: true,
+				},
+				success: (data:any): void => {
+					Em.set(post, 'upvoteCount', data.upvoteCount);
+				},
+				error: (err:any): void => {
+					// @TODO: handle errors
```
kvas-damian: When are we going to resolve this ```TODO```?

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	actions: {
+		upvote(post: typeof App.DiscussionPostModel): void {
+			var hasUpvoted: boolean,
+				method: string,
+				oldUpvoteCount: number = Em.get(post, 'upvoteCount'),
+				postId: number = Em.get(post, 'id');
+
+			if (this.upvotingInProgress[postId] || Em.get(post, '_embedded') === undefined ||
+				Em.get(post._embedded, 'userData') === undefined) {
```
kvas-damian: Do we need 2 separarate conditions to check is it's ```undefined```?

front/scripts/main/models/DiscussionPostModel.ts
```diff
@@ -52,22 +52,23 @@ App.DiscussionPostModel.reopenClass({
 				threadId: threadId
 			});
 
-			Em.$.ajax({
+			Em.$.ajax(<JQueryAjaxSettings>{
 				url: 'https://' + M.prop('servicesDomain') +
 					 `/discussion/${wikiId}/threads/${threadId}` +
 					 '?responseGroup=full&sortDirection=descending&limit=' + postInstance.replyLimit,
 				dataType: 'json',
+				xhrFields: {
```
kvas-damian: Maybe we could create Util method for discussion ajax calls?
This way we could avoid code repetition for ```dataType``` and ```xhrFields```.

front/templates/main/components/discussion-upvote.hbs
```diff
@@ -0,0 +1,12 @@
...
+		{{svg 'upvote' viewBox='0 0 30 30' class='icon upvote'}}
+	{{else}}
+		{{svg 'upvote-reply' viewBox='0 0 30 30' class='icon upvote-reply'}}
+	{{/if}}
+</a>
+{{#if firstPost}}
+	{{post.upvoteCount}}
+{{else}}
+	<small>{{post.upvoteCount}}</small>
+{{/if}}
```
kvas-damian: What do you think about creating separate template for reply?

front/scripts/main/components/PostDetailComponent.ts
rogatty: `this.get('post.createdBy.name')`

front/scripts/main/components/PostReplyComponent.ts
```diff
@@ -1,17 +1,15 @@
...
-	content: null,
-	timestamp: null,
-	upvoteCount: null,
-	authorUrl: Em.computed('author', function (): string {
+	post: null,
+	authorUrl: Em.computed('post', function (): string {
 		return M.buildUrl({
 			namespace: 'User',
-			title: this.get('author.name')
+			title: this.get('post').createdBy.name
```
rogatty: `this.get('post.createdBy.name')`

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rogatty: You can just use `Em.get(post, '_embedded.userData')`

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rogatty: If it's not going to be done in this PR then please put a ticket number in this comment.

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+				return null;
+			}
+
+			this.upvotingInProgress[postId] = true;
+			hasUpvoted = Em.get(post._embedded.userData[0], 'hasUpvoted');
+			method = (hasUpvoted ? 'delete' : 'post');
+
+			// assuming the positive scenario, the change in the front-end is done here
+			Em.set(post, 'upvoteCount', oldUpvoteCount + (hasUpvoted ? -1 : 1));
```
hakubo: why use Em.set instead of post.set?

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+					withCredentials: true,
+				},
+				success: (data:any): void => {
+					Em.set(post, 'upvoteCount', data.upvoteCount);
+				},
+				error: (err:any): void => {
+					// @TODO: handle errors
+
+					Em.set(post, 'upvoteCount', oldUpvoteCount);
+					Em.set(post._embedded.userData[0], 'hasUpvoted', !Em.get(post._embedded.userData[0], 'hasUpvoted'));
```
hakubo: post.toggleProperty

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+App.DiscussionUpvoteMixin = Em.Mixin.create({
+	upvotingInProgress: {},
+
+	actions: {
+		upvote(post: typeof App.DiscussionPostModel): void {
+			var hasUpvoted: boolean,
+				method: string,
+				oldUpvoteCount: number = Em.get(post, 'upvoteCount'),
+				postId: number = Em.get(post, 'id');
```
hakubo: post.get

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
```diff
@@ -0,0 +1,54 @@
...
+
+			if (this.upvotingInProgress[postId] || Em.get(post, '_embedded') === undefined ||
```
hakubo: why do we have _embedded? 
What does this underscore mean?

front/scripts/main/models/DiscussionPostModel.ts
```diff
@@ -52,22 +52,23 @@ App.DiscussionPostModel.reopenClass({
 				threadId: threadId
 			});
 
-			Em.$.ajax({
+			Em.$.ajax(<JQueryAjaxSettings>{
 				url: 'https://' + M.prop('servicesDomain') +
```
hakubo: weird mix of normal strings and sting templates.
can you use just sting templates?
additionally they do support multiline strings

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
kvas-damian: @hakubo this is response from service that uses HAL+json as a response. ```_embedded``` is keyword for this format: http://stateless.co/hal_specification.html

front/scripts/main/components/DiscussionUpvoteComponent.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="../app.ts" />
+/// <reference path="../mixins/DiscussionUpvoteActionSendMixin.ts" />
+'use strict';
+
+App.DiscussionUpvoteComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixin, {
+	classNames: ['small-4', 'columns', 'upvote', 'count'],
+
+	post: null,
+	classNameBindings: ['hasUpvoted'],
```
hakubo: this would look better on line 7

front/scripts/main/components/DiscussionUpvoteComponent.ts
```diff
@@ -0,0 +1,16 @@
...
+	hasUpvoted: Em.computed('post._embedded.userData.@each.hasUpvoted', function (): boolean {
+		if (this.get('post._embedded.userData') === undefined) {
```
hakubo: Can you write more about what this check guard us against?

front/scripts/main/components/DiscussionUpvoteComponent.ts
```diff
@@ -0,0 +1,16 @@
...
+			return false
+		}
+		return this.get('post')._embedded.userData[0].hasUpvoted;
```
hakubo: so hasUpvoted is dependant on all userData and then returns hasUpvoted value for first?
can you write more about that logic here?

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
hakubo: doh! Thanks

rogatty: https://github.com/201-created/ember-data-hal-9000
One day! ;)

front/scripts/main/components/PostDetailComponent.ts
rafalkalinski: buildUrl improved (and uTests added)

test/specs/front/scripts/mercury/utils/buildUlr.js
```diff
@@ -0,0 +1,89 @@
+QUnit.module('buildUrl tests setup',{
```
kvas-damian: What do you think about merging those tests with exsiting test cases for this method?
https://github.com/Wikia/mercury/blob/dev/test/specs/front/scripts/baseline/mercury/utils/buildUrl.js

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: Since it's not in this story scope:
https://wikia-inc.atlassian.net/browse/SOC-1343

test/specs/front/scripts/mercury/utils/buildUlr.js
rafalkalinski: done

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: With the new information - we don't need this todo there :)

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: `Uncaught TypeError: post.toggleProperty is not a function`

rogatty: If `post` is `App.DiscussionPostModel` instance then it should have this method, see http://emberjs.com/api/classes/Ember.Object.html#method_toggleProperty

front/scripts/main/models/DiscussionPostModel.ts
rafalkalinski: https://wikia-inc.atlassian.net/browse/SERVICES-802

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: It is not the App.DiscussionPostModel instance. It is the part of it, but not the instance itself.

rogatty: In that case `upvote(post: typeof App.DiscussionPostModel): void {` isn't true, right?

rafalkalinski: hmm... right :/

garthwebb: Checkmate!

front/templates/main/components/discussion-upvote.hbs
rafalkalinski: done

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: post does not have this method here

front/scripts/main/components/DiscussionUpvoteComponent.ts
rafalkalinski: actually this is enough:
`return this.get('post._embedded.userData[0].hasUpvoted');`
And there is no need for any checks before.

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: done

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: Post does not have this method here.
I will work on that...

front/scripts/main/models/DiscussionPostModel.ts
rafalkalinski: It's done like that in every place in the Dicsussion app.
I've filled a ticket about changing the way how we handle creating the service urls:
https://wikia-inc.atlassian.net/browse/SOC-1343

front/scripts/main/components/DiscussionUpvoteComponent.ts
rafalkalinski: done

front/scripts/main/components/DiscussionUpvoteComponent.ts
rafalkalinski: It the user is not authorized, there is no 'userData' properties.
I've changed the way how the get below this works and removed the check, cos it was not needed any more.

front/scripts/main/components/DiscussionUpvoteComponent.ts
rafalkalinski: When the user is authorized, the UserData property points to an array with only one element (it is always only one element - thaks to HAL+json), which (among others) has hasUpvoted property.

front/scripts/main/components/DiscussionUpvoteComponent.ts
rafalkalinski: Nope... Actually it is needed.

front/scripts/main/components/DiscussionUpvoteComponent.ts
rafalkalinski: So the version
`this.get('post._embedded.userData[0].hasUpvoted')`
doesn't work.
I wrote:
`if (Em.isArray(this.get('post._embedded.userData'))) {
			return this.get('post._embedded.userData')[0].hasUpvoted;
		}
		return false`

front/scripts/main/mixins/DiscussionUpvoteMixin.ts
rafalkalinski: Until we will figure it out how (and if) we want to change the model, i put there 'any' as a type.
One way or another, that's for sure not in the scope of this story.

front/scripts/main/components/DiscussionUpvoteComponent.ts
hakubo: doesn't it have to be:
`this.get('post._embedded.userData.[0].hasUpvoted')`

rafalkalinski: `this.get('post._embedded.userData.[0].hasUpvoted')
doesn't work as same as 
`this.get('post._embedded.userData[0].hasUpvoted')

front/scripts/main/components/DiscussionUpvotePostComponent.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference path="../app.ts" />
+/// <reference path="../mixins/DiscussionUpvoteComponentMixin.ts" />
+/// <reference path="../mixins/DiscussionUpvoteActionSendMixin.ts" />
+'use strict';
+
+App.DiscussionUpvotePostComponent = Em.Component.extend(
+	App.DiscussionUpvoteComponentMixin,
+	App.DiscussionUpvoteActionSendMixin, {
+		svgBaseName: 'upvote'
```
bkoval: Is there a need to set it here? Since there are seperate handlebars ftemplates for these 2 components, you could just use svg helper directly with the name in the handlebars.

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
```diff
@@ -0,0 +1,48 @@
...
+				'/discussion/' + Em.get(post, 'siteId') + '/votes/post/' + Em.get(post, 'id'),
+				dataType: 'json',
+				xhrFields: {
+					withCredentials: true,
+				},
+				success: (data: any): void => {
+					Em.set(post, 'upvoteCount', data.upvoteCount);
+				},
+				error: (): void => {
+					Em.set(post._embedded.userData[0], 'hasUpvoted', hasUpvoted)
```
bkoval: Shouldn't you use quotes for the first argument? Just to assume Ember will resolve any "uncaught because undefined" errors?

front/scripts/main/mixins/DiscussionUpvoteComponentMixin.ts
```diff
@@ -0,0 +1,17 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+/**
+ * Handles sending upvote action outside from the component.
+ */
+App.DiscussionUpvoteComponentMixin = Em.Mixin.create({
+	classNames: ['small-5', 'columns', 'upvote', 'count'],
```
bkoval: I'm not sure but I think we're not supporting classes like 'small-5' form style-guide, @mklucsarits is the best source of knowledge here.

front/scripts/main/components/DiscussionUpvotePostComponent.ts
rafalkalinski: Yes there is a need to set it here. It is set to 'upvote' in the DiscussionUpvotePostComponent , and to 'upvote-reply' in the DiscussionUpvoteReplyComponent, and the DiscussionUpvoteComponentMixin, which is used by the both components creates the svgName depended on svgBaseName and on the hasUpvoted property.

front/scripts/main/mixins/DiscussionUpvoteComponentMixin.ts
```diff
@@ -0,0 +1,17 @@
...
+
+	post: null,
+	svgName: Em.computed('post._embedded.userData.@each.hasUpvoted', function (): string {
+		if (Em.isArray(this.get('post._embedded.userData'))) {
+			return this.svgBaseName + (this.get('post._embedded.userData')[0].hasUpvoted ? '-activated' : '');
```
bkoval: Wow, this concatenation doesn't feel right :) How about toggling this through handlebars?

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
rafalkalinski: At this point there is always something in the post._embedded.userData[0] object.

front/scripts/main/mixins/DiscussionUpvoteComponentMixin.ts
mklucsarits: Yes, it's alright to use these. We're starting to use the grid now more extensively in Mercury beginning with v1.5.0 of the style guide. However 1.5.0 does not affect the small-# classes, so this is fine and won't be affected by changes to the grid in 1.5.0.

front/templates/main/components/forum-wrapper.hbs
```diff
@@ -1,13 +1,10 @@
...
 		postId=item.id
-		author=item._embedded.firstPost.0.createdBy
```
mklucsarits: These properties were used to abstract away the API's naming of parameters. What is the reason for changing that?

front/templates/main/components/post-detail.hbs
```diff
@@ -1,24 +1,37 @@
...
+	<div class="discussion-content" {{action 'goToPost' postId}}>
+		{{{post.rawContent}}}
 	</div>
+{{/if}}
+<div class="row">
+	{{discussion-upvote-post
+		post=post
+		upvote='upvote'
+	}}
+	{{#if isDetailsView}}
```
mklucsarits: Can you simplify this block so there is not so much repeated markup for only a difference in the action?

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
```diff
@@ -0,0 +1,48 @@
...
+				method: string,
+				postId: number = Em.get(post, 'id');
+
+			if (this.upvotingInProgress[postId] || Em.get(post, '_embedded.userData') === undefined) {
+				return null;
+			}
+
+			this.upvotingInProgress[postId] = true;
+			hasUpvoted = Em.get(post._embedded.userData[0], 'hasUpvoted');
+			method = (hasUpvoted ? 'delete' : 'post');
```
mklucsarits: No need for parentheses here.

front/templates/main/components/post-detail.hbs
bkoval: We were struggling with figuring out better solution here. This one is the most readable for now and I propose to leave it until we're upgraded to Ember 2.0 with possibility to use closure actions.

vforge: @bkoval :+1:

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
rafalkalinski: At this point I want to fire the trigger exactly on this particular property (post._embedded.userData[0]), because on the other end the component observes and reacts on the change there.

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
rafalkalinski: I put them around the ternary operator to improve readability. This is a simple example, but with a longer one, if your IDE highlights the opening and closing parenthesis it is easier to read it.

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
bkoval: :+1:

front/templates/main/components/forum-wrapper.hbs
bkoval: I agree with that, we spoke with @rogatty and agreed with the fact that we should do as much data sanitizing as possible in the model to keep components as decoupled as possible. We discussed with @garthwebb a possibility of creating a technical debt epic in Jira, this would be a good material for a ticket there.

rafalkalinski: I can see you point. And you're right. I just assumed that if I pass the whole item._embedded.firstPost.0  object here, it would be simpler to remove additional parameters which duplicates the information from the object.
On the other hand, we shouldn't rely on what we get from the source, but only on the model - it should contain sanitized and standardized data from the external source or sources.
@kvas-damian what do you think?

front/scripts/main/components/DiscussionUpvotePostComponent.ts
rafalkalinski: I changes how the svg icon is handled and that code is removed

front/scripts/main/mixins/DiscussionUpvoteComponentMixin.ts
rafalkalinski: It's not there anymore

bkoval:
LGTM :church: :v: :accept:

vforge:
LGTM!

front/templates/main/components/forum-wrapper.hbs
kvas-damian: agree with @bkoval, this change is not in the scope of this ticket, we can leave as it is for now.

Ideally we should standardize and sanitize data in model and then pass it to post detail or post reply components. Finally agree with @mklucsarits  post detail and post reply components shouldn't be dependent on API response structure. The same should be applied to ```ForumWrapperComponent```.

kvas-damian:
LGTM

PR #1388, opened 2015-09-18
SOC-1294 Implement rails and the wiki badge

garthwebb:
This adds responsive left and right rails as well as adding the first rail content, the wiki badge.

https://wikia-inc.atlassian.net/browse/SOC-1294

PR #1397, opened 2015-09-22
SOC-1293 "Back to the wiki" link for Desktop Discussion

bkoval:
@garthwebb @mklucsarits Take a closer look, please.

I refactored the Rails code (removed components, left them only in route templates) in order to be able to bubble actions from components higher, to controllers / routes. Refactoring commit -> https://github.com/Wikia/mercury/commit/3810f57dbbbf1fd1b62c53e353591eb7d5a33217

Problem with creating components in a loop is a fact that you cannot bind outside actions to actions inside components for a specific component. This is the only way components can communicate with higher application levels in Ember.

I discussed this solution with the X-Wing members and it seems like it's least constful (minimal number of files and readable markup) but I would love to know your opinion - maybe we can achieve the same effect in a more elegant way?


front/scripts/main/components/BackButtonComponent.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.BackButtonComponent = Em.Component.extend({
```
rogatty: `BackButtonComponent` sounds like something general but this implementation is pretty specific for a single use case. Maybe some other name?

bkoval: OH yeah, I was meant tobadd discussion- prefix, good catch!

garthwebb:
@bkoval , I think its ok.  Like I said in a previous PR comment, I like being more explicit and this is more explicit.  I don't actually remember now why I though a component was a good idea ...

mklucsarits:
LGTM

bkoval:
@garthwebb I'm closing this one since it's already merge to SOC-1294 (see #1388)

PR #1403, opened 2015-09-23
Dat 2996 - add item reordering functionality to infobox builder

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-2996

### DIFF
--- a/front/scripts/main/models/InfoboxBuilderModel.ts
+++ b/front/scripts/main/models/InfoboxBuilderModel.ts
@@ -214,17 +214,26 @@ App.InfoboxBuilderModel = Em.Object.extend({
 	 * @param {DataItem|ImageItem|TitleItem} item
 	 */
 	moveItem(offset: number, item: DataItem|ImageItem|TitleItem) {
+		var position = this.get('infoboxState').indexOf(item);
+
+		if (this.isValidMove(position, offset)) {
+			this.get('infoboxState').removeAt(position);
+			this.get('infoboxState').insertAt(position + offset, item);
+		}
+	},
+
+	/**
+	 * @desc checks if move is valid based on item current position in the infoboxState and the move offset
+	 * @param {Number} position
+	 * @param {Number} offset
+	 * @returns {Boolean}
+	 */
+	isValidMove(position: number, offset: number): boolean {
 		var lastItemIndex = this.get('infoboxState').length -1,
-			position = this.get('infoboxState').indexOf(item),
 			newPosition = position + offset;
 
-		if (
-			position > 0 && offset < 0 && newPosition >= 0 ||
-			position < lastItemIndex && offset > 0 && newPosition <= lastItemIndex
-		) {
-			this.get('infoboxState').removeAt(position);
-			this.get('infoboxState').insertAt(newPosition, item);
-		}
+		return position > 0 && offset < 0 && newPosition >= 0 ||
+			position < lastItemIndex && offset > 0 && newPosition <= lastItemIndex;
 	},
 
 	/**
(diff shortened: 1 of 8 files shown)


front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -195,26 +194,44 @@ App.InfoboxBuilderModel = Em.Object.extend({
...
+		this.get('infoboxState').removeObject(item);
 		this.resetEditMode();
 	},
 
 	/**
+	 * @desc moves item in infoboxState by given offset
+	 * @param {Number} offset
+	 * @param {DataItem|ImageItem|TitleItem} item
+	 */
+	moveItem(offset: number, item: DataItem|ImageItem|TitleItem) {
```
RafLeszczynski: I'll think about unit tests for this method

rogatty: Would it make sense and not be too hard to abstract some logic from https://github.com/Wikia/mercury/blob/dev/front/scripts/main/mixins/CuratedContentEditorSortableItemsMixin.ts and reuse it in both places?

RafLeszczynski: @rogatty: You're right. Our initial plan was to implement drag'n drop / sortable and AFAIK this is still something we definitely need to do maybe not as a part of first MVP but for sure with the second release. Base on the PO decision we will either invest in drag'n drop / sortable that could be used across Mercury or refactor this code as you suggested.

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -195,26 +194,44 @@ App.InfoboxBuilderModel = Em.Object.extend({
...
+		var lastItemIndex = this.get('infoboxState').length -1,
+			position = this.get('infoboxState').indexOf(item),
+			newPosition = position + offset;
+
+		if (
+			position > 0 && offset < 0 && newPosition >= 0 ||
+			position < lastItemIndex && offset > 0 && newPosition <= lastItemIndex
+		) {
+			this.get('infoboxState').removeAt(position);
+			this.get('infoboxState').insertAt(newPosition, item);
```
SebastianMarzjan: Is this set of remove and insert operations atomic? (thinking about race condition on rapid clicks)

front/scripts/main/models/InfoboxBuilderModel.ts
```diff
@@ -195,26 +194,44 @@ App.InfoboxBuilderModel = Em.Object.extend({
...
+	 * @param {DataItem|ImageItem|TitleItem} item
+	 */
+	moveItem(offset: number, item: DataItem|ImageItem|TitleItem) {
+		var lastItemIndex = this.get('infoboxState').length -1,
+			position = this.get('infoboxState').indexOf(item),
+			newPosition = position + offset;
+
+		if (
+			position > 0 && offset < 0 && newPosition >= 0 ||
+			position < lastItemIndex && offset > 0 && newPosition <= lastItemIndex
```
SebastianMarzjan: Could this expression be extracted to something along the lines of isValidMove ? :)

front/scripts/main/models/InfoboxBuilderModel.ts
RafLeszczynski: Interesting question. Actually it would be a real edge case but it's worth investigation. I spoke with @rogatty about it and it seams that we need to check the Ember core. These sam solution is implemented for Curated Content so I would leave it as it is now, so some research and if needed fix in future.

SebastianMarzjan:
Tests seem to be failing now

RafLeszczynski:
There is something wrong with mercury server tests. Runner hits timeout. I've spoked with @tomasznapieralski and here created a new issue in Jira to investigate it.

SebastianMarzjan:
LGTM

PR #1406, opened 2015-09-23
XW-177 | Rename explore the wiki to featured categories in curated content editor

bognix:
https://wikia-inc.atlassian.net/browse/XW-177


front/scripts/main/components/CuratedContentEditorImageSearchComponent.ts
```diff
@@ -22,9 +22,19 @@ App.CuratedContentEditorImageSearchComponent = Em.Component.extend(
 		searchPlaceholder: Em.computed((): string =>
 			i18n.t('app.curated-content-editor-search-images-placeholder')
 		),
-		label: Em.computed('model.label', function() {
+		label: Em.computed('model.label', 'isFeatured', 'isSection', function(): string {
```
bognix: I have mixed feelings about this solution (it's not very dry) and I leaning toward extracting header to another component. (This message is only shown in header).
Although it would add a lot of complexity in passing goBack action back and forth.
Hence, those mixed feelings

Warkot: Maybe a mixin could DRY your code? :wink:

vforge: I second that - can you please create `CuratedContentEditorLabel` mixin with that functionality and just re-use it inside `CuratedContentEditorItemFormComponent` ?

Warkot: https://github.com/Wikia/mercury/blob/XW-424/front/scripts/main/mixins/CuratedContentEditorLabelsMixin.ts

bognix: sounds perfect! Although I'll create a ticket for moving header to another compontent

hakubo:
LGTM

rogatty:
LGTM

front/locales/de/main.json
```diff
@@ -121,7 +121,7 @@
...
     "curated-content-editor-block-help-text":"Ziehe die Einträge in der von dir bevorzugten Reihenfolge herüber. Tippe auf einen Eintrag, um dir die Konfigurationsmöglichkeiten anzusehen.",
-    "curated-content-editor-explore-the-wiki-description":"Mithilfe empfohlener Kategorien können die Benutzer deine Community auf hohem Level unter Verwendung der Mediawiki-Kategorien erforschen. Kategorien sollten in logische Gruppen wie „Filme\" oder „Comics\" aufgeteilt werden. Wenn eine Kategorie nicht in eine bestimmte Gruppe passt, empfehlen wir, sie dem nachfolgenden Abschnitt „Optionale Kategorien\" zuzuordnen.",
+    "curated-content-editor-featured-categories-description":"Mithilfe empfohlener Kategorien können die Benutzer deine Community auf hohem Level unter Verwendung der Mediawiki-Kategorien erforschen. Kategorien sollten in logische Gruppen wie „Filme\" oder „Comics\" aufgeteilt werden. Wenn eine Kategorie nicht in eine bestimmte Gruppe passt, empfehlen wir, sie dem nachfolgenden Abschnitt „Optionale Kategorien\" zuzuordnen.",
```
Warkot: Just be aware that changing the keys here won't change much if you don't upload these files to Crowdin. Crowdin CLI by default uploads only the EN file. To upload also the translations, you have to run `crowdin upload` with `-t` parameter. However, I'm afraid it is/was buggy, because the strings that were not translated — they were in English in the translation files — after upload were marked as valid translations (which is not the default behavior surprisingly). But don't worry, you did good, because the string was not changed and was already translated before. In this case Crowdin offers translating with one click by bringing back the translation from Translation Memory. I already did that today, so Crowdin has the fresh version and it shouldn't break after someone else uploads to this file… hopefully. :wink:

Warkot:
LGTM :cookie:

PR #1409, opened 2015-09-23
ADEN-2464 Fix externaltest on mobile

nandy-andy:
Recently changes with using `x-original-host` probably broke `externaltest.*` and `showcase.*` domain aliases. Since those are quite edge cases and aren't separated machines we decided to treat it as edge case in Mercury server code.

This is the only idea we came up with as a quick fix. We can try it out tomorrow.

// cc @bognix @vforge @rogatty @hakubo

### DIFF
--- a/server/lib/Utils.ts
+++ b/server/lib/Utils.ts
@@ -126,7 +126,7 @@ export function clearHost (host: string): string {
 	// We use two special domain prefixes for Ad Operation and Sales reasons
 	// They behave similar to our staging prefixes but are not staging machines
 	// Talk to Ad Engineering Team if you want to learn more
-	var adDomainAliases: Array<string> = [ 'externaltest', 'showcase' ];
+	var adDomainAliases: Array<string> = ['externaltest', 'showcase'];
 
 	host = host.split(':')[0]; // get rid of port
 	Object.keys(adDomainAliases).forEach(function (key){
--- a/server/lib/Utils.ts
+++ b/server/lib/Utils.ts
@@ -117,13 +117,22 @@ export function getWikiDomainName (localSettings: LocalSettings, hostName: strin
 }
 
 /**
- * @desc Removes the port from hostname
+ * @desc Removes the port from hostname as well as ad domain aliases
  *
  * @param {string} host
  * @returns {string}
  */
 export function clearHost (host: string): string {
-	return host.split(':')[0]; //get rid of port
+	var adDomainAliases: Array<string> = [ 'externaltest', 'showcase' ];
+
+	host = host.split(':')[0]; // get rid of port
+	Object.keys(adDomainAliases).forEach(function (key){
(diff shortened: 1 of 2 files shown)


server/lib/Utils.ts
```diff
@@ -117,13 +117,22 @@ export function getWikiDomainName (localSettings: LocalSettings, hostName: strin
...
  * @param {string} host
  * @returns {string}
  */
 export function clearHost (host: string): string {
-	return host.split(':')[0]; //get rid of port
+	var adDomainAliases: Array<string> = [ 'externaltest', 'showcase' ];
+
+	host = host.split(':')[0]; // get rid of port
+	Object.keys(adDomainAliases).forEach(function (key){
+		if (host.indexOf(adDomainAliases[key]) === 0) {
```
bognix: what about hosts with www at the beginning?
Not sure if it can get there but I know it's possible to have it

nandy-andy: Good catch! But currently `externaltest.*` doesn't work as `preview.*` or `verify.*` and it redirects to production URL. I'll learn if it's a desired way of working or not but I don't think it affects in anyway this implementation. Those staging prefixes or weird domain aliases (as I call `externaltest` and `showcase`) are always at the beginning:
* http://verify.muppet.wikia.com/wiki/Muppet_Wiki
* http://verify.www.muppet.wikia.com/wiki/Muppet_Wiki
* http://externaltest.muppet.wikia.com/wiki/Muppet_Wiki
* http://externaltest.www.muppet.wikia.com/wiki/Muppet_Wiki

rogatty: Yup, `www` isn't at the beginning if there is env in URL. The same goes for sandboxes.

server/lib/Utils.ts
```diff
@@ -117,13 +117,22 @@ export function getWikiDomainName (localSettings: LocalSettings, hostName: strin
...
 /**
- * @desc Removes the port from hostname
+ * @desc Removes the port from hostname as well as ad domain aliases
  *
  * @param {string} host
  * @returns {string}
  */
 export function clearHost (host: string): string {
-	return host.split(':')[0]; //get rid of port
+	var adDomainAliases: Array<string> = [ 'externaltest', 'showcase' ];
```
rogatty: THIS IS NO MEDIAWIIIKIII! ;)

rogatty:
Can you add some comment explaining why we need this? Apart from that LGTM

nandy-andy:
Sure I can! I thought my variable name was self-descriptive :smiley_cat:

rogatty:
LGTM

vforge:
LGTM

nandy-andy:
Yeah, sorry about it @mklucsarits! Luckily @vforge has just fixed it: https://github.com/Wikia/mercury/pull/1428#issue-108322392 THANK YOU! :smile_cat:

PR #1410, opened 2015-09-23
Hg 887 Fixing custom web components in FF

lizlux:
Firefox, ember and polymer don't play well together when the `tagName` property of a component is a custom element. This PR wraps the custom elements in ember view divs and removes `position:static` from the top-bar element, which seems to fix the jerky UI on scroll.


kenkouot:
LGTM

mklucsarits:
@lizlux Does this fix Safari as well?

lizlux:
@mklucsarits it unbreaks safari but it doesn't look 100% right in safari atm.

PR #1411, opened 2015-09-23
HG-888 Set width of global nav content to grid max-width

mklucsarits:
https://wikia-inc.atlassian.net/browse/HG-888

This updates the width of the content for the new global nav component to be the max-width of the breakpoint grid, which is declared with the ```row``` CSS class.

@lizlux @kenkouot


lizlux:
This LGTM. I can apply something similar to the style guide.

lizlux:
Turns out I ended up applying these styles to the component itself, so we won't need this.

PR #1418, opened 2015-09-24
INT-203: Google Custom Search for Japan Homepage

perjg:
Simple integration of google custom search api.

Uses the "stand-alone search result page" example here as a base:
https://developers.google.com/custom-search/docs/tutorial/implementingsearchbox


homepage/server/views/_layouts/search.html
```diff
@@ -0,0 +1,19 @@
...
+  <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1, minimal-ui">
+
+  <title>{{title}}</title>
+
+  <link rel="stylesheet" href="/css/app.css">
+</head>
+<body>
+{{{content}}}
+
+<script src="/vendor/jquery/dist/jquery.min.js"></script>
```
RafLeszczynski: Please consider concatenating and minifying these JS assets to reduce the number of requests

homepage/server/facets/search.js
```diff
@@ -0,0 +1,17 @@
...
+ * Wikia (Japan) Homepage
+ *
+ * @author Per Johan Groland <pgroland@wikia-inc.com>
+ */
+
+var util = require('../util');
+
+function companyInfo(request, reply) {
+	var data = {
+		title: 'ウィキア・ジャパン'
```
RafLeszczynski: shouldn't this hardcoded string be stored in i18n file or any other place where you store messages for wikia japan homepage?

homepage/server/util.js
```diff
@@ -53,7 +53,7 @@ exports.getSignupUrl = function () {
 };
 
 // todo: look up this data in user session first
-exports.renderWithGlobalData = function (request, reply, data, view) {
+exports.renderWithGlobalData = function (request, reply, data, view, options) {
```
RafLeszczynski: Please consider using single hash map object as function argument if you need to pass so many arguments to the function call

homepage/server/views/_layouts/search.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
RafLeszczynski: The content of this html file is a duplication of https://github.com/Wikia/mercury/blob/INT-203/homepage/server/views/_layouts/default.html except additional `<script src="/js/search.js"></script>`

Please consider reusing code

homepage/server/views/search.html
```diff
@@ -0,0 +1,11 @@
+{{> nav}}
+
+<div>
+  <gcse:searchresults-only>
+    <section class="search-results-loading">
+      <p>検索結果ロード中</p>
```
RafLeszczynski: shouldn't the message string be passed to the view as argument?

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
+/// <reference path="../../../../typings/jquery/jquery.d.ts" />
+
+'use strict';
+
+function processArguments() : any {
```
RafLeszczynski: is this function registered in global scope?

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
...
+				continue;
+			}
+			params[nv[0]] = nv[1] || true;
+		}
+	}
+
+	return params;
+}
+
+function fillSearchTextBox() : void {
```
RafLeszczynski: is this function registered in global scope?

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
...
+	var params : any = processArguments();
+
+	if ($(document).width() < 710) {
+		$('#searchWikiaTextMobile').val(decodeURIComponent(params.q) || '');
```
RafLeszczynski: `processArguments()` can return empty `{}`. If so `params.q` will be `undefined`. Passing ` `undefined` to   `decodeURIComponent()` will cause `Uncaught ReferenceError:`

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
...
+	} else {
+		$('#searchWikiaText').val(decodeURIComponent(params.q) || '');
+	}
+}
+
+(function() : void {
+	// Google custom search injection
+	// https://developers.google.com/custom-search/docs/tutorial/implementingsearchbox
+	var cx : string = '005745855109319432328:coaj7jf_wgs',
```
RafLeszczynski: Please consider not using acronyms for variable names.
https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#acronyms

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
...
+		gcse : HTMLScriptElement = document.createElement('script'),
+		s : HTMLScriptElement;
+
+	gcse.type = 'text/javascript';
+	gcse.async = true;
+	gcse.src = (document.location.protocol === 'https:' ? 'https:' : 'http:') +
+		'//www.google.com/cse/cse.js?cx=' + cx;
+
+	s = document.getElementsByTagName('script')[0];
+	s.parentNode.insertBefore(gcse, s);
```
RafLeszczynski: If for some reason there won't be any `<script>` tags on the page this statement will cause `Uncaught TypeError: Cannot read property 'parentNode' of undefined`

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
...
+	return params;
+}
+
+function fillSearchTextBox() : void {
+	var params : any = processArguments();
+
+	if ($(document).width() < 710) {
+		$('#searchWikiaTextMobile').val(decodeURIComponent(params.q) || '');
+	} else {
+		$('#searchWikiaText').val(decodeURIComponent(params.q) || '');
```
RafLeszczynski: If `fillSearchTextBox` will be called multiple times please consider caching selectors

homepage/front/styles/app.scss
```diff
@@ -10,9 +10,19 @@ $visual-grid-opacity: .02;
 // Warnings for Bourbon deprecations
```
RafLeszczynski: just a side note. I see that `app.scss` is not minified on production. Please consider minifying CSS files for production deployment

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,52 @@
...
+		}
+	}
+
+	return params;
+}
+
+function fillSearchTextBox() : void {
+	var params : any = processArguments();
+
+	if ($(document).width() < 710) {
```
RafLeszczynski: Please consider storing this number in const value so it could be reused in other places. I assume this is of o the breakpoint values.

homepage/server/views/_layouts/search.html
perjg: I've created a separate issue to do this: INT-213

homepage/front/styles/app.scss
perjg: See INT-213.

homepage/front/scripts/search/search.ts
perjg: The functions inside search.ts processArguments and fillSearchTextBox are local, and are not accessed outside this file.

homepage/front/scripts/search/search.ts
perjg: The functions inside search.ts processArguments and fillSearchTextBox are local, and are not accessed outside this file.

homepage/front/scripts/search/search.ts
perjg: There are always script tags in the root server side template, so I don't think this will ever happen.

homepage/front/scripts/search/search.ts
perjg: It is only called on page load.

homepage/server/facets/search.js
perjg: I18N isn't yet integrated for the Japan Homepage app.
It is on my todo, I've added an issue for it: https://wikia-inc.atlassian.net/browse/INT-214

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	return params;
+}
+
+function fillSearchTextBox() : void {
+	var params : any = processArguments();
+
+	if (params.hasOwnProperty('q')) {
+		if ($(document).width() < mobileBreakpoint) {
+			$('#searchWikiaTextMobile').val(decodeURIComponent(params.q) || '');
```
RafLeszczynski: I believe you don't need the `|| ''` in this statement. Cause `decodeURIComponent(params.q)` will always return something and if `parms.q = ''` it will return `''`

homepage/front/scripts/search/search.ts
RafLeszczynski: Ok, but please consider handling this edge case gracefully, so that user won't get JS Errors in console. It could be an improvement done in a separate story just to make this code bullet proof. If you do decide to do it as a separate story please add TODO comment with a link to Jira issue in the code.

homepage/front/scripts/search/search.ts
RafLeszczynski: Correct me if I'm wrong. In `front/main/main.ts` both `search()` and `parallax()` are defined as global functions. You can access them directly on `ja.wikia.com`.
screen shot 2015-09-28 at 10 36 45
In `front/search/search.ts` both `processArguments()` and `fillSearchTextBox()` are defined the same way so unless these functions are wrapped in a closure during build step they will also be accessible from global scope.

homepage/server/facets/search.js
RafLeszczynski: Great, please consider adding TODO comment here with link Jira issue

homepage/front/scripts/main/main.ts
```diff
@@ -3,7 +3,10 @@
 
 'use strict';
 
-var parallaxWindow : JQuery = $('#js-parallax-window');
+declare var getGlobals : any;
+
+var parallaxWindow : JQuery = $('#js-parallax-window'),
+	mobileBreakpoint : number = getGlobals().mobileBreakpoint;
```
RafLeszczynski: as `parallaxWindow` is a global variable:
screen shot 2015-09-28 at 10 58 58
`mobileBreakpoint` will also pollute global scope. If you don't need to access these variables from other files please wrap them in closure. If you need to use them in other places consider creating modules with public API or put htese vars behind namespace to limit global scope pollution.

homepage/front/scripts/search/search.ts
perjg: For clarity, I'll wrap the contents of search.ts in a closure.

I have not really made any effort to isolate namespaces for the functionality on the homepage, since there are so few functions, but to minimize errors going forward, I agree that it makes sense that other isolated functionality that is meant to be local is wrapped in a closure.

homepage/front/scripts/search/search.ts
perjg: Good point... and especially now that I added the hasOwnProperty, the code won't even be evaluated unless params.q exists :)

homepage/front/scripts/main/main.ts
perjg: mobileBreakpoint should be global.

parallaxWindow doesn't need to be global. I'd like to refactor this part of the code later, so I'll add a comment to that effect in the code

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,63 @@
...
+	var mobileBreakpoint : number = getGlobals().mobileBreakpoint;
+
+	function processArguments() : any {
+		var params : any = {},
+			parts : Array<string>;
+
+		if (location.search) {
+			parts = location.search.substring(1).split('&');
+
+			for (var i = 0; i < parts.length; i++) {
```
dianafa: one var per function

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,63 @@
...
+				var nv = parts[i].split('=');
```
dianafa: one var per function

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,63 @@
...
+			if ($(document).width() < mobileBreakpoint) {
+				$('#searchWikiaTextMobile').val(decodeURIComponent(params.q));
+			} else {
+				$('#searchWikiaText').val(decodeURIComponent(params.q));
+			}
+		}
+	}
+
+	function loadSearch() : void {
+		console.log(mobileBreakpoint);
```
dianafa: do we need this here?

homepage/front/scripts/main/main.ts
```diff
@@ -3,7 +3,11 @@
 
 'use strict';
 
-var parallaxWindow : JQuery = $('#js-parallax-window');
+declare var getGlobals : any;
+
+// TODO: Refactor into module. Only mobileBreakpoint needs to be global
```
RafLeszczynski: please put in link to jira ticket for this refactoring

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,63 @@
...
+			} else {
+				$('#searchWikiaText').val(decodeURIComponent(params.q));
+			}
+		}
+	}
+
+	function loadSearch() : void {
+		// Google custom search injection
+		// https://developers.google.com/custom-search/docs/tutorial/implementingsearchbox
+		var searchKey : string = '006230450596576500385:kcgbfm7zpa8',
```
RafLeszczynski: this key is used here and in https://github.com/Wikia/mercury/blob/INT-195/front/scripts/main/routes/SearchResultsRoute.ts#L20

Please consider storing in a one place to avoid code duplication.

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,63 @@
...
+			googleCustomSearch : HTMLScriptElement = document.createElement('script'),
```
RafLeszczynski: Please consider using similar pattern for loading JS asset as here: https://github.com/Wikia/mercury/blob/INT-195/front/scripts/main/routes/SearchResultsRoute.ts#L25 

For example `$.getScript()`

homepage/front/scripts/search/search.ts
perjg: Do you have any suggestions for where to put shared config like this?

Right now mercury/config contains files used by the mercury app only. I can add a file in there if that is ok.

In either case I will make a new CR after this one is merged to change to using a config files for both places at the same time.

RafLeszczynski: `mercury/config` is OK and if you want to do it as separate task please put the TODO comment with the link to Jira ticket

RafLeszczynski:
Also please consider adding JSDoc comments to each function listing its `@desc` list of `@param` with their types and value it `@returns`

perjg:
Created INT-217 to follow up on documentation/refactoring issues.

RafLeszczynski:
@perjg if there are any TODO in the code that you plan to address in the future, please link them with appropriate Jira issue by adding comment in the code. For example:
`// TODO: This requires refactoring because ......... More info here: link_to_Jira.html`

perjg:
@RafLeszczynski I've added INT-222 as a placeholder for creating a config file for values that are shared between Japan HP and Mercury.

Please note that the searchKey from your comment is probably not the best candidate for this. Currently, we plan to do an A/B test for Google Custom Search integration. If the A/B test is favorable to integrating GCS long-term, we will probably wrap the API in a service and remove the requirement for having the key embedded in the front end.

RafLeszczynski:
@perjg If these changes are temporary (for A/B test only) please add appropriate comment and create Jira issue to remove this code after test is done. Add link to this Jira issue in comment

perjg:
To clarify: A/B test is for the mercury part, so in the future the code may be left for just the homepage.

In other words, it is not relevant to this PR.

perjg:
Will merge Tuesday if there are no further comments.

homepage/server/util.js
```diff
@@ -52,8 +52,12 @@ exports.getSignupUrl = function () {
 	return localSettings.signupUrl;
 };
 
+exports.getMobileBreakpoint = function () {
+	return 710;
```
hakubo: This looks a bit over engineered to me.  Can we make it simpler?

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,57 @@
...
+declare var getGlobals : any;
+
+(function() : void {
+	var mobileBreakpoint : number = getGlobals().mobileBreakpoint;
+
+	function processArguments() : any {
+		var params : any = {},
+			parts : Array<string>,
+			i : number,
+			nv : Array<string>;
```
hakubo: Please use more descriptive vars

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,57 @@
...
+
+		if (location.search) {
+			parts = location.search.substring(1).split('&');
+
+			for (i = 0; i < parts.length; i++) {
```
hakubo: Consider forEach here

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,57 @@
...
+
+		return params;
+	}
+
+	function fillSearchTextBox() : void {
+		var params : any = processArguments();
+
+		if (params.hasOwnProperty('q')) {
+			if ($(document).width() < mobileBreakpoint) {
+				$('#searchWikiaTextMobile').val(decodeURIComponent(params.q));
```
hakubo: Is it safe from XSS?

homepage/server/util.js
perjg: The breakpoint for mobile is required by the front end javascript in a couple of places.

It is included in the globals sent to the front-end, and the value is hardcoded in getMobileBreakpoint. Not sure how I can make it any simpler.

homepage/front/scripts/search/search.ts
perjg: It should be: The q argument is used by the Google Custom Search script to read the search term.

rogatty: @perjg please do some testing to be sure.

perjg: @rogatty @hakubo The cse.js script injects some other google custom search related scripts and css, and uses encodeURIComponent() to encode everything, so it looks fine to me.

homepage/server/facets/globals.js
```diff
@@ -9,10 +9,12 @@ var util = require('../util');
 function globals(request, reply) {
 	var loginUrl = util.getLoginUrl(),
 		signupUrl = util.getSignupUrl(),
+		mobileBreakpoint = util.getMobileBreakpoint(),
 		script = 'function getGlobals () {' +
```
hakubo: can we create a json out of that?
creating a function this way does not look like a healthy solution

perjg: @hakubo Currently globals sent from the backend to the front end are generated in the backend and put into a script globals.js, where they are all loaded from in the frontend.

I agree that this isn't a very nice solution, and I was thinking of doing something like this:

* Add a /globals route to the backend app that returns a json file with the current value of the globals
* Load globals async on page load and cache
* Use cached globals whenever they are requested

Is this what you mean, or am I misunderstanding?

perjg: @hakubo I refactored the globals here: https://github.com/Wikia/mercury/pull/1534

perjg:
I'd like to go ahead and merge this now.

Globals refactoring is here: 
https://github.com/Wikia/mercury/pull/1534

Will add the extra global added here in the same way once that PR has been merged.

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,57 @@
...
+(function() : void {
+	var mobileBreakpoint : number = getGlobals().mobileBreakpoint;
+
+	function processArguments() : any {
+		var params : any = {},
+			argParts : Array<string>,
+			argValues : Array<string>,
+			i : number;
+
+			if (location.search) {
```
hakubo: indentation is off,
please run some autoformatter on the code

homepage/front/scripts/search/search.ts
```diff
@@ -0,0 +1,57 @@
...
+			}
+		}
+	}
+
+	function loadSearch() : void {
+		// Google custom search injection
+		// https://developers.google.com/custom-search/docs/tutorial/implementingsearchbox
+		// TODO: Consider keeping searchKey in separate config file as it is currently shared with Mercury
+		var searchKey : string = '006230450596576500385:kcgbfm7zpa8',
+			url : string = (document.location.protocol === 'https:' ? 'https:' : 'http:') +
```
hakubo: From Paul Irish:
http://www.paulirish.com/2010/the-protocol-relative-url/

It used to be better to ask for // relative protocol

now it's rather recommended to always use https if available

perjg:
We are currently running an A/B test for Google Search on Oasis and Mercury only. This PR will not be merged until at least after the A/B test completes, at which point we will decide what to do about Japan HP.

hakubo:
@perjg Hey, we will be merging ES6 soon and this PR will not be mergable. Do we want to keep that PR? Should we update it to ES6? Thanks

perjg:
@hakubo We will not integrate google search with Japan homepage for the time being, so I am closing this PR. Will open a new one if based on this branch if we need it later.

PR #1423, opened 2015-09-24
SOC-1118 Decouple share icons from share header

mklucsarits:
These are changes I made prior to the work for adding share icons to discussion pages in locations that aren't the header (which would happen on desktop view). This separates the share icons from the share header, so that we can re-use the ShareFeatureComponent around the site wherever it's needed, which helps us to use consistent icons and sharing logic.

The goal here is that there is no resulting change in UX or overall logic for the user.

@garthwebb 
/cc @inez @rogatty


front/scripts/main/controllers/ApplicationController.ts
```diff
@@ -25,8 +25,8 @@ App.ApplicationController = Em.Controller.extend(App.LoadingSpinnerMixin, App.Al
 	lightboxType: null,
 	lightboxModel: null,
 	lightboxVisible: false,
-	// Controls the appearance of the share-feature component
-	enableSharingHeader: false,
+	// Controls the appearance of the share-header component
+	enableShareHeader: false,
```
mklucsarits: I renamed this property to match up with the share-header component.

rogatty:
@mklucsarits sorry for missing this one. Please don't merge without explicit LGTM - https://github.com/Wikia/guidelines/blob/master/Process/PullRequests.md

Release was delayed and we had to fix selenium tests because of this PR (CSS class that test was depending on was changed). If someone reviewed it maybe we would catch it sooner.

mklucsarits:
Sorry @rogatty, I'll fix this today.

PR #1424, opened 2015-09-24
SOC-1291 Use of contributors component and avatar component

armonr:
* Use the user avatar component in discussion pages using raw html
* Addition of Discussion contributors component to the hero unit
* Addition of Discussion contributors component to left rail

/CC @mklucsarits @garthwebb @bkoval 

https://wikia-inc.atlassian.net/browse/SOC-1291


front/scripts/main/models/DiscussionPostModel.ts
```diff
@@ -17,7 +18,7 @@ App.DiscussionPostModel = Em.Object.extend({
 				url: 'https://' + M.prop('servicesDomain') + '/discussion/' +
 					 this.wikiId + '/threads/' + this.threadId +
 					 '?responseGroup=full' +
-					 '&sortDirection=descending' +
+					 '&sortDirection=descending&sortKey=creation_date' +
```
mklucsarits: Was this left in by accident?

front/styles/main/component/_discussion-hero-unit.scss
```diff
@@ -16,6 +17,22 @@ $discussion-hero-unit-text-color: $color-white;
...
+	.discussion-contributors {
+		.avatar {
+			img {
+				height: $discussion-hero-unit-avatar-dimension;
+				width: $discussion-hero-unit-avatar-dimension;
+			}
+		}
+
+		.discussion-contributors-more {
+			font-size: rem-calc(12);
```
mklucsarits: You can use $type-min here instead for 12rem size font. Reference https://github.com/Wikia/style-guide/blob/dev/src/scss/lib/_settings.scss for all variables you can use from the style guide.

front/styles/main/module/_discussion.scss
```diff
@@ -72,6 +72,20 @@ $horizontal-padding: 15px;
 		}
 	}
 
+	.discussion-contributors {
```
mklucsarits: Just my opinion, because this file is getting too large and unfocussed: I think all the styles for the DiscussionContributorsComponent should go in /front/styles/main/components/_discussion-contributors.scss, even if they are part of another container or component. Other component styles in this file, like post-detail and post-reply, should ideally be moved to their own files also.

front/scripts/main/models/DiscussionPostModel.ts
```diff
@@ -68,7 +71,20 @@ App.DiscussionPostModel.reopenClass({
 						replies.reverse();
 					}
 
+					replies.forEach(function (reply: any) {
```
garthwebb: Is there anywhere common we could put this chunk of code?  I'm not really sure where such things would go but it seems like we could have this be in only one place.

armonr: Yes, that'd be great. I was not sure where this would go. What do you think, @mklucsarits?

garthwebb:
Other than dup code question, LGTM

front/scripts/main/models/DiscussionPostModel.ts
mklucsarits: I think the only thing you could do to share functionality between two models is create a mixin. I personally don't think it's worth it here, but if you want to, you could extract all the contributors logic to a mixin used by these two models.

mklucsarits:
LGTM

front/scripts/main/models/DiscussionPostModel.ts
armonr: Thanks, it's good to know for future reference. I'll leave this as is for now.

armonr:
Closing in favor of https://github.com/Wikia/mercury/pull/1469

PR #1433, opened 2015-09-25
XW-435 Use layoutName instead of deprecated templateName

rogatty:
https://wikia-inc.atlassian.net/browse/XW-435

### DIFF
--- a/front/scripts/main/components/WikiaMapComponent.ts
+++ b/front/scripts/main/components/WikiaMapComponent.ts
@@ -3,18 +3,18 @@
 
 App.WikiaMapComponent = App.ImageMediaComponent.extend({
 	classNames: ['wikia-map'],
+	templateName: 'components/wikia-map',
 
 	url: null,
 	title: null,
 	imageSrc: null,
 	id: null,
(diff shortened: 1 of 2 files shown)


front/scripts/main/components/WikiaMapComponent.ts
```diff
@@ -3,18 +3,18 @@
 
 App.WikiaMapComponent = App.ImageMediaComponent.extend({
 	classNames: ['wikia-map'],
+	templateName: 'components/wikia-map',
```
rogatty: I didn't change this to `layoutName` because this component extends `ImageMediaComponent` and setting `layoutName` would override the one from parent - this is not what we want to do.

Using `templateName` like this isn't deprecated - no warning in the console. I've found this info here: http://stackoverflow.com/questions/23190815/emberjs-extending-a-custom-component-doesnt-get-added-properly-to-a-container#comment35484681_23191409

vforge: Can we add this as a comment to the code?

vforge:
LGTM

PR #1437, opened 2015-09-28
INT-212: Don't return failure message when invalid cookie is set (Jap…

perjg:
…an Homepage)


rogatty:
Been there, done that, LGTM

PR #1438, opened 2015-09-28
INT-195: Add route for showing google custom search results in Mercury

perjg:
For INT-195, we are adding Google Custom Search to Mercury for A/B testing in Japan.

This commit only covers a new route with code to show google custom search results inline.
Integration with optimizely and transition into this route when pressing ENTER in the search text box (SideNavComponent.ts) will be handled in a separate PR.


rogatty:
How about:
```
App.SearchResultsRoute = Em.Route.extend({
	// Don't let the script to start loading multiple times (user opens the route, goes back, opens it again)
	googleCustomSearchLoadingInitialized: false,

	// Return a promise and resolve only after script is loaded - this way the route won't load before it happens
	beforeModel(): Em.RSVP.Promise {
		// FIXME: Add a check if the script is already initialized (use some global probably?)
		if (!this.get('googleCustomSearchLoadingInitialized')) {
			return this.loadGoogleCustomSearch();
		}
		
		// Just resolve an empty promise if the script is already loaded (subsequent transitions to the route)
		return Em.RSVP.Promise.resolve();
	},
	
	loadGoogleCustomSearch(): void {
		var searchKey = '006230450596576500385:kcgbfm7zpa8',
			url = (document.location.protocol === 'https:' ? 'https:' : 'http:') +
			'//www.google.com/cse/cse.js?cx=' + searchKey;
			
		this.set('googleCustomSearchLoadingInitialized', true);
		return Em.$.getScript(url);
	}
});

// Controller is as thin as possible - it should preferably consist of just query params related code
App.SearchResultsController = Em.Controller.extend({
	queryParams: ['q'],
	q: null
});
```

front/scripts/main/controllers/SearchResultsController.ts
```diff
@@ -0,0 +1,24 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../baseline/mercury.d.ts" />
+'use strict';
+
+App.SearchResultsController = Em.Controller.extend({
+	queryParams: ['q'],
+	q: null,
+
+	init: function() : void {
```
RafLeszczynski: Please consider avoiding code duplication. 
https://github.com/Wikia/mercury/blob/INT-203/homepage/front/scripts/search/search.ts#L40
You could abstract this function and reuse it in both places.

front/templates/main/searchResults.hbs
```diff
@@ -0,0 +1,6 @@
+<div>
```
RafLeszczynski: consider avoiding code duplication between this file and https://github.com/Wikia/mercury/blob/INT-203/homepage/server/views/search.html

perjg:
@rogatty I've rewritten the CR based on your suggestion. I didn't add any global state variable, as the google script should handle loading and unloading gracefully itself.

front/scripts/main/routes/SearchResultsRoute.ts
```diff
@@ -0,0 +1,27 @@
+/// <reference path="../app.ts" />
+/// <reference path="../../../../typings/ember/ember.d.ts" />
+
+App.SearchResultsRoute = Em.Route.extend({
+	// Don't let the script to start loading multiple times (user opens the route, goes back, opens it again)
+	googleCustomSearchLoadingInitialized: false,
+
+	// Return a promise and resolve only after script is loaded - this way the route won't load before it happens
```
rogatty: Comments for functions should be in JSDoc format, not `//`
https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#comments

front/scripts/main/routes/SearchResultsRoute.ts
```diff
@@ -0,0 +1,27 @@
...
+	beforeModel(): Em.RSVP.Promise {
+		if (!this.get('googleCustomSearchLoadingInitialized')) {
+			return this.loadGoogleCustomSearch();
+		}
+
+		return new Em.RSVP.Promise((resolve: Function, reject: Function): void => {
+			resolve();
```
rogatty: `return Em.RSVP.Promise.resolve();` should be enough.

rogatty:
@perjg you're not supposed to merge PRs without explicit "LGTM". Please read https://github.com/Wikia/guidelines/blob/master/Process/PullRequests.md

front/scripts/main/routes/SearchResultsRoute.ts
perjg: Using
    return Em.RSVP.Promise.resolve();
Gives this compile error from TypeScript:
    front/scripts/main/routes/SearchResultsRoute.ts(16,26): 2339 Property 'resolve' does not exist on type 'typeof Promise'.

Running 'Em.RSVP.Promise.resolve();' in console works though, so perhaps there is a mismatch in the typings file for Ember?

rogatty: Yup, might be the case. Can you update the typing, please? Don't worry too much about it being clean, anything that will stop errors is fine, we're going to drop TS soon anyway.

perjg: I checked the file available on DefinitelyTyped, and it is the same version
as the one in Mercury right now:
https://github.com/borisyankov/DefinitelyTyped/blob/master/ember/ember.d.ts

Both are version 1.11.3.


Sincerely,
  Per Johan Groland

PR #1444, opened 2015-09-28
XW-417 Revert hotfix as there is a proper fix now on VCL

bognix:
https://wikia-inc.atlassian.net/browse/XW-417

This reverts commit 65c4c8b96d2cfc68adfe2272e37fe48b9c323ed3


rogatty:
LGTM

PR #1452, opened 2015-09-29
ADEN-2476 | Expose article categories to adsContext

fraszczakszymon:
We would like to have article categories in ads context

### DIFF
--- a/front/scripts/main/models/ArticleModel.ts
+++ b/front/scripts/main/models/ArticleModel.ts
@@ -183,7 +183,9 @@ App.ArticleModel.reopenClass({
 			}
 
 			if (source.adsContext) {
-				source.adsContext.targeting.mercuryPageCategories = data.categories;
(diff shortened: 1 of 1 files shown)


front/scripts/main/models/ArticleModel.ts
```diff
@@ -183,6 +183,7 @@ App.ArticleModel.reopenClass({
 			}
 
 			if (source.adsContext) {
+				source.adsContext.targeting.mercuryPageCategories = data.categories;
```
rogatty: Are you sure that when `source.adsContext` exists there is always `source.adsContext.targeting`? If not then it will fail.

fraszczakszymon: It should, but you're right - I'll wrap it with `if`

jacekjursza:
lgtm

PR #1457, opened 2015-09-29
SOC-1355 | Auth fix redirect from /join to /register

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1355

During the redirect we dropped query params which are useful on register page. From now redirect preserves querystring when redirecting to ```/register```
ping @lizlux


rafalkalinski:
LGTM

bkoval:
LGTM :santa:

PR #1460, opened 2015-09-29
HG-879 - redo theme bar behavior to match design spec

katanka:
https://wikia-inc.atlassian.net/browse/HG-879

This adds the ability for the discussion header to only display the theme bar.


front/templates/main/discussion/post.hbs
```diff
@@ -1,3 +1,10 @@
+{{discussion-header
+	showContent=false
+	showSortComponent='showSortComponent'
+	hideSortComponent='hideSortComponent'
+	sortBy=sortBy
+	sortMessageKey=sortMessageKey
```
rogatty: Please keep attrs first and then actions, all sorted alphabetically.

front/templates/main/discussion/post.hbs
```diff
@@ -1,3 +1,10 @@
+{{discussion-header
```
mklucsarits: Please note my comments in the closed PR about this (https://github.com/Wikia/mercury/pull/1376)

front/scripts/main/mixins/HeadroomMixin.ts
```diff
@@ -62,6 +62,9 @@ App.HeadroomMixin = Em.Mixin.create({
 	),
 
 	initHeadroom(headroomOptions: any, offset: number): void {
+		if(this.get('headroomEnabled') === false){
```
bkoval: Spaces after "if" and before the curly bracket are missing :)

front/scripts/main/mixins/HeadroomMixin.ts
```diff
@@ -62,6 +62,9 @@ App.HeadroomMixin = Em.Mixin.create({
 	),
 
 	initHeadroom(headroomOptions: any, offset: number): void {
+		if (this.get('headroomEnabled') === false) {
```
mklucsarits: This seems strange, but I might be missing something. When would you include the HeadroomMixin but not want it enabled? Why ever use it if it's not enabled?

front/styles/main/component/_discussion-header.scss
```diff
@@ -11,7 +11,7 @@ $discussion-header-color-text: #fff;
 	z-index: $z-6; // drop-downs from top-bar component (which is $z-7) display over the discussion header
 
 	&.un-pinned {
-		margin-top: $theme-bar-height;
+		margin-top: $theme-bar-height+$top-bar-height;
```
lizlux: there should be spaces around the `+`. Not sure where/if it's explicitly state in the coding conventions but there's an example of it [here](https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#multi-line-statements).

front/scripts/main/mixins/HeadroomMixin.ts
katanka: AFAIK there's no way to conditionally extend a mixin in Ember. Some components, such as the  new global nav and the discussion header, need to use Headroom on some pages and not on others. So, this property allows a component to sometimes use headroom and sometimes not.

front/templates/main/discussion/post.hbs
katanka: It seems much simpler to just modify the discussion header component than to create an entirely new theme bar component.

front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -12,6 +12,8 @@ App.DiscussionHeaderComponent = Em.Component.extend(App.HeadroomMixin, {
 	}),
 	overlay: null,
 
+	headroomEnabled: true,
+
```
lizlux: Please add `showContent` as a default property of this component and give it a value of `true`

front/templates/main/discussion/post.hbs
lizlux: yeah on the one hand it's a little funny to be adding these toggles to display content and apply headroom, but I like that all the logic for the discussion header and theme bar is in one place. 

The new top-nav will also be taking advantage of the `headroomEnabled` boolean b/c on some pages the top bar will show/hide and others it won't. It's only on discussions for now though, where it's always shown.

mklucsarits: It makes no sense to me why you'd want to add the overhead of using this component on pages that you don't need it. We may as well go back to using a simple `<hr class="theme">` instead.

PR #1462, opened 2015-09-29
SOC-1118 Sharing icons for discussion posts on desktop

mklucsarits:
These changes add the sharing icons for discussion posts on desktop, and updates the current sharing icons with new SVGs.

Related tickets:
https://wikia-inc.atlassian.net/browse/SOC-1118 and https://wikia-inc.atlassian.net/browse/SOC-1222


front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -13,6 +13,53 @@ App.PostDetailComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixi
...
+		this.hideShareTimeout = window.setTimeout(() => {
+			$shareFeature.hide();
+		}, 5000);
+	},
+
+	didInsertElement: function () {
+		if (!this.get('isDetailsView')) {
+			var $shareFeature = this.$('.share-feature');
+
+			this.$('.toggle-share')
```
bkoval: I believe we should avoid using jQuery as much as possible. Ember is giving us a pretty flexible platform for binding events using actions and we should be consistent in using it wherever we can.

This code could be replaced with actions bound to markup in handlebars and their handlers added here, in the "actions" object.

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -13,6 +13,53 @@ App.PostDetailComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixi
...
+	// URL passed to the ShareFeatureComponent for sharing a post
+	sharedUrl: Em.computed(function () {
+		return Em.getWithDefault(Mercury, 'wiki.basePath', window.location.origin)
+			+ '/d/p/' + this.get('postId');
+	}),
+
+	showShareComponent: ($shareFeature: JQuery) => {
+		$shareFeature.show();
+		this.hideShareTimeout = window.setTimeout(() => {
+			$shareFeature.hide();
```
bkoval: I'd consider using CSS classes + transitions with delay for this action to drop calls to jQuery. WHat do you think?

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -13,6 +13,53 @@ App.PostDetailComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixi
...
+	// Timeout used for auto-hiding the sharing icons
+	hideShareTimeout: null,
+
+	// URL passed to the ShareFeatureComponent for sharing a post
+	sharedUrl: Em.computed(function () {
+		return Em.getWithDefault(Mercury, 'wiki.basePath', window.location.origin)
+			+ '/d/p/' + this.get('postId');
+	}),
+
+	showShareComponent: ($shareFeature: JQuery) => {
```
bkoval: Please, consider using ember bindings instead of using jQuery directly.

front/svg/symbols/main/share-google.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
bkoval: Please, run the SVG's through an optimizer https://jakearchibald.github.io/svgomg/ (this one can be over 30% smaller)

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -13,6 +13,53 @@ App.PostDetailComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixi
...
+		$shareFeature.show();
+		this.hideShareTimeout = window.setTimeout(() => {
```
rogatty: Please don't use `setTimeout`, see http://emberjs.com/api/classes/Ember.run.html#method_later

front/scripts/main/components/PostDetailComponent.ts
rogatty: +1, this is hacking against the framework.

mklucsarits: As far as I know we can't bind multiple events and actions to a single element, so this is the approach I was forced to use.

garthwebb: It should be possible.  I know you didn't want the shareComponent to hide or show itself since you wanted it to be usable in places where it was always expanded to show all share options, but I think it makes sense to have a component that takes an `isCollapsible` parameter that triggers this behavior if `true` or always open if `false`.  Or if you don't like that, I think its still possible from outside via the PostDetail component.

Way 1: Remote control via PostDetail:

* Define `click()` and `mouseEnter()` methods in shareComponent.
* Have them both do a `this.sendAction('interact')`
* Include the shareComponent in `post-details.hbs` as
```
{{share-feature interact='toggleShowHide'}}
```
* In `PostDetailsComponent.ts` define a `toggleShowHide` in the `actions` section that will get called whenever a click or mouseEnter happens.  It can toggle a bound variable that you can use in the `post-detail.hbs` template in an `{{#if` block.
* Handle the `mouseLeave` similarly

Way 2 : Full service shareComponent
* Use the same handlers as above within ShareComponent.ts but have them check the `isCollapsible` property to see if they should be paid attention to.

bkoval: @garthwebb that's right! I'm not sure which version (1/2) would I pick but it's definitely an approach worth considering here!

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -9,30 +9,29 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
 	sharedUrl: null,
 
-	didInsertElement: function (): void {
-		if (this.get('sharedUrl') === null) {
-			// If not set, the shared URL is the current page
-			this.set('sharedUrl', Em.computed('title', function (): string {
-				return Em.getWithDefault(Mercury, 'wiki.basePath',
-					window.location.origin) + window.location.pathname;
-			}));
+	computedSharedUrl: Em.computed('title', 'sharedUrl', function (): string {
```
garthwebb: It feels a little funny to include `computed` in the name, as it would if the `sharedUrl` property was called `sharedUrlProperty`.  Since `computedSharedUrl` is never used in the template, maybe this should just be a function instead, like `getSharedUrl`.  Then the other computed properties (e.g. lineShare) would have dependencies on `sharedUrl`

front/scripts/main/components/PostDetailComponent.ts
mklucsarits: @garthwebb One problem is that the "click" and "mouseEnter" events are for an ```{{svg}}`` icon and not for the share-feature component. So I would have to add extraneous markup to wrap the SVG helper in an element that contains an ```{{action}}``` helper. Then there is the issue of that wrapping element only having one event for the action (click or mouseenter, but not both). Alternatively I could create a component to wrap just the SVG icon, but that seems ridiculous to me. It seems I'll be hacking against the framework, as Igor put it, either way, as Ember does not provide a practical solution for this scenario.

front/scripts/main/components/ShareFeatureComponent.ts
mklucsarits: I'm not sure what to rename this to. If we call it ```getSharedUrl```, then we'll have ```get('getSharedUrl')``` which also looks odd.

garthwebb:
LGTM

bkoval:
LGTM :santa:

front/styles/main/module/_discussion.scss
```diff
@@ -119,6 +119,14 @@ $horizontal-padding: 15px;
 			}
 		}
 
+		.share {
+			text-align: right;
+
+			> * {
```
vforge: As a general rule asterisk are very slow anti-pattern - have you considered using `FlexBox` or similar technique instead of fiddling with `text-align`? Also: this will probably break/look wrong on RTL.

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -13,9 +13,42 @@ App.PostDetailComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixi
...
+	// Whether the share-feature component is visible inside this component
+	isShareFeatureVisible: false,
+
+	// Timeout used for auto-hiding the sharing icons
+	hideShareTimeout: null,
+
+	// URL passed to the ShareFeatureComponent for sharing a post
+	sharedUrl: Em.computed('postId', function () {
+		return Em.getWithDefault(Mercury, 'wiki.basePath', window.location.origin)
+			+ '/d/p/' + this.get('postId');
```
hakubo: What is the reasoning behind having one letter route? What user problem this tries to solve?

front/scripts/main/components/PostDetailComponent.ts
```diff
@@ -13,9 +13,42 @@ App.PostDetailComponent = Em.Component.extend(App.DiscussionUpvoteActionSendMixi
...
+				}, 5000);
+			}
+		},
+
+		hideShareComponent(): void {
+			this.set('isShareFeatureVisible', false);
+		},
+
+		cancelHideShareComponent(): void {
+			Em.run.cancel(this.hideShareTimeout);
```
hakubo: you should cancel that event also onDestroy of component. Or is it handled somehow differently?

front/styles/main/module/_discussion.scss
hakubo: +1

front/scripts/main/components/PostDetailComponent.ts
vforge: I was gonna ask for this too. +1

garthwebb: @hakubo its a product decision.  It doesn't solve any particular problem except that we want a short URL prefix so that more of the slug after this is visible.  The URL structure isn't  part of this ticket so if you have any other questions about this, please send an email to Cristina and I.

hakubo: K, thanks!

front/scripts/main/components/PostDetailComponent.ts
mklucsarits: Thanks, I've added this a few lines above.

PR #1463, opened 2015-09-30
INT-195: Add search action

perjg:
Add new action "search" to enable transition into SearchResultsRoute.

This will be activated in a separate PR from enter() in SideNavComponent like so:

	if (<optimizely ab test for google search is set>) {
		this.sendAction('toggleVisibility', false);
		this.send('searchCancel');

		this.sendAction('search', value);
	} else {
	 window.location.assign('%@Special:Searchsearch=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
	}

### DIFF
--- a/front/templates/main/application.hbs
+++ b/front/templates/main/application.hbs
@@ -23,6 +23,7 @@
 	toggleSideNav=(action 'toggleSideNav')
 	toggleSmartBanner=(action 'toggleSmartBanner')
 	toggleUserMenu=(action 'toggleUserMenu')
+	search=(action 'search')
 }}
 	{{outlet}}
 {{/application-wrapper}}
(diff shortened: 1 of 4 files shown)


front/templates/main/application.hbs
```diff
@@ -23,6 +23,7 @@
 	toggleSideNav=(action 'toggleSideNav')
 	toggleSmartBanner=(action 'toggleSmartBanner')
 	toggleUserMenu=(action 'toggleUserMenu')
+	search=(action 'search')
```
rogatty: Please keep it in alphabetic order (properties first, then actions).

front/templates/main/components/application-wrapper.hbs
```diff
@@ -36,6 +36,7 @@
 		shouldBeVisible=sideNavVisible
 		loadRandomArticle=attrs.loadRandomArticle
 		toggleVisibility=attrs.toggleSideNav
+		search=attrs.search
```
rogatty: Please keep it in alphabetic order (properties first, then actions).

rogatty:
Apart from alphabetic order - LGTM

PR #1465, opened 2015-09-30
XW-287 Load User Language through userinfo call, not through WikiVariables

bkoval:
https://wikia-inc.atlassian.net/browse/XW-287

There is already a ticket (https://wikia-inc.atlassian.net/browse/XW-423) to implement user language properly in Mercury. WikiVariables won't be involved.


front/scripts/mercury/modules/Trackers/Internal.ts
```diff
@@ -52,7 +52,7 @@ module Mercury.Modules.Trackers {
 				config: InternalTrackingConfig = {
 					c: mercury.wiki.id,
 					x: mercury.wiki.dbName,
-					lc: mercury.wiki.language.user,
+					lc: M.prop('userLanguage'),
```
bkoval: This is the only place in code I found that was using user language. 

It might be a bit problematic, as an async request to userinfo resource might come after the first calls to Internal tracker, meaning that userLanguage would be undefined. However, It kind of makes sense since we don't know the user language yet (until the userinfo call is done), and defaulting to english before the call would just make tracking inconsistent.

front/scripts/main/CurrentUser.ts
```diff
@@ -50,20 +51,23 @@ App.CurrentUser = Em.Object.extend({
...
-					uiprop: 'rights',
+					uiprop: 'rights|options',
 					format: 'json'
 				},
 				dataType: 'json',
 				success: (result: QueryUserInfoGroupsRightsResponse): void => {
 					var rights = Em.get(result, 'query.userinfo.rights');
 
+					this.language = Em.getWithDefault(result, 'query.userinfo.options.language', 'en');
+					M.prop('userLanguage', this.language);
```
bkoval: Is there any more sufficientway than M.prop to expose this variable?

front/scripts/main/CurrentUser.ts
```diff
@@ -50,20 +51,23 @@ App.CurrentUser = Em.Object.extend({
 		this._super();
 	},
 
-	loadRights(): Em.RSVP.Promise {
+	loadUserData(): Em.RSVP.Promise {
```
rogatty: How about `loadUserInfo` to be consistent with the API?

front/scripts/main/CurrentUser.ts
```diff
@@ -50,20 +51,23 @@ App.CurrentUser = Em.Object.extend({
...
 					meta: 'userinfo',
-					uiprop: 'rights',
+					uiprop: 'rights|options',
 					format: 'json'
 				},
 				dataType: 'json',
 				success: (result: QueryUserInfoGroupsRightsResponse): void => {
 					var rights = Em.get(result, 'query.userinfo.rights');
 
+					this.language = Em.getWithDefault(result, 'query.userinfo.options.language', 'en');
```
rogatty: Use `this.set`

Also clean it up, please - here you set `language` and below you resolve `rights`. Preferably `init` would look something like this:
```
this
	.loadUserInfo()
	.then(this.setRights)
	.then(this.setOptions);
```

I think we should set all user options - why download everything and then throw it away?

front/scripts/mercury/modules/Trackers/Internal.ts
bognix: What  if user is not logged in? Previously it would fallback to content lang (that's how $wgLang works) but now it's going to be undefined or I'm missing some context  here?

front/scripts/main/CurrentUser.ts
```diff
@@ -50,25 +45,49 @@ App.CurrentUser = Em.Object.extend({
 		this._super();
 	},
 
-	loadRights(): Em.RSVP.Promise {
+	loadUserLanguage(result: QueryUserInfoGroupsRightsResponse): Em.RSVP.Promise {
```
rogatty: Wrong type or just its name.

front/scripts/main/CurrentUser.ts
```diff
@@ -50,25 +45,49 @@ App.CurrentUser = Em.Object.extend({
...
+		return new Em.RSVP.Promise(function (resolve: Function, reject: Function): void {
+			var language = Em.getWithDefault(result, 'query.userinfo.options.language', 'en');
+
+			this.set('language', language);
+			M.prop('userLanguage', language);
+
+			resolve(result);
+		}.bind(this));
```
rogatty: Can you use `() => {}` instead?

front/scripts/mercury/modules/Trackers/Internal.ts
```diff
@@ -52,7 +52,7 @@ module Mercury.Modules.Trackers {
 				config: InternalTrackingConfig = {
 					c: mercury.wiki.id,
 					x: mercury.wiki.dbName,
-					lc: mercury.wiki.language.user,
+					lc: M.prop('userLanguage') || mercury.wiki.language.content,
```
vforge: Nope, please see https://github.com/Wikia/app/blob/dev/resources/wikia/modules/tracker.js#L142

front/scripts/main/CurrentUser.ts
```diff
@@ -33,16 +35,9 @@ App.CurrentUser = Em.Object.extend({
...
-						rights[right] = true;
-					});
-
-					this.set('rights', rights);
-				})
+			this.loadUserInfo()
+				.then(this.loadUserLanguage.bind(this))
+				.then(this.loadUserRights.bind(this))
 				.catch((err: any): void => {
 					Em.Logger.warn('Couldn\'t load current user rights', err);
```
vforge: It's not only `rights` now, right? :)

vforge:
LGTM

PR #1467, opened 2015-09-30
XW-229 show message when article redirects to empty title

bognix:
https://wikia-inc.atlassian.net/browse/XW-229

related PR: https://github.com/Wikia/app/pull/8646


front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -71,6 +71,15 @@ App.ArticleRoute = Em.Route.extend({
 			this.notifyPropertyChange('cleanTitle');
 		},
 
+		didTransition(): void {
```
rogatty: `afterModel` maybe?

bognix: in afterModel I would need to make the alert permanent because it's before didTransition in ApplicationRoute and all notifications are removed. 
On the other hand it doesn't make sense to make this notification permanent since it's strictly related with one particular page

rogatty: Still, we shouldn't put model related logic here because of the weird implementation of alerts. IMO it doesn't belong here.

bognix: how about getting model's property in afterModel, setting it as route's property and then read it in didTransiton?

rogatty: How about doing everything in `afterModel` and setting `expiry`? Let's not create complicated logic for an edge case.

bognix: not sure about the edge case if it occurs more than 29,151 every day in Kibana (and AFAIK Kibana is sampled)

front/scripts/main/models/ArticleModel.ts
```diff
@@ -169,7 +170,8 @@ App.ArticleModel.reopenClass({
 					media: App.MediaModel.create({
 						media: article.media
 					}),
-					categories: article.categories
+					categories: article.categories,
+					redirectEmptyTarget: source.redirectEmptyTarget
```
vforge: Shouldn't it be `redirectEmptyTarget: article.redirectEmptyTarget` as `redirectEmptyTarget` is a part of `article` object? (see (1))

vforge:
One comment, LGTM if fixed

tomasznapieralski:
LGTM

vforge:
https://github.com/Wikia/mercury/pull/1467#discussion_r40915685 fixed, LGTM now

front/scripts/main/models/ArticleModel.ts
rogatty: Should we default it to `false` if `undefined` (cached API responses)?

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -54,6 +56,8 @@ App.ArticleRoute = Em.Route.extend({
 
 		// Reset query parameters
 		model.set('commentsPage', null);
+
+		this.set('redirectEmptyTarget', model.redirectEmptyTarget);
```
rogatty: `model.get('redirectEmptyTarget')`

rogatty:
LGTM

PR #1470, opened 2015-09-30
SOC-1194 - Not Found page for Discussions

armonr:
Adds a new 404 page at `/404`

/CC @mklucsarits @garthwebb @bkoval 

https://wikia-inc.atlassian.net/browse/SOC-1194


rafalkalinski:
Correct me if I'm wrong, but I don't think, we should redirect user to the /404 if the url addres is 'Not Found'. It would be good to stay at the same address user wanted to visit and display the message 'Not found' there.
@kvas-damian @garthwebb @bkoval ?

kvas-damian:
I agree with @rafalkalinski, additionally ```/404``` returns status 200 which is against SEO.

front/styles/main/layout/_page-wrapper.scss
```diff
@@ -25,6 +24,12 @@
...
 }
 
-.site-head.theme-bar ~ .page-wrapper {
-	margin-top: 60px;
+@media #{$desktop-range} {
+	.page-wrapper {
+		margin-top: 50px;
+	}
+
+	.site-head.theme-bar ~ .page-wrapper {
```
bkoval: Out of curiosity: why do we need to check for page-wrapper as a sibling of theme-bar? I guess there's a good reason for it :)

front/styles/main/module/_discussion.scss
```diff
@@ -76,6 +76,26 @@ $horizontal-padding: 15px;
 		}
 	}
 
+	.error-content {
+		background-color: $color-white;
+		border-radius: 8px;
+		text-align: center;
+
+		.error-message {
+			color: $aluminum;
```
bkoval: Is this a foundation variable? If yes, we probably should avoid calling them directly as there are plans to evaluate dropping foundation in styleguide.

front/styles/main/module/_discussion.scss
```diff
@@ -76,6 +76,26 @@ $horizontal-padding: 15px;
...
+			font-size: rem-calc(14);
+		}
+
+		.error-action {
+			font-size: rem-calc(12);
+			margin-top: 8px;
+
+			a {
+				color: $oil;
```
bkoval: Same here :)

front/templates/main/discussion/notFound.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="minimized discussion-hero-unit"></div>
+<div class="discussion-page row">
+	<div class="rail large-2 columns">
+		{{community-badge}}
+		{{discussion-back-button}}
+	</div>
+	<hr class="theme desktop-hidden">
+	<div class="discussion post small-12 large-8 columns">
+		<div class="error-container">
+			<div class="error-content">
```
bkoval: Wow, is there any way to simplify the markup a bit? There are 3 layers of divs containing just another div inside :)

bkoval:
@armonr Are there a lot of changes to i18n here? It's not visible at the first sight as the diff is pretty big, maybe because of changed whitespacing?

rogatty:
@bkoval after merging the latest dev it should stop being a problem.

bkoval:
@rogatty thanks!

@rafalkalinski Yeah, that feels more suitable to me as well.

front/styles/main/layout/_page-wrapper.scss
armonr: I just moved this rule here, but maybe @mklucsarits can shed light on it. https://github.com/Wikia/mercury/commit/fc4b7f53df8f7230ce524aaf0ae329b9fb336d06#diff-85a9c48ab225cb8722f97bdcea440979R30

mklucsarits: I honestly don't even remember. And with the changes removing page-wrapper usage on discussion pages, and changing implementation of the theme bar after top-bar integration, this style might not be used anymore.

garthwebb:
@rafalkalinski, @armonr, re: redirecting to the 404 page.  I agree with @rafalkalinski; if you've mistyped a URL, its really annoying to have to type in the whole URL again rather than correct the mistake.

armonr:
@garthwebb & @rafalkalinski: Yes, agreed. I have been working on that; ember transitioning is proving very tricky to me in this case; but stay tuned for the fix!

rafalkalinski:
The story was done as a part of SOC-1193

PR #1472, opened 2015-10-01
INT-215: Add Universal Analytics script to Japan HP

perjg:
Loads the Google Universal Analytics script into Japan HP async.
Same procedure as for main Mercury app.


homepage/server/views/_partials/tracking.html
```diff
@@ -0,0 +1,23 @@
+<!-- Universal Analytics -->
```
RafLeszczynski: Please consider reusing: https://github.com/Wikia/mercury/blob/dev/server/views/_partials/tracking.hbs to avoid code duplication.

RafLeszczynski: Additionally if you adding a code snippet from a third party source (example: GA) please add comment with link describing what it does, why it is required and that it contents shouldn't be changed. In this case adding this would be enough: https://developers.google.com/analytics/devguides/collection/analyticsjs/

perjg:
@RafLeszczynski I've switched to share the tracking template with Mercury.
(To do this I also had to switch the extension for the templates to .html to match Mercury).

Going forward, I'd like to be cautious when including templates directly from Mercury, as it will make the HP app vulnerable to hard to notice regressions. (For example, if somebody now rewrites the tracking template in way that is not compatible with the Japan HP, they will not notice)

RafLeszczynski:
@perjg that's good point. In order to prevent such a situation from happening please consider writing a  test for it.

perjg:
Will merge Tuesday if there are no further comments.

PR #1477, opened 2015-10-01
XW-461 add mw-content class to all modules on main page

bognix:
add mw-content class to all modules on main page not main page itself so handleLink function won't handle link to main page on section and category view

JIRA ticket: https://wikia-inc.atlassian.net/browse/XW-461

### DIFF
--- a/front/scripts/main/components/WikiaStatsComponent.ts
+++ b/front/scripts/main/components/WikiaStatsComponent.ts
@@ -2,7 +2,7 @@
 'use strict';
 
 App.WikiaStatsComponent = Em.Component.extend({
-	classNames: ['wikia-stats'],
+	classNames: ['wikia-stats', ],
 	items: Em.computed('model', function (): any {
 		return [
 			{'label': 'app.pages-label', 'value': this.get('model.articles')},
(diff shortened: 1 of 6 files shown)


front/scripts/main/components/WikiaStatsComponent.ts
```diff
@@ -2,7 +2,7 @@
 'use strict';
 
 App.WikiaStatsComponent = Em.Component.extend({
-	classNames: ['wikia-stats'],
+	classNames: ['wikia-stats', ],
```
rogatty: :no_good:

Warkot: O RLY? https://github.com/Wikia/guidelines/pull/91

Warkot: Or did you mean missing `'mw-content'`? :confused:

bognix: I wanted to add it there and then I changed my mind and forgot about removing the coma. 
It's not needed there since wikia stats is a static component and does't have to handle any interactions

PR #1479, opened 2015-10-01
HG-879 - redo theme bar behavior to match design spec

katanka:
https://wikia-inc.atlassian.net/browse/HG-879

This adds the ability for the discussion header to only display the theme bar.


front/scripts/main/components/DiscussionHeaderComponent.ts
```diff
@@ -12,6 +12,9 @@ App.DiscussionHeaderComponent = Em.Component.extend(App.HeadroomMixin, {
 	}),
 	overlay: null,
 
+	headroomEnabled: true,
```
lizlux: We actually don't technically need this here b/c it's a property of the headroom mixin.

front/templates/main/discussion/post.hbs
```diff
@@ -1,4 +1,7 @@
-<div class="minimized discussion-hero-unit"></div>
+{{discussion-header
+	showContent=false
+	headroomEnabled=false
+}}<div class="minimized discussion-hero-unit"></div>
```
lizlux: `<div...` should be on a new line.

lizlux:
LGTM

PR #1482, opened 2015-10-02
INT-195: A/B test for Google Custom Search integration

perjg:
Code to enable showing search results in Mercury via Google Custom Search.

We do this as an A/B test.

Also includes code to make sure the search is limited to the active community.

### DIFF
--- a/front/scripts/main/components/SideNavComponent.ts
+++ b/front/scripts/main/components/SideNavComponent.ts
@@ -59,15 +59,15 @@ App.SideNavComponent = Em.Component.extend({
 				},
 				variationNumber = Mercury.Utils.VariantTesting.getExperimentVariationNumber(experimentIds);
 
-			// Use Google Search
 			if (variationNumber === 2) {
+				// Use Google Search
 				// Hide SideNav
 				this.sendAction('toggleVisibility', false);
 				this.send('searchCancel');
 
 				this.sendAction('search', value);
-				// Use Wikia Search
 			} else {
+				// Use Wikia Search
 				window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
 			}
 		}
--- a/front/scripts/main/components/SideNavComponent.ts
+++ b/front/scripts/main/components/SideNavComponent.ts
@@ -45,13 +45,28 @@ App.SideNavComponent = Em.Component.extend({
 		},
 
 		/**
-		 * TODO: Refactor, use api
+		 * @desc Handler for enter in search box
 		 *
-		 * Temporary solution for enter on search, will be refactored to be a route in mercury
+		 * Running A/B test to switch between using MediaWiki Special:Search and Google Custom Search
+		 *
 		 * @param value of input
 		 */
 		enter(value = ''): void {
-			window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
+			var experimentIds = { // Experiment id from Optimizely
+					prod: '3571301500',
+					dev: '3579160288'
+				},
+				variationNumber = Mercury.Utils.VariantTesting.getExperimentVariationNumber(experimentIds);
+
+			if (variationNumber && variationNumber === 2) { // Use Google Search
+				// Hide SideNav
+				this.sendAction('toggleVisibility', false);
+				this.send('searchCancel');
+
+				this.sendAction('search', value);
+			} else { // Use Wikia Search
+				window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
+			}
 		}
 	},
 
--- a/front/scripts/main/components/SideNavComponent.ts
+++ b/front/scripts/main/components/SideNavComponent.ts
@@ -52,19 +52,22 @@ App.SideNavComponent = Em.Component.extend({
 		 * @param value of input
 		 */
 		enter(value = ''): void {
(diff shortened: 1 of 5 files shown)


front/scripts/main/components/SideNavComponent.ts
```diff
@@ -45,13 +45,28 @@ App.SideNavComponent = Em.Component.extend({
...
 		 */
 		enter(value = ''): void {
-			window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
+			var experimentIds = { // Experiment id from Optimizely
+					prod: '3571301500',
+					dev: '3579160288'
+				},
+				variationNumber = Mercury.Utils.VariantTesting.getExperimentVariationNumber(experimentIds);
+
+			if (variationNumber && variationNumber === 2) { // Use Google Search
```
dianafa: isn't the `if (variationNumber === 2) ` sufficient here?
Also, let's try do avoid EOL comments.

front/scripts/main/routes/ApplicationRoute.ts
```diff
@@ -149,7 +149,14 @@ App.ApplicationRoute = Em.Route.extend(Em.TargetActionSupport, App.TrackClickMix
 		},
 
 		search: function (searchString : string) {
-			this.transitionTo('searchResults', {queryParams: {q: searchString}});
+			// Only search in current community
+			var basePath : string  = Em.get(Mercury, 'wiki.basePath');
+			basePath = basePath.replace(/.*?:\/\//g, ""); // remove protocol from basePath
```
dianafa: I would be really nice to create a separate method for removing protocols from paths and write unit test for this! :)

RafLeszczynski: @perjg do you need just a `location.hostname` or `location.hostname + location. pathname` ?
@rogatty do we have helper method for this in mercury?

perjg: I need just the base domain name, which is wiki.basePath with the protocol removed.

I couldn't find anywhere with the protocol removed from Em.get(Mercury, 'wiki')

RafLeszczynski: @perjg If you just need the `hostname` you can get it from `window.location.hostname` without using any RegEx. If this is not available anywhere in Mercury as a helper method or a property, consider adding this so next person who will need the same functionality could reuse it. 

@rogatty: if we don't already have it where would be the best place to add it?

@perjg another thing is the using of RegEx. If you need to so something like this `basePath.replace(/.*?:\/\//g, "");` Please consider moving in to separate method and write unit test for it.

front/scripts/main/controllers/SearchResultsController.ts
```diff
@@ -3,6 +3,7 @@
 'use strict';
 
 App.SearchResultsController = Em.Controller.extend({
-	queryParams: ['q'],
+	queryParams: ['q', 'site'],
```
rogatty: Why would you use additional query param that keeps the current site url (`http://starwars.wikia.com/search?site=starwars.wikia.com` :see_no_evil:)? Also there is no need to pass it from `ApplicationRoute` as it's accessible from anywhere.

You can put this in `SearchResultsRoute` instead:
```
setupController(controller: any, model: any, transition: EmberStates.Transition): void {
	var site = whateverMethodYouChoose();
	
	controller.set('site', site);
}
```

front/scripts/main/components/SideNavComponent.ts
```diff
@@ -45,13 +45,31 @@ App.SideNavComponent = Em.Component.extend({
...
+			// Use Google Search
+			if (variationNumber === 2) {
+				// Hide SideNav
+				this.sendAction('toggleVisibility', false);
+				this.send('searchCancel');
+
+				this.sendAction('search', value);
+				// Use Wikia Search
+			} else {
+				window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
```
Warkot: You could use ES6 string templates `` `string {$expression}` ``.

front/scripts/main/routes/ApplicationRoute.ts
rogatty: The best place to add generic functions like that would be `/front/scripts/mercury/utils/`.

Altough in this situation I think that `window.location.hostname` is exactly what @perjg needs.

front/scripts/main/components/SideNavComponent.ts
```diff
@@ -45,13 +45,31 @@ App.SideNavComponent = Em.Component.extend({
...
+				variationNumber = Mercury.Utils.VariantTesting.getExperimentVariationNumber(experimentIds);
+
+			// Use Google Search
+			if (variationNumber === 2) {
+				// Hide SideNav
+				this.sendAction('toggleVisibility', false);
+				this.send('searchCancel');
+
+				this.sendAction('search', value);
+				// Use Wikia Search
```
rogatty: Please move this comment one line lower, doesn't make much sense here.

front/scripts/main/routes/ApplicationRoute.ts
perjg: @rogatty @RafLeszczynski You're right, what I need is exactly available in window.location.hostname.

I'll just drop the RegEx, but I will keep /front/script/mercury/utils/ in mind as a location for general utility functionality.

perjg:
@rogatty @dianafa @Warkot Is it ok to merge this now?

rogatty:
LGTM

PR #1486, opened 2015-10-02
XW-112 Upgrade to Ember 2.0

rogatty:
https://wikia-inc.atlassian.net/browse/XW-112

### DIFF
--- a/front/scripts/main/components/LightboxImageComponent.ts
+++ b/front/scripts/main/components/LightboxImageComponent.ts
@@ -294,23 +294,21 @@ App.LightboxImageComponent = Em.Component.extend(App.ArticleContentMixin, {
 			});
 
 			image.addEventListener('error', (): void => {
-				this.update(null);
+				this.update('', true);
 			});
 		}
 	},
 
 	/**
-	 * @desc Updates img with its src and sets media component to visible state
-	 * If null is given it means that error occured when trying to load the image
+	 * @desc Updates img with its src or displays error
 	 *
 	 * @param {string} imageSrc
+	 * @param {boolean} loadingError
 	 */
-	update(imageSrc: string): void {
-		var loadingError = (imageSrc === null);
-
+	update(imageSrc: string, loadingError: boolean = false): void {
 		if (!this.get('isDestroyed')) {
 			this.setProperties({
-				imageSrc: imageSrc || '',
+				imageSrc,
 				isLoading: false,
 				loadingError,
 				visible: true
--- a/front/scripts/main/components/LightboxImageComponent.ts
+++ b/front/scripts/main/components/LightboxImageComponent.ts
@@ -170,7 +170,9 @@ App.LightboxImageComponent = Em.Component.extend(App.ArticleContentMixin, {
 			direction: Hammer.DIRECTION_ALL
 		});
 
-		this.loadUrl();
+		Em.run.scheduleOnce('afterRender', this, (): void => {
+			this.loadUrl();
+		});
 	},
 
 	/**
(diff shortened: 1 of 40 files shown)


front/scripts/main/components/LightboxImageComponent.ts
```diff
@@ -292,23 +294,28 @@ App.LightboxImageComponent = Em.Component.extend(App.ArticleContentMixin, {
 			});
 
 			image.addEventListener('error', (): void => {
-				this.set('loadingError', true);
-				this.hideLoader();
+				this.update(null);
```
bognix: does it have to be null or any falsy value?

front/scripts/main/components/LightboxMediaComponent.ts
```diff
@@ -74,17 +74,14 @@ App.LightboxMediaComponent = Em.Component.extend(App.ThirdsClickMixin, {
...
+		Em.observer('model', 'currentMedia', function (): void {
+			this.updateHeader();
+			this.updateFooter();
 
-		this.sendAction('setQueryParam', 'file', M.String.normalizeToUnderscore(this.get('currentMedia.title')));
-	}).on('didInsertElement'),
-
-	didInsertElement(): void {
-		// This is needed for keyDown event to work
-		this.$().focus();
```
bognix: this is not needed any more?

front/scripts/main/components/LightboxMediaComponent.ts
```diff
@@ -162,17 +161,20 @@ App.LightboxMediaComponent = Em.Component.extend(App.ThirdsClickMixin, {
 			header = (this.get('currentGalleryRef') + 1) + ' / ' + this.get('galleryLength');
 		}
 
-		this.sendAction('setHeader', header);
+		Em.run.scheduleOnce('afterRender', this, (): void => {
```
bognix: why does it have to be done on afterRender?

test/specs/front/scripts/main/components/CuratedContentComponent.js
```diff
@@ -1,4 +1,5 @@
 moduleForComponent('curated-content', 'CuratedContentComponent', {
+	unit: true,
 	setup: function () {
```
bognix: inconsistency in new lines => https://github.com/Wikia/mercury/pull/1486/files#diff-156feb0a26e9cdf9880f3972037f93ceR5

front/scripts/main/components/LightboxImageComponent.ts
rogatty: `update` is now used for two things:
- displaying error, like in here
- updating `src` if image is loaded properly

Do you think it would be better if it defaulted to `null` or if we had two separate methods? I just wanted to keep it DRY.

front/scripts/main/components/LightboxMediaComponent.ts
rogatty: Didn't work at the first place ;) When we work on Mercury@desktop we can tackle that again.

front/scripts/main/components/LightboxMediaComponent.ts
rogatty: I've refactored it a little bit and added a comment, see https://github.com/Wikia/mercury/commit/a4ee9083ab4e9b660873084ac5b7712d7d8fcd48

Let me know if it's still unclear.

front/scripts/main/components/LightboxImageComponent.ts
bognix: So loading error should only appear when imageSrc is null?
What about undefined, false and empty string? Are those values correct? Should we display an error when those values are passed to function?

Passing null and handling null in some special way makes maintainability, reusability and readability of code a little bit more difficult for me. I believe it's better to stick to one possible type (string) and pass empty string there.

On the other hand if empty string is one of possible values then I would vote for having two functions, one for updating props and second one for showing an error. Or rearrange code in a way that it's DRY and null is not a special case value

hakubo:
LGTM

bognix:
LGTM

bognix:
LGTM

tomasznapieralski:
LGTM!

PR #1487, opened 2015-10-02
[SERVICES-860] Don't check if user is blocked when validating token

ArturKlajnerok:
We are introducing user blocking mechanism to `/info` resource to prevent blocked users from taking any actions on services level. This will be a default  behavior and we don't want it to force it in external clients like Mercury or MW. That's way an additional `noblockcheck` parameter will disable that feature and is introduced here.

More details in SERVICES-851 and in subtasks. Reference PR for MW : https://github.com/Wikia/app/pull/8665

/cc @Wikia/services-team @kvas-damian @bkoval @perjg @lizlux

### DIFF
--- a/server/lib/HeliosSession.ts
+++ b/server/lib/HeliosSession.ts
@@ -50,7 +50,7 @@ module HeliosSession {
 				}
 
 				Wreck.get(
-					authUtils.getHeliosUrl('/info') + '?code=' + accessToken,
+					authUtils.getHeliosUrl('/info') + '?code=' + accessToken + '&noblockcheck=1',
 					callback
 				);
 			}
--- a/homepage/server/auth.js
+++ b/homepage/server/auth.js
@@ -55,7 +55,7 @@ Auth.prototype.login = function (username, password) {
 
 Auth.prototype.info = function (token) {
 	var url = this.baseUrl + 'info?' +
-		'code=' + token;
+		'code=' + token + '&noblockcheck=1';
 
 	return requestWrapper(url);
 };


server/lib/HeliosSession.ts
```diff
@@ -50,7 +50,7 @@ module HeliosSession {
 				}
 
 				Wreck.get(
-					authUtils.getHeliosUrl('/info') + '?code=' + accessToken,
+					authUtils.getHeliosUrl('/info') + '?code=' + accessToken + '&noblockcheck=1',
```
drsnyder: It seems odd that the same code is in here twice. Is there any way to consolidate?

lizlux:
Not sure familiar with this code but this seems fine to me.

rogatty:
LGTM

perjg:
Checked the change for homepage (homepage/server/auth.js)

LGTM.

server/lib/HeliosSession.ts
kvas-damian: JA homepage is completely separate app, that's why it required separate session handler.

kvas-damian:
LGTM

drsnyder:
Are there automated tests for the functionality provided by the authentication verification? More specifically are there automated tests that make sure these changes didn't break anything?

bkoval:
@drsnyder We do have automated tests suite for Mercury available in Jenkins - see http://qa-s2.wikia-prod:8080/view/Mercury/ 

It runs before each release of Mercury, so basically every day, but they can be triggered manually (although it takes around 2 hours to run them).

There are test cases separately for new login and reg, I bet @ludwikkazmierczak has more info about how to run them.

PR #1490, opened 2015-10-05
INT-213: Minify css files for Japan homepage

perjg:
Update to sass gulp task to also minify css and create css sourcemap.


homepage/gulp/tasks/sass.js
```diff
@@ -6,6 +6,8 @@
 var gulp = require('gulp'),
 	sass = require('gulp-sass'),
 	autoprefixer = require('gulp-autoprefixer'),
+	minifyCss = require('gulp-minify-css'),
```
RafLeszczynski: `gulp-sass` has minifying already build-in so you don't need another dev dependency for it. Check how it's done in Mercury:
https://github.com/Wikia/mercury/blob/dev/gulp/tasks/sass.js#L22
Different options are used for `prod` and `dev` environment:
https://github.com/Wikia/mercury/blob/dev/gulp/options/prod.js#L10
https://github.com/Wikia/mercury/blob/dev/gulp/tasks/sass.js#L22

In addition please consider reusing existing code as much as you can (for example sharing gulp tasks between Mercury and JA Home Page). If you have any question or need help with it please ask @hakubo

perjg: @RafLeszczynski As far as I can see, mercury uses outputStyle: 'compressed' for sass, which I also use for for the homepage. Is there some other setting I am missing?

@RafLeszczynski @hakubo I think sharing more of the gulp tasks between mercury and homepage is a good idea, but the structure of the gulp tasks right now doesn't make that easy. Perhaps we can create a separate Jira issue to discuss how to do this? Also, I don't think that should prevent me from doing tweaks and adjustments to the current homepage gulp scripts as in this PR.

RafLeszczynski: @perjg you are using `gulp-sass` without any custom set options. `gulp-sass` uses the same defaults as `node-sass` which for outputStyles is `nested` https://github.com/sass/node-sass#outputstyle. Later you are using `gulp-minify-css`. I was suggesting not using `gulp-minify-css` as you can accomplish everything you need with gulp-sass. In addition having different settings for dev and production environment would be great - similar to what mercury have. Even better would be sharing the same task between mercury and homepage and just passing different configuration with paths and other required options for each of them.

perjg: @RafLeszczynski I have the options set up for sass in my config files already, but it wasn't working, so I added gulp-minify-css instead.

However, now I found the error: I was simply not passing the options correctly to sass in the gulp script.

I'll close this down and create a new simpler CR with just the change in the sass gulp file.

perjg: New CR here: https://github.com/Wikia/mercury/pull/1499

perjg:
Use this CR instead: https://github.com/Wikia/mercury/pull/1499

PR #1408, opened 2015-09-23
allow for trailing comma in ts

hakubo:
as per
https://github.com/Wikia/guidelines/pull/91

PR #1494, opened 2015-10-05
SOC-1292 Add back to forum link to post details page

garthwebb:
Allow users to get back to the list of posts when they are on a post detail page.

https://wikia-inc.atlassian.net/browse/SOC-1292


front/templates/main/discussion/forum.hbs
```diff
@@ -13,7 +13,12 @@
 <div class="discussion-page row">
 	<div class="rail large-3 columns">
 		{{community-badge}}
-		{{discussion-back-button}}
+		<div class="separator"></div>
```
mklucsarits: Why not ```<hr>```?

bkoval: +1 to hr :)

bkoval:
LGTM except the div <-> hr thing :)

front/styles/main/layout/_discussion-rail.scss
```diff
@@ -4,3 +4,10 @@
 	}
 }
 
+$discussion-rail-separator-color: #e6e6e6;
+
+.rail .separator {
```
mklucsarits: Sorry to nitpick, but can it just be ```.rail hr```? Then you don't need the classname that was needed for the div.

bkoval:
LGTM :octopus:

front/scripts/main/routes/DiscussionPostRoute.ts
```diff
@@ -38,6 +38,9 @@ App.DiscussionPostRoute = Em.Route.extend(App.UseNewNavMixin, App.DiscussionRout
 		didTransition(): boolean {
 			this.controllerFor('application').set('noMargins', true);
 			return true;
-		}
+		},
+		goToForum(forumId: number): void {
+			this.transitionTo('discussion.forum', forumId);
+		},
```
armonr: I think this will cause a warning

front/scripts/main/models/DiscussionPostModel.ts
```diff
@@ -2,7 +2,8 @@
 
 App.DiscussionPostModel = Em.Object.extend({
 	wikiId: null,
-	threadId: null,
+	postId: null,
+	forumId: null,
```
armonr: Should these properties (and the ones below) be inside `init()`, [As @rogatty pointed out](https://dockyard.com/blog/2015/09/18/ember-best-practices-avoid-leaking-state-into-factories)?

front/scripts/main/routes/DiscussionPostRoute.ts
garthwebb: What will cause a warning?  Did you try it and see a warning?

front/scripts/main/models/DiscussionPostModel.ts
garthwebb: I don't think these will be affected since these are scalar, null-type default values, rather than an array that could be added to, or a meaningful default that could be assumed to be valid by an instance.

front/scripts/main/routes/DiscussionPostRoute.ts
rogatty: I think that @armonr meant the trailing comma here. It shouldn't cause a warning anymore as we've changed tslint rules, see https://github.com/Wikia/mercury/pull/1408

front/scripts/main/models/DiscussionPostModel.ts
bognix: I believe it's safe for nulls and ints, I'm not so sure about array below https://github.com/Wikia/mercury/pull/1494/files#diff-6ca9940dddda8e8ce9d94b5f94033310R9.

Check out the paragraph about "Leaking state into the class" it mentioned article.
What is more I suggest using Ember.A() instead of [] -  https://one.wikia-inc.com/wiki/User_blog:Bogna_Knychala/Fun_With_Ember:_JS_Arrays we already have this problem with Curated Content Editor and Infobox builder

front/scripts/main/routes/DiscussionPostRoute.ts
armonr: NVM, then. Thanks!

front/scripts/main/models/DiscussionPostModel.ts
armonr: Scalars are not immutable, are they? I still suspect that these properties are class level rather than instance level. @rogatty?

rogatty: @armonr scalar properties don't leak, check the demo I've prepared: http://emberjs.jsbin.com/wolivahucu/1/edit?js,console

There is also another good article - http://www.cerebris.com/blog/2012/03/06/understanding-ember-object/ - in which you can read:
>When declaring objects or arrays in your classes, you'll typically want to initialize them along with each instance in the init() function. In this way, each of your objects will receive its own unique instances of objects and arrays.

@bognix yes, it's not safe to use `replies` and `contributors` like that.

mklucsarits:
LGTM

bkoval:
LGTM after resolving https://github.com/Wikia/mercury/pull/1494#discussion-diff-41336845

garthwebb:
@bkoval, I think @rogatty cleared those concerns when he said that scalars don't link.

front/scripts/main/models/DiscussionPostModel.ts
armonr: Thanks for clarifying!

PR #1495, opened 2015-10-06
SOC-1387 Conform to top-bar component styling in disc pages

armonr:
This was a blocking issue when trying to test discussion pages. This changeset introduces a global wrapping class for the "body" which houses all the content on the page after the top-bar component in the header of the page. Additionally adds minor fixes specific to mobile and desktop.

Use https://github.com/Wikia/mercury/pull/1495/files?w=1 view

/CC @lizlux @bkoval @garthwebb @mklucsarits 

https://wikia-inc.atlassian.net/browse/SOC-1387


front/styles/main/component/_discussion-community-badge.scss
```diff
@@ -1,3 +1,5 @@
+$z-header-badge-image: $z-6;
```
rogatty: Current convention is to put it here: https://github.com/Wikia/mercury/blob/dev/front/styles/main/_wikia-variables.scss#L6

mklucsarits:
The code is fine, but at some point we need to rethink how page layout should work in Mercury going forward, between mobile and desktop. First we removed use of the ```.page-wrapper``` div on discussion pages, and now we essentially just replaced it with ```.site-body```. That seems silly to me. Also, having this ```.site-body``` wrapper in multiple templates violates DRY and should be in a parent template.

bkoval:
Agreed. Let's set up a discussion about the cleanest way to use <ercury on both mobile and desktop, but let's not block this ticket with it.

mklucsarits:
Alright, LGTM.

PR #1496, opened 2015-10-06
INT-215: Revert Reuse tracking template from Mercury as shared templa…

perjg:
…tes are not available after deployment.

Follow-up to https://github.com/Wikia/mercury/pull/1472.

It is not possible to re-use the templates because the homepage is run as a separate app and copied to its own location, which does not include the mercury templates.

### DIFF
--- a/homepage/server/index.js
+++ b/homepage/server/index.js
@@ -45,7 +45,7 @@ server.views({
 	},
 	path: path.resolve(__dirname, 'views'),
 	layoutPath: path.resolve(__dirname, 'views/_layouts'),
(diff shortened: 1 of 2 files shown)


perjg:
Merging on Friday if there are no comments.

rogatty:
LGTM as a quick fix. Altough we should think about a way to share common code between apps in this repo.

RafLeszczynski:
I agree with @rogatty. @perjg please create a new issue in Jira and speak with @hakubo about this issue.

perjg:
@rogatty @hakubo @RafLeszczynski Created INT-227 as a starting point for discussion about refactoring for better code sharing.

PR #1497, opened 2015-10-06
INT-218: Show avatar for logged in users for Japan HP

perjg:
Show avatar when user is logged in.

Previously, the home page used the attributes api for retrieving user name from the helios user id. This api does not return full path for the avatar.

In order to get _both_ the user name and the full path of the avatar in one api call, switched to use /api/v1/User/Details/ instead.


homepage/front/styles/theme/_icons.scss
```diff
@@ -1,7 +1,8 @@
 .icon {
-	&.login {
+	&.avatar {
+		border-radius: 50%;
 		height: rem-calc(45);
```
RafLeszczynski: Please consider moving all magic number that are repeated throughout this file into sass variables and reuse them in appropriate places

homepage/front/styles/theme/_icons.scss
```diff
@@ -10,14 +11,20 @@
...
-	&.toc {
-		height: rem-calc(13);
-		width: rem-calc(18);
+	&.close-side-menu {
+		height: rem-calc(15);
+		width: rem-calc(15);
+	}
+
+	&.login {
+		height: rem-calc(45);
```
RafLeszczynski: please try avoiding style duplication. All style declarations for `.login`are used in `.avatar`. Please consider grouping selectors for common style declarations or defining common classNames used for these styles

homepage/server/util.js
```diff
@@ -68,11 +69,22 @@ exports.renderWithGlobalData = function (request, reply, data, view) {
...
 
-		return auth.getUserName(data);
+		return auth.getUserInfo(data);
 	}).then(function (data) {
-		request.log('info', 'Retrieved user name for logged in user: ' + data.value);
+		var userName, avatarUrl;
 
-		renderView(true, data.value);
+		// Data is in first element of items array. In case of error, items element does not exist and is ignored
+		try {
```
RafLeszczynski: consider doing this instead of `try catch`:

```javascript
if (data.items && data.items.length) {
    userName  = data.items[0].name;
    avatarUrl = data.items[0].avatar
}
```

perjg: That's what I had originally, but I decided I liked the try catch variant better.

homepage/front/styles/theme/_icons.scss
perjg: @RafLeszczynski The search and close styles in the same file are also the same (copied from Mercury styles).

I assumed this was done to keep some semantic information in the class names. Is this against the style guide?

RafLeszczynski: try this:
```css
.avatar,
.login {
     // common style declarations
}

```

perjg:
Will Merge on Thursday if there are no further comments.

homepage/front/styles/theme/_icons.scss
```diff
@@ -1,23 +1,18 @@
...
-		height: rem-calc(45);
-		padding: rem-calc(12);
-		width: rem-calc(45);
-	}
-
-	&.chevron {
-		height: rem-calc(7);
-		width: rem-calc(12);
-	}
+	$nav-icon-height: rem-calc(45);
```
RafLeszczynski: consider this: `$nav-icon-size: rem-calc(45);`

homepage/server/auth.js
```diff
@@ -9,15 +9,16 @@ var Promise = require('bluebird'),
 	localSettings = require('../config/localSettings').localSettings;
 
 function Auth() {
-	this.baseUrl = localSettings.helios.host;
-	if (this.baseUrl.charAt(this.baseUrl.length - 1) !== '/') {
-		this.baseUrl += '/';
+	function appendSlash(str) {
```
hakubo: why don't we keep those strings in this form in localSettings?

hakubo: also there is standard library to handle urls in node

perjg: @hakubo The strings may also come from environment variables (on preview/production).

I wanted both formats to be acceptable for robustness.

perjg: @hakubo 
This also works:
```
	this.baseUrl = url.resolve(localSettings.helios.host + '/', '.');
	this.servicesUrl = url.resolve(localSettings.servicesUrl + '/', '.');
	this.apiUrl = url.resolve(localSettings.apiUrl + '/', '.');
```

I can change it to be like that if you prefer, but I think it is less obvious what this code does.

perjg: @hakubo Is it ok to keep the code as is for now, or do you want me to change it?
Otherwise I will merge the PR.

homepage/config/localSettings.base.js
```diff
@@ -19,6 +19,7 @@ var deepExtend = require('deep-extend'),
 		//devboxDomain: Utils.stripDevboxDomain(process.env.HOST || process.env.LOGNAME),
 		port: process.env.PORT || 8111,
 		servicesUrl: 'https://services.wikia.com/',
+		apiUrl: 'http://wikia.com/api/v1/',
```
hakubo: what will happen when we move to https?

homepage/server/auth.js
hakubo: Sorry for a lag.

since this is something that we have control over can we support one format and stick to it.
I might not see something and if I do please correct me but why do want to pay the price of maintaining two formats where one is sufficient?

perjg: @hakubo My reasoning was simply that some of these variables come from environment variables on staging/production, and I do not have control over how they are set or access to change them, so I simply wanted to add this extra check for robustness.

homepage/config/localSettings.base.js
perjg: Should I use https? It seems to just redirect to http, so I'm not sure if that is considered good practice for api calls.

hakubo: @gbenson-wikia should we use https here? It sounds to me like this would be more future proof.

homepage/server/auth.js
hakubo: @perjg 
I think you should store apiUrl, servicesUrl and baseUrl in what ever form they come.
then when you use them
instead of

`this.apiUrl + 'User/Details/?ids=' + heliosInfoResponse.user_id;`

you should use:
```
url.resolve(this.apiUrl, 'User/Details?' + querystring.stringify({ids: heliosInfoResponse.user_id}))
```

what do you think?

homepage/config/localSettings.base.js
gbenson-wikia: services.wikia.com is the new API stack, and has https built in from the ground up.  wikia.com/api/v1 is the Nirvana API, which runs off of MediaWiki.  Unfortunately, the Nirvana API can't run in https at the moment because it's called directly from other non-secure MW pages, which themselves can't be converted to https yet without having a massive negative impact on ads.

That said, if the https endpoint already redirects to http, then using https is preferred.  This is so that when (if ever) the Nirvana APIs are able to be changed to use https, you'll already be set without needing additional work in the future.

rogatty: >if the https endpoint already redirects to http, then using https is preferred

But this adds an overhead of redirect (Chrome shows me 330ms). And we would do this to all users only to not have to worry about changing this single line when we move toward https (if ever!). I don't think it's worth it.

hakubo: if performance is at stake how about
relative protocol here, would that work?
`//wikia.com/api/v1/`

homepage/server/auth.js
perjg: @hakubo I like the idea of moving the check to where the variable is actually being used and using the querystring library.

However, if the intention is to work both in the case this.apiUrl has a trailing slash and when it does not, your suggestion cannot be used directly:

For example, given:

```javascript
var querystring = require('querystring'),
	url = require('url'),
	heliosUrlSlash = 'https://services.wikia.com/helios/',
	heliosUrlNoSlash = 'https://services.wikia.com/helios';
```

Then we get different results:

```javascript
> url.resolve(heliosUrlSlash, 'User/Details?' + querystring.stringify({ids: 'user_id'}));
'https://services.wikia.com/helios/User/Details?ids=user_id'
> url.resolve(heliosUrlNoSlash, 'User/Details?' + querystring.stringify({ids: 'user_id'}));
'https://services.wikia.com/User/Details?ids=user_id'
> url.resolve(heliosUrlSlash + '/', 'User/Details?' + querystring.stringify({ids: 'user_id'}));
'https://services.wikia.com/helios//User/Details?ids=user_id'
```

To work around the double slash and to make sure that the relative path doesn't change, I can use `this.baseUrl = url.resolve(localSettings.helios.host + '/', '.');` when assigning the variable. 

Right now, only baseUrl may come from a environment variable, so in the assignement I can use:

```javascript
function Auth() {
	this.baseUrl     = url.resolve(localSettings.helios.host + '/', '.');
	this.servicesUrl = localSettings.servicesUrl;
	this.apiUrl      = localSettings.apiUrl;
}
```

This gets rid of the appendSlash function. Combined with using url.resolve and querystring for all the member functions in the class, I think this will be a good solution.

homepage/config/localSettings.base.js
perjg: @hakubo @rogatty I'd prefer to just continue using http:// for this.

The reason I am using the /api/v1/ endpoint here instead of the services url is that services does not return the full url of the avatar.

This gives me all info I need in one api call, including full avatar url
http://wikia.com/api/v1/User/Details/?ids=26468778

This gives me just the base filename of the avatar, so it is not really possible to use:
https://services.wikia.com/user-attribute/user/26468778/attr/avatar

If the services endpoint is updated to give the full url of the avatar, I can get rid of the /api/v1/ call entirely, and just use the services endpoint (I guess this is the ultimate goal of the services api).

Except for the missing full path for avatar, I can get all info I need with this call:
https://services.wikia.com/user-attribute/user/26468778/

However, the nesting of the reply json will mean I have to do quite a bit of post-processing to extract the data I need, whereas the /api/v1/ call just gives me everything I need in a flat json reply.

homepage/server/auth.js
hakubo: thanks :)

homepage/config/localSettings.base.js
gbenson-wikia: @jcellary can probably assist you with forming the proper avatar URL using the service, which can help simplify things moving forward.

jcellary: Hey @perjg actually we're in the middle of migrating all avatars to use a new format in which the above mentioned call to user-attribute will give you the full url. We're at 25% and by the end of the week we should be at 100% - that means you will always get the full url starting next week (if nothing unexpected happens).

Sample new url format: 

curl https://services.wikia.com/user-attribute/user/5440760/attr/avatar

{"_links":{"self":{"href":"/user/5440760/attr/avatar"}},"name":"avatar","value":"http://static.wikia.nocookie.net/5d08a116-085d-4d2c-afdf-315c573318ed"}

perjg: @jcellary Thanks, that is good news. I'll wait for that change to go live and then amend my code to drop the /api/v1/ call entirely.

perjg: @jcellary When I call the url above now, I get the full path of the avatar. That means it is ok for me to integrate now, right?

jcellary: Hi @perjg, yes it should be all migrated now. It finished just this weekend.

perjg:
@jcellary @hakubo @rogatty Last commit switches from calling /v1/User/Details/ to retrieve user name and avatar url to using the services api to retrieve both (chained promise).

hakubo:
@perjg Hey, we will be working on moving server side to ES6 soon. When do you plan to merge it?

perjg:
@hakubo This change will be merged once it is ok-ed by QA.
It is inside the homepage app, so I don't think it will be affected by the Mercury ES6 switch.

PR #1498, opened 2015-10-06
INT-200: Fix for text getting cut off in featured communities on Japa…

perjg:
…n HP

Use bigText jquery plugin to dynamically adjust text size to make sure it is never cut off for featured communities.

Also refactored gulp build-combined task to include jquery, slick and bigText plugins in combined.js

### DIFF
--- a/homepage/front/scripts/main/main.ts
+++ b/homepage/front/scripts/main/main.ts
@@ -29,7 +29,9 @@ $(function() : void {
 		variableWidth: true
 	});
 
-	// Dynamically adjust text size to show community title without text break
+	// Dynamically adjust text size to show community title without text break.
+	// bigText adjusts the size programatically and strips off css padding, so it is
+	// necessary to add it in explicitly afterwards
 	var headings = $('.grid-heading');
 	headings.bigText({maximumFontSize: 20, verticalAlign: 'top'});
 	headings.css({padding: '.1rem'});
--- a/homepage/front/scripts/main/main.ts
+++ b/homepage/front/scripts/main/main.ts
@@ -5,7 +5,7 @@
 
 // bigText JQuery plugin
 interface JQuery {
-	bigText(options : any) : any;
+	bigText(options: any): any;
 }
 
 var parallaxWindow : JQuery = $('#js-parallax-window');
@@ -30,8 +30,9 @@ $(function() : void {
 	});
 
 	// Dynamically adjust text size to show community title without text break
-	$('.grid-heading').bigText({maximumFontSize: 20, verticalAlign: 'top'});
-	$('.grid-heading').css({padding: '.1rem'});
+	var headings = $('.grid-heading');
+	headings.bigText({maximumFontSize: 20, verticalAlign: 'top'});
+	headings.css({padding: '.1rem'});
 });
 
 function parallax() : void {
--- a/homepage/front/scripts/main/main.ts
+++ b/homepage/front/scripts/main/main.ts
@@ -3,6 +3,11 @@
 
 'use strict';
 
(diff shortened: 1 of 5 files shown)


homepage/front/scripts/main/main.ts
```diff
@@ -3,6 +3,11 @@
 
 'use strict';
 
+// bigText JQuery plugin
+interface JQuery {
+	bigText(options : any) : any;
```
rogatty: Convention is to put spaces like this: `bigText(options: any): any;`

homepage/front/scripts/main/main.ts
```diff
@@ -23,6 +28,10 @@ $(function() : void {
 		centerMode: false,
 		variableWidth: true
 	});
+
+	// Dynamically adjust text size to show community title without text break
+	$('.grid-heading').bigText({maximumFontSize: 20, verticalAlign: 'top'});
+	$('.grid-heading').css({padding: '.1rem'});
```
rogatty: You're searching for the same element twice here.

RafLeszczynski: Please consider caching `.grid-heading` selector

perjg:
Merging on Friday if there are no further comments.

homepage/front/scripts/main/main.ts
```diff
@@ -23,6 +28,11 @@ $(function() : void {
 		centerMode: false,
 		variableWidth: true
 	});
+
+	// Dynamically adjust text size to show community title without text break
+	var headings = $('.grid-heading');
+	headings.bigText({maximumFontSize: 20, verticalAlign: 'top'});
+	headings.css({padding: '.1rem'});
```
RafLeszczynski: Please consider dividing CSS from JS. In this case you could put styles behind css class name and add in to `headings` in JS

perjg: That won't work: The bigText library works by rewriting the contents of the .grid-heading class, so anything I put in padding will be overwritten by the call to bigText. In other words, the headings.css call at the end is a workaround to get the padding back to what I want, as that isn't a parameter I can put into the bigText call.

RafLeszczynski: @perjg: I don't how `bigText` works but if it is overwriting `padding` than is should be possible to customise it by passing custom `padding` in `options` or you can set `padding` in CSS using higher specificity selector. As you said doing it in JS is a workaround which doesn't seam right to me. Maybe you could accomplish font resizing using just CSS - relative font units and breakpoints or units relative to viewport dimension: https://drafts.csswg.org/css-values/#viewport-relative-lengths ?

perjg: @RafLeszczynski I tried to use vw units like you suggest, but browser support is not great, and it doesn't work with the built-in browser in common Japanese Android phones.

The bigText library sets padding to zero internally, and does not allow passing an override for padding as a parameter. The only way to accomplish what I want with this library is thus to call .css afterwards to reinstate the padding I want. This gives exactly the visual result I want, and works in old browsers too.

RafLeszczynski: At least add appropriate comment before `headings.css({padding: '.1rem'});` plus move magic numbers to local variables

perjg: @RafLeszczynski Sure, I'll add comments to clarify.

Do you mean move '.1rem' into a variable called padding and then call
headings.css({padding: padding}); (or alternatively var padding = '.1rem'; headings.css(padding); )

It is only used in that one place, so I am not sure it will make a difference to readability...

PR #1499, opened 2015-10-06
INT-213: Minify css for production for Japan HP

perjg:
Options for minimizing css were not set correctly in the sass gulp task.


rogatty:
LGTM

PR #1502, opened 2015-10-08
INT-195: Fix for incorrect variation number for Google Custom Search …

perjg:
…A/B test

Using variant numbers 0 and 1 rather than 1 and 2.


perjg:
@hakubo Can you have a quick look at this. I'd like to merge it quickly as it will prevent my upcoming A/B test from running correctly.

Right now, the A/B test is not activated in Optimizely, so the return value will be null, so GSE will not be activated until I actually activate the test in Optimizely. The plan is to test it tomorrow on sandbox.

rogatty:
LGTM
Remember that Mercury releases are on hold, not sure for how long.

PR #974, opened 2015-06-24
HG-734 Replace assets path for ember-main.hbs only

rogatty:
https://github.com/Wikia/mercury/pull/962 caused issue where auth pages didn't have assets loaded (tried to load from https://www.wikia.com/mercury-static/styles/auth/app-2bd62818.css). We had to rollback release-94 because of that.

This is a quick fix to unblock release, I've tested it on sandbox-qa04.

@kenkouot please take a look at the root of this issue, thanks.

/cc @kvas-damian @hakubo

PR #1505, opened 2015-10-08
XW-440 Use mercury.nocookie.net for assets defined in partials

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-440


rogatty:
LGTM

gulp/options/prod.js
```diff
@@ -50,7 +50,10 @@ module.exports = {
 		}
 	},
 	replace: {
-		selector: '**/ember-main.hbs',
+		selector: {
+			layouts: '**/_layouts/*.hbs',
```
rogatty: Nope, we already had a P2 because of this change, see https://github.com/Wikia/mercury/pull/974

gulp/options/prod.js
```diff
@@ -50,7 +50,10 @@ module.exports = {
 		}
 	},
 	replace: {
-		selector: '**/ember-main.hbs',
```
hakubo: how about:
`selector: '**/*.hbs'`
or would that capture too much?

hakubo:
LGTM

PR #1507, opened 2015-10-09
XW-252 | load optimizely after ember starts

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-252

### DIFF
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -59,7 +59,23 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'optimizely',
+	initialize () {
+		App.deferReadiness();
+
+		if (M.prop('optimizelyScript') !== undefined && !M.getQueryParam('noexternals')) {
+			Em.$.getScript(M.prop('optimizelyScript')).then(
+				App.advanceReadiness()
+			);
+		} else {
+			App.advanceReadiness();
+		}
+	}
+});
+
+App.initializer({
 	name: 'preload',
+	after: 'optimizely',
 	initialize: (container: any, application: any) => {
 		var $window = $(window);
 
--- a/front/scripts/main/app.ts
+++ b/front/scripts/main/app.ts
@@ -61,14 +61,14 @@ App.initializer({
 App.initializer({
 	name: 'optimizely',
 	initialize () {
-		App.deferReadiness();
+		var optimizelyScript =  M.prop('optimizelyScript');
 
-		if (M.prop('optimizelyScript') !== undefined && !M.getQueryParam('noexternals')) {
-			Em.$.getScript(M.prop('optimizelyScript')).then(
-				App.advanceReadiness()
-			);
-		} else {
-			App.advanceReadiness();
+		if (!Em.isEmpty(optimizelyScript) && !M.getQueryParam('noexternals')) {
+			App.deferReadiness();
+
+			Em.$.getScript(optimizelyScript).always(() => {
+				App.advanceReadiness();
+			});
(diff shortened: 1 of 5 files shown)


front/scripts/main/app.ts
```diff
@@ -59,7 +59,23 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'optimizely',
+	initialize () {
+		App.deferReadiness();
+
+		if (M.prop('optimizelyScript') !== undefined && !M.getQueryParam('noexternals')) {
```
rogatty: Put `M.prop('optimizelyScript')` in a variable and use `!Em.isEmpty(optimizelyScript)`. We don't want to load `null` or empty string.

front/scripts/main/app.ts
```diff
@@ -59,7 +59,23 @@ App.initializer({
...
+			Em.$.getScript(M.prop('optimizelyScript')).then(
```
rogatty: This is weird use of `then`. I think it should be
```
.then(() => {
    App.advanceReadiness();
});
```

server/lib/Utils.ts
```diff
@@ -199,7 +199,11 @@ export function createServerData(localSettings: LocalSettings, wikiDomain: strin
 		mediawikiDomain: getWikiDomainName(localSettings, wikiDomain),
 		apiBase: localSettings.apiBase,
 		environment: getEnvironmentString(env),
-		cdnBaseUrl: getCDNBaseUrl(localSettings)
+		cdnBaseUrl: getCDNBaseUrl(localSettings),
+		optimizelyScript:
+			(localSettings.optimizely.enabled) ?
+			localSettings.optimizely.scriptPath + localSettings.optimizely.account + '.js' :
+			undefined
```
rogatty: If you want something to be `undefined` then you shouldn't define it at first place.

server/views/_layouts/ember-main.hbs
```diff
@@ -72,6 +72,9 @@
 			{{#if error}}
 				M.provide('error', {{{error}}});
 			{{/if}}
+			{{#if server.optimizelyScript}}
+                M.prop('optimizelyScript', '{{{server.optimizelyScript}}}');
```
rogatty: Wrong indentation.

front/scripts/main/app.ts
```diff
@@ -59,7 +59,23 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'optimizely',
+	initialize () {
+		App.deferReadiness();
```
bognix: can this be done inside if statement?
This way you don't have to advanceRediness inside "else"

front/scripts/main/app.ts
rogatty: Also, what if the request fails? Currently Ember won't load. Maybe we should use `.always()` instead?
http://api.jquery.com/jQuery.ajax/#jqXHR

front/scripts/main/app.ts
```diff
@@ -59,7 +59,23 @@ App.initializer({
 });
 
 App.initializer({
+	name: 'optimizely',
+	initialize () {
```
bognix: missing return type

front/scripts/main/app.ts
```diff
@@ -59,7 +59,23 @@ App.initializer({
...
+		var optimizelyScript =  M.prop('optimizelyScript');
```
rogatty: Too... much... space :see_no_evil:

rogatty:
LGTM

PR #1509, opened 2015-10-13
Fixed scoping for table:not for article rendering

perjg:
The original css was causing problems for google custom search rendering. Added .article-content to scope as suggested by @rogatty


rogatty:
Code LGTM but let us test it today.

perjg:
@rogatty Got it... I won't merge until I see another comment here.

perjg:
I tested with the change on sandbox, and it fixes my google custom search issue.

bognix:
I tested infoboxes and tables. Everything looks fine :+1:

PR #1513, opened 2015-10-14
SOC-1273 Introduce discussionsSplashPageConfig app links in front end + App promotion component

armonr:
* Makes Wikia config data originally added for splash pages reusable and accessible from server data
* Creates an App promotion component to be able to link to each Wikia's app
* places the app promo component into the right rail of post and forum pages

/CC @bkoval @mklucsarits


front/svg/symbols/main/apple-store-logo.svg
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
mklucsarits: Please minimize the markup in these SVG files using https://jakearchibald.github.io/svgomg/ or another tool.

front/styles/main/component/_discussion-app-promotion.scss
```diff
@@ -0,0 +1,23 @@
+$text-color: #999;
```
mklucsarits: This variable doesn't seem necessary for only one use.

front/scripts/main/components/DiscussionAppPromotionComponent.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionAppPromotionComponent = Em.Component.extend({
+	classNames: ['discussion-app-promotion'],
+
+	wikiaConfig: Mercury.Utils.prop('wikiaConfig'),
+
+	androidAppLink: Em.computed('wikiaConfig', function () {
+		return this.get('wikiaConfig') ? this.get('wikiaConfig').androidAppLink : null;
```
bkoval: Getters in Ember are asynchronous and we're not sure how much time can it take to obtain a variable. I would consider assigning it's result to a local variable e.g. var wikiaConfig = this.get('wikiaConfig'). What do you think?

front/styles/main/component/_discussion-app-promotion.scss
bkoval: Hmmm, I would actually consider leaving it here, as we'll soon be porting all the colors for discussions to use color pallettes (for dark and light themes) - ping @rafalkalinski and having colors in variables makes them easier to trace and change. 

I would also consider changing the name to be more specific, as we concatenate all the scss files in app.scss and this variable might overwrite something defined in the files imported above and used in the files imported below this one. How about $app-promotion-text-color?

front/svg/symbols/main/apple-store-logo.svg
bkoval: :+1:

front/svg/symbols/main/google-play-logo.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="129px" height="45px" viewBox="0 0 129 45" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sketch="http://www.bohemiancoding.com/sketch/ns">
+    <!-- Generator: bin/sketchtool 1.4 (305) - http://www.bohemiancoding.com/sketch -->
+    <title>Google-Play-Button</title>
```
bkoval: https://github.com/Wikia/mercury/pull/1513/files#r41968086

server/facets/showApplication.ts
```diff
@@ -9,17 +9,22 @@ import Tracking = require('../lib/Tracking');
...
 import Logger = require('../lib/Logger');
 import localSettings = require('../../config/localSettings');
+import wikiaConfig = require('../../config/discussionsSplashPageConfig');
 
 function showApplication (request: Hapi.Request, reply: Hapi.Response): void {
 	var wikiDomain = Utils.getCachedWikiDomainName(localSettings, request),
 		wikiVariables = new MW.WikiRequest({wikiDomain: wikiDomain}).wikiVariables(),
-		context: any = {};
+		context: any = {},
+		hostName: string = request.info.host.replace(
```
bkoval: What do you think about moving this logic to a method in /server/lib/utils.ts?

server/facets/showApplication.ts
```diff
@@ -9,17 +9,22 @@ import Tracking = require('../lib/Tracking');
...
+			/^(?:(?:verify|preview|sandbox-[^.]+)\.)?([a-z\d.]*[a-z\d])\.(?:wikia|[a-z\d]+\.wikia-dev)?\.com/,
+			'$1');
 
 	// TODO: These transforms could be better abstracted, as such, this is a lot like prepareArticleData
 	context.server = Utils.createServerData(localSettings, wikiDomain);
 	context.queryParams = Utils.parseQueryParams(request.query, []);
 	context.localSettings = localSettings;
 	context.userId = request.auth.isAuthenticated ? request.auth.credentials.userId : 0;
+	context.wikiaConfig = wikiaConfig[hostName] ? wikiaConfig[hostName] : {};
```
bkoval: If you need just play store / iTunes links you could create a new object here and pass just these two properties, what do you think? I believe the less bytes we load into our html response the better :)

front/scripts/main/components/DiscussionAppPromotionComponent.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="../app.ts" />
+'use strict';
+
+App.DiscussionAppPromotionComponent = Em.Component.extend({
+	classNames: ['discussion-app-promotion'],
+
+	wikiaConfig: Mercury.Utils.prop('wikiaConfig'),
```
rogatty: `M.prop('wikiaConfig')` is the same thing.

front/scripts/main/components/DiscussionAppPromotionComponent.ts
```diff
@@ -0,0 +1,16 @@
...
+
+	androidAppLink: Em.computed('wikiaConfig', function () {
```
rogatty: `androidAppLink: Em.computed.oneWay('wikiaConfig.androidAppLink')`
See http://emberjs.com/api/classes/Ember.computed.html#method_oneWay

front/templates/main/components/discussion-app-promotion.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if androidAppLink}}
+<div class="discussion-app-join-text">{{i18n "discussion.discussions-app-join"}}</div>
```
rogatty: Please indent, see https://github.com/Wikia/guidelines/tree/master/Handlebars#indentation

rogatty: Also fix the quotes https://github.com/Wikia/guidelines/tree/master/Handlebars#quotes

server/views/_layouts/ember-main.hbs
```diff
@@ -48,6 +48,7 @@
 			M.prop('userId', {{{userId}}});
 			M.prop('cookieDomain', {{{json localSettings.authCookieDomain}}});
 			M.prop('servicesDomain', {{{json localSettings.servicesDomain}}});
+			M.prop('wikiaConfig', {{{json wikiaConfig}}});
```
rogatty: What if `wikiaConfig` isn't there? I think that currently it'll will cause fatal error on article pages.

server/facets/showApplication.ts
rogatty: Also it's not "wikiaConfig", it's just "discussionsSplashPageConfig".
Do we need to send it to all users? Can we limit it to just discussion route?

server/facets/showApplication.ts
rogatty: It looks like something that could use a unit test.

server/facets/showApplication.ts
bkoval: :+1:

front/scripts/main/components/DiscussionAppPromotionComponent.ts
armonr: I'm not sure if the `M` alias is defined here yet.

server/facets/showApplication.ts
armonr: I was tempted to rename `discussionsSplashPageConfig` also to `wikiaConfig`, because looking at the config file, there is nothing specific in there on discussions.
Also, I believe this is going to be removed soon anyways when read-only discussion is released.
But anyway, what would be a better way to do this if we still believe it is worth changing.

armonr: Also the response size change is already negligible and given the single-page nature of the app, this may sound like an optimization we may not need to do just yet.

server/views/_layouts/ember-main.hbs
armonr: Will this not be always populated by https://github.com/Wikia/mercury/pull/1513/files#diff-612dba71afb0074b67bb74b28aba6f89R27?

server/facets/discussions/landingPage.ts
```diff
@@ -1,16 +1,14 @@
...
 
-	function getConfigFromUrl (url: string): WikiaDiscussionsConfig {
+	function getConfigFromUrl(host: string): WikiaDiscussionsConfig {
 		var domain: string;
 
-		domain = url.replace(
-			/^(?:(?:verify|preview|sandbox-[^.]+)\.)?([a-z\d.]*[a-z\d])\.(?:wikia|[a-z\d]+\.wikia-dev)?\.com/,
-			'$1'
-		);
+		domain = Utils.getWikiaSubdomain(host);
```
bkoval: Thank you!

server/facets/showApplication.ts
bkoval: I'm sorry, I know I'm acting like a super-harsh performance police here, but I believe removing additional 9 properties that won't be use anywhere from the first pageview HTML document makes sense, especially given the fact that just 2 of them will be used, so it should be pretty quick to extract them :) I feel like we should do data sanitizing as soon as possible (especially on the backend side) as the bigger the HTML gets, the longer the user has to wait for the first impression of the page.

BTW. You're right that it's going to be removed, but I think it's not going to happen with the read-only release, but later - with releasing contribution tools.

test/specs/server/lib/Utils.js
```diff
@@ -91,6 +91,48 @@ test('getWikiName', function () {
 	});
 });
 
+test('getWikiaSubdomain', function () {
```
bkoval: Thanks for adding unit tests, awesome! :trophy:

server/views/_layouts/ember-main.hbs
mklucsarits: The showApplication handler is not used by all pages, and not article pages.

front/scripts/main/components/DiscussionAppPromotionComponent.ts
mklucsarits: It should be defined there. Let's stick with ```M.prop``` whenever possible so it makes for easy ```grep```ing.

test/specs/server/lib/Utils.js
mklucsarits: :+1:

front/scripts/main/components/DiscussionAppPromotionComponent.ts
rogatty: It's defined in https://github.com/Wikia/mercury/blob/dev/front/scripts/baseline/mercury.ts which is inlined in the initial HTML response so before Ember is even downloaded. Does it not work?

server/views/_layouts/ember-main.hbs
rogatty: You set `wikiaConfig` only for `showApplication` facet. There is also `showArticle` and others where it doesn't exist. Simply wrapping it in `{{#if wikiaConfig}}` should be enough for now (we have to clean it up anyway).

front/templates/main/components/discussion-app-promotion.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if androidAppLink}}
```
rogatty: This condition seems strange. If we have android link then we want to display android and ios one?

server/views/_layouts/ember-main.hbs
armonr: I did not know about separate handlers. Will fix!

server/facets/showApplication.ts
rogatty: >I was tempted to rename discussionsSplashPageConfig also to wikiaConfig, because looking at the config file, there is nothing specific in there on discussions.

This file is mostly what `getWikiVariables` returns, see http://walkingdead.wikia.com/wikia.php?controller=MercuryApi&method=getWikiVariables

I don't know why is there so much duplicated data but I think we shouldn't send it to users twice.

rogatty: Oh, I've checked and here is why:

Discussions landing page doesn't make a request for WikiVariables (probably for performance reasons) so it needed those values to be kept somewhere.

We shouldn't use this file anywhere else, at all. Android and iOS app links should be added to WikiVariables instead as we already request this resource and it's a perfect place for data like this.

armonr: @rogatty: That sounds good, but `wikiVariables` is not added to context in showApplications. How can it be accessed on the frontend?

bkoval:
LGTM after resolving the unit tests issue :)

server/facets/showApplication.ts
rogatty: It's there, see https://github.com/Wikia/mercury/blob/dev/server/facets/showApplication.ts#L29
It can be accessed by `Em.get(Mercury, 'wiki')`

bkoval: @rogatty This is a temporary feature. These links will disappear from the right rail as soon as we're good to go with contribution experience for discussions (and we're already working on it) so I would assume that adding 2 new properties to WikiVariables is out of scope here, especially that:

1) We have a limited number of communities we're shipping discussions to (in the first batch)
2) Two new fields in the API mean bigger number of bytes that needs to be sent between MW and HG in order to show the first render of the page
3) Again, we'll kill this feature as soon as we have contribution experience

rogatty: @bkoval makes sense, but IMO it shouldn't be named `wikiaConfig`. We already have wiki variables ;) Let's not mix them up.

server/views/_layouts/ember-main.hbs
```diff
@@ -48,6 +48,9 @@
 			M.prop('userId', {{{userId}}});
 			M.prop('cookieDomain', {{{json localSettings.authCookieDomain}}});
 			M.prop('servicesDomain', {{{json localSettings.servicesDomain}}});
+			{{#if discussionsSplashPageConfig}}
+				M.prop('wikiaConfig', {{{json discussionsSplashPageConfig}}});
```
rogatty: @armonr forgot about this one? ;)

armonr: Good Catch. Thanks, @rogatty! I pushed this directly to dev 9197095

PR #1518, opened 2015-10-14
XW-474 handle error when wikiVariables or article are null

bognix:
https://wikia-inc.atlassian.net/browse/XW-474


server/lib/MediaWiki.ts
```diff
@@ -217,6 +217,17 @@ export function fetch (url: string, host: string = '', redirects: number = 1, he
 
 				reject(err);
 			} else if (response.statusCode === 200) {
+				if (payload === null) {
+					payload = {
+						exception: {
+							message: 'Empty response, URL: ' + url,
+							code: response.statusCode,
```
vforge: You can write code directly, it's always `200`

server/facets/showArticle.ts
```diff
@@ -73,11 +73,19 @@ function showArticle (request: Hapi.Request, reply: Hapi.Response): void {
 		article
 			.getFull()
 			.then((result: any) => {
-				Utils.redirectToCanonicalHostIfNeeded(localSettings, request, reply, result.wiki);
-				onArticleResponse(request, reply, result.exception, result, allowCache);
+				if (result.exception) {
+					throw Error(result.exception.message);
```
vforge: What about when there's no `message`?

server/lib/Article.ts
```diff
@@ -30,49 +30,30 @@ export class ArticleRequestHelper {
...
+					return {
+						server: Utils.createServerData(localSettings, this.params.wikiDomain),
+						wiki: wikiVariables,
+						article: article.data
+					};
+				}).catch(MediaWiki.ArticleRequestError, (err) => {
+					err.wikiVars = wikiVariables;
+					throw err;
+				});
+			}).catch((err) => {
```
hakubo: `err => throw err`

server/lib/MediaWiki.ts
```diff
@@ -78,7 +78,13 @@ export class WikiRequest extends BaseRequest {
 		return this
 			.fetch(url)
 			.then((wikiVariables: any) => {
-				return Promise.resolve(wikiVariables.data);
+				if (wikiVariables && wikiVariables.data) {
+					return Promise.resolve(wikiVariables.data);
+				} else {
+					return Promise.reject("Empty Wiki Variables");
```
rogatty: How about throwing here too?

server/lib/MediaWiki.ts
```diff
@@ -268,3 +277,27 @@ export function createUrl (wikiDomain: string, path: string, params: any = {}):
...
+WikiVariablesRequestError.prototype = Object.create(Error.prototype);
+
+export class ArticleRequestError {
+	private message;
+	private code;
+	private details;
+	private data;
+	constructor (errAttrs) {
+		this.message = errAttrs.exception.message;
+		this.code = errAttrs.exception.code;
```
rogatty: Are you sure we'll always have `errAttrs.exception`? There is this line in `wrapResult.ts`: `result.status = err ? (err.code || err.exception.code || err.statusCode || 500) : 200;` and it's there for a reason, I guess.

bognix: I'm still working on this part. I created it this way at this point so I can fix other parts of code.
Any input how to handle it is more than welcome

bognix: wrapResult is only applied to Mercury api responses. This exception is thrown when MW response fails. Logic from wrapResult is not applied here

server/facets/showArticle.ts
```diff
@@ -55,9 +57,11 @@ function showArticle (request: Hapi.Request, reply: Hapi.Response): void {
 			})
 			.spread((wikiVariables: any, article: any): void => {
 				article = article.data;
-				onArticleResponse(request, reply, article.exception, {
+				reply.code = article.exception ? article.exception.code || article.exception.statusCode || 500 : 200;
```
hakubo: This line is hard to follow

server/facets/showArticle.ts
```diff
@@ -74,10 +78,31 @@ function showArticle (request: Hapi.Request, reply: Hapi.Response): void {
...
+						server: error.server
+					};
+				reply.code = error.code || error.statusCode || 500;
+				Logger.info('Article has errors');
+
+				Utils.redirectToCanonicalHostIfNeeded(localSettings, request, reply, error.wiki);
+				onArticleResponse(request, reply, result, allowCache);
+			})
+			.catch((error: any): void => {
+				Logger.error('SOMETHING IS REALLY BROKEN: Error when trying to serve an article ', error);
```
hakubo: I don't think we should shout on our logs ☺

PR #1520, opened 2015-10-14
SOC-1225 | Localized social networks in ShareFeatureComponent

bkoval:
@garthwebb @rafalkalinski @armonr Please, take a look :)
https://wikia-inc.atlassian.net/browse/SOC-1225
Also, pinging @rogatty and @inez 

What it does?
 - Allows using different social networks based on user / browser language
 - Adds icons for most popular social networks in our top contries
 - Adds share urls and social networks config for these countries
 - Introduces new directory of SVG files (/svg/symbols/social)
 - Adds a method for obtaining browser language


front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -9,6 +9,57 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
 	title: '',
 	sharedUrl: null,
 
+	socialNetworks: {
+		'en': [
+			'facebook',
```
bkoval: We were discussing creating a hashmap keyed with social network names instead of countries with @rogatty , but this was the only way I figured out that was preserving an exact order of social networks. If you have any ideas about simplifying / shortening this, let me know.

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -9,6 +9,57 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+			'twitter',
+			'reddit',
+			'tumblr'
+		],
+		'jp': [
+			'facebook',
+			'twitter',
+			'g+',
```
inez: I think it would be beneficial to keep parity between symbolic names of different networks and function names that generate URLs: e.g. instead of "g+" and "google" just have "google" and "google" - perhaps "gplus" makes most sense.

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+			+ encodeURIComponent(this.get('title'));
+	},
+
+	meneame(): string {
+		return 'https://www.meneame.net/submit.php?url=' + encodeURIComponent(this.get('computedSharedUrl'));
+	},
+
+	actions: {
+		share(network: string): void {
+			var urlGetter: Function = this.get(network),
```
inez: It's not very obvious what's going on here. Please add comment explaining it or write it in more self-explanatory way

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -11,7 +13,29 @@ App.LanguagesMixin = Em.Mixin.create({
...
-	})
+	}),
+
+	getBrowserLanguage(): string {
+		var lang = navigator.language || navigator.browserLanguage;
+		if (!lang) {
+			return this.get('defaultLanguage');
+		} else {
+			lang = lang.dasherize();
+			if (lang !== 'pt-br') {
```
inez: Add comment why there is exception for "pt-br".

front/scripts/main/components/ShareFeatureComponent.ts
armonr: A slightly different way you can do this is to have a separate map for the social networks and reference them from your share map `socialNetworks`. The advantage would be more flexibility if you want to later on, say, add properties for each social network such as a share-url template. So It'd look like:
```js
var networks: {
'fb': 'facebook',
'tw': 'twitter',
...
},
socialNetworks: {
'en': ['fb', 'tw', ...],
'jp': [...],
...
};
```
And when looping through this, you get the value/value map from `networks`.

armonr:
Is `vk.svg` missing?

bkoval:
Good catch @armonr ! I was supposed to name the network vkontakte, as the svg. But after @inez suggestion I will redefine all of these to be more consistent.

front/scripts/main/components/ShareFeatureComponent.ts
rafalkalinski: That's a good pattern, but in this case it would looked like overengineering :)

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
 		return sharedUrl;
 	}),
 
-	lineShare: Em.computed('title', 'computedSharedUrl', function (): string {
-		return 'http://line.me/R/msg/text/?' + encodeURIComponent(this.get('title') + ' ' + this.get('computedSharedUrl'));
+	currentSocialNetworks: Em.computed('currentUser.language', function (): void {
```
rogatty: Seems like it's returning something.

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
 
-	googleShare: Em.computed('computedSharedUrl', function (): string {
+	google(): string {
 		return 'https://plus.google.com/share?url=' + encodeURIComponent(this.get('computedSharedUrl'));
-	}),
+	},
+
+	reddit(): string {
+		return 'http://www.reddit.com/submit?url=' + encodeURIComponent(this.get('computedSharedUrl')) + '&title='
+			+ encodeURIComponent(this.get('title'));
```
rogatty: `+` at the end of line, see https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#multi-line-statements

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
 		return sharedUrl;
 	}),
 
-	lineShare: Em.computed('title', 'computedSharedUrl', function (): string {
-		return 'http://line.me/R/msg/text/?' + encodeURIComponent(this.get('title') + ' ' + this.get('computedSharedUrl'));
+	currentSocialNetworks: Em.computed('currentUser.language', function (): void {
+		var lang = this.getLanguage(),
+			socialNetworks = this.get('socialNetworks');
+			return socialNetworks[lang] ? socialNetworks[lang] : socialNetworks['en'];
```
rafalkalinski: `socialNetworks[lang] || socialNetworks['en']`

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
 	}),
 
-	facebookShare: Em.computed('computedSharedUrl', function (): string {
+	line(): string {
+		return 'http://line.me/R/msg/text/?' + encodeURIComponent(this.get('title') + ' ' + this.get('computedSharedUrl'));
```
rogatty: How about moving `encodeURIComponent` to `computedSharedUrl` and `encodedTitle`? I see a lot of `encodeURIComponent` in this file.

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+	meneame(): string {
+		return 'https://www.meneame.net/submit.php?url=' + encodeURIComponent(this.get('computedSharedUrl'));
+	},
+
+	actions: {
+		share(network: string): void {
+			var urlGetter: Function = this.get(network),
+				link: string;
+
+			if (typeof urlGetter !== 'function') {
```
rafalkalinski: Is it necessary if we have urlGetter type set as Function?

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+				return;
+			}
+
+			link = urlGetter.bind(this)();
```
rafalkalinski: Would it work if you pass the context in the urlGetter initialization (in the first line of the method)?
If yet, then this line could look much more... natural :)

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+		return 'https://www.meneame.net/submit.php?url=' + encodeURIComponent(this.get('computedSharedUrl'));
+	},
+
+	actions: {
+		share(network: string): void {
+			var urlGetter: Function = this.get(network),
+				link: string;
+
+			if (typeof urlGetter !== 'function') {
+				return;
```
rogatty: You probably want to log it too - `Em.Logger.xxx`

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,82 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+			}
+
+			link = urlGetter.bind(this)();
+
+			if (typeof link === 'string') {
```
rafalkalinski: Is it necessary when we have link type set as string?

front/scripts/main/components/ShareFeatureComponent.ts
rogatty: Yes, TS types aren't used in the browser. They're just for static analysis, before compilation.

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -11,7 +13,29 @@ App.LanguagesMixin = Em.Mixin.create({
...
+				lang = lang.split('-')[0];
+			}
+			return lang;
+		}
+	},
+
+	getLanguage(): string {
+		if (Em.get('currentUser', 'isAuthenticated')) {
```
rafalkalinski: Wouldn't it be enough just to:
`return Em.get('currentUser', 'language') || this.getBrowserLanguage();

front/scripts/main/components/ShareFeatureComponent.ts
bkoval: Agreed. This structure would work perfectly if we'd need any more data than just the network name. In this case, it will be fairly easy to port the existing structure to this one if we need more data. What do you think, @armonr ?

front/scripts/main/components/ShareFeatureComponent.ts
bkoval: Agreed, I will try to be more consistent in naming.

front/scripts/main/components/ShareFeatureComponent.ts
bkoval: But wouldn't it mean two unneeded encoding operations on each transition?

front/scripts/main/components/ShareFeatureComponent.ts
bkoval: Agreed, I'll do so.

front/scripts/main/components/ShareFeatureComponent.ts
bkoval: Setting a type doesn't make you sure the string will actually exist in the browser.

front/scripts/main/mixins/LanguagesMixin.ts
bkoval: No. Em.get('currentUser', 'language') is defaulting to 'en' for anons.

front/scripts/main/mixins/LanguagesMixin.ts
rogatty: Are you sure we don't need another exception for `zh-tw`? https://github.com/Wikia/mercury/blob/dev/crowdin-main.conf

front/svg/symbols/social/facebook.svg
```diff
@@ -0,0 +1 @@
+<svg width="12" height="24" viewBox="0 0 12 24" xmlns="http://www.w3.org/2000/svg"><title>Facebook-Icon-Active</title><g fill="none" fill-rule="evenodd"><path d="M-10-3.5h32v32h-32z"/><path d="M11.063 4.413H9.04c-1.62 0-1.89.81-1.89 1.89v2.562h3.913l-.404 4.048H7.42v10.39H3.374v-10.39H0V8.865h3.373V5.897C3.373 2.524 5.397.5 8.365.5c1.35 0 2.564.135 2.968.135v3.778h-.27z" fill="#4760A5"/></g></svg>
```
rogatty: Did you run these svgs through optimizer? AFAIK `<title>` shouldn't be here.

front/templates/main/components/share-feature.hbs
```diff
@@ -1,16 +1,5 @@
-{{#if isJapaneseBrowser}}
-	<a {{action 'trackClick' 'share' 'line' preventDefault=false}} href="{{lineShare}}" target="_blank" >
-		{{svg 'share-line' role='img' class='icon line'}}
+{{#each currentSocialNetworks as |item|}}
+	<a {{action 'share' item}} href="#">
```
rogatty: Is this `href="#"` useful anyhow?

front/templates/main/components/share-feature.hbs
```diff
@@ -1,16 +1,5 @@
-{{#if isJapaneseBrowser}}
-	<a {{action 'trackClick' 'share' 'line' preventDefault=false}} href="{{lineShare}}" target="_blank" >
-		{{svg 'share-line' role='img' class='icon line'}}
+{{#each currentSocialNetworks as |item|}}
```
rogatty: How about `network` instead of general `item`?

server/views/_layouts/ember-main.hbs
```diff
@@ -87,6 +87,7 @@
 	<body>
 		{{> qualaroo}}
 		<!-- @include ../../../www/front/svg/common.svg -->
+		<!-- @include ../../../www/front/svg/social.svg -->
```
hakubo: :(

We have to come up with better solution for this.
As this will make whole app fatter just for the sake of some social icons that I might never use

front/scripts/main/components/ShareFeatureComponent.ts
hakubo: I think @armonr solution is something that gzip will do  for you anyway

as you mentioned it's ok for now, I'd say

armonr: Sure, ok as is.

inez:
Not a big deal but it seems that each language will have at least three different networks displayed, and for each network encodeURIComponent on the URL gonna be called (generating exactly same result) - so this could be possibly optimized.

front/svg/symbols/social/facebook.svg
bkoval: I did run them through an optimizer.

front/templates/main/components/share-feature.hbs
bkoval: Pointer on desktop.

bkoval:
@inez encodeURIComponent is not going to be called until the network icon is actually clicked.

server/views/_layouts/ember-main.hbs
bkoval: Any quick-term solutions? If not, we coul create a spike to investigate how to handle SVG-s in a better way.

rogatty: We already have https://wikia-inc.atlassian.net/browse/XW-93

front/scripts/main/components/ShareFeatureComponent.ts
rogatty: Yeah, probably. If you think we shouldn't do it then maybe a simple function? IMO we should keep it DRYer.

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,89 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
 		return sharedUrl;
 	}),
 
-	lineShare: Em.computed('title', 'computedSharedUrl', function (): string {
-		return 'http://line.me/R/msg/text/?' + encodeURIComponent(this.get('title') + ' ' + this.get('computedSharedUrl'));
+	currentSocialNetworks: Em.computed('currentUser.language', function (): string[] {
+		var lang = this.getLanguage(),
+			socialNetworks = this.get('socialNetworks');
+			return socialNetworks[lang] || socialNetworks['en'];
```
kvas-damian: What do you think about defining ```defaultLanguage``` in one place for entire platform?

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -19,21 +70,89 @@ App.ShareFeatureComponent = Em.Component.extend(App.TrackClickMixin, App.Languag
...
+		 * Obtains a shared url getter and executes it to get a shared url with a current page details
+		 * In this case, handler should be named after the string in the config object at the top of the file
+		 * @param network
+		 */
+		share(network: string): void {
+			var urlGetter: Function = this.get(network),
+				link: string;
+
+			if (typeof urlGetter !== 'function') {
+				Em.Logger.warn('Shared Url getter for ' + network + ' does not exist');
```
kvas-damian: IMO we should use higher severity. ```error``` works better for me.

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -22,33 +70,101 @@ App.ShareFeatureComponent = Em.Component.extend(
...
+
+		meneame():string {
+			return 'https://www.meneame.net/submit.php?url=' + encodeURIComponent(this.get('computedSharedUrl'));
+		},
+
+		actions: {
+			/**
+			 * Obtains a shared url getter and executes it to get a shared url with a current page details
+			 * In this case, handler should be named after the string in the config object at the top of the file
+			 * @param network
```
hakubo: can you add type?

front/templates/main/components/share-feature.hbs
```diff
@@ -1,16 +1,3 @@
...
-</a>
-<a {{action 'trackClick' 'share' 'twitter' preventDefault=false}} href="{{twitterShare}}" target="_blank">
-	{{svg 'share-twitter' role='img' class='icon twitter'}}
-</a>
-{{#if isJapaneseBrowser}}
-	<a {{action 'trackClick' 'share' 'google' preventDefault=false}} href="{{googleShare}}" target="_blank">
-		{{svg 'share-google' role='img' class='icon google'}}
-	</a>
-{{/if}}
+{{#each currentSocialNetworks as |item|}}
```
hakubo: `s/item/currentNetwork/`

front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -22,33 +70,101 @@ App.ShareFeatureComponent = Em.Component.extend(
...
+				this.get('computedSharedUrl'));
+		},
+
+		facebook():string {
 			return 'http://www.facebook.com/sharer/sharer.php?u=' + encodeURIComponent(this.get('computedSharedUrl'));
-		}),
+		},
 
-		twitterShare: Em.computed('computedSharedUrl', function (): string {
+		twitter():string {
```
mklucsarits: Will these URLs still update appropriately to the correct URL between article transitions, now that you've removed the dependency on updates to `computedSharedUrl`? We don't want to share a URL for the wrong article.

bkoval: They will only be calculated after clicking, so they're always going to get a right url :)

hakubo:
Hey,

I see no LGTM after 10 latest commits - was this code reviewed?

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -11,10 +13,32 @@ App.LanguagesMixin = Em.Mixin.create({
 		return lang === 'ja';
 	}),
 
-	isJapaneseWikia: Ember.computed(function (): boolean {
+	isJapaneseWikia: Em.computed(function (): boolean {
 		return Em.get(Mercury, 'wiki.language.content') === 'ja';
 	}),
 
+	getBrowserLanguage(): string {
```
hakubo: this could be a computed property

front/scripts/main/mixins/LanguagesMixin.ts
```diff
@@ -11,10 +13,32 @@ App.LanguagesMixin = Em.Mixin.create({
...
+			lang = lang.dasherize();
+			//pt-br is the only one supported share-feature language with dash and 5 characters
+			if (lang !== 'pt-br') {
+				lang = lang.split('-')[0];
+			}
+			return lang;
+		}
+	},
+
+	getLanguage(): string {
```
hakubo: this could be a computed property

front/styles/main/theme/_share-icons.scss
```diff
@@ -0,0 +1,90 @@
+.share-feature {
+	.facebook {
+		background-image: url('data:image/svg+xml,%3Csvg%20width%3D%2212%22%20height%3D%2223%22%20viewBox%3D%220%200%2012%2023%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%3Ctitle%3EFacebook-Icon%3C%2Ftitle%3E%3Cg%20fill%3D%22none%22%20fill-rule%3D%22evenodd%22%3E%3Cpath%20d%3D%22M-10-4h32v32h-32z%22%2F%3E%3Cpath%20d%3D%22M11.063%203.913H9.04c-1.62%200-1.89.81-1.89%201.89v2.562h3.913l-.404%204.048H7.42v10.39H3.374v-10.39H0V8.365h3.373V5.397C3.373%202.024%205.397%200%208.365%200c1.35%200%202.564.135%202.968.135v3.778h-.27z%22%20fill%3D%22%23999%22%2F%3E%3C%2Fg%3E%3C%2Fsvg%3E%0A');
```
rogatty: Can you add a comment in this file why we need those svgs twice (svg files and inline)?

bkoval:
Yes, it was. We focused on testing as it's a very time-sensitive thing now and forgot to add our "LGTM"s here. We also talked about this offline with @rogatty

PR #1526, opened 2015-10-16
INT-229: Unit test for edit/add photo in mercury. 

mari-mcmurtrie:
Unit test, not integration test. But the code is depending on file input so maybe it has to be integration test...


mari-mcmurtrie:
@hakubo @kvas-damian this is my unit test attempt. I made this branch from INT-170. Test file is "ArticleContributionComponent.js". Only the last test is failing because of file input (I have to fake it but do not know why).

rogatty:
For easier comparison: https://github.com/Wikia/mercury/compare/INT-170...INT-229

front/scripts/main/components/ArticleContributionComponent.ts
```diff
@@ -20,16 +22,35 @@ App.ArticleContributionComponent = Em.Component.extend({
...
 			M.track({
 				action: M.trackActions.click,
 				category: 'sectioneditor',
-				label: 'addPhoto',
+				label: 'add-photo-no-auth',
 				value: this.get('section')
 			});
+
+			windowObj = windowObj || window; //Injecting window obj for unit test.
+			windowObj.location.href = href;
```
rogatty: How about creating new function `loadPage` that would wrap this exact operation? You could mock it easily in the unit test and check the param that it receives.

mari-mcmurtrie:
@Igor can you take a look? Thanks!!

test/specs/front/scripts/main/components/ArticleContributionComponent.js
```diff
@@ -0,0 +1,82 @@
+moduleForComponent('article-contribution', 'ArticleContributionComponent', {
+	unit: true,
+	setup: function () {
+		M.track = function () {};
+	}
+});
+
+FakeUser = Em.Object.extend({
```
rogatty: `var FakeUser` maybe?

test/specs/front/scripts/main/components/ArticleContributionComponent.js
```diff
@@ -0,0 +1,82 @@
...
+				section: section,
+				sectionId: sectionId,
+				title: title,
+				uploadFeatureEnabled: uploadFeatureEnabled,
+			}
+		});
+		component.openLocation = openLocationSpy;
+		component.send('select');
+	});
+	ok(openLocationSpy.calledOnce);
```
rogatty: Don't you want to check what `href` it was called with? There is some logic that can be tested too.

rogatty:
LGTM, thanks for cleaning it up.

test/specs/front/scripts/main/components/ArticleContributionComponent.js
mari-mcmurtrie: I do not want to check href since the redirect param's domain differs depending on where the test runs (can be sandbox, localhost, etc). I "could" test a part of url which does not change though.

mari-mcmurtrie:
I merged this test into INT-170, and release them together.

PR #1528, opened 2015-10-16
SOC-1174 | Unified discussion URL for front end calls

Bloodmoon24:
https://wikia-inc.atlassian.net/browse/SOC-1174

Please CR @bkoval  @kvas-damian  @rafalkalinski


server/views/_layouts/ember-main.hbs
```diff
@@ -48,6 +48,7 @@
 			M.prop('userId', {{{userId}}});
 			M.prop('cookieDomain', {{{json localSettings.authCookieDomain}}});
 			M.prop('servicesDomain', {{{json localSettings.servicesDomain}}});
+            M.prop('discussionBaseRoute', {{{json localSettings.discuss.baseAPIPath}}});
```
rogatty: Use tabs instead of spaces.

front/scripts/mercury/utils/discussionLink.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference path="../../baseline/mercury.d.ts" />
+
+module Mercury.Utils {
+	export function getDiscussionLink(): string {
```
bkoval: How about getDiscussionServiceUrl? Currently the name doesn't suggest that it's a service url :)

kvas-damian:
In the ticket last Acceptance Criteria stands as following:
> A function should be created that allows construction of the full URL, including path and query parameters.

Can we cover that by our change?

front/scripts/baseline/mercury/utils/buildUrl.ts
```diff
@@ -106,4 +119,8 @@ module Mercury.Utils {
 		// host string (ie. "mercury:8000"), it will be left unmodified and returned as-is.
 		return host;
 	}
+
+	export function getDiscussionServiceUrl(query: any, path: string = ''): string {
```
kvas-damian: IMO it's more intuitive if you put ```path``` param first.

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
```diff
@@ -27,8 +27,8 @@ App.DiscussionRouteUpvoteMixin = Em.Mixin.create({
 
 			Em.$.ajax(<JQueryAjaxSettings>{
 				method: method,
-				url: 'https://' + M.prop('servicesDomain') +
-				'/discussion/' + Em.get(post, 'siteId') + '/votes/post/' + Em.get(post, 'id'),
+				url: M.getDiscussionServiceUrl() + '/' +
```
kvas-damian: While new method accepts ```path``` please use it.

front/scripts/baseline/mercury/utils/buildUrl.ts
Bloodmoon24: Indeed, but IDE had problem with it, said that I have to put 'em in that order. Shall I ignore it?

front/scripts/main/mixins/DiscussionRouteUpvoteMixin.ts
Bloodmoon24: Done

Bloodmoon24:
Suggestions from above have been implemented

front/scripts/baseline/mercury/utils/buildUrl.ts
kvas-damian: Why your IDE is telling you how you should code? ;)
I think we still should put ```path``` first.

rogatty: You get warnings because you use a default parameter, see http://www.typescriptlang.org/Handbook#functions-optional-and-default-parameters
>Just as with optional parameters, default parameters must come after required parameters in the parameter list.

You can use `getDiscussionServiceUrl(path: string, query: any)`

Bloodmoon24: But then I've to check inside if path is not null, for example:
```pathCheck= path ? path : '' ```
to prevent "undefined". So I decided to make both of them default. The advantage of this solution is that, if there is no query string you don't have to pass empty object in each function call. Is that a proper way?

kvas-damian:
Code LGTM,
there are still unit tests to be written.

Bloodmoon24:
Unit tests have been added

kvas-damian:
LGTM
@rogatty can you take a look while @bkoval is out of office?

rogatty:
LGTM

PR #1531, opened 2015-10-16
ADEN-2486 | Add Flite tag support for mercury

fraszczakszymon:
Make it possible to use `<flite>` tag in articles' content on mercury.

Related PR: https://github.com/Wikia/app/pull/8806

### DIFF
--- a/front/scripts/main/components/ArticleContentComponent.ts
+++ b/front/scripts/main/components/ArticleContentComponent.ts
@@ -239,6 +239,7 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 				twitter: 'WidgetTwitterComponent',
 				vk: 'WidgetVKComponent',
 				polldaddy: 'WidgetPolldaddyComponent',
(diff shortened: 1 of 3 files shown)


nandy-andy:
Update the description, please :)

nandy-andy:
LGTM :+1:

@bognix @rogatty @vforge do you guys wanna take a look? :smile_cat:

vforge:
So basically that will load as many flite <scripts> as tags?

nandy-andy:
That's correct @vforge! And it'll add as many `<script />` tags and `` tags as `<flite />` tags on a page. 

But keep in mind it's a "staff tool" and it's usually used as one per a page since our Community Development team puts a quiz on a page with the tag. Also, the `<script />` tag which is being added provides some data used later by the logic of `combo.js` script, so having one of it per page with many flite tags isn't that trivial, I'm afraid :disappointed:

rogatty:
It's a shame we can't use some API instead of `<script>` tag attributes :(
Apart from that LGTM.

PR #1534, opened 2015-10-19
INT-217: Return front end globals for Japan HP via json file

perjg:
Refactoring of how globals are sent from backend to frontend.

Previously, this was done by generating a .js file with the globals embedded as variables.

With this change, the globals are returned as json retrieved via a promise and wrapped in a front end class.

### DIFF
--- a/homepage/front/scripts/main/globals.ts
+++ b/homepage/front/scripts/main/globals.ts
@@ -1,0 +1,29 @@
+/// <reference path="../../../../typings/jquery/jquery.d.ts" />
+/// <reference path="../../../../typings/rsvp/rsvp.d.ts" />
+
+'use strict';
+
+class Globals {
+	cachedData: any;
+	promise: RSVP.Promise;
+
+	constructor() {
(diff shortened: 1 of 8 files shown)


perjg:
Will merge on Friday if there are no objections.

homepage/front/scripts/main/globals.ts
```diff
@@ -0,0 +1,29 @@
...
+'use strict';
+
+class Globals {
+	cachedData: any;
+	promise: RSVP.Promise;
+
+	constructor() {
+		this.cachedData = null;
+
+		this.promise = new RSVP.Promise((resolve, reject) : void => {
```
hakubo: I don't get what you're trying to achieve here

why do you wrap $.get in promise 
and then don't use it at all?

perjg:
@hakubo Sorry about that, I simply meant to wrap the $.get in a promise, but left the code incomplete.
Follow-up here: https://github.com/Wikia/mercury/pull/1553

PR #1536, opened 2015-10-19
INT-234: Fixed position for Japan homepage nav on desktop

perjg:
Global nav should be fixed to top of screen in desktop mode.

### DIFF
--- a/homepage/front/styles/component/_nav.scss
+++ b/homepage/front/styles/component/_nav.scss
@@ -25,6 +25,10 @@ header.navigation {
 	width: 100%;
 	z-index: 999;
 
+	@include media ($horizontal-bar-mode) {
+		position: fixed;
+	}
+
 	.navigation-wrapper {
 		@include clearfix;
 		@include outer-container;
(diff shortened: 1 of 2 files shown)


homepage/front/styles/component/_nav.scss
```diff
@@ -25,6 +25,10 @@ header.navigation {
 	width: 100%;
 	z-index: 999;
 
+	@include media ($horizontal-bar-mode) {
```
RafLeszczynski: Please put `@include` before declaring local properties for given selector

perjg: @RafLeszczynski This seems to be in conflict with the rules in styleguide/.scss-lint.yml 

If I move the include up to right under header.navigation I get this error:

homepage/front/styles/component/_nav.scss:26 [W] DeclarationOrder: Expected item on line 26 to appear before line 22. Rule sets should be ordered as follows: `@extends`, `@includes` without `@content`, properties, `@includes` with `@content`, nested rule sets

perjg: so according to this, @include with no properties should come before properties, but @include with content should go after.

RafLeszczynski: You're right, my mistake

perjg:
@RafLeszczynski Thanks for clarifying.

Unless there are other issues, I will merge this tomorrow (Friday)

PR #1539, opened 2015-10-19
XW-516 | Move typings to JSDoc

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-516


front/scripts/baseline/mercury/utils/isPrimitive.ts
```diff
@@ -1,4 +1,8 @@
 module Mercury.Utils {
+	/**
+	 * @param {any} val
+	 * @returns {boolean} isPrimitive
```
hakubo: I think you don't specify a var name in `@returns` as outside the function the var name have no effect

front/scripts/baseline/mercury/utils/loadScript.ts
```diff
@@ -2,11 +2,21 @@
 /// <reference path='../../mercury.d.ts' />
 
 module Mercury.Utils {
+	/**
+	 * @throws URIError exception
+	 * @param {ErrorEvent} error
+	 * @returns {void}
```
hakubo: this I think should be undefined, there is on `void ` type in JS

hakubo: also if we'd say that JSDoc are mendatory for every function
I'd say we should rethink if we want to enforce
`@returns {undefined}` 

sounds like noise

and if all functions would have JSDoc
lack of returns would mean that this function does not return anything

front/scripts/baseline/mercury/utils/state.ts
```diff
@@ -10,6 +10,13 @@ interface ObjectProperties {
 }
 
 module Mercury.Utils {
+	/**
+	 * @param {string} [str='']
```
hakubo: I don't see any default for str in function body

front/scripts/baseline/mercury/utils/state.ts
```diff
@@ -10,6 +10,13 @@ interface ObjectProperties {
...
+	 * @param {string|object} [ns=window]
+	 * @param {any} val
```
hakubo: I think in JSDoc this should be {?object} which means any object or null

front/scripts/main/routes/CuratedContentEditorBlockAddItemRoute.ts
```diff
@@ -4,10 +4,19 @@
 'use strict';
 
 App.CuratedContentEditorBlockAddItemRoute = Em.Route.extend({
+	/**
+	 * @returns {CuratedContentEditorItemModel}
```
hakubo: Can't find @typedef of it

front/scripts/main/routes/CuratedContentEditorRoute.js
```diff
@@ -4,7 +4,7 @@ App.CuratedContentEditorRoute = Em.Route.extend(
 		/**
 		 * @returns {void}
 		 */
-		beforeModel() {
+			beforeModel() {
```
rogatty: Indentation is broken in this file.

front/scripts/auth/common/AuthLogger.ts
```diff
@@ -1,24 +1,36 @@
...
-	status: number;
-	response: any;
-	heliosUrl: string;
-	clientUrl: string;
-}
-
+/**
+ * ClickStreamPayload
+ * @typedef {object} ClickStreamPayload
+ * @property {Object[]} events
```
bognix: capital Object

front/scripts/baseline/mercury/utils/isPrimitive.ts
```diff
@@ -1,4 +1,8 @@
 module Mercury.Utils {
+	/**
+	 * @param {any} val
```
bognix: replace any with *

front/scripts/baseline/mercury/utils/loadScript.ts
bognix: the problem is that eslint can either force you to add jsdocs everywhere or nowhere.
You can't do something like: don't force jsdocs when there is no return value.

But don't get me wrong, I totally agree with you, it introduces unnecessary noise

front/scripts/baseline/mercury/utils/loadScript.ts
```diff
@@ -2,11 +2,21 @@
...
+	 */
 	function loadError (error: ErrorEvent): void {
 		var scriptTag: HTMLScriptElement = <HTMLScriptElement> error.target;
 		throw new URIError('The script ' + scriptTag.src + ' is not accessible.');
 	}
 
+	/**
+	 * @param {string} src
+	 * @param {Function} [fOnload]
```
bognix: fOnload does't look like optional parameter

front/scripts/baseline/mercury/utils/state.ts
```diff
@@ -10,6 +10,13 @@ interface ObjectProperties {
 }
 
 module Mercury.Utils {
+	/**
+	 * @param {string} [str='']
+	 * @param {string|object} [ns=window]
```
bognix: ns does't look like it optional parameter

front/scripts/main/CurrentUser.ts
```diff
@@ -56,6 +63,10 @@ App.CurrentUser = Em.Object.extend({
 		M.prop('userLanguage', userLanguage);
 	},
 
+	/**
+	 * @param {QueryUserInfoResponse} result
```
bognix: This type is not defined anywhere.
Are you sure JSDoc understands it?

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -5,289 +5,370 @@
...
-			    $contributionComponent: JQuery = null;
-
-			headers.forEach((header: ArticleSectionHeader): void => {
-				$contributionComponent = this.createArticleContributionComponent(header.section);
-				$sectionHeader = this.$(header.element);
-				$sectionHeader.prepend($contributionComponent).addClass('short-header');
-				$contributionComponent.wrap('<div class="icon-wrapper"></div>');
+		/**
+		 * Generates table of contents data based on h2 elements in the article
+		 * TODO: Temporary solution for generating Table of Contents
```
bognix: use @TODO - it's official JSDoc notation

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -5,289 +5,370 @@
...
+					height: parseInt(element.getAttribute('height'), 10),
+					imgWidth: element.offsetWidth,
+					media: media,
+				}).createElement();
+
+			return component.$().attr('data-ref', ref);
+		},
+
+		/**
+		 * @param {App.ArticleModel} model
```
bognix: is ArticleModel defined anywhere - in JSDocs?

tomasznapieralski:
front and server passed locally for me

PR #1541, opened 2015-10-19
VE 1960: Mobile rendering of image collections

pauloslund:
This is based off of https://github.com/Wikia/app/tree/DAT-3340
Tested here: http://visualeditor.paulo.wikia-dev.com/wiki/1383649460634

### DIFF
--- a/front/scripts/main/components/ArticleContentComponent.ts
+++ b/front/scripts/main/components/ArticleContentComponent.ts
@@ -168,16 +168,16 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 	},
 
 	replaceImageCollectionPlaceholdersWithComponents(model: typeof App.ArticleModel): void {
-		var $placeholders = this.$('.pi-image-collection');
-		var articleMedia = model.get('media');
-		var index: number;
+		var $placeholders = this.$('.pi-image-collection'),
+			articleMedia = model.get('media'),
+			index: number;
 
 		// TODO: Allow us to pass in the number we want to process
 		var numberToProcess = $placeholders.length;
 
 		for (index = 0; index < numberToProcess; index ++) {
-			var element = $placeholders.eq(index);
-			var collectionMedia: typeof App.ArticleMedia[] = [];
+			var element = $placeholders.eq(index),
+				collectionMedia: typeof App.ArticleMedia[] = [];
 
 			var refs = element.data('refs')
 				.split(',')
--- a/front/scripts/main/components/ArticleContentComponent.ts
+++ b/front/scripts/main/components/ArticleContentComponent.ts
@@ -33,6 +33,7 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 				this.replaceInfoboxesWithInfoboxComponents();
 				this.replaceMapsWithMapComponents();
 				this.replaceMediaPlaceholdersWithMediaComponents(this.get('media'), 4);
+				this.replaceImageCollectionPlaceholdersWithComponents(this.get('media'));
 				this.replaceWikiaWidgetsWithComponents();
 				this.handleWikiaWidgetWrappers();
 				this.handlePollDaddy();
@@ -164,6 +165,43 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 		for (index = 0; index < numberToProcess; index++) {
 		    $mediaPlaceholders.eq(index).replaceWith(this.createMediaComponent($mediaPlaceholders[index], model));
 		}
+	},
+
+	replaceImageCollectionPlaceholdersWithComponents(model: typeof App.ArticleModel): void {
+		var $placeholders = this.$('.pi-image-collection');
+		var articleMedia = model.get('media');
+		var index: number;
+
+		// TODO: Allow us to pass in the number we want to process
+		var numberToProcess = $placeholders.length;
+
+		for (index = 0; index < numberToProcess; index ++) {
+			var element = $placeholders.eq(index);
+			var collectionMedia: typeof App.ArticleMedia[] = [];
+
+			var refs = element.data('refs')
+				.split(',')
+				.compact()
+				.filter((ref: any): boolean => {
+					return ref.length > 0;
+				});
+
+			refs.forEach((ref: number): void => {
+				var image = model.find(ref);
+				image.ref = articleMedia.length;
+				collectionMedia.push(image);
+			});
+
+			articleMedia.push(collectionMedia);
+
+			var component = this.createChildView(App.InfoboxImageCollectionComponent, {
+				media: collectionMedia
+			}).createElement();
+
+			element.replaceWith(component.$());
+		}
+
+		model.set('media', articleMedia);
 	},
 
 	replaceMapsWithMapComponents(): void {
--- a/front/scripts/main/components/ArticleContentComponent.ts
+++ b/front/scripts/main/components/ArticleContentComponent.ts
@@ -179,9 +179,7 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 				refs = $element.data('refs')
 					.split(',')
 					.compact()
-					.filter((ref: any): boolean => {
-						return ref.length > 0;
-					});
+					.filter((ref: any): boolean => ref.length > 0);
 
 			refs.forEach((ref: number): void => {
 				var image = model.find(ref);
--- a/front/scripts/main/components/ArticleContentComponent.ts
+++ b/front/scripts/main/components/ArticleContentComponent.ts
@@ -170,21 +170,18 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 	replaceImageCollectionPlaceholdersWithComponents(model: typeof App.ArticleModel): void {
 		var $placeholders = this.$('.pi-image-collection'),
 			articleMedia = model.get('media'),
-			index: number;
-
-		// TODO: Allow us to pass in the number we want to process
-		var numberToProcess = $placeholders.length;
+			index: number,
+			numberToProcess: number = $placeholders.length;
 
 		for (index = 0; index < numberToProcess; index ++) {
-			var element = $placeholders.eq(index),
-				collectionMedia: typeof App.ArticleMedia[] = [];
-
-			var refs = element.data('refs')
-				.split(',')
-				.compact()
-				.filter((ref: any): boolean => {
-					return ref.length > 0;
-				});
+			var $element = $placeholders.eq(index),
+				collectionMedia: typeof App.ArticleMedia[] = [],
+				refs = $element.data('refs')
+					.split(',')
+					.compact()
+					.filter((ref: any): boolean => {
+						return ref.length > 0;
+					});
 
 			refs.forEach((ref: number): void => {
 				var image = model.find(ref);
@@ -198,7 +195,7 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 				media: collectionMedia
 			}).createElement();
 
-			element.replaceWith(component.$());
+			$element.replaceWith(component.$());
 		}
 
 		model.set('media', articleMedia);
--- a/front/scripts/main/components/InfoboxImageCollectionComponent.ts
+++ b/front/scripts/main/components/InfoboxImageCollectionComponent.ts
@@ -17,9 +17,7 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 		return this.get('activeRef') < (this.get('collectionLength') - 1);
 	}),
 
-	hasPreviousImage: Em.computed('activeRef', function(): boolean {
-		return this.get('activeRef') > 0;
-	}),
+	hasPreviousImage: Em.computed.gt('activeRef', 0),
 
 	computedHeight(media: typeof App.ArticleMedia): number {
 		var windowWidth: number = this.get('viewportDimensions.width'),
@@ -62,14 +60,14 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 		this.set('media', mediaArray);
 	},
 
-	loadImages(galleryRef: number = 0): void {
+	loadImages(): void {
 		var image: ArticleMedia,
 			cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
 			width: number = this.get('viewportDimensions.width'),
 			height: number,
 			collectionLength = this.get('collectionLength');
 
-		for (; galleryRef < collectionLength ; galleryRef ++) {
+		for (var galleryRef = 0; galleryRef < collectionLength ; galleryRef ++) {
 			image = this.get('media').get(galleryRef);
 			height = this.computedHeight(image);
 
@@ -93,16 +91,16 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 
 	actions: {
 		switchImage(direction: string) {
-			var activeRef = this.get('activeRef'),
+			var oldRef = this.get('activeRef'),
 				refDirection = (direction === 'next') ? 1 : -1,
-				nextRef =  activeRef + refDirection,
+				newRef =  oldRef + refDirection,
 				media = this.get('media'),
-				activeImage = media.get(activeRef),
-				nextImage = media.get(nextRef);
+				oldImage = media.get(oldRef),
+				newImage = media.get(newRef);
 
-			activeImage.set('isActive', false);
-			nextImage.set('isActive', true);
-			this.set('activeRef', nextRef);
+			oldImage.set('isActive', false);
+			newImage.set('isActive', true);
+			this.set('activeRef', newRef);
 		}
 	}
 });
--- a/front/scripts/main/components/InfoboxImageCollectionComponent.ts
+++ b/front/scripts/main/components/InfoboxImageCollectionComponent.ts
@@ -2,11 +2,46 @@
 /// <reference path="./MediaComponent.ts" />
 'use strict';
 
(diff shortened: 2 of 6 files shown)


front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
+/// <reference path="../app.ts" />
```
inez: All indentations in this file seems wrong.

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -166,6 +167,43 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
...
+	replaceImageCollectionPlaceholdersWithComponents(model: typeof App.ArticleModel): void {
+		var $placeholders = this.$('.pi-image-collection'),
+			articleMedia = model.get('media'),
+			index: number;
+
+		// TODO: Allow us to pass in the number we want to process
+		var numberToProcess = $placeholders.length;
+
+		for (index = 0; index < numberToProcess; index ++) {
+			var element = $placeholders.eq(index),
```
inez: Is "element" a jQuery selection? Then probably should be called "$element".

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -166,6 +167,43 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
...
+				collectionMedia: typeof App.ArticleMedia[] = [];
+
+			var refs = element.data('refs')
```
inez: Try to have one var statement per scope.

front/templates/main/components/infobox-image-collection.hbs
```diff
@@ -0,0 +1,14 @@
...
+    {{#if image.caption}}
+      <hgroup class="pi-hero-title-wrapper pi-item-spacing">
+        <h3 class="pi-hero-caption">{{image.caption}}</h3>
+      </hgroup>
+    {{/if}}
+    <img src="{{image.thumbUrl}}" class="{{if image.load 'loaded'}}" data-ref="{{image.ref}}" data-gallery-ref="{{image.galleryRef}}">
+  </div>
+{{/each}}
+<div class="image-collection-actions">
+  <button class="action-previous {{previousButtonClass}}" {{action "switchImage" "previous"}}>{{svg 'chevron-linked' viewBox='0 0 12 7'}}</button>
```
inez: What do you think about having methods "havePreviousImage" and "haveNextImage" in the component instead of "previousButtonClass" and "nextButtonClass" - and then using them here but in the if statement around the each button? I think that would be more generic.

front/templates/main/components/infobox-image-collection.hbs
```diff
@@ -0,0 +1,14 @@
...
+  <button class="action-next {{nextButtonClass}}" {{action "switchImage" "next"}}>{{svg 'chevron-linked' viewBox='0 0 12 7'}}</button>
```
inez: I think single quotes should be used around "switchImage" and "next". (and it's ridiculous that I used double quotes when typing it)

test/specs/front/scripts/main/components/InfoboxImageCollectionComponent.js
```diff
@@ -0,0 +1,69 @@
+var originalThumbnailerGetThumbURL = Mercury.Modules.Thumbnailer.getThumbURL;
```
inez: NO WAY! You've got tests!

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    //high image- image higher than square. Use top-crop-down mode.
+    if (windowWidth < computedHeight) {
+      return windowWidth;
+    }
+
+    return computedHeight;
+  },
+
```
dianafa: Some parts of this code are copy of `InfoboxImageMediaComponent` parts, maybe we could extend this component and just add additional functionalities?

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+    return this.get('activeRef') > 0 ? 'visible' : '';
+  }),
+
+  thumbWidth: Em.computed('viewportDimensions.width', function(): number {
+    return this.get('viewportDimensions.width');
+  }),
+
+  computedHeight(media: typeof App.ArticleMedia): number {
+    var windowWidth: number = this.get('viewportDimensions.width'),
+      imageAspectRatio: number = this.get('imageAspectRatio'),
```
dianafa: it should be a computed property in order to get updated each one of mentioned values get changed. That's why (like for example here: https://github.com/Wikia/mercury/blob/dev/front/scripts/main/components/ImageMediaComponent.ts#L39) we need to pass all variables we use `get` on in order to be able to recompute this property, when for example width of the screen will change.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.loadImages();
+  },
+
+  actions: {
+    switchImage(direction: string) {
+      var activeRef = this.get('activeRef'),
+        refDirection = (direction === 'next') ? 1 : -1,
+        nextRef =  activeRef + refDirection,
+        activeImage = this.get('media').get(activeRef),
+        nextImage = this.get('media').get(nextRef);
```
dianafa: `<nitpick>` here you can call `this.get('media.activeRef')`. Or you can "cache" `this.get('media')` in `media` variable and then just call it's properties (it's one var more) `</nitpick>`

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  loadImages(galleryRef: number = 0): void {
+    var image: ArticleMedia,
+      cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
+      width: number = this.get('thumbWidth'),
+      height: number,
+      collectionLength = this.get('collectionLength');
+
+    for (; galleryRef < collectionLength ; galleryRef ++) {
+      image = this.get('media').get(galleryRef);
```
dianafa: `this.get('media.galleryRef')`

please remember about the apostrophes in `get()` and `set()` :)

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+'use strict';
+
+App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixin, {
+	classNames: ['pi-image-collection'],
+	layoutName: 'components/infobox-image-collection',
+
+	limitHeight: true,
+	imageAspectRatio: 16 / 9,
+
+	activeRef: 0,
```
inez: Why some properties are separated with extra new line and other are not?

inez:
@pauloslund jenkins-Mercury-Front  is failing - is this related to your changes?

rogatty:
@inez yup:
```
front/scripts/main/components/InfoboxImageCollectionComponent.ts(18,10): 2322 Type 'boolean' is not assignable to type 'string'.
Build contains some typescript errors/warnings
+ echo BUILD:FAILURE
```

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+	collectionLength: Em.computed('media', function(): number {
+		return this.get('media').length;
+	}),
+
+	hasNextImage: Em.computed('activeRef', 'collectionLength', function(): string {
```
rogatty: You return boolean here.

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -166,6 +167,40 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
 		}
 	},
 
+	replaceImageCollectionPlaceholdersWithComponents(model: typeof App.ArticleModel): void {
```
rogatty: `typeof App.ArticleMedia`?

front/templates/main/components/infobox-image-collection.hbs
```diff
@@ -0,0 +1,14 @@
+{{#each media as |image|}}
+  <div class="pi-item {{if image.isActive 'active'}}">
```
rogatty: Please use tabs in hbs files.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+		return this.get('activeRef') < (this.get('collectionLength') - 1);
+	}),
+
+	hasPreviousImage: Em.computed('activeRef', function(): string {
```
rogatty: You return boolean here.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+		return this.get('activeRef') > 0;
+	}),
+
+	thumbWidth: Em.computed('viewportDimensions.width', function(): number {
```
bognix: how about http://emberjs.com/api/classes/Ember.computed.html#method_alias

bognix: you introduced this property but you are only reading it's value.
In one place you get: viewportDimensions.width - https://github.com/Wikia/mercury/pull/1541/files#diff-5ee805c5b14b7c7235968c8a8687bc16R30
and in other you get: thumbWidth - https://github.com/Wikia/mercury/pull/1541/files#diff-5ee805c5b14b7c7235968c8a8687bc16R30

Can you use only one? Maybe introducing thumbWidth is redundant?

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	computedHeight(media: typeof App.ArticleMedia): number {
+		var windowWidth: number = this.get('viewportDimensions.width'),
+			imageAspectRatio: number = this.get('imageAspectRatio'),
+			imageWidth: number = media.width || windowWidth,
+			imageHeight: number = media.height,
+			maxWidth: number = Math.floor(imageHeight * imageAspectRatio),
+			computedHeight: number = imageHeight;
+
+		if (windowWidth < imageWidth) {
```
bognix: could you combine this if statement with this one: https://github.com/Wikia/mercury/pull/1541/files#diff-5ee805c5b14b7c7235968c8a8687bc16R48.
You are checking the same thing in both

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+	},
+
+	setup(): void {
+		var mediaArray = Em.A(),
+			emptyGif = this.get('emptyGif');
+
+		this.get('media').forEach((image: ArticleMedia, index: number) => {
+			image.galleryRef = index;
+			image.thumbUrl = emptyGif;
+			image.isActive = (index === this.get('activeRef')) ? true : false;
```
bognix: how about simply:
```
image.isActive = (index === this.get('activeRef'))
```

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,114 @@
...
+
+			mediaArray.pushObject(Em.Object.create(image));
+		});
+
+		this.setProperties({
```
bognix: use this.set if you are setting only one property

rogatty:
It seems like a lot of this code is copied from other components. Is it possible to extract some common logic to a mixin?

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -166,6 +167,40 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
...
+		var $placeholders = this.$('.pi-image-collection'),
+			articleMedia = model.get('media'),
+			index: number,
+			numberToProcess: number = $placeholders.length;
+
+		for (index = 0; index < numberToProcess; index ++) {
```
hakubo: consider using `forEach`,

front/scripts/main/components/ArticleContentComponent.ts
```diff
@@ -166,6 +167,40 @@ App.ArticleContentComponent = Em.Component.extend(App.AdsMixin, App.PollDaddyMix
...
+			var $element = $placeholders.eq(index),
+				collectionMedia: typeof App.ArticleMedia[] = [],
+				refs = $element.data('refs')
+					.split(',')
+					.compact()
+					.filter((ref: any): boolean => {
```
hakubo: ```(ref: any): boolean => ref.length > 0```

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,108 @@
...
+
+	collectionLength: Em.computed('media', function(): number {
+		return this.get('media').length;
+	}),
+
+	hasNextImage: Em.computed('activeRef', 'collectionLength', function(): boolean {
+		return this.get('activeRef') < (this.get('collectionLength') - 1);
+	}),
+
+	hasPreviousImage: Em.computed('activeRef', function(): boolean {
```
hakubo: `Em.computed.gt('activeRef', 0)`

front/styles/main/module/article/_portable-infobox.scss
```diff
@@ -57,6 +59,45 @@ $hero-caption-line-height: 18px;
...
 
+// infobox gallery
+.pi-image-collection {
+	position: relative;
+	text-align: left;
+
+	.pi-item {
+		display: none;
+
+		&.active { display: block; }
```
hakubo: we try to have properties on new line
https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,108 @@
...
+			image.thumbUrl = emptyGif;
+			image.isActive = (index === this.get('activeRef'));
+
+			mediaArray.pushObject(Em.Object.create(image));
+		});
+
+		this.set('media', mediaArray);
+	},
+
+	loadImages(galleryRef: number = 0): void {
```
inez: I don't see anywhere call to this method that would be passing any parameters (galleryRef).

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,108 @@
...
+	},
+
+	load(): void {
+		this.setup();
+		this.loadImages();
+	},
+
+	actions: {
+		switchImage(direction: string) {
+			var activeRef = this.get('activeRef'),
```
inez: @pauloslund I see that you are using "nextRef" to store a reference to a next item to be displayed - which sometimes is a next item from the collection and sometimes is a previous item from the collection. Would calling those references oldRef and newRef make more sense and would help avoid confusion with "next" and "prev" as UI elements?

front/scripts/main/components/ArticleContentComponent.ts
pauloslund: `App.ArticleModel` is correct here, as we are getting the media from the article itself

front/scripts/main/components/ArticleContentComponent.ts
pauloslund: Initially written with the idea that we would be able to process less than all of the placeholders which is why the for loop was used. Since that is no longer the case this may eventually get refactored to a `forEach` statement

front/scripts/main/components/ArticleContentComponent.ts
rogatty: Not really.
`ArticleModel` structure is:
```
{
  content: string;
  media: typeof App.MediaModel;
  ...
}
```
`MediaModel` is created here https://github.com/Wikia/mercury/blob/dev/front/scripts/main/models/ArticleModel.ts#L188-L190 and is an array of `ArticleMedia` objects. So when you do `var image = model.find(ref);` you're calling `MediaModel#find`, not `ArticleModel#find` which would give you an article.

Yeah, I see `replaceMediaPlaceholdersWithMediaComponents` uses incorrect type too :see_no_evil:
Unfortunately article media is much more complicated than it should be. We'll try to clean it up.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	loadImages(): void {
+		var image: ArticleMedia,
+			cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
+			width: number = this.get('viewportDimensions.width'),
+			height: number,
+			collectionLength = this.get('collectionLength');
+
+		for (var galleryRef = 0; galleryRef < collectionLength ; galleryRef ++) {
+			image = this.get('media').get(galleryRef);
```
dianafa: what about performing `media = this.get('media')` only once, at the begining of the function and then call in loop only `media.get('galleryRef')`?
Also, we have a convention to use one var per function if this is possible.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -0,0 +1,106 @@
...
+
+		this.set('media', mediaArray);
+	},
+
+	loadImages(): void {
+		var image: ArticleMedia,
+			cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
+			width: number = this.get('viewportDimensions.width'),
+			height: number,
+			collectionLength = this.get('collectionLength');
```
dianafa: please try to add types where it's possible; here `string` for `cropMode` and `number` for `collectionLength`

PR #1542, opened 2015-10-20
GA tracking for google custom search

perjg:
We are doing an A/B test for Google Custom Search on Japanese wikias.

We want to track this using the Site Search panel in GA, and this requires that we send an extra pageview impression to GA, which I add in the PR.

Documentation here: 
https://support.google.com/analytics/answer/1012264?hl=en

### DIFF
--- a/front/scripts/main/routes/SearchResultsRoute.ts
+++ b/front/scripts/main/routes/SearchResultsRoute.ts
@@ -9,6 +9,9 @@ App.SearchResultsRoute = Em.Route.extend({
 		this._super(controller, model, transition);
 		// Only search in current community
 		controller.set('site', window.location.hostname);
+
+		// Send extra tracking info to GA to track search usage
+		M.trackGoogleSearch(window.location.href + 'search?q=' + controller.q);
 	},
 
 	/*
--- a/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
+++ b/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
@@ -212,7 +212,9 @@ module Mercury.Modules.Trackers {
 
 		/**
 		 * Tracks usage of Google Custom Search
-		 * @param queryParam
+		 * 
+		 * @param {string} queryParam
+		 * @returns {undefined}
 		 */
 		trackGoogleSearch (queryParam: string): void {
 			this.tracked.forEach((account:GAAccount) => {
--- a/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
+++ b/front/scripts/mercury/modules/Trackers/UniversalAnalytics.ts
@@ -209,5 +209,16 @@ module Mercury.Modules.Trackers {
 				ga(`${prefix}send`, 'pageview');
 			});
 		}
+
+		/**
+		 * Tracks usage of Google Custom Search
+		 * @param queryParam
+		 */
+		trackGoogleSearch (queryParam: string): void {
+			this.tracked.forEach((account:GAAccount) => {
+				var prefix = this.getPrefix(account);
+				ga(`${prefix}send`, 'pageview', queryParam);
+			});
+		}
 	}
 }
--- a/front/scripts/mercury/utils/track.ts
+++ b/front/scripts/mercury/utils/track.ts
@@ -34,6 +34,7 @@ interface TrackerInstance {
 	new(): TrackerInstance;
 	track: TrackFunction;
 	trackPageView: (context?: TrackContext) => void;
+	trackGoogleSearch: (queryParam: string) => void;
 	updateTrackedUrl: (url: string) => void;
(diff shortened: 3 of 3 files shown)


perjg:
@hakubo @rogatty Does anyone have time to have a quick look at this? We need this to get data to GA for our Google Custom Search A/B test.

front/scripts/main/routes/SearchResultsRoute.ts
```diff
@@ -9,6 +9,9 @@ App.SearchResultsRoute = Em.Route.extend({
 		this._super(controller, model, transition);
 		// Only search in current community
 		controller.set('site', window.location.hostname);
+
+		// Send extra tracking info to GA to track search usage
+		M.trackGoogleSearch(window.location.href + 'search?q=' + controller.q);
 	},
```
vforge: Consider using backticks and `${window.location.href}search?q=${controller.q}`

vforge:
@perjg I've fixed up JSDocs and left you small suggestion

front/scripts/main/routes/SearchResultsRoute.ts
```diff
@@ -9,6 +9,9 @@ App.SearchResultsRoute = Em.Route.extend({
 		this._super(controller, model, transition);
 		// Only search in current community
 		controller.set('site', window.location.hostname);
+
+		// Send extra tracking info to GA to track search usage
+		M.trackGoogleSearch(window.location.href + 'search?q=' + controller.q);
```
rogatty: I think this belongs in `activate` hook, see http://emberjs.com/api/classes/Ember.Route.html#method_activate

vforge:
Also did you thought about adding custom dimension with google query param instead of sending another PV to GA?
First PV sent to GA disables anything further - at least till `page` param is changed, that's why we have to use `updateTrackedUrl` in our SPA page view code.

And on the other side passing third param for GA's `send` method is viable only in multiple-page sites and this approach is discouraged by Google.

rogatty:
Did you analyze https://github.com/Wikia/mercury/blob/dev/front/scripts/mercury/utils/variantTesting.ts#L75? Isn't it enough to send custom dimension with google tracking? Sending additional page view seems like something that isn't consistent with the rest of app.

front/scripts/main/routes/SearchResultsRoute.ts
perjg: @rogatty How do I access the controller from inside the activate hook? I need it to extract the q parameter.

perjg:
Thanks for the comments. I looked into using updateTrackedUrl instead of sending a new page view.

In order for the search to show up in the GA Site Search panel, the data sent to GA must include
the search parameter, in this case q.

Example:
  http://.../search?q=hello 

However, if I do this:
  M.updateTrackedUrl(`${window.location.href}search?q=${controller.q}`);

Then inside UniversalAnalytics.ts -> updateTrackedUrl, the parameter is not in the string and only
the base url is included. http://.../search

perjg:
I'm not sure if I understand how to use custom dimensions for my use case.
As far as I can see, custom dimensions are set up on application load (in App.initializer).

I need to send both the parameter name (dimension?) and the actual search term when loading the search results page.

front/scripts/main/routes/SearchResultsRoute.ts
rogatty: You can use `this.paramsFor('searchResults').q`, see http://emberjs.com/api/classes/Ember.Route.html#method_paramsFor

rogatty:
`M.updateTrackedUrl` strips query params when it uses `location.pathname`. One of the solutions could be passing additional flag to this method which would decide if we want to strip them or not. I personally don't know why the current solution works like it does.

It would be the best if you consult it with a person that knows exactly what should we send to UA and when. /cc @harnash 

And yeah, Optimizely's custom dimensions aren't helpful for your case as it doesn't really matter that this is AB test.

front/scripts/main/routes/SearchResultsRoute.ts
hakubo: +1

hakubo:
42 lines of code that normally should take about 1
`ga('send', 'pageview', '/search_results.php?q=keyword');`

clearly we did something wrong with our tracking code set up

for now it LGTM
we do plan on refactoring this part of code as it becomes hard to manage

perjg:
@hakubo Got it, will merge now.
After the A/B test is over, I will reevaluate this part of the code to see if we need to keep it or not. I'll revert if needed.

harnash:
@hakubo + other: Sorry for taking so long to response. I think current approach by setting query/search parameter is best since it is consistent with MW side and it allows us to populate Behaviour/Site search views. With custom dimensions GA will not be aware that user is searching and we would have two places to look for that data (Behaviour view and custom dimension).

PR #1544, opened 2015-10-20
ADEN-2569 Fix krux implementation

fraszczakszymon:
Those changes should prevent occurring error in console `Krux is not a function` - issue connected with SP detection on mobile.

Details:
Previously `window.Krux` object was used only to load krux module from `Mercury.Modules.Trackers.Krux` (kind of temporary variable?). But `window.Kurx` is a queue-function ([in app code](https://github.com/Wikia/app/blob/dev/resources/wikia/modules/krux.js#L2)) and previous implementation causes an issue when response from krux is relatively slow.


jacekjursza:
LGTM ( @dianafa -what do you think? )

dianafa:
LGTM (if it works :D)

PR #1547, opened 2015-10-22
INT-231: Remove some links from Japan HP footer in mobile view

perjg:
Some of the links in Japan HP footer do not work on mobile, and should only show in desktop mode.


hakubo:
lgtm

PR #1549, opened 2015-10-23
Track the first didRender instead of didInsertElement

rogatty:
Track the first `didRender` instead of `didInsertElement` to get more reliable results.
Change `firstContent` to `appRendered` because the former is misleading.


hakubo:
LGTM

hakubo:
https://github.com/Wikia/mercury/pull/1549

PR #1550, opened 2015-10-23
VE-1998: Make sure article videos are clickable

pauloslund:
Restructure media in collection so the container is the clickable element


inez:
@pauloslund Could you please add information why this change actually solved the problem?

inez:
LGTM

PR #1552, opened 2015-10-23
VE-1999: Change image crop mode on portrait images

pauloslund:
This PR also introduces code that fixes VE-2001 due to setting visible=true

### DIFF
--- a/front/scripts/main/components/InfoboxImageCollectionComponent.ts
+++ b/front/scripts/main/components/InfoboxImageCollectionComponent.ts
@@ -4,6 +4,7 @@
 
 App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixin, {
 	classNames: ['pi-image-collection'],
+	classNameBindings: ['visible'],
 	layoutName: 'components/infobox-image-collection',
 	limitHeight: true,
 	imageAspectRatio: 16 / 9,
@@ -71,6 +72,10 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 			image = this.get('media').get(galleryRef);
 			height = this.computedHeight(image);
 
+			if (image.height > image.width) {
+				cropMode = Mercury.Modules.Thumbnailer.mode.topCropDown;
+			}
+
 			var thumbUrl = this.getThumbURL(image.url, {
 				mode: cropMode,
 				width: width,
@@ -87,6 +92,7 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 	load(): void {
 		this.setup();
 		this.loadImages();
+		this.set('visible', true);
 	},
 
 	actions: {
--- a/front/scripts/main/components/InfoboxImageCollectionComponent.ts
+++ b/front/scripts/main/components/InfoboxImageCollectionComponent.ts
@@ -63,12 +63,12 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 
 	loadImages(): void {
 		var image: ArticleMedia,
-			cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
 			width: number = this.get('viewportDimensions.width'),
 			height: number,
 			collectionLength = this.get('collectionLength');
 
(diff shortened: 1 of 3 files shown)


front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -71,6 +72,10 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 			image = this.get('media').get(galleryRef);
 			height = this.computedHeight(image);
 
+			if (image.height > image.width) {
```
inez: @pauloslund Is that really what you want? My understanding is that by default you want to use zoomCrop, and use topCropDown only if height>width - but you define default outside of the loop - then change it inside the loop (on the first image where height>width) and that new default gonna apply to all images after that one.

inez: Actually it's also not clear for me what it iterates over - over collections in the article or over images in the collection. I think "galleryRef" variable name is bit confusing.

pauloslund: Good catch on the scoping issue. This is iterating over images in the collections.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -62,15 +63,19 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
...
 		var image: ArticleMedia,
-			cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
 			width: number = this.get('viewportDimensions.width'),
 			height: number,
 			collectionLength = this.get('collectionLength');
 
 		for (var galleryRef = 0; galleryRef < collectionLength ; galleryRef ++) {
+			var cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop;
 			image = this.get('media').get(galleryRef);
 			height = this.computedHeight(image);
```
inez: ```cropMode = image.height > image.width ? Mercury.Modules.Thumbnailer.mode.topCropDown : Mercury.Modules.Thumbnailer.mode.zoomCrop```

front/scripts/main/components/InfoboxImageCollectionComponent.ts
inez: There was a comment from @dianafa about moving call to this.get('media') to outer scope, making sure you know about it

inez: Can you think of some var name other than galleryRef that would make more sense in this scope?

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -62,15 +63,19 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
 
 	loadImages(): void {
 		var image: ArticleMedia,
-			cropMode = Mercury.Modules.Thumbnailer.mode.zoomCrop,
 			width: number = this.get('viewportDimensions.width'),
 			height: number,
 			collectionLength = this.get('collectionLength');
 
 		for (var galleryRef = 0; galleryRef < collectionLength ; galleryRef ++) {
```
dianafa: could we use here `this.get('media').forEach((image, index) => {})`?
We wouldn't need ```collectionLength = this.get('collectionLength');``` then.

Also, as content of this loop is growing, maybe it would be nice to consider moving this stuff to a separate function.

front/scripts/main/components/InfoboxImageCollectionComponent.ts
```diff
@@ -61,32 +62,31 @@ App.InfoboxImageCollectionComponent = App.MediaComponent.extend(App.ViewportMixi
...
+		this.get('media').forEach((image: ArticleMedia, index: number) => {
+			var cropMode = image.height > image.width ? Mercury.Modules.Thumbnailer.mode.topCropDown : Mercury.Modules.Thumbnailer.mode.zoomCrop,
+				height: number = this.computedHeight(image),
+				thumbUrl: string = this.getThumbURL(image.url, {
+					mode: cropMode,
+					width: width,
+					height: height
+				});
+
+				console.log(thumbUrl);
```
inez: @pauloslund ^^^

PR #1563, opened 2015-10-27
i18n | New translations for auth and auth-front

bkoval:
Fresh translations for the auth project.

@Warkot @armonr Please take a look :)

### DIFF
--- a/front/locales/zh/auth-front.json
+++ b/front/locales/zh/auth-front.json
@@ -11,7 +11,7 @@
     "password_exceeds_max_length":"密碼不能超過50個字元",
     "password_equal_name":"用戶名和密碼不匹配",
     "user_already_linked":"這個Wikia帳戶已經連接到Facebook",
-    "fb_account_already_linked":"這個Facebook帳戶已經連接到另一個Wikia用戶帳戶",
+    "fb_account_already_linked":"這個Facebook帳戶已經連接到另一個Wikia帳戶",
     "birthdate_invalid":"出生日期無效",
     "wrong-credentials":"我們無法辨識這些登入資料。請再試一次或註冊一個新帳戶。",
     "server-error":"處理你的請求時出錯。請稍後再試。"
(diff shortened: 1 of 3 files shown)


Warkot:
I'm not an expert on Chinese, but I trust it, as it was downloaded from Crowdin. Therefore it LGTM. :+1:

front/locales/zh/auth-front.json
```diff
@@ -11,7 +11,7 @@
     "password_exceeds_max_length":"密碼不能超過50個字元",
     "password_equal_name":"用戶名和密碼不匹配",
     "user_already_linked":"這個Wikia帳戶已經連接到Facebook",
-    "fb_account_already_linked":"這個Facebook帳戶已經連接到另一個Wikia用戶帳戶",
+    "fb_account_already_linked":"這個Facebook帳戶已經連接到另一個Wikia帳戶",
```
armonr: Just for fun, in crowdin looks like this. `这个Facebook帐户已经连接到另一个Wikia用户帐户`. But if that's what gets downloaded, that is what it is and LGTM too :)

Warkot:
And now for something completely different…
Why most of the keys in `auth-front.json` are written with **snake_case** instead of **kebab-case**?

front/locales/zh/auth-front.json
Warkot: Indeed, in Crowdin Chinese simplified is:
```
这个Facebook帐户已经连接到另一个Wikia用户帐户
```
However, Chinese traditional is:
```
這個Facebook帳戶已經連接到另一個Wikia帳戶
```

armonr: Interesting. It makes me want to simply learn traditional Chinese

bkoval:
Talked to @warkot offline about i18n keys, basically they're compliant with errors returned by Helios and snake-casing is making them easier to map.

PR #1564, opened 2015-10-28
INT-89: Check user login state before allowing edits

perjg:
Checks the user's login state before allowing edits.

Logged in users are always allowed.
Anonymous edits are allowed unless disableAnonymousEditing is set to true
Any wikia marked as a COPPA wikia always requires login to edit regardless of settings.

Depends on: https://github.com/Wikia/app/pull/8896


front/scripts/main/components/ArticleWrapperComponent.ts
```diff
@@ -86,6 +86,27 @@ App.ArticleWrapperComponent = Em.Component.extend(
...
+		 * @returns True if edit is allowed
+		 */
+		isEditAllowed: Em.computed(function(): boolean {
+			var isCoppaWiki = Em.get(Mercury, 'wiki.isCoppaWiki'),
+				disableAnonymousEditing = Em.get(Mercury, 'wiki.disableAnonymousEditing'),
+				isLoggedIn = Em.get(Mercury, 'currentUser.isAuthenticated');
+
+			if (isLoggedIn) {
+				return true;
+			}
```
hakubo: I think we use: 
} else {

front/scripts/main/components/ArticleWrapperComponent.ts
```diff
@@ -86,6 +86,26 @@ App.ArticleWrapperComponent = Em.Component.extend(
 			return !Em.get(Mercury, 'wiki.disableAnonymousUploadForMercury');
 		}),
 
+		/**
+		 * For section editor, checks if the user is allowed to edit
+		 * - Logged in users are always allowed to edit
+		 * - Wikias with disableAnonymousEditing set need login to edit
+		 * - Coppa wikias (for wikias directed at children) always require login
+		 *
+		 * @returns True if edit is allowed
```
rogatty: `@returns {boolean} True if edit is allowed`

front/templates/main/components/article-contribution.hbs
```diff
@@ -1,7 +1,13 @@
 <div class="edit-section">
-	<span {{action 'edit'}}>
-		{{svg 'pencil' role='img' class='icon pencil'}}
-	</span>
+	{{#if isEditAllowed}}
+		<span {{action 'edit'}}>
+			{{svg 'pencil' role='img' class='icon pencil'}}
+		</span>
+	{{else}}
+		<span {{action 'select'}}>
```
rogatty: `select` action tracks `label: 'add-photo-no-auth'`. I don't think we want to do that when user clicks on the pencil. Also `select` should be renamed to something else as it contains no code related to selecting anything.

perjg:
INT-246 builds directly on the changes made here.
I've squashed all the commits here and moved them to a new PR that also includes an extra commit with changes for INT-246:

https://github.com/Wikia/mercury/pull/1605

PR #1565, opened 2015-10-28
INT-245: Workaround for broken z-index in Safari

perjg:
This is a workaround for what appears to be a bug in Safari related to z-order.

See screenshot here:
https://wikia-inc.atlassian.net/browse/INT-245

### DIFF
--- a/homepage/front/styles/app.scss
+++ b/homepage/front/styles/app.scss
@@ -43,6 +43,7 @@ section {
 	margin-bottom: 3em;
 	padding-left: 3em;
 	padding-right: 3em;
+	-webkit-transform: translate3d(0,0,0); // workaround for broken z-index in safari
 
 	@include media($mobile) {
 		padding-left: .1em;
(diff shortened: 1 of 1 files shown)


homepage/front/styles/app.scss
```diff
@@ -43,6 +43,7 @@ section {
 	margin-bottom: 3em;
 	padding-left: 3em;
 	padding-right: 3em;
+	-webkit-transform: translate3d(0,0,0); // workaround for broken z-index in safari
```
bkoval: If you use mercury's build step, there's an autoprefixer for CSS, so if you declare it as transform: translate3d(0,0,0); and autoprefixer will copy it into prefixed properties automatically.

Also, I believe the commenting rule is to move comments to the line above the code line they describe.

Besides that, LGTM

perjg:
We are currently running an A/B test for Google Search on Oasis and Mercury only. This PR will not be merged until at least after the A/B test completes, at which point we will decide what to do about Japan HP.

bkoval:
LGTM

hakubo:
lgtm

PR #1567, opened 2015-10-28
Share Feature | Trigger this.attrs handlers only if they exist

bkoval:
Since the functionality was not broken, it was just throwing after opening a social network share url, this should be enough to prevent it.

@hakubo @rogatty @Warkot

### DIFF
--- a/front/scripts/main/components/ShareFeatureComponent.ts
+++ b/front/scripts/main/components/ShareFeatureComponent.ts
@@ -42,14 +42,18 @@ App.ShareFeatureComponent = Em.Component.extend(
 		 * @returns {undefined}
 		 */
 		mouseEnter(): void {
-			this.attrs.onMouseEnter();
+			if (this.attrs && typeof this.attrs.onMouseEnter === 'function') {
+				this.attrs.onMouseEnter();
+			}
 		},
 
 		/**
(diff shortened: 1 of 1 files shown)


front/scripts/main/components/ShareFeatureComponent.ts
```diff
@@ -42,14 +42,18 @@ App.ShareFeatureComponent = Em.Component.extend(
 		 * @returns {undefined}
 		 */
 		mouseEnter(): void {
-			this.attrs.onMouseEnter();
+			if (this.attrs && typeof this.attrs.onMouseEnter === 'function') {
```
rogatty: I think that we don't have to check for `this.attrs` as it's Ember's internal property and is always set to at least `null`.

Warkot: If it was `null`, it would break. →
```
object = {attrs: null};
console.log(object.attrs.stuff);
VM160:2 Uncaught TypeError: Cannot read property 'stuff' of null(…)
```

bkoval: @Warkot is right, null is not an object descendant, so this code would break.

Warkot:
LGTM

PR #1578, opened 2015-11-02
XW-584 Fix article meta tags

rogatty:
https://wikia-inc.atlassian.net/browse/XW-584
https://wikia-inc.atlassian.net/browse/XW-619

* updates `title`, `canonical` and `description` meta tags on `ArticleRoute#didTransition`
* uses `url` property from article model to build the canonical url instead of replacing chars in article title


front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -124,5 +126,56 @@ App.ArticleRoute = Em.Route.extend({
...
 			return true;
 		}
+	},
+
+	/**
+	 * @TODO this can be much simpler using ember-cli-meta-tags
+	 *
+	 * @returns {void}
+	 */
+	updateHead(): void {
```
hakubo: what do you think about some simple unit tests for all 3 below?

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -124,5 +126,56 @@ App.ArticleRoute = Em.Route.extend({
...
+		this.updateCanonicalLinkTag();
+		this.updateDescriptionMetaTag();
+	},
+
+	/**
+	 * @returns {void}
+	 */
+	updateTitleTag(): void {
+		var model: typeof App.ArticleModel = this.modelFor('article'),
+			htmlTitleTemplate = Em.getWithDefault(Mercury, 'wiki.htmlTitleTemplate', '$1 - Wikia');
```
bognix: are you sure about getWithDefault here? I'm not sure what sort of API response comes from MW but getWithDefault would only work in case htmlTitleTemplate is not set. It won't work when value is null.
Maybe simple || is safer?

bognix:
how about some tests for this? I know this is going to be replaced by ember addon some day but some day can happen in a year from now.

Warkot:
LGTM, but, as mentioned, adding few tests would be nice. :+1:

rogatty:
@hakubo @bognix @Warkot I've added unit tests.

front/scripts/main/routes/ArticleRoute.ts
```diff
@@ -124,5 +126,59 @@ App.ArticleRoute = Em.Route.extend({
...
+		this.updateCanonicalLinkTag(model);
+		this.updateDescriptionMetaTag(model);
+	},
+
+	/**
+	 * @param {App.ArticleModel} model
+	 * @returns {void}
+	 */
+	updateTitleTag(model: typeof App.ArticleModel): void {
+		var	defaultHtmlTitleTemplate = '$1 - Wikia',
```
hakubo: tab?

hakubo:
lgtm

PR #1582, opened 2015-11-02
SOC-1552 | Can't upvote post that are lazy-loader

rafalkalinski:
After scrolling down, older posts are loaded. When trying to upvote loaded posts, nothing happen.
https://wikia-inc.atlassian.net/browse/SOC-1552

Fix is simple: I've added credentials to the loadPage request - after that server sends us all needed data for the upvoting.


kvas-damian:
LGTM

PR #1583, opened 2015-11-02
XW-651 Fix formatting in ES6 files

rogatty:
https://wikia-inc.atlassian.net/browse/XW-651

Auto formatting was broken in WebStorm 10, it's fixed in WebStorm 11.

/cc @vforge


hakubo:
LGTM

PR #1589, opened 2015-11-03
Enabling discussion under flag

bkoval:
It preserves the splash pages as they were working before, but if 'enableDiscussion' flag is set to true, this code will render 404 on communities that don't support discussions.

@kvas-damian @rafalkalinski


rafalkalinski:
LGTM

kvas-damian:
LGTM

PR #1591, opened 2015-11-03
XW-514 ES6 on front end

rogatty:
https://wikia-inc.atlassian.net/browse/XW-514


rogatty:
Moved changes to separate PRs which target `es6-front` feature branch.

PR #1594, opened 2015-11-04
Discussions | Enabling discussions on featured Wikias

bkoval:
@kvas-damian @rafalkalinski Let's do it!

BTW. there's a ticket for removing splash pages whatsoever (https://wikia-inc.atlassian.net/browse/SOC-1575) - we're just leaving them here in case of any possible quick reverts.


kvas-damian:
LGTM

PR #1597, opened 2015-11-04
XW-514 ESLint fixes

rogatty:


hakubo:
LGTM

.eslintrc
```diff
@@ -45,7 +45,7 @@
     "no-sparse-arrays": 2,
     "no-unreachable": 2,
     "use-isnan": 2,
-    "valid-jsdoc": [2, {"requireParamDescription": false}],
+    "valid-jsdoc": [2, {"requireParamDescription": false, "requireReturnDescription": false}],
```
rogatty: It doesn't give us any value if we have return description:
```
/**
 * @returns {boolean} is valid
 */
function isValid() {...}
```
but it surely is a PITA for devs.

.eslintrc
```diff
@@ -98,7 +98,7 @@
     "no-useless-call": 2,
     "no-useless-concat": 2,
     "no-void": 2,
-    "no-warning-comments": 1,
+    "no-warning-comments": 0,
```
rogatty: It displayed warnings for all TODO comments which we have a lot. Let's not pollute warnings because people will ignore them.

.eslintrc
```diff
@@ -112,7 +112,7 @@
     "no-delete-var": 2,
     "no-label-var": 2,
     "no-shadow-restricted-names": 2,
-    "no-shadow": [2, {"builtinGlobals": true, "hoist": "functions"}],
+    "no-shadow": [0, {"builtinGlobals": true, "hoist": "functions"}],
```
rogatty: Because of this rule ESLint displayed errors for function `setQueryParam(name, value)` because `name` and `value` are already declared somewhere else in the outer scope. We may discuss how much value is in it for us later but right now it would force us to complicate a lot of variable names.

.eslintrc
```diff
@@ -147,7 +147,7 @@
     "jsx-quotes": 0,
     "key-spacing": [2, {"beforeColon": false, "afterColon": true}],
     "lines-around-comment": [0, { "beforeBlockComment": true}],
-    "linebreak-style": [2, "unix"],
+    "linebreak-style": [0, "unix"],
```
rogatty: It caused errors for all files as I'm using Windows. Git is converting linebreaks to Unix style anyway on every push.

.eslintrc
```diff
@@ -159,7 +159,7 @@
     "no-mixed-spaces-and-tabs": 2,
     "no-multiple-empty-lines": [2, {"max": 2}],
     "no-nested-ternary": 1,
-    "no-negated-condition": 2,
+    "no-negated-condition": 0,
```
rogatty: It makes sense to force it as some point but right now we would have to refactor multiple condition blocks and this can lead to unexpected bugs.

.eslintrc
```diff
@@ -181,7 +181,7 @@
     "space-after-keywords": [2, "always"],
     "space-before-keywords": [2, "always"],
     "space-before-blocks": 2,
-    "space-before-function-paren": [2, { "anonymous": "never", "named": "never" }],
+    "space-before-function-paren": [2, { "anonymous": "always", "named": "never" }],
```
rogatty: https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#functions

PR #1609, opened 2015-11-06
XW-514 Convert models to ES6

rogatty:
https://wikia-inc.atlassian.net/browse/XW-514

Git displays diff as I've replaced files and not changed them. To see the proper diff look at https://github.com/Wikia/mercury/commit/597d1f5fb4f5c4fd02744020b9600a8284a97ddb


front/scripts/main/models/ArticleEditModel.js
```diff
@@ -69,21 +65,23 @@ App.ArticleEditModel.reopenClass(App.ArticleEditMixin, {
 					rvsection: sectionIndex,
 					format: 'json'
 				}
-			})
-			.done((resp): void => {
-				var pages: any,
-					revision: any;
+			}).done((resp) => {
```
bognix: can it be replaced with `then()` so it's consistent with rest of file

front/scripts/main/models/ArticleModel.js
```diff
@@ -144,21 +116,23 @@ App.ArticleModel.reopenClass({
 			article.data.article.content = $.trim($('#preloadedContent').html());
 		}
 
-		delete Mercury['article'];
+		Mercury.article = null;
```
bognix: http://www.smashingmagazine.com/2012/11/writing-fast-memory-efficient-javascript/#de-referencing-misconceptions null is better than delete but if nothing reads it gc would clean it up either way

front/scripts/main/models/WikiaInYourLangModel.js
```diff
@@ -0,0 +1,78 @@
...
+
+			Em.$.getJSON(
+				M.buildUrl({path: '/wikia.php'}),
+				{
+					controller: 'WikiaInYourLangController',
+					method: 'getNativeWikiaInfo',
+					format: 'json',
+					targetLanguage: browserLang
+				}
+			).done((resp) => {
```
bognix: again `done` and `fail` which could be `then`

front/scripts/main/models/ArticleEditModel.js
rogatty: It would change the flow so I wouldn't do it right now.

PR #1617, opened 2015-11-09
XW-514 | convert mixins to modules

tomasznapieralski:


front/scripts/main/mixins/ads.js
```diff
@@ -1,4 +1,6 @@
-App.AdsMixin = Em.Mixin.create({
+import Ember from 'ember';
+
+const AdsMixin = Ember.Mixin.create({
```
hakubo: should we export default here?

front/scripts/main/mixins/article-content.js
```diff
@@ -0,0 +1,37 @@
+import Ember from 'ember';
+import ArticleContentListeners from '/app/-article-content-mixin';
```
hakubo: `'/app/-article-content-mixin'` seems wrong

front/scripts/main/mixins/visible.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+import VisibilityStateManager from '/app/-visibility-state-manager';
```
hakubo: `'/app/-visibility-state-manager'`

PR #1624, opened 2015-11-09
XW-514 Convert Mercury modules to ES6

rogatty:


front/scripts/mercury/modules/Thumbnailer.js
```diff
@@ -0,0 +1 @@
+export const Thumbnailer = window.Vignette;
```
hakubo: lol :)

hakubo: maybe we should just use the global? :)

front/scripts/mercury/modules/Trackers/BaseTracker.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Base class for trackers that have to append their scripts like Comscore or Quantserve
+ */
+export class BaseTracker {
```
hakubo: should that have default?

front/scripts/mercury/modules/Trackers/Comscore.js
```diff
@@ -0,0 +1,40 @@
+import {BaseTracker} from 'BaseTracker';
```
hakubo: I think if BaseTracker had a default you could do 
`import BaseTracker from 'BaseTracker';`

right?

front/scripts/mercury/modules/Trackers/Quantserve.js
```diff
@@ -0,0 +1,43 @@
+import {BaseTracker} from 'BaseTracker';
+import {prop} from '../../../baseline/mercury/utils/state';
```
hakubo: should prop be default in state?
I can see it is used in most places

front/scripts/mercury/modules/Thumbnailer.js
rogatty: It's exactly the same situation as with https://github.com/Wikia/mercury/blob/dev/front/scripts/mercury/utils/load.ts
They were introduced to not couple with external libraries but IMO we should get rid of both of them. Especially that Vignette comes with AMD / CJS implementation.

front/scripts/mercury/utils/track.js
```diff
@@ -26,278 +11,246 @@ interface TrackContext {
...
+
+/**
+ * @param {TrackContext} data
+ * @returns {void}
+ */
+export function setTrackContext(data) {
+	this.context = data;
+}
+
+export {actions as trackActions};
```
hakubo: why do we change name here?

front/scripts/mercury/utils/track.js
```diff
@@ -1,24 +1,9 @@
...
- * @property {*} Mercury
- */
-
-interface TrackContext {
-	//article title
-	a: string;
-	//namespace
-	n: number;
-}
+import {Comscore} from '../modules/Trackers/Comscore';
```
hakubo: consider adding default exports

front/scripts/mercury/utils/track.js
rogatty: Someone had this great idea long time ago and now it's used in 69 places. Do you think we should fix it now?

front/scripts/mercury/modules/Thumbnailer.js
rogatty: Anyway, let's do this but not right now. It's big enough already.

front/scripts/mercury/modules/Trackers/Quantserve.js
rogatty: Not sure which way is better. We want to rewrite this `M.prop` anyway so I would leave it as it is for now.

front/scripts/main/router.js
```diff
@@ -1,3 +1,5 @@
+import Ember from 'Ember';
```
tomasznapieralski: ```import Ember from 'ember';```

front/scripts/main/app.js
```diff
@@ -1,8 +1,11 @@
+import Ember from 'Ember';
```
tomasznapieralski: ```import Ember from 'ember';```

front/scripts/main/CurrentUser.js
```diff
@@ -1,3 +1,4 @@
+import Ember from 'Ember';
```
tomasznapieralski: ```import Ember from 'ember';```

front/scripts/mercury/modules/Ads.js
```diff
@@ -0,0 +1,294 @@
+import Krux from 'Trackers/Krux';
+import UniversalAnalytics from 'Trackers/UniversalAnalytics';
+import load from '../utils/load';
+
```
tomasznapieralski: i'm not sure but ember-migrator is adding ```import Ember from 'ember';``` even if `Ember` is not used, so maybe we should putt it always?

front/scripts/mercury/utils/track.js
```diff
@@ -26,278 +11,245 @@ interface TrackContext {
...
-			return !!(M.prop('isGASpecialWiki') || Mercury.wiki.isGASpecialWiki);
-		} catch (e) {
-			// Property doesn't exist
-			return false;
-		}
+/**
+ * @returns {boolean}
+ */
+function isSpecialWiki() {
+	try {
```
hakubo: why do we have a `try/catch` block here?

front/scripts/mercury/utils/track.js
```diff
@@ -26,278 +11,245 @@ interface TrackContext {
...
-	 * @returns {void}
-	 */
-	export function trackGoogleSearch (queryParam: string) {
-		var trackers: any = Mercury.Modules.Trackers;
+		if (typeof Tracker.prototype.trackPageView === 'function') {
+			const instance = new Tracker(isSpecialWiki());
 
-		if (M.prop('queryParams.noexternals')) {
-			return;
+			console.info('Track pageView:', tracker);
```
hakubo: console?

front/scripts/mercury/utils/track.js
```diff
@@ -26,278 +11,245 @@ interface TrackContext {
...
-		if (M.prop('queryParams.noexternals')) {
-			return;
-		}
+		if (typeof Tracker.prototype.trackGoogleSearch === 'function') {
+			const instance = new Tracker(isSpecialWiki());
 
-		Object.keys(trackers).forEach((tracker: string): void => {
-			var Tracker = trackers[tracker],
-				instance: TrackerInstance;
+			console.info('Track Google Search:', tracker);
```
hakubo: console! :)

front/scripts/mercury/modules/Ads.js
hakubo: no, you should only import things that are used

rogatty: Also this is a common Mercury module, nothing to do with Ember.

front/scripts/mercury/utils/track.js
rogatty: Probably so it doesn't blow up when there is no `Mercury.wiki` global. I agree that it can be refactored.

front/scripts/mercury/utils/track.js
hakubo: I can see this was copied

rogatty: What else? We don't have `Ember.Logger` here.

hakubo:
lgtm

PR #1626, opened 2015-11-10
Don't include domain name for google search tracking

perjg:
The extra pageview impression when performing google custom search to track search should only include the path and not the domain.


rogatty:
LGTM

PR #1628, opened 2015-11-10
XW-514 Move to ES6 from TS in our front code

hakubo:
HERE BE DRAGONS


front/scripts/main/components/forum-wrapper.js
```diff
@@ -1,9 +1,8 @@
-/// <reference path="../app.ts" />
-/// <reference path="../mixins/DiscussionUpvoteActionSendMixin.ts" />
-'use strict';
+import Ember from 'ember';
+import DiscussionUpvoteActionSendMixin from '../mixins/discussion-upvote-action-send.js';
 
-App.ForumWrapperComponent = Em.Component.extend(
-	App.DiscussionUpvoteActionSendMixin,
+const ForumWrapperComponent = Ember.Component.extend(
```
hakubo: I think we should go with 
```
});
```

- looks cleaner
- you already have name of the component in form of a file name
- we don't have problem with const :)
- that is what ember-cli generates :)

hakubo:
I think we should go with

});
looks cleaner
you already have name of the component in form of a file name
we don't have problem with const :)
that is what ember-cli generates :)

PR #1631, opened 2015-11-10
XW-714 | import M and Mercury functions

tomasznapieralski:


front/scripts/main/app.js
```diff
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import {prop} from '../baseline/mercury/utils/state.js';
 import * as trackPerf from '../mercury/utils/trackPerf.js';
 import {getQueryParam} from '../mercury/utils/queryString.js';
+import {integrateOptimizelyWithUA} from '../mercury/utils/variantTesting.js';
```
hakubo: lets stick with no extension

tomasznapieralski: i know :) there is ticket about removing js extension, it will remove all at once

front/scripts/main/app.js
```diff
@@ -1,10 +1,12 @@
 import Ember from 'ember';
-import {prop} from '../baseline/mercury/utils/state.js';
-import * as trackPerf from '../mercury/utils/trackPerf.js';
-import {getQueryParam} from '../mercury/utils/queryString.js';
-import Ads from '../mercury/modules/Ads.js';
-import UniversalAnalytics from '../mercury/modules/Trackers/UniversalAnalytics.js';
-import LinkComponent from 'mixins/link-component.js';
+import Mercury from '../mercury/Mercury';
+import {prop} from '../baseline/mercury/utils/state';
+import * as trackPerf from '../mercury/utils/trackPerf';
```
Warkot: Importing `*` to the local scope makes it loose its origin.

front/scripts/main/components/post-detail.js
```diff
@@ -1,5 +1,7 @@
 import Ember from 'ember';
-import DiscussionUpvoteActionSendMixin from '../mixins/discussion-upvote-action-send.js';
+import Mercury from '../../mercury/Mercury';
+import DiscussionUpvoteActionSendMixin from '../mixins/discussion-upvote-action-send';
+import {buildUrl} from '../../baseline/mercury/utils/buildUrl';
```
Warkot: What about adding `export default` to `buildUrl`, so you can write without `{}`, as it's the most frequently imported function from this module?

front/scripts/main/mixins/main-page-route.js
```diff
@@ -1,4 +1,6 @@
 import Ember from 'ember';
+import {normalizeToWhitespace} from '../../mercury/utils/string';
+import {prop} from '../../baseline/mercury/utils/state';
```
Warkot: Same as mentioned before about `buildUrl` – maybe we could add `export default prop` to `state`?

front/scripts/main/app.js
Warkot: …or not. Nevermind. :wink:

Warkot:
I'm still not sure how to handle these `export default` statements. We can think about it later together. For now it LGTM.

PR #1641, opened 2015-11-16
INT-254: Send user id to qualaroo

perjg:
Mercury qualaroo tracker does not send user id (marked as TODO in the code)


hakubo:
LGTM

PR #1642, opened 2015-11-16
SOC-1566 | Use WhoAmI service instead of Helios for access_token validation

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1566

### DIFF
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -26,6 +26,9 @@ var localSettings: LocalSettings = {
 	discuss: {
 		baseAPIPath: 'discussion',
 	},
+	whoAmIService: {
+		path: '/whoami'
+	},
 	ironSecret: 'TEST_SECRET_REPLACE_THIS',
 	// NOTE: On your devbox, use your eth0 address in able to bind route to something accessible
 	host: process.env.HOST,
--- a/config/localSettings.base.ts
+++ b/config/localSettings.base.ts
@@ -27,7 +27,8 @@ var localSettings: LocalSettings = {
 		baseAPIPath: 'discussion',
 	},
 	whoAmIService: {
-		path: '/whoami'
+		path: '/whoami',
+		timeout: 3000
 	},
 	ironSecret: 'TEST_SECRET_REPLACE_THIS',
 	// NOTE: On your devbox, use your eth0 address in able to bind route to something accessible
--- a/server/app.ts
+++ b/server/app.ts
@@ -13,7 +13,7 @@ import Caching       = require('./lib/Caching');
 import Hapi          = require('hapi');
 import Logger        = require('./lib/Logger');
 import Utils         = require('./lib/Utils');
-import HeliosSession = require('./lib/HeliosSession');
+import WikiaSession = require('./lib/WikiaSession');
 import cluster       = require('cluster');
 import localSettings = require('../config/localSettings');
 import path          = require('path');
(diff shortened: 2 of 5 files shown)


server/app.ts
```diff
@@ -81,7 +81,7 @@ server.register(plugins, (err: any) => {
 	}
 });
 
-server.auth.scheme('helios', HeliosSession.scheme);
+server.auth.scheme('helios', WikiaSession.scheme);
```
bkoval: How about switching scheme name form 'helios' to 'wikia'?

server/lib/WikiaSession.ts
```diff
@@ -31,18 +31,18 @@ module HeliosSession {
...
+							Logger.error('WhoAmI connection error: ', {
 								err: err,
 								parseError: parseError
 							});
-							return reply(Boom.unauthorized('Helios connection error'));
+							return reply(Boom.unauthorized('WhoAmI connection error'));
 						}
 
-						if (parsed.error) {
+						if (parsed.status === 401) {
```
bkoval: What if it's going to be 500 or 400? Shouldn't we return Boom.unauthorized everytime when status !== 200?

server/lib/WikiaSession.ts
```diff
@@ -50,7 +50,13 @@ module HeliosSession {
 				}
 
 				Wreck.get(
-					authUtils.getHeliosUrl('/info') + '?code=' + accessToken + '&noblockcheck=1',
+					authUtils.getWhoAmIUrl(),
+					{
+						timeout: 3000,
```
bkoval: This value could be configurable when moved to localSettings.whoAmIService :)

server/lib/WikiaSession.ts
kvas-damian: parsed.status != HTTP response status.
HTTP errors are handled few lines above.

bkoval:
LGTM :santa:

server/lib/WikiaSession.ts
rafalkalinski: +1

server/lib/WikiaSession.ts
```diff
@@ -50,7 +50,13 @@ module HeliosSession {
...
+						headers: {
+							Cookie: 'access_token=' + encodeURIComponent(accessToken)
```
rafalkalinski: Isn't it enough to set 
xhrFields: {
	withCredentials: true,
},
?

kvas-damian: It isn't frontent bro...

rafalkalinski: whatever...

rafalkalinski:
LGTM

config/localSettings.base.ts
```diff
@@ -26,6 +26,10 @@ var localSettings: LocalSettings = {
 	discuss: {
 		baseAPIPath: 'discussion',
 	},
+	whoAmIService: {
```
rogatty: Why the `Service` prefix? We don't use it after `helios`. Would be nice to have some standard here.

kvas-damian: After looking again at ```helios``` and ```discuss``` I found that they're not enough descriptive. That's why I decided to start using ```Service``` suffix.
Changing all of them is not in scope of this ticket.

rogatty: Fair enough. Do we have a ticket to clean them up?

kvas-damian: https://wikia-inc.atlassian.net/browse/SOC-1631

PR #1652, opened 2015-11-18
XW-744 fix link component not transpiled to es5

bognix:


front/scripts/main/mixins/link-component.js
```diff
@@ -1,4 +1,4 @@
-Ember.LinkComponent.reopen({
+export default Ember.LinkComponent.reopen({
```
rogatty: Please add an explanatory comment in this file.

rogatty:
LGTM

PR #1654, opened 2015-11-18
Use Clousure syntax for 2D array

vforge:
This will be fixed in later versions of JSDoc
See https://github.com/jsdoc3/jsdoc/issues/705


rogatty:
LGTM

PR #1658, opened 2015-11-18
XW-715 Fix unit tests after switch to ES6

rogatty:
This PR changes two things:
1. Switches from PhantomJS 1.9.8 to PhantomJS 2.0.1
2. Fixes unit tests

Ad 1.
Switch to https://github.com/rogatty/karma-phantomjs2-launcher which has only one change - it uses https://github.com/rogatty/phantomjs2. The latter has two changes:
- It uses PhantomJS version `2.0.1-development` instead of `2.0.0-20141016`
- It has hardcoded URL (`https://github.com/gitromand/phantomjs/releases/download/2.0.1/phantomjs-2.0.1-development-linux-x86_64.zip`) to download a binary that works in our environment. The official repo doesn't currently support Ubuntu 12.04, see https://github.com/bprodoehl/phantomjs/releases/


rogatty:
LGTM :smiling_imp:

bognix:
As I understand this is temporary solution. What has to happen so we are no longer dependent on forked versions of libraries? It would be nice to put it in some comment in code why we use this not official version.

As temporary solution LGTM

PR #1680, opened 2015-11-26
INT-272 | Make sure nav is centered on screen

perjg:
https://wikia-inc.atlassian.net/browse/INT-272


mari-mcmurtrie:
LGTM if you want fixed value (px).

PR #1687, opened 2015-11-26
i18n | New translations for discussion

bkoval:
@kvas-damian @rafalkalinski It seems like translation files for discussions for a couple of languages are not in sync with the source one (english). Look at ja, de etc.

I would consider merging these translations to dev to get them in sync again.

This would also fix https://wikia-inc.atlassian.net/browse/SOC-1625


tomasznapieralski:
What's the status of this PR? Is it waiting for CR or is it in progress?

PR #1689, opened 2015-11-26
XW-518 | ES6 server-side

vforge:
https://wikia-inc.atlassian.net/browse/XW-518


README.md
```diff
@@ -3,29 +3,28 @@
...
-        localSettings = baseLocalSettings.extendSettings({
-            devboxDomain: 'joe',
-            loggers: {
-                console: 'debug'
-            },
-            port: 8000 // 7000 if running on devbox
-    });
-
-    export = localSettings;
+	import baseExtendSettingsDev from './localSettings.dev';
```
rogatty: It would look better without tab.

README.md
```diff
@@ -3,29 +3,28 @@
...
+	import baseExtendSettingsProd from './localSettings.base';
+
+	const baseExtendSettings = process.env.WIKIA_ENVIRONMENT === 'dev' ? baseExtendSettingsDev : baseExtendSettingsProd,
+		localSettings = baseExtendSettings({
+			devboxDomain: 'joe',
+			loggers: {
+				console: 'debug'
+			},
+			port: 8000 // 7000 if running on devbox
```
rogatty: I think we should just leave 7000 as devboxes are now the default.

server/lib/Utils.js
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/**
+ * Get environment as string
+ *
+ * @param {String} environment
+ * @returns {string}
+ */
+export function getEnvironmentString(environment) {
+	return getEnvironmentString(environment).toLowerCase();
```
rogatty: This seems strange, maybe it should be `getEnvironment(environment)`?

config/authLocaleSettings.js
```diff
@@ -1,7 +1,9 @@
-var deepExtend: any = require('deep-extend');
+/* eslint dot-notation:0 */
 
-var localeSettings: any = {
-	'en' : {
+import deepExtend from 'deep-extend';
+
+const authLocaleSettings = {
```
kvas-damian: Don't you think that this file should be shared across entire mercury platform?
In that case ```auth``` prefix doesn't make sense here.

PR #1691, opened 2015-11-27
SOC-1451 | Displaying deleted posts and replies

kvas-damian:
https://wikia-inc.atlassian.net/browse/SOC-1452
https://wikia-inc.atlassian.net/browse/SOC-1451


front/scripts/main/models/DiscussionPostModel.ts
```diff
@@ -103,15 +105,16 @@ App.DiscussionPostModel.reopenClass({
...
-						postCount: data.postCount,
+						forumId: data.forumId,
 						id: data.id,
-						pivotId: pivotId,
+						isDeleted: data.isDeleted,
 						page: 0,
-						title: data.title
+						pivotId: pivotId,
+						postCount: data.postCount,
+						replies: replies,
```
hakubo: you can use object shorthand
{
replies
}

front/styles/main/component/_share-feature.scss
```diff
@@ -2,6 +2,10 @@
 	display: block;
 	text-align: right;
 
+	.is-deleted & {
```
hakubo: we used to have .hidden class
maybe we could reuse that?

front/styles/main/module/_discussion.scss
```diff
@@ -65,6 +66,47 @@ body.discussions {
...
+	.post-reply,
+	.post-detail {
+		position: relative;
+
+		&.is-deleted {
+			&::before {
+				border-color: $discussion-deleted-mark-color transparent transparent transparent;
+				border-style: solid;
+				border-width: $deleted-mark-size $deleted-mark-size 0 0;
+				content: " ";
```
hakubo: silly: 
I think you can omit the space there

front/styles/main/theme/_dark.scss
```diff
@@ -82,20 +82,33 @@ body.dark-theme {
...
+		.post-reply,
+		.post-detail {
+			&.is-deleted,
+			&.is-parent-deleted {
+				.discussion-content {
+					color: $dark-discussion-deleted-content-color;
+				}
+
+				.count {
+					a, span {
```
hakubo: This nesting here went wild!
.dark-theme .discussion .post-detail.is-deleted .count a {
}

Can we flatten it somehow?
I think we should add scss linter to our build

front/styles/main/component/_share-feature.scss
kvas-damian: We're not hiding entire component - it's only functionality that doesn't make sense when component is deleted.

front/styles/main/theme/_dark.scss
kvas-damian: Probably it won't be a problem if finally introduce sass builder on the fly instead of build step - we talked about it already.
Right now we have to override selectors to apply dark theme colors so they needs to be more specific.

front/scripts/main/models/DiscussionPostModel.ts
vforge: you *should* use it - eslint will notice that ;)

PR #1695, opened 2015-11-27
XW-716 | Setup UA account only once

rogatty:
https://wikia-inc.atlassian.net/browse/XW-716

We create new tracker instance every time `mercury/utils/track`'s `track()` or `trackPageView()` is called.
Google wants us to call some methods just once per account.

We should rethink this whole concept but it's out of this ticket's scope.


hakubo:
LGTM

tomasznapieralski:
LGTM

PR #1697, opened 2015-11-30
INT-188 | Use common style for edit page / upload photo header

perjg:
Restyling of the header used for edit page and upload photo to use the same style as the main page curated content editor.

https://wikia-inc.atlassian.net/browse/INT-188

### DIFF
--- a/front/scripts/main/components/article-edit.js
+++ b/front/scripts/main/components/article-edit.js
@@ -11,7 +11,7 @@ export default App.ArticleEditComponent = Ember.Component.extend(
 		}),
 
 		adjustTextareaHeight: Ember.on('didInsertElement', () => {
-			this.$('textarea').css('height', this.$(window).height() - this.$('.sub-head').outerHeight());
+			Ember.$('textarea').css('height', Ember.$(window).height() - Ember.$('.sub-head').outerHeight());
 		}),
 
 		actions: {
--- a/front/scripts/main/components/article-edit.js
+++ b/front/scripts/main/components/article-edit.js
@@ -11,7 +11,7 @@ export default App.ArticleEditComponent = Ember.Component.extend(
 		}),
 
 		adjustTextareaHeight: Ember.on('didInsertElement', () => {
-			Ember.$('textarea').css('height', Ember.$(window).height() - Ember.$('.edit-head').outerHeight());
+			Ember.$('textarea').css('height', Ember.$(window).height() - Ember.$('.sub-head').outerHeight());
 		}),
 
 		actions: {
--- a/front/scripts/main/components/article-edit.js
+++ b/front/scripts/main/components/article-edit.js
@@ -11,7 +11,7 @@ export default App.ArticleEditComponent = Ember.Component.extend(
 		}),
 
 		adjustTextareaHeight: Ember.on('didInsertElement', () => {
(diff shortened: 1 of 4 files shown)


front/scripts/main/components/article-edit.js
```diff
@@ -11,7 +11,7 @@ export default App.ArticleEditComponent = Ember.Component.extend(
 		}),
 
 		adjustTextareaHeight: Ember.on('didInsertElement', () => {
-			Ember.$('textarea').css('height', Ember.$(window).height() - Ember.$('.edit-head').outerHeight());
+			Ember.$('textarea').css('height', Ember.$(window).height() - Ember.$('.sub-head').outerHeight());
```
hakubo: this.$ gives you a jquery that is searching within the component

perjg:
@hakubo Ok to merge this now?

front/scripts/main/components/article-edit.js
```diff
@@ -11,7 +11,7 @@ export default App.ArticleEditComponent = Ember.Component.extend(
 		}),
 
 		adjustTextareaHeight: Ember.on('didInsertElement', () => {
-			Ember.$('textarea').css('height', Ember.$(window).height() - Ember.$('.edit-head').outerHeight());
+			this.$('textarea').css('height', this.$(window).height() - this.$('.sub-head').outerHeight());
```
rogatty: Are you sure this works? I think that `this.$(window)` will fail as window isn't a part of the component.

rogatty:
One thing and it LGTM. Please test it before merging :)

hakubo:
One thing and it LGTM. Please test it before merging :) :)

perjg:
@hakubo @rogatty You're right, this.$ doesn't work in this context, so I reverted that commit.

There is Ember.$, this.$ and global $, I find the difference quite confusing :)

rogatty:
`$` and `Ember.$` are basically the same thing, it's good practice to use `Ember.$`. See http://stackoverflow.com/questions/19808742/what-is-the-difference-between-and-ember.

`this.$()` gives you jQuery element of the current component so you don't traverse through the whole DOM.

If you want to access something inside the component use `this.$()`. If you want to access something from the global scope use `Ember.$`. In your case it would be:
```
this.$('textarea').css('height', Ember.$(window).height() - Ember.$('.sub-head').outerHeight());
```

front/templates/main/components/article-edit.hbs
```diff
@@ -1,12 +1,6 @@
...
-	<button {{action 'publish'}} class="large publish" disabled={{publishDisabled}}>{{i18n 'app.edit-page-publish'}}</button>
-
-	<div class="edit-head-title">
-		{{! TODO: Ideally the title would be passed to the message so translators can choose the order (see HG-663) }}
-		{{i18n 'app.edit-page-heading'}} <span>{{model.title}}</span>
-	</div>
+<header class="sub-head">
+	<a {{action 'back'}} class="sub-head--cancel">{{svg 'back-arrow' role='img' class='icon'}}</a>
+	<button {{action 'publish'}} class="sub-head--done">{{i18n 'app.edit-page-publish'}}</button>
+	<div class="sub-head--title">{{i18n 'app.edit-page-heading'}}<span>{{model.title}}</span></div>
```
rogatty: @perjg there should be a space before `<span>` or it should have a margin.

PR #1698, opened 2015-11-30
Fix scripts-test-module task to compile only the files we want it to

rogatty:
Compile to ES6 modules only scripts from `auth`, `main` and `mercury` directories. The rest is handled in `scripts-front` task.

The secret here was the `base: './front/scripts/'` option which I found in `scripts-server` task.


tomasznapieralski:
LGTM

PR #1699, opened 2015-11-30
XW-728 | add PAD slot to main pages

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-728

### DIFF
--- a/front/styles/main/module/_ads.scss
+++ b/front/styles/main/module/_ads.scss
@@ -12,8 +12,8 @@
 	display: none !important;
 }
 
-.main-page-pad-slot {
-  margin-bottom: 36px;
+.main-page-pad-slot .ad-slot {
+	margin-bottom: 36px;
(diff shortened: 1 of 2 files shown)


front/styles/main/module/_ads.scss
```diff
@@ -12,6 +12,10 @@
 	display: none !important;
 }
 
+.main-page-pad-slot {
+  margin-bottom: 36px;
```
rogatty: Formatting

front/styles/main/module/_ads.scss
```diff
@@ -12,6 +12,10 @@
 	display: none !important;
 }
 
+.main-page-pad-slot {
```
hakubo: what if it does not load? will it be applied?

vforge:
LGTM, Tested

PR #1705, opened 2015-12-01
INT-281: Change url for Star Wars banner button

perjg:


perjg:
Merging immediately as this is an urgent request.

PR #1706, opened 2015-12-01
ADEN-2707 | Added window.wgNow - adTracker dependency

fraszczakszymon:
In order to make `ext.wikia.adEngine.adTracker` work properly we need to add `window.wgNow` with date when page was initialised (like it's on oasis).

https://wikia-inc.atlassian.net/browse/ADEN-2707

### DIFF
--- a/server/views/_layouts/ember-main.hbs
+++ b/server/views/_layouts/ember-main.hbs
@@ -34,6 +34,7 @@
 
 		<style><!-- @include ../../../www/front/styles/main/baseline.css --></style>
 
+		<script>var wgNow = new Date();</script>
 		{{> nirvana }}
 		<script>
 			<!-- @include ../../../www/front/scripts/baseline.js -->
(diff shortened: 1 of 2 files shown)


server/views/_layouts/ember-main.hbs
```diff
@@ -34,6 +34,7 @@
 
 		<style><!-- @include ../../../www/front/styles/main/baseline.css --></style>
 
+		<script>var wgNow = new Date();</script>
```
rogatty: Currently we keep those legacy workarounds in `server/views/_partials/nirvana.hbs`, please move it there.

Obviously it isn't the perfect name or place to keep it as you haven't found it. How about creating new file `front/scripts/baseline/mediawiki-shame.js` and moving all of it there? You don't need to change any logic, it will be inlined in the HTML automatically.

fraszczakszymon: Fixed :+1: 

I think I wouldn't find mediawiki-shame.js as well. I was looking for the first html file and I didn't know about nirvana. Now I will, thanks! :)

rogatty:
LGTM

nandy-andy:
LGTM :+1: And thanks @rogatty for follow-up ticket!

hakubo:
LGTM

PR #1710, opened 2015-12-01
PLATFORM-1564 | Image Review service

RafalWilinski:
@pchojnacki decided to mount image review frontend layer in mercury so there it is. Almost working. Any code review is highly appreciated.

Note:
-gulp/tasks/watch.js is modified because on my devbox command 'npm run dev' fails
-image uses /register route temporary because /imagereview is not yet added as exception in chef-repo

https://wikia-inc.atlassian.net/browse/PLATFORM-1564


front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,20 @@
+import App from '../app';
+import ImageReviewRoute from '../routes/image-review-route';
+
+export default App.ImageReviewItemComponent = Ember.Component.extend({
+    editorLayout: 'image-review-item',
+
+    actions: {
+        setOk() {
+			this.set('model.status', 0);
```
hakubo: odd spacing
also
it would be best if component was responsible for displaying data and sending actions up instead of manipulating data directly - this leads to much cleaner responsibilities

front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,20 @@
+import App from '../app';
+import ImageReviewRoute from '../routes/image-review-route';
```
hakubo: I think it's unused

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,27 @@
+import App from '../app';
+import ImageReviewModel from '../models/image-review-model';
+
+export default App.ImageReviewComponent = Ember.Component.extend({
+	classNames: ['image-review'],
+	isLoading: false,  //Is it needed?
```
hakubo: for now yes...
we have a ticket to fix that
I broke it - I'm sorry :(

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,27 @@
...
+
+	actions: {
+		reviewImages() {
+			console.log('reviewImages()');
+
+			this.get('model.images').map(function(x) {
```
hakubo: again it would be best if comoponents did not deal with data
also it looks like you intend to call each 
map is to create an array iterating over a different array

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,27 @@
+import App from '../app';
+import ImageReviewModel from '../models/image-review-model';
+
+export default App.ImageReviewComponent = Ember.Component.extend({
+	classNames: ['image-review'],
+	isLoading: false,  //Is it needed?
+
+	actions: {
+		reviewImages() {
+			console.log('reviewImages()');
```
hakubo: oops :)

front/scripts/main/mixins/image-review-mixin.js
```diff
@@ -0,0 +1,11 @@
+import App from '../app';
+
+export default App.ImageReviewMixin = Ember.Mixin.create({
```
hakubo: what is the purpose of this mixin?
also this renderTemplate looks very typical
I don't think you actually need it

front/scripts/main/models/image-review-item-model.js
```diff
@@ -0,0 +1,8 @@
+import App from '../app';
+import ObjectUtilitiesMixin from '../mixins/object-utilities';
+
+export default App.ImageReviewItemModel = Ember.Object.extend(App.ObjectUtilitiesMixin, {
```
hakubo: you import ObjectUtilitiesMixin so you can use it directly

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,128 @@
+import App from '../app';
+
+export default App.ImageReviewModel = Ember.Object.extend({
+    sessionId: null,
```
hakubo: spaces

front/scripts/main/routes/image-review-route.js
```diff
@@ -0,0 +1,44 @@
...
+						type: 'warning'
+					});
+					this.transitionTo('mainPage');
+				}
+				return true;
+			},
+
+			getMoreImages(sessionId) {
+				console.log("Getting more images for ID: "+sessionId);
+				ImageReviewModel.getImages(sessionId)
```
hakubo: you should use model that you have already created for this route
you can get that via:
`this.modelFor(this.routeName)`

front/scripts/main/components/image-review-item.js
RafalWilinski: This is quite weird, in webstorm looks ok. I'll fix it.

hakubo: you can enable there to highlight whitespaces - this will show you that there are spaces and tabs mixed

front/scripts/main/mixins/image-review-mixin.js
RafalWilinski: Well, yes. I was copying CuratedContent very strictly and didn't realize that this part is not necessary.

front/scripts/main/components/image-review.js
RafalWilinski: Can you elaborate on that :D ?

front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default App.ImageReviewItemComponent = Ember.Component.extend({
+    editorLayout: 'image-review-item',
```
rogatty: You don't need it, it was used in curated content editor to switch between different steps of process using `CuratedContentEditorLayoutMixin`.

front/scripts/main/controllers/image-review-controller.js
```diff
@@ -0,0 +1,14 @@
+import App from '../app';
+
+//Disclaimer: I'm not sure whether in current architecture of ImageReview it's necessary to have controller
+//I guess it's not
```
rogatty: It's not, you can delete this file.

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,116 @@
...
+	images: []
```
rogatty: This one is dangerous, please read "Leaking state into the class" part of https://dockyard.com/blog/2015/09/18/ember-best-practices-avoid-leaking-state-into-factories

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,23 @@
...
+import ImageReviewModel from '../models/image-review-model';
+
+export default App.ImageReviewComponent = Ember.Component.extend({
+	classNames: ['image-review'],
+	isLoading: false,  //Is it needed?
+
+	actions: {
+		reviewImages() {
+			this.get('model.images').forEach(function(imageItem) {
+				ImageReviewModel.reviewImage(imageItem.contractId, imageItem.imageId, imageItem.status);
```
rogatty: You probably don't want to operate on model from the component level. It's a good practice to send action up and handle it in the route.

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,116 @@
...
+					reject(data);
+				}
+			});
+		});
+	},
+
+	sanitize(rawData, contractId) {
+		var images = [];
+
+		if (rawData.length) {
```
vforge: I think it's not needed as line below won't be executed if length is zero anyway

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,116 @@
...
+			});
+		}
+
+		return App.ImageReviewModel.create({
+			images: images,
+			contractId: contractId
+		});
+	},
+
+	getServiceUrl() {
```
vforge: Use getter here :)
``get getServiceUrl() {
...
```

Then you can call it as a value directly, ie. `App.ImageReviewModel.getServiceUrl` instead of ``App.ImageReviewModel.getServiceUrl()`

File: front/scripts/main/routes/image-review-route.js [hunk 16]
```diff
@@ -0,0 +1,47 @@
...
+				});
+				this.transitionTo('mainPage');
+			}
+			return true;
+		},
+
+		getMoreImages(sessionId) {
+			ImageReviewModel.getImages(sessionId)
+				.then(function (data) {
+					// TODOś
```
vforge: I know it's WiP, but 3 comments:

1) TODO**ś** ?
2) This is proper syntax:
```
// @todo XXXXXXX
```
3) It's nice practice to put future ticket number when writing `@todo`

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="img-container">
+    <img id="{{model.imageId}}" src="{{img_src}}" width="230">
+</div>
+<a href="{{img_href}}" target="_blank" class="internal sprite details magnify" title="Go to image page"></a>
```
vforge: Are we sure we want to open another tab/window? In general it's a bad practice, especially on mobile.

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,10 @@
...
+<p> ImageID: {{model.imageId}} </p>
+<form>
+	<label title="Mark as OK"><input {{action 'setOk'}} type="radio" name="{{model.imageId}}" value="2" checked="">OK</label>
```
vforge: Just for curiosity, where's `value="1"` ?

gulp/tasks/watch.js
```diff
@@ -80,12 +80,12 @@ gulp.task('watch', ['build'], function () {
...
 		log('File changed:', gutil.colors.green(event.path), 'Restarting server');
 
-		server.restart(function () {
-			if (event.path.match('front')) {
-				reload(path);
-
-				log('Updating browser');
-			}
-		});
+		//server.restart(function () {
```
vforge: Please add `@todo` comment (or any other with explanation) to temporary changes.

server/routes.ts
```diff
@@ -144,19 +144,23 @@ authenticatedRoutes = [
 			]
 		}
 	},
+	//{
```
vforge: Please add `@todo` comment (or any other with explanation) to temporary changes.

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,116 @@
...
+				}
+			});
+		});
+	},
+
+	reviewImage(contractId, imageId, flag) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: App.ImageReviewModel.getServiceUrl() +
+				contractId + '/image/' + imageId + '?status=' + flag,
```
vforge: Use backticks

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,116 @@
...
+					reject(data);
+				}
+			});
+		});
+	},
+
+	getImages(contractId) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: App.ImageReviewModel.getServiceUrl() + contractId + '/image',
```
vforge: Use backticks

front/scripts/main/router.js
```diff
@@ -82,6 +82,10 @@ App.Router.map(function () {
 		});
 	});
 
+	this.route('imageReview', {
```
vforge: Please add @todo comment (or any other with explanation) to temporary changes.

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="img-container">
+    <img id="{{model.imageId}}" src="{{img_src}}" width="230">
```
vforge: It's very bad to hard-code with/height of objects. What about different screen resolutions?

RafalWilinski: I know it's very bad but it's copy-pasted from current image review :D

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,24 @@
+import App from '../app';
+import ImageReviewModel from '../models/image-review-model';
+
+export default App.ImageReviewComponent = Ember.Component.extend({
+	classNames: ['image-review'],
+	isLoading: false,
+
+	actions: {
+		reviewImages() {
+			// @todo Move this logic from model to route
```
rogatty: Component - sends action to route saying it wants to modify the model.
Route - keeps model instance, calls model's method.
Model - all the model logic, like this loop.

front/scripts/main/models/image-review-model.js
```diff
@@ -0,0 +1,115 @@
...
+				}
+			});
+		});
+	},
+
+	reviewImage(contractId, imageId, flag) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: App.ImageReviewModel.getServiceUrl +
+				contractId + `/image/` + imageId + `?status=` + flag,
```
rogatty:
```
`/image/${imageId}?status=${flag}`
```
but ESLint will tell you that anyway ;)

front/scripts/main/components/image-review.js
RafalWilinski: Like this? 

```
actions: { reviewImages() { ImageReviewRoute.reviewImages(); } }
```

In ImageReviewRoute
```
reviewImages() {
	ImageReviewModel.reviewImages(this.get('images'));
}
```

And in model: 
```
reviewImages(images) {
		images.forEach(function(imageItem) {
			reviewImage(imageItem.contractId, imageItem.imageId, imageItem.status);
		});
	},
```

rogatty: More like this:

ImageReviewComponent
```
actions: {
  reviewImage() { this.sendAction('reviewImages'); }
}
```

ImageReviewRoute
```
actions: {
  const model = this.modelFor('imageReview'); // get route's model which comes from model()
  model.reviewImages(); // or ImageReviewModel.reviewImages(model)
}
```

RafalWilinski:
Because of my misunderstanding of contractId, image-review service has been removed. It appears that contractId is rather package identifier than session identifier and it needs to be fetched each time you want to review more pictures.

front/scripts/main/components/radio-button.js
```diff
@@ -0,0 +1,15 @@
+import App from '../app';
+
+export default App.RadioButtonComponent = Ember.Component.extend({
+	tagName: 'input',
+	type: 'radio',
+	attributeBindings: ['checked', 'name', 'type', 'value'],
+
+	checked: function () {
+		return this.get('value') === this.get('groupValue');
+	}.property('value', 'groupValue'),
```
bognix: we use this notation in our codebase
`fullName: Ember.computed('firstName', 'lastName', function() {
    return this.get('firstName') + ' ' + this.get('lastName');
  })`

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,142 @@
...
+	startSession(onlyFlagged) {
+		let url = App.ImageReviewModel.getServiceUrl;
+
+		if (onlyFlagged) {
+			url = `${url}?status=FLAGGED`;
+		}
+
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: url,
```
bognix: use object shorthand notation:
```
url,
dataType: 'json'
```

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,142 @@
...
+				dataType: 'json',
+				method: 'POST',
+				xhrFields: {
+					withCredentials: true
+				},
+				success: (data) => {
+					resolve(App.ImageReviewModel.getImages(data.id));
+				},
+				error: (data) => {
+					reject(data);
```
bognix: you can change it to this:
```
error: (data) => reject(data)
```

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+export default App.ImageReviewRoute = Ember.Route.extend({
+
+	onlyFlagged: false,
+
+	renderTemplate(controller, model) {
+		this.render('image-review', {controller, model});
+	},
+
+	model() {
+		console.log('Flagged? '+ this.get('onlyFlagged'));
```
bognix: remove console.log

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+		this.set('isLoading', true);
+		return App.ImageReviewModel.startSession(this.get('onlyFlagged'));
+	},
+
+	actions: {
+		error(error) {
+			if (error.status === 401) {
+				this.modelFor(this.routeName).addAlert({
```
bognix: you won't see this error because on every transitions errors are cleared (didTransition inside routes/application.js).
You can add persistent: true flag (see error inside routes/main-page-section.js)

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+					message: 'Unauthorized, you don\'t have permissions to see this page',
```
bognix: what about i18n here? You should add this message to locales/en/main.json and use key here.

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,33 @@
+.image-review {
```
bognix: do you need to nest this selectors?

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
+<h2> Image Review </h2>
+<p> Click on images to mark them for deletion or as questionable (for staff review). When you're done with the batch, click "Review" below to get the next batch.</p>
```
bognix: i18n

bognix:
Please run eslint on this branch.
I'm not sure about release process for this feature (EN communities/users) only but it's always good pattern to add all messages to main.json file and use keys for messages in code

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,33 @@
...
+	  font-size: 0.5em;
+	}
+
+	.zero-margin {
+	  margin: 0;
+	}
+
+	.image-review-item {
+	  .REJECTED {
+		background-color: red;
```
vforge: Please use `#rgb` or `#rrggbb` colors instead ;)

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
...
+		{{/each}}
+    </li>
+</ul>
+
+{{#modal-dialog isVisible=model.isModalVisible}}
+	<img src="http://vignette-poz.wikia-dev.com/"{{model.modalImageUUID}}/>
+	<p>{{model.modalImageUUID}}</p>
+{{/modal-dialog}}
+
+<button class="review-images-and-get" {{action 'reviewAndGetMoreImages'}}>Review & get next batch</button>
```
vforge: use `&amp;` for ampersand instead of `&` sign in HTML

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
...
+<ul class="image-review-list">
+    <li class="image-review-item">
+		{{#each model.images as |item|}}
+			{{image-review-item model=item targetObject=this showModal='showModal'}}
+		{{/each}}
+    </li>
+</ul>
+
+{{#modal-dialog isVisible=model.isModalVisible}}
+	<img src="http://vignette-poz.wikia-dev.com/"{{model.modalImageUUID}}/>
```
vforge: Hard-coding URL?

RafalWilinski:
I think everything has been fixed here.

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,133 @@
...
+	endSession() {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: `${App.ImageReviewModel.getServiceUrl}/`,
+				xhrFields: {
+					withCredentials: true
+				},
+				dataType: 'json',
+				method: 'DELETE',
+				success: (data) => {
```
bognix: you can use the same shorthand here as in the "error" callback

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,133 @@
...
+	reviewImage(contractId, imageId, flag) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: `${App.ImageReviewModel.getServiceUrl}/${contractId}/image/${imageId}?status=${flag}`,
+				xhrFields: {
+					withCredentials: true
+				},
+				dataType: 'json',
+				method: 'PUT',
+				success: (data) => {
```
bognix: you can use the same shorthand here as in the "error" callback
 Add a line note

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,133 @@
...
+	addImage(contractId, imageId) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: `${App.ImageReviewModel.getServiceUrl}/${contractId}/image?imageId=${imageId}&status=UNREVIEWED`,
+				xhrFields: {
+					withCredentials: true
+				},
+				dataType: 'json',
+				method: 'POST',
+				success: (data) => {
```
bognix: you can use the same shorthand here as in the "error" callback
 Add a line note

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
+<h2> {{i18n 'app.image-review-head'}} </h2>
+<p> {{i18n 'app.image-review-description'}} </p>
+
+<h1> {{i18n 'app.image-review-images-to-review'}} </h1>
+
+<ul class="image-review-list">
+    <li class="image-review-item">
+		{{#each model.images as |item|}}
+			{{image-review-item model=item targetObject=this showModal='showModal'}}
```
bognix: where do you use targetObject?

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
...
+		{{/each}}
+    </li>
+</ul>
+
+{{#modal-dialog isVisible=model.isModalVisible}}
+	<img src={{model.baseImgUrl}} {{model.modalImageUUID}}/>
```
bognix: you should use " " around src value

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,125 @@
+import App from '../app';
+
+export default App.ImageReviewModel = Ember.Object.extend({
+
+	init() {
+		this.isModalVisible = false;
+		this.modalImageUUID = '123';
```
bognix: can this be null or not set by default? '123' looks odd

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,125 @@
+import App from '../app';
+
+export default App.ImageReviewModel = Ember.Object.extend({
+
+	init() {
+		this.isModalVisible = false;
+		this.modalImageUUID = null;
+		this.images = this.get('images');
+		this.baseImgUrl = 'http://vignette-poz.wikia-dev.com/';
```
bognix: this url is only valid on poz devboxes. Are you sure we want to hardcode it?

RafalWilinski:
I wondering whether I should rename CuratedContentThumbnailMixin to something more generic, what do you think?

rogatty:
@RafalWilinski I think you shouldn't use it at all as it's tightly coupled with curated content. There are things like `block` property or `imageCrop` data. If I were you I would just extract what you need from it.

rogatty:
Please read https://dockyard.com/blog/2015/10/29/ember-best-practice-stop-bubbling-and-use-closure-actions and use closure actions instead of bubbling them manually like this:
```
showModal(id) {
	this.sendAction('showModal', id);
}
```

front/locales/en/main.json
```diff
@@ -143,7 +143,14 @@
     "curated-content-editor-main-page": "Main Page",
     "curated-content-editor-new-featured-content": "New Featured Content",
     "curated-content-editor-new-section": "New Section",
-    "curated-content-editor-new-category": "New Category"
+    "curated-content-editor-new-category": "New Category",
+    "image-review-head": "Image Review",
+    "image-review-description": "Click on images to mark them for deletion or as questionable (for staff review). When you're done with the batch, click 'Review' below to get the next batch.",
+    "image-review-images-to-review": "Images to review: ",
```
rogatty: I think there should be no space at the end.

front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,21 @@
+import App from '../app';
+import Thumbnailer from '../../mercury/modules/Thumbnailer';
+
+export default App.ImageReviewItemComponent = Ember.Component.extend({
```
rogatty: Please remove all references to `App`. After loader.js was merged you should just do:
```
```

front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,21 @@
...
+	thumbnailer: Thumbnailer,
+	imageWidth: 256,
```
rogatty: Seems like it's `imageSize`

front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,21 @@
...
+	emptyGif: 'data:image/gif;base64,R0lGODlhEAAJAIAAAP///////yH5BAEKAAEALAAAAAAQAAkAAAIKjI+py+0Po5yUFQA7',
+	thumbnailUrl: Ember.computed('model.fullSizeImageUrl', function () {
+		if (this.get('model.fullSizeImageUrl')) {
```
rogatty: You use it twice, so how about a const?

front/scripts/main/components/image-review-item.js
```diff
@@ -0,0 +1,21 @@
+import App from '../app';
+import Thumbnailer from '../../mercury/modules/Thumbnailer';
+
+export default App.ImageReviewItemComponent = Ember.Component.extend({
+	thumbnailer: Thumbnailer,
```
rogatty: I think this is obsolete, you already use `Thumbnailer.mode.scaleToWidth` directly below and it makes sense.

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,20 @@
...
+		reviewAndGetMoreImages() {
+			this.sendAction('reviewAndGetMoreImages');
+		},
+
+		showModal(imageUrl) {
+			this.sendAction('showModal', imageUrl);
+		},
+
+		getFlaggedOnly() {
+			this.sendAction('getFlaggedOnly');
```
rogatty: You can get rid of this bubbling using closure actions.

front/scripts/main/components/radio-button.js
```diff
@@ -0,0 +1,15 @@
+import App from '../app';
+
+export default App.RadioButtonComponent = Ember.Component.extend({
+	tagName: 'input',
+	type: 'radio',
+	attributeBindings: ['checked', 'name', 'type', 'value'],
+
+	checked: Ember.computed('value', 'groupValue', function () {
+		return this.get('value') === this.get('groupValue');
```
rogatty: I don't understand how it works. In `change` hook you set `groupValue` to the same thing as `value` and here you compare them. Won't it always be `true`, then?

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,126 @@
+import App from '../app';
+
+export default App.ImageReviewModel = Ember.Object.extend({
+
+	init() {
+		this.isModalVisible = false;
+		this.modalImageUrl = null;
+		this.images = this.get('images');
```
rogatty: What is it for?

front/scripts/main/models/image-review-item.js
```diff
@@ -0,0 +1,10 @@
+import App from '../app';
+import ObjectUtilitiesMixin from '../mixins/object-utilities';
+
+export default App.ImageReviewItemModel = Ember.Object.extend(ObjectUtilitiesMixin, {
+	imageId: null,
+	thumbnailUrl: null,
```
rogatty: I think it's not needed.

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,126 @@
...
+	},
+	reviewImages(images) {
```
rogatty: This method doesn't use `this` so it can be moved below, right?

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,126 @@
...
+					status: 'ACCEPTED'
+				});
+			}
+			// else skip because is reviewed already
+		});
+
+		return App.ImageReviewModel.create({images, contractId});
+	},
+
+	getServiceUrl: 'https://services-poz.wikia-dev.com/image-review/contract'
```
rogatty: I think what @vforge mentioned by making it a getter wasn't to change the name but to use https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Functions/get altough I'm not sure about the actual syntax you could use here.

rogatty: Oh, and this URL shouldn't be hardcoded here, especially that it will work only from Poznań devboxes. See how discussions do that (using `servicesDomain` from localSettings).

front/scripts/main/router.js
```diff
@@ -82,6 +82,10 @@ Router.map(function () {
...
+		path: '/imagereview'
```
rogatty: How about `image-review`? It's the common web standard that we plan to introduce anyway.

RafalWilinski: This will require change in chef-repo but ok, I'll try to force that change.

front/scripts/main/components/radio-button.js
RafalWilinski: To be honest with you I also don't have clue how does it work, it's copied from StackOverflow which I know it's terrible practice but it's working. I'll try to provide clear solution asap.

front/scripts/main/router.js
rogatty: I don't think you need to "force" that change. Just open a PR ;)

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+				this.modelFor(this.routeName).addAlert({
+					message: i18n.t('app.image-review-error-no-access-permissions'),
+					type: 'warning',
+					persistent: true
+				});
+				this.transitionTo('mainPage');
+			} else {
+				Ember.Logger.error(error);
+				this.modelFor(this.routeName).addAlert({
+					message: i18n.t('app.image-review-error-other'),
```
rogatty: Can you keep it DRY? Those two blocks are almost identical, just put the message in a variable.

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+		error(error) {
+			if (error.status === 401) {
+				this.modelFor(this.routeName).addAlert({
+					message: i18n.t('app.image-review-error-no-access-permissions'),
+					type: 'warning',
+					persistent: true
+				});
+				this.transitionTo('mainPage');
+			} else {
+				Ember.Logger.error(error);
```
rogatty: You return true at the end so it will bubble up to `routes/application#error` where we already log errors. Isn't it duplicated right now?

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+					persistent: true
+				});
+				this.transitionTo('mainPage');
+			}
+			return true;
+		},
+
+		reviewAndGetMoreImages() {
+			this.set('onlyFlagged', false);
+			const model = this.modelFor('imageReview');
```
rogatty: Please put `const` at the top of the scope.

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,56 @@
...
+
+			model.reviewImages(model.images);
+			this.refresh();
+		},
+
+		showModal(imageUrl) {
+			const model = this.modelFor('imageReview');
+
+			model.set('modalImageUrl', imageUrl);
+			model.set('isModalVisible', true);
```
rogatty: It seems wrong to have this logic in the route and set property controlling UI in the model. Can this action be moved to the component instead?

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,26 @@
+.image-review-list {
+  @media screen and (min-width:500px) {
```
rogatty: Please use tabs.

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,26 @@
...
+	-webkit-column-count: 2;
+	-moz-column-count: 2;
```
rogatty: We use autoprefixer so this shouldn't be needed.

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,26 @@
...
+	column-count: 4;
+  }
+  .caption {
+	font-size: 0.5em;
+  }
+  .zero-margin {
+	margin: 0;
+  }
+  .image-review-item {
+	.REJECTED {
```
rogatty: Seems strange, can we use lowercase?

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="img-container {{model.status}}">
+	<img id="{{model.imageId}}" src="{{thumbnailUrl}}" tabindex = "-1" {{action (action this.attrs.showModal model.fullSizeImageUrl)}}>
```
rogatty: Too much space in there :see_no_evil:

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,11 @@
...
+
+	<p class="caption zero-margin"> ImageID: {{model.imageId}} </p>
```
rogatty: Those spaces seem unnecessary. Shouldn't we also localize `ImageID`?

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,11 @@
...
+
+	<form>
+		{{radio-button value='ACCEPTED' groupValue=model.status}} OK
```
rogatty: I think it would be cleaner if we passed this text as a `label` attribute and render inside the component. It should probably be localized too (yes, we localize everything! :smiling_imp:).

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
+<h2> {{i18n 'app.image-review-head'}} </h2>
```
rogatty: Spaces, spaces everywhere :see_no_evil:

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
...
+<p> {{i18n 'app.image-review-description'}} </p>
+
+<h1> {{i18n 'app.image-review-images-to-review'}} </h1>
```
rogatty: It probably shouldn't be `h1`.

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
...
+
+<ul class="image-review-list">
+    <li class="image-review-item">
```
rogatty: Please use tabs.

package.json
```diff
@@ -24,7 +24,6 @@
     "bluebird": "2.10.2",
     "boom": "2.7.0",
     "bunyan": "1.3.4",
-    "bunyan-syslog": "^0.3.0",
```
rogatty: A mistake? We use it in server.

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
...
+		{{#each model.images as |item|}}
+			{{image-review-item model=item showModal=(action 'showModal')}}
+		{{/each}}
+    </li>
+</ul>
+
+{{#modal-dialog isVisible=model.isModalVisible}}
+	<img src="{{model.modalImageUrl}}"/>
+	<p>{{model.modalImageUUID}}</p>
```
rogatty: Please don't use model for this. It's strictly a UI thing.

vforge:
Please fix linter:
```
> eslint .


/var/lib/jenkins/workspace/Mercury-PR-Checks/front/scripts/main/components/image-review.js
  11:42  error  Missing semicolon  semi

/var/lib/jenkins/workspace/Mercury-PR-Checks/front/scripts/main/routes/image-review.js
  10:10  error  "ImageReviewModel" is not defined  no-undef

/var/lib/jenkins/workspace/Mercury-PR-Checks/server/facets/showDiscussions.js
  2:35  warning  Multiple spaces found before 'from'  no-multi-spaces
```

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,127 @@
...
+				xhrFields: {
+					withCredentials: true
+				},
+				dataType: 'json',
+				method: 'GET',
+				success: (data) => {
+					if (Ember.isArray(data)) {
+						resolve(ImageReviewModel.sanitize(data, contractId));
+					} else {
+						reject('Invalid data was returned from Image Review API');
```
vforge: What about i18n this message?

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,127 @@
...
+		return ImageReviewModel.create({images, contractId});
+	},
+
+	reviewImages(images) {
+		images.forEach((imageItem) => {
+			ImageReviewModel.reviewImage(imageItem.contractId, imageItem.imageId, imageItem.status);
+		});
+	},
+
+	getServiceUrl: 'https://services-poz.wikia-dev.com/image-review/contract'
```
vforge: Wouldn't be better to put this into `localSettings`? Also, this URL is valid only for POZ devboxes.

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,22 @@
+.image-review-list {
+	@media screen and (min-width:500px) {
```
vforge: Are you sure we want to introduce custom media queries? We would probably be  better using floats or flexed layout for this.

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="img-container {{model.status}}">
+	<img id="{{model.imageId}}" src="{{thumbnailUrl}}" tabindex = "-1"
+		{{action (action this.attrs.showModal model.fullSizeImageUrl)}}>
+	<form>
+		{{radio-button value='ACCEPTED' groupValue=model.status}} OK
```
vforge: Why we don't i18n "OK", "Del" and "Q" ? Also, I would suggest adding tooltips to those to explain the actions.

front/styles/main/component/_image-review.scss
```diff
@@ -0,0 +1,22 @@
...
+	@media screen and (min-width:800px) {
+		column-count: 4;
+	}
+	.caption {
+		font-size: 0.5em;
+	}
+	.zero-margin {
+		margin: 0;
+	}
+	.image-review-item {
```
vforge: This nesting isn't necessary (you can move it outside of `.image-review-list`.

front/templates/main/components/image-review-item.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="img-container {{model.status}}">
+	<img id="{{model.imageId}}" src="{{thumbnailUrl}}" tabindex = "-1"
```
vforge: Tab Index should be unique on all elements. Also, why we want it anyway here?

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
+<h2>{{i18n 'app.image-review-head'}}</h2>
+<p>{{i18n 'app.image-review-description'}}</p>
+
+<h4>{{i18n 'app.image-review-images-to-review'}}</h4>
+
+<ul class="image-review-list">
+	<li class="image-review-item">
```
vforge: Are you sure this shouldn't be inside `each` block? In your version you'll have:
```
ul
  li
    div
    div
    div
```

front/templates/main/components/image-review-item.hbs
RafalWilinski: Because I wanted to make images unselectable by pressing 'tab' key.

front/scripts/main/router.js
RafalWilinski: PR on it's way: https://github.com/Wikia/chef-repo/pull/7651/files

rogatty: Not sure if you already planned to do it but remember to do the same change for US devboxes and production vcl before / during / after the release of this feature.

server/routes.js
```diff
@@ -239,6 +239,14 @@ let routes,
 				cache: routeCacheConfig
 			}
 		},
+		{
+			method: 'GET',
+			path: '/imagereview',
```
rogatty: You probably want to change this one too.

front/scripts/main/components/image-review.js
```diff
@@ -0,0 +1,13 @@
+export default Ember.Component.extend({
+	classNames: ['image-review'],
+	isLoading: false,
```
rogatty: This is the default, no need to write it here.

front/scripts/main/controllers/image-review.js
```diff
@@ -0,0 +1,13 @@
+export default Ember.Controller.extend({
+	application: Ember.inject.controller(),
```
rogatty: You don't seem to use it, is it needed?

front/scripts/main/models/image-review-item.js
```diff
@@ -0,0 +1,5 @@
+const ImageReviewItemModel = Ember.Object.extend({
+	status: 'accepted'
```
rogatty: Sometimes you use lowercase, sometimes uppercase. Can it be unified? IMO it should be lowercase everywhere. If you need to send uppercase to the service I would sanitize in the model just before sending.

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,51 @@
+import ImageReviewModel from '../models/image-review';
+
+export default Ember.Route.extend({
+	onlyFlagged: false,
+
+	beforeModel(transition) {
+		if (transition.queryParams.fullscreen === 'true') {
+			this.controllerFor('application').set('fullPage', true);
```
rogatty: It would be cleaner to do this in `didTransition` hook. It has nothing to do with the model, after all. Also you should unset it when leaving this route, how about using `willTransition` as in `routes/courated-content-editor`?

front/scripts/main/routes/image-review.js
```diff
@@ -0,0 +1,51 @@
...
+		}
+	},
+
+	renderTemplate(controller, model) {
+		this.render('image-review', {controller, model});
+	},
+
+	model() {
+		this.set('isLoading', true);
```
rogatty: From what I see you use this property in the component. Setting it here doesn't change anything.

front/scripts/main/models/image-review-item.js
RafalWilinski: I'm already using lowercase everywhere.

rogatty: Oh, right, there's uppercase only in the service urls :+1:

front/scripts/main/routes/image-review.js
RafalWilinski: Neither `didTransition` nor `willTransition` hook fire when entering site directly (like from an iframe).

RafalWilinski: Ok, `didTransition` as action is triggered however it's argument is undefined. `willTransition` still doesn't work when entering website.

front/scripts/main/components/radio-button.js
```diff
@@ -0,0 +1,15 @@
...
+	}),
+
+	change() {
+		this.set('groupValue', this.get('value'));
```
rogatty: `groupValue` is an attribute passed to this component in [the parent template](https://github.com/Wikia/mercury/pull/1710/files#diff-92236f820d9378b4862f9663741d6710R4) and here we change its value. Effectively it means that we mutate route's model down here in a component which isn't really [Data Down, Actions Up](https://dockyard.com/blog/2015/10/14/best-practices-data-down-actions-up).

front/scripts/main/components/radio-button.js
rogatty: Actually it does make sense, see [this comment](https://github.com/Wikia/mercury/pull/1710/files#r48061853).

front/scripts/main/components/radio-button.js
```diff
@@ -0,0 +1,15 @@
+import App from '../app';
+
+export default App.RadioButtonComponent = Ember.Component.extend({
+	tagName: 'input',
+	type: 'radio',
```
rogatty: What do you think about using button group instead of radio buttons? I've read (and agree) that [button groups give user better experience](http://www.usability247.com/blog/radio-buttons-touchscreen-usability/), especially that we want to be mobile first.

front/templates/main/components/image-review.hbs
```diff
@@ -0,0 +1,23 @@
+<h2>{{i18n 'app.image-review-head'}}</h2>
```
rogatty: I believe this tool should have the same kind of header as Curated Content Editor. Altough it might be a candidate for a separate PR.

front/scripts/main/components/radio-button.js
RafalWilinski: As far as I know, goal of this PR is to create thing that looks almost the same like the old imagereview but leverages new tech. Buttons might be good however they have to give clear feedback about current image state. 

I thought about something like this: http://imgur.com/S9umwUl

Button background color reflects current state (in the picture current image is marked as inappropriate)

RafalWilinski:
- [ ] radio buttons replaced with button groups, buttons showing review status by background-color
- [ ] same subheader as in curated content with hiding via fullscreen queryParam

front/locales/en/main.json
```diff
@@ -143,7 +143,17 @@
     "curated-content-editor-main-page": "Main Page",
     "curated-content-editor-new-featured-content": "New Featured Content",
     "curated-content-editor-new-section": "New Section",
-    "curated-content-editor-new-category": "New Category"
+    "curated-content-editor-new-category": "New Category",
+    "image-review-head": "Image Review",
```
rogatty: CC editor is a really bad example here :see_no_evil: 
How about moving Image Review's strings to a separate file? See https://github.com/Wikia/mercury/tree/dev/front/locales/en for examples and https://github.com/Wikia/mercury/blob/dev/front/templates/main/components/discussion-connection-error.hbs#L2 for usage.

@Warkot how would this change the Crowdin process?

Warkot: If you talk about moving CC i18n, we could upload the new files initially including their translations. I'm not sure about Crowdin CLI, but it should work smoothly via their webpage. There's also Translation Memory, which proposes already existing translations if the base (non-translated) string is not changed.

Adding new file to Crowdin, however, should be close to trivial—just remember to add new configuration file in apps root dir. From there act as usual.

If in doubt, I offer my help! :helicopter:

front/scripts/main/models/image-review.js
```diff
@@ -0,0 +1,128 @@
...
+ImageReviewModel.reopenClass({
+
+	startSession(onlyFlagged) {
+		const options = {};
+
+		if (onlyFlagged) {
+			options.status = 'FLAGGED';
+		}
+
+		return new Ember.RSVP.Promise((resolve, reject) => {
```
rogatty: After https://github.com/Wikia/mercury/pull/1797 is merged it would be great to reuse `ajaxCall` util. It's already in testing and shouldn't take long.

RafalWilinski: As far as I understand this ajax-call utility always resolves promise with `this`. https://github.com/Wikia/mercury/blob/b0aa4e42177a9f9a5509816fabd4a82fcf35a905/front/scripts/main/utils/ajax-call.js

@rogatty My model heavily relies on resolving promises with data or even chaining them (StartSession if succeed launches another promise (getImages)). Should I change logic behind my model and route or extend ajax-call util to have an ability to reject/resolve with custom data? Second option seems more forward-looking.

rogatty: You're right, I missed it. Let's leave it for now as we're going to introduce https://github.com/instructure/ic-ajax anyway.

front/scripts/main/components/image-review.js
jcellary: My javascript is very poor, but my guess is we don't use console.log in production code :) it's for debugging and development only

front/templates/main/components/image-review-item.hbs
jcellary: Just because someone did something wrong, doesn't mean we need to repeat this mistake :)

rogatty:
LGTM

PR #1712, opened 2015-12-02
INT-248: Global rollout of section editor

perjg:
This PR enables the Mercury section editor globally.
It simply removes the whitelist check in place now: All communities will have section editor enabled unless wiki.disableMobileSectionEditor is set to true.

No other rule changes.

### DIFF
--- a/front/scripts/main/components/article-wrapper.js
+++ b/front/scripts/main/components/article-wrapper.js
@@ -72,36 +72,16 @@ export default App.ArticleWrapperComponent = Ember.Component.extend(
 
 		/**
 		 * Checks if contribution component should be enabled
-		 * http://clashofclans.wikia.com/        db: clashofclans
-		 * http://de.clashofclans.wikia.com/     db: declashofclans
-		 * http://zh.clashofclans.wikia.com/     db: zhclashofclans723
-		 * http://fr.clashofclans.wikia.com/     db: frclashofclans
-		 * http://ru.clashofclans.wikia.com/     db: ruclashofclans
-		 * http://es.clash-of-clans.wikia.com/   db: esclashofclans727
-		 * http://pt-br.clashofclans.wikia.com/  db: ptbrclashofclans
 		 *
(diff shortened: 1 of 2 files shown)


perjg:
Need to update tests to match this PR too, will push another commit

owend:
Review looks good, once the tests are updated this can be merged, and it looks like it should also be merged to release branch for thursday 12/3 release.

perjg:
Fixed the tests. Merging as this needs to be ready for tomorrow’s release.

Ignoring the remaining linter error for now:
It complains about using function instead of =>, and this file still has lots of remaining uses of function for callbacks.

However, some of the callbacks use ‘this’ - I’m not sure how Ember.Component.extend binds 'this', and whether or not that plays well together with => (i.e. will this still be the same as the one bound by Ember.Component.extend).

@rogatty @igor If you can help me clarify I can clean this up in another PR.

rogatty:
@perjg linter complains only about this one function:
```
/var/lib/jenkins/workspace/Mercury-PR-Checks/JOB-575/front/scripts/main/components/article-wrapper.js
  78:51  error  Unexpected function expression  prefer-arrow-callback
```
It didn't complain before because there was `this` used in it. After your change it's a very simple function that doesn't use `this` and it could (should) be converted to an arrow one.

bognix:
@perjg if you don't know whether to use arrow function or not read this https://github.com/Wikia/guidelines/blob/master/JavaScript/es6Guidelines.md#most-important-enabled-rules and https://github.com/Wikia/guidelines/blob/master/JavaScript/es6Guidelines.md#arrow-functions-gotchas. If it doesn't explain it to you let me know and I'll update the doc with needed information

perjg:
@rogatty @bognix Thanks for the clarification, and thanks for cleaning it up in a separate PR.

I thought the linter rule would complain about all the uses of function instead of => in the file, so I was nervous about introducing possible regressions.

PR #1713, opened 2015-12-02
XW-810 | re-add tracking to share toolbar

Warkot:


vforge:
LGTM

front/scripts/main/components/share-feature.js
```diff
@@ -199,6 +199,8 @@ export default App.ShareFeatureComponent = Ember.Component.extend(
 			 * @returns {void}
 			 */
 			share(network) {
+				this.trackClick('share', network);
+
 				const urlGetter = this.get(network);
```
rogatty: This should be on top.

PR #1716, opened 2015-12-02
XW-688 | Use ES6 and babel for Japan Homepage

hakubo:
https://wikia-inc.atlassian.net/browse/XW-688

Thanks @perjg !

We'll take it from here


homepage/.eslintrc
```diff
@@ -0,0 +1,26 @@
+{
```
hakubo: it would be awesome if it was the same as the one in front

homepage/front/loader.js
```diff
@@ -0,0 +1,188 @@
+var define, requireModule, require, requirejs;
```
hakubo: bower should install it actually I think

homepage/front/scripts/main/globals.js
```diff
@@ -0,0 +1,25 @@
+class Globals {
```
hakubo: I think class here is not needed and this could be just enclosed in a module

homepage/front/scripts/main/main.js
```diff
@@ -0,0 +1,100 @@
+
+
+let Globals = require('./globals'),
```
hakubo: import globals from './globals';

hakubo: probably we need also some app.js that would boot the app by requiring everything

homepage/front/scripts/main/main.js
```diff
@@ -0,0 +1,100 @@
...
+function search() {
+	let searchText = encodeURI($('#searchWikiaText').val());
+
+	if (!searchText) {
+		// search button for mobile is different element
+		searchText = encodeURI($('#searchWikiaTextMobile').val());
+	}
+
+	if (searchText) {
+		let searchUrl = 'http://ja.wikia.com/Special:Search?search=';
```
vforge: Use backticks.

homepage/package.json
```diff
@@ -52,6 +57,7 @@
     "multipipe": "^0.1.2",
     "phantomjs": "^1.9.18",
     "pretty-hrtime": "^1.0.0",
-    "qunitjs": "^1.18.0"
+    "qunitjs": "^1.18.0",
+    "systemjs-builder": "^0.14.11"
   }
 }
```
bognix: remove gulp-tslint from package.json

homepage/front/scripts/main/globals.js
```diff
@@ -0,0 +1,23 @@
+/* eslint no-new: 0 */
+
+let cachedData = {};
+
+/**
+ * @returns {void}
+ */
+export function loadGlobalData() {
+	new RSVP.Promise((resolve) => {
+		$.get('/globals', (data) => {
```
bognix: can you use the arrow function shorthand here?
```
$.get('/globals', (data) => resolve(data))
.then(...
```

homepage/front/scripts/main/globals.js
```diff
@@ -0,0 +1,23 @@
...
+			resolve(data);
+		}).then((data) => {
```
bognix: same here

homepage/package.json
Warkot: What about `gulp-jshint`?

Warkot: And `gulp-uglify`?

bognix: if it's not used - go for it

Warkot:
Hey @perjg, could you take a look at this? :smiley:

homepage/front/scripts/main/globals.js
```diff
@@ -0,0 +1,19 @@
+/* eslint no-new: 0 */
+
+let cachedData = {};
+
+/**
+ * @returns {void}
+ */
+export function loadGlobalData() {
+	new RSVP.Promise((resolve) => {
```
hakubo: this promise here looks weird... do we need that
$.get does return promise as well

also

this cachedData will become private to globals.js right?
should it get a getter?

perjg: My bad... it should be sufficient to just return $.get here.

perjg:
@Warkot @hakubo I've checked, and everything seems to work smoothly. I will check if it is ok to merge now, and get back to you.

We are currently planning to make a new revision of the Japan HP, see here: INT-257
Ideally I'd like to merge the ES6 change before we start work on that. It is a perfect point to do the ES6 switch, since the new revision will need to be tested extensively before release anyway, and it will give me time to find any minor issues I might have missed this time around.

I also have a branch for activating Google Search A/B test on Japan homepage:
https://github.com/Wikia/mercury/commits/INT-285-2

The original issue is INT-203. We will probably start this A/B test after new year, so it makes sense for me to rewrite that PR in ES6 too.

(In case we need to make any hotfixes to the homepage during the code freeze, I will
branch from the last point before ES6 and use that for emergency releases.)

Warkot:
Therefore, @perjg, this PR returns into your hands. :open_hands: Check again if the code works after the last commit and then merge it when you see it fit. If any X-Wing assistance is needed (with this or anything else we can help), just ping us! :smiley:

PR #1719, opened 2015-12-02
XW-613 | Download scripts in parallel

rogatty:
https://wikia-inc.atlassian.net/browse/XW-613

Loading script is copied from http://www.html5rocks.com/en/tutorials/speed/script-loading/#toc-dom-rescue


front/scripts/baseline/mercury/utils/loadScripts.js
```diff
@@ -0,0 +1,64 @@
+if (typeof window.M === 'undefined') {
```
bkoval: What if M is null / 0 / '0' / false? Why not if (!window.M) ? :)

front/scripts/baseline/mercury/utils/loadScripts.js
```diff
@@ -0,0 +1,64 @@
...
+	 *
+	 * @param {string[]} scripts
+	 * @returns {void}
+	 */
+	M.loadScripts = function (scripts) {
+		let src = scripts.shift(),
+			scriptElement;
+
+		// loop through our script urls
+		while (src) {
```
hakubo: I think for each would be clearer here. What is the benefit of while in this case?

front/scripts/baseline/mercury/utils/loadScripts.js
```diff
@@ -0,0 +1,64 @@
...
+				pendingScripts.push(scriptElement);
+				// listen for state changes
+				scriptElement.onreadystatechange = stateChange;
+				// must set src AFTER adding onreadystatechange listener
+				// else we’ll miss the loaded event for cached scripts
+				scriptElement.src = src;
+			} else {
+				// fall back to defer
+				/* eslint no-useless-concat:0 */
+				document.write(`<script src="${src}" defer><` + `/script>`);
```
hakubo: What is this concatenation for?

rogatty: http://stackoverflow.com/questions/27517750/uncaught-syntaxerror-unexpected-token-illegal-script-in-script-scrip

front/scripts/baseline/mercury/utils/loadScripts.js
rogatty: I'm not sure why, just copied it from other utils... @vforge? @bognix?
It's a temporary solution anyway, we plan to rethink the baseline scripts when moving towards Ember CLI.

hakubo:
LGTM

tomasznapieralski:
LGTM

front/scripts/baseline/mercury/utils/loadScripts.js
rogatty: I've created a ticket to clean this up https://wikia-inc.atlassian.net/browse/XW-820

vforge: @bkoval Why it should be? Alternatively we can check if it's `=== "object"`

bkoval: Just wondering. Thing is, you want an object here, so the most natural check is to find out if it's not an object. Third party libs or change in how you initialize window.M might expose this line as not foolproof.

!== "object" would be a much safer check, +1 for that.

PR #1723, opened 2015-12-03
XW-94 prefetch optimizely

bognix:


rogatty:
Do we need all those commits? Would be nice to squash them :)

PR #1725, opened 2015-12-03
XW-767 Fix tracking for videos

rogatty:
https://wikia-inc.atlassian.net/browse/XW-767

Use standard tracking events for video tracking.


hakubo:
LGTM

PR #1727, opened 2015-12-04
XW | show notifications when there is space in article title

bognix:
@rogatty @perjg 

Details in the ticket: https://wikia-inc.atlassian.net/browse/INT-282

### DIFF
--- a/front/scripts/main/controllers/article-edit.js
+++ b/front/scripts/main/controllers/article-edit.js
@@ -27,7 +27,7 @@ export default App.ArticleEditController = Ember.Controller.extend({
 	handlePublishSuccess() {
 		let title = this.get('model.title');
 
-		if (title.match(/\s/)) {
+		if (title.indexOf(' ')) {
 			title = normalizeToUnderscore(title);
 		}
 
--- a/front/scripts/main/controllers/article-edit.js
+++ b/front/scripts/main/controllers/article-edit.js
@@ -1,6 +1,7 @@
 import App from '../app';
 import ArticleEditModel from '../models/article-edit';
 import {track, trackActions} from '../../mercury/utils/track';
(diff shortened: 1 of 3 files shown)


tomasznapieralski:
LGTM

vforge:
LGTM

rogatty:
LGTM

front/scripts/main/controllers/article-edit.js
```diff
@@ -24,7 +25,11 @@ export default App.ArticleEditController = Ember.Controller.extend({
 	 * @returns {void}
 	 */
 	handlePublishSuccess() {
-		const title = this.get('model.title');
+		let title = this.get('model.title');
+
+		if (title.match(/\s/)) {
```
Warkot: I was doing some research on this. I understand that checking here can save unnecessary running two global `replace`s in `normalizeToUnderscore` function. Good idea! :+1:

But here you actually don't need a match array which is returned by `match`, you only want to know an answer to a yes/no question. If you had a more complex RegEx, you could use [`test()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) for that. But you're looking for a simple string—a space character, so using `indexOf()` [is more reasonable](http://jsperf.com/exec-vs-match-vs-test-vs-search/99)—moreover, the further from the beginning of the string is the first space, the stronger it outperforms the RegExp methods.

vforge: :+1:

perjg:
@bognix Thanks for the fix. The same problem happens for add photo, so I pushed another commit that fixes the problem in the same way for add photo too.

bognix:
LGTM

vforge:
LGTM

hakubo:
LGTM me frenzy! :)

PR #1730, opened 2015-12-04
INT-183 | As a global Wikia user, I want to see the Wikia-in-your-lang features on mobile

mari-mcmurtrie:
We show "wikia in your language" alert if there is an equivalent wikia in user's language. For example, if a user's language is Japanese and he/she loads starwars.wikia.com (English wikia), then we show the alert saying "there is ja.starwars.wikia.com" <- not literally this text but similar.

This feature has been on for Japanese users only since this August. We want this to be available for all locales. Ticket: https://wikia-inc.atlassian.net/browse/INT-183


mari-mcmurtrie:
@perjg will you CR when you have chance? If you need what it is, then I will explain on Monday, thanks!

front/scripts/main/components/wikia-in-your-lang.js
```diff
@@ -85,9 +85,9 @@ export default App.WikiaInYourLangComponent = Ember.Component.extend(
 				 * 2,592,000,000 = 30 days
 				 */
 				notDismissed = !value || (now - value > 2592000000),
-				isJaOnNonJaWikia = this.get('isJapaneseBrowser') && !this.get('isJapaneseWikia');
+				isUserLangDifferentFromWikiLang = this.get('getBrowserLanguage') !== Ember.get(Mercury, 'wiki.language.content');
```
hakubo: I can't find `getBrowserLanguage` property
also I would name it
`browserLanguage`

hakubo: oh, that probably comes from LanguagesMixin

hakubo:
apart from on property name, LGTM

front/scripts/main/components/wikia-in-your-lang.js
rogatty: `getBrowserLanguage` is a function, not property, so it should be called by `this.getBrowserLanguage()`
https://github.com/Wikia/mercury/blob/dev/front/scripts/main/mixins/languages.js#L25

mari-mcmurtrie:
@perjg Added condition to this ticket. Originally, we opened this feature to all locales but now we limit to ['zh', 'ko', 'vi', 'ru', 'ja'];. Will you CR? Thanks!

front/scripts/main/components/wikia-in-your-lang.js
```diff
@@ -84,9 +84,23 @@ export default Ember.Component.extend(
...
 				notDismissed = !value || (now - value > 2592000000),
-				isJaOnNonJaWikia = this.get('isJapaneseBrowser') && !this.get('isJapaneseWikia');
+				notSameLanguage = this.isUserLangDifferentFromWikiLang();
 
-			return notDismissed && isJaOnNonJaWikia;
+			return notDismissed && notSameLanguage;
 		},
+
+		/**
+		 * @return {integer}
```
hakubo: looks more like a boolean

mari-mcmurtrie: Sorry, yes. I had this function returning int initially, and then I changed the func but not return data type. Thanks for checking.

PR #1731, opened 2015-12-07
pre-git installation as dev-dependency

hakubo:
You can see documentation here: https://www.npmjs.com/package/pre-git

For now I have disabled commitizen, but that could nicely enforce commit messages like
`XW-XXX | message with ease`

This way we can be sure'er that our code is more stable.

@Wikia/x-wing @kvas-damian @mari-mcmurtrie what do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "pre-git": {
       "commit-msg": "validate-commit-msg",
       "pre-commit": [
+        "npm run linter",
         "npm run test-front",
         "npm run test-server"
       ],
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
   },
   "config": {
     "pre-git": {
-      "commit-msg": "validate-commit-msg",
       "pre-commit": [
         "npm run linter",
         "npm run test-front",
@@ -95,7 +94,8 @@
       ],
       "post-commit": [],
       "post-merge": [
-        "npm update"
+        "npm update",
+        "bower update"
       ]
     }
   }
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "linter": "eslint .",
     "build": "rm -rf www && gulp build --type prod --nosync",
     "build-coverage": "rm -rf www && gulp build --nosync --testing",
-    "build-test": "rm -rf www && gulp build --type prod --nosync --testing"
+    "build-test": "rm -rf www && gulp build --type prod --nosync --testing",
+    "commit": "commit-wizard"
   },
   "repository": "git://github.com/Wikia/mercury.git",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


bognix:
LGTM, only one side note. With comitizen enabled you wouldn't commit this changes because you don't have ticket, hence there is no way you would create message like `XW-XXX | ...`

And one more question, is it going to run `bower/npm update` after every merge or do I have to run some special command?

rafalkalinski:
Hi,
Why do we need this for each commit? Isn't enough to run the linter on PRs?

bkoval:
Is it really needed to run the tests after every single commit? I realize this will enforce developers to write self-contained, non-breaking commits, but this might:
 - end up in extensive use of -n option for committing, which feels like it's going to be the case for small commits
 - force developers to use the same environment for committing and maintaining Mercury (personally, I'm using git on my local machine, which I keep in sync with devbox running Mercury - this allows me to use full power of my IDE and yet keep Mercury environment in sync with the newest dev machines setup)

vforge:
LGTM, neat idea!

@bognix we can always create some messages like literally `XW-XXX | Hotfix blahblablah` ;)

package.json
```diff
@@ -14,7 +14,8 @@
     "linter": "eslint .",
     "build": "rm -rf www && gulp build --type prod --nosync",
     "build-coverage": "rm -rf www && gulp build --nosync --testing",
-    "build-test": "rm -rf www && gulp build --type prod --nosync --testing"
+    "build-test": "rm -rf www && gulp build --type prod --nosync --testing",
+    "commit": "commit-wizard"
```
vforge: Is that needed anymore?

bognix:
@v but this would enforce to create ticket for everything. For most of the time it's good, but sometimes it's worth just to code what you have in your mind without any additional "paper work" - for instance something like this PR.

I'm with Bart when it comes to development process. I don't run `bower/npm update` on my local machine and I don't want it to run on every merge.

I have ambivalent feelings about this ticket. Some might find this useful but I'm not sure if we should force it.

kvas-damian:
@bkoval +1
I'm using git from my laptop and server tests are failing there for some reason...

tomasznapieralski:
Instead of adding this to `package.json` i propose using git hooks.
We can define default commit message template, to start with branch name followed by |
i.e. while typing `git commit` on `XW-321` branch you will see:

```
XW-312 | 

# Add commit messege after pipe above
```
What do you think?

vforge:
@bogna I was thinking of commit named `XW-XXX | Lorem ipsum` (with exes); but right now I'm thinking about `XW | Lorem ipsum` just to put Team's name in it.

PR #1737, opened 2015-12-08
INT-286 | Adding a missing template file so that wikia-in-your-lang pop up works

mari-mcmurtrie:
Current Wikia-in-your-lang is not working on Mercury because of this missing template file. We want to release the fix asap. The fix was originally implemented as a part of https://github.com/Wikia/mercury/pull/1730 (INT-183) but I separate a release so this alone can go out asap.


mari-mcmurtrie:
@perjg will you CR? We want to release this before code freeze which is 12/10.

perjg:
LGTM

PR #1738, opened 2015-12-08
INT-285 | Enable optimizely for Japan Homepage

perjg:
Added Optimizely script.

According to the Optimizely docs, the script should be embedded in <head>
It requires jquery to be included first, so I removed jquery from the combined.js script and include it manually before the optimizely script.

https://wikia-inc.atlassian.net/browse/INT-285


tomasznapieralski:
@perjg this PR related to https://wikia-inc.atlassian.net/browse/INT-285 ?

perjg:
@tomasznapieralski Yes, this is a PR for https://wikia-inc.atlassian.net/browse/INT-285.

PR #1742, opened 2015-12-08
XW-144 | Improve critical CSS

rogatty:
https://wikia-inc.atlassian.net/browse/XW-144

Load critical CSS as a blocking resource and then load the rest as a non blocking one.


front/styles/main/theme/index.scss
```diff
@@ -1,3 +1,4 @@
-@import 'icons';
-@import 'overrides';
+//@import 'icons';
+//@import 'overrides';
+@import 'dark';
```
rogatty: Nope, this is loaded lazily in mixins/theme.js

front/styles/main/app.scss
```diff
@@ -1,12 +1,7 @@
-// DO NOT CHANGE:
-// load order of files below is intentional
-@import '../../vendor/wikia-style-guide/src/scss/main';
-@import 'settings';
-@import 'foundation-imports';
-@import 'wikia-variables';
-@import 'mixin/index';
+@import 'shared_imports';
```
rogatty: `shared-imports`

front/styles/main/component/critical.scss
```diff
@@ -0,0 +1,3 @@
+@import 'site-head';
```
rogatty: This is here and in index.scss, should stay here only.

front/styles/main/module/index.scss
```diff
@@ -22,4 +22,5 @@
 @import 'site-footer';
 @import 'wikia-map';
 @import 'wikia-widget';
-@import 'wiki/index';
+//@import 'wiki/index';
+@import 'wiki/login';
```
rogatty: What do we do with wiki/index.scss? Right now it's not used but still exists. Maybe it should consist of this single line?

vforge: `wiki/index.scss` has been removed

front/styles/main/component/critical.scss
vforge: Imports has been fixed (share-header extends site-head)

PR #1739, opened 2015-12-08
SOC-1505 | Displaying message to blocked user when opening editor

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-1505

@kvas-damian @rafalkalinski 

Targeting SOC-1739 as it contains needed fixes for editor behavior.

PR #1743, opened 2015-12-08
SOC-1505 | Displaying message to blocked user when opening editor

bkoval:
Original pull request - https://github.com/Wikia/mercury/pull/1739

https://wikia-inc.atlassian.net/browse/SOC-1505


front/scripts/main/models/discussion-forum.js
```diff
@@ -7,6 +7,7 @@ export default App.DiscussionForumModel = DiscussionBaseModel.extend(DiscussionD
 	pageNum: null,
 	posts: null,
 	totalPosts: 0,
+	isRequesterBlocked: true,
```
kvas-damian: Do you think we should assume that user is blocked by default?

rafalkalinski: +1 for the question

bkoval: Yeah. If we don't get the sufficient response from the API we'll redirect to the error page with no editor anyway. Assuming false-negative in this case should be safe then.

kvas-damian:
LGTM

rafalkalinski:
LGTM

rafalkalinski:
LGTM

rafalkalinski:
https://github.com/Wikia/mercury/pull/1794

PR #1746, opened 2015-12-08
XW | Environment Fix

bognix:
Correct environment value returned on sandbox
```
>>> M.prop('environment');
>>> "sandbox"
```

@rogatty @vforge


vforge:
LGTM

PR #1748, opened 2015-12-08
DAT-3555 | set max-width to arbitrary value

idradm:
https://wikia-inc.atlassian.net/browse/DAT-3555

### DIFF
--- a/front/styles/main/module/article/_portable-table.scss
+++ b/front/styles/main/module/article/_portable-table.scss
@@ -1,6 +1,6 @@
 $portable-table-cell-horizontal-padding: 56px;
 $portable-table-cell-vertical-padding: 12px;
-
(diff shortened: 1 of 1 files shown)


front/styles/main/module/article/_portable-table.scss
```diff
@@ -28,6 +28,11 @@ $portable-table-cell-vertical-padding: 12px;
 table[data-portable="true"] {
 	font-size: $type-smallest;
 
+	// overrides default img styles for proper rendering inside tables
+	img {
+		max-width: $portable-table-cell-horizontal-padding * 3;
```
SebastianMarzjan: Given the fact that it's a product of a var and an arbitrary multiplier, I suggest moving it to a separate var at the top

RafLeszczynski:
LGTM

PR #1750, opened 2015-12-09
INT-294 | Update text for Star Wars campaign

perjg:


perjg:
Merging this minor content change immediately.

PR #1751, opened 2015-12-09
INT-284 | Switched from nodemon to gulp-develop server for Japan HP

perjg:
Originally used nodemon to monitor and restart the node process on file changes for Japan HP. At some point it started to malfunction causing infinite restarts (likely due to a nodemon version mismatch).

To work around this, I switched to using gulp-develop-server instead. This should match the main Mercury app better too.

https://wikia-inc.atlassian.net/browse/INT-284


homepage/gulp/tasks/watch.js
```diff
@@ -25,13 +21,12 @@ gulp.task('watch', ['build-combined'], function () {
...
-		ext: 'js html',
-		env: { 'NODE_ENV': options.env },
-		tasks: ['lint'],
-	}).on('start', function () {
-		reload(path);
+	server.listen({ path: paths.server.homepage.script });
+
+	// Restart server
+	gulp.watch(paths.server.homepage.watch, function (event) {
+		console.log('Restarting server due to file change: ' +  event.path);
```
mari-mcmurtrie: Do you want to console.log() ? Is this server-side code?

mari-mcmurtrie:
Not sure about console.log but other than that, it is LGTM.

perjg:
This is for the gulp tasks that run when you do 'npm run dev' - so the console.log is there to show a notification whenever the server side code is automatically restarted.

PR #1769, opened 2015-12-10
INT-279: JP homepage - login icon missing

mari-mcmurtrie:
I found that there is an extra "/" in request to check user login status. The URL was "https://services.wikia.com/helios//info?code=..." (notice "//" before "info"). I am not 100% sure if this is a cause of failure because I cannot hit the URL from my development environment: I always got 403 Forbidden no matter what. I talked to service team and they say the given url is now only allowed internally, that I am still investigating what it means "internal". 

@perjg how do we test Japan HP, given that URL to hit helios is production URL (no switch between dev or production). Do you stage this in sandbox?

### DIFF
--- a/homepage/config/localSettings.base.js
+++ b/homepage/config/localSettings.base.js
@@ -8,9 +8,15 @@ var deepExtend = require('deep-extend'),
 		authCookieDomain: 'wikia.com',
 		// Targeted environment [prod|preview|verify|dev|testing]
 		//environment: Utils.getEnvironment(process.env.WIKIA_ENVIRONMENT),
-		// Login
+
 		helios: {
-			host: process.env.HELIOS_HOST,
+			path: '/auth',
+			usernameMaxLength: 50,
+			passwordMaxLength: 50
+		},
+		whoAmIService: {
+			path: '/whoami',
+			timeout: 3000
 		},
 		host: process.env.HOST,
 		logger: 'syslog',
--- a/homepage/front/styles/theme/_icons.scss
+++ b/homepage/front/styles/theme/_icons.scss
@@ -1,16 +1,17 @@
 .icon {
-	$nav-icon-size: rem-calc(45);
+	$nav-icon-size: rem-calc(28);
 
 	&.avatar {
+		border: #c7d1d8 1px solid;
 		border-radius: 50%;
 		height: $nav-icon-size;
-		padding: rem-calc(6);
+		margin-top: 3px;
 		width: $nav-icon-size;
 	}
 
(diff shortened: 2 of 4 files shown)


tomasznapieralski:
LGTM

perjg:
The extra '/' was put in there to handle this.baseUrl both with and without trailing '/' (it is set from an environment variable).

```
var url = require('url')

var test1 = 'http://www.ulv.no/test';
var test2 = 'http://www.ulv.no/test/';

url.resolve(test1 + '/', 'hello')
 -> 'http://www.ulv.no/test/hello'

url.resolve(test1, 'hello')
 -> 'http://www.ulv.no/hello'

url.resolve(test2 + '/', 'hello')
 -> 'http://www.ulv.no/test//hello'
url.resolve(test2, 'hello')
 -> 'http://www.ulv.no/test/hello'
```

If this is the cause of the problem, I would expect it to always fail, but the issue shows up only for some users and not all the time.

mari-mcmurtrie:
Unfortunately, this did not fix the issue. I am still developing.

kvas-damian:
@mari-mcmurtrie It looks that the resource you're trying to use was removed.

See: https://wikia-inc.atlassian.net/browse/SERVICES-581

BTW for access_token validation it's recommended to use whoAmI service.

perjg:
@kvas-damian Do you know where I can find information on the whoAmI service?

kvas-damian:
When I was switching mercury to WhoAmI swagger documentation was broken. You can follow our implementation here: https://github.com/Wikia/mercury/blob/d315857546dbc9d61c134e63a2221633b49fa0fd/server/lib/WikiaSession.js#L57

perjg:
@mari-mcmurtrie I've deployed your change to sandbox-japan (http://sandbox-japan.ja.wikia.com/) and it works well.

One minor thing is that your code uses the Wreck library for the GET request, but the rest of the code uses the request library. It doesn't affect functionality, but for consistency we should use a single library.

To fix that, I pushed a new commit that changes back to using the request library. I also added one extra minor fix where the default avatar icon is used as a fallback should any of the api calls fail for any reason.

Please have a look at my new commit.

mari-mcmurtrie:
Thanks @perjg . I did not know how to set cookie with existing library. Thanks for fixing it. I think default icon url was never set so I set it. I stage this in preview again and check.

perjg:
@mari-mcmurtrie LGTM on your latest fixes.
I checked on preview.ja.wikia.com, and it works well.

PR #1778, opened 2015-12-15
INT-206: Re-style search box in top nav for Japan HP

mari-mcmurtrie:
1) change the text to say "すべてのコミュニティを検索"
2) fix the distorted search icon (looks stretched)
3) center both in the middle of the search box (currently it looks like it's lower than the center)

Ticket: https://wikia-inc.atlassian.net/browse/INT-206


perjg:
LGTM

PR #1781, opened 2015-12-15
XW | change systemjs/builder version to 0.14.11

bognix:
Fixing this issue - https://github.com/systemjs/builder/issues/386 - broke our code.
It changed order of loading our dependencies (app was loaded after model/user.js and CurrentUser, instead being loaded before them)


tomasznapieralski:
LGTM

rogatty:
LGTM

PR #1782, opened 2015-12-15
VE-2116: Change the usage of addmediatemporary and addmediapermanent APIs

inez:
Specifically this change is about retrieving token first and then passing it to the API call. It's going to be required after merging: https://github.com/Wikia/app/pull/9315

This change can be merged and released before the other one is - and that's the right way of doing it.


inez:
@mari-mcmurtrie I think you worked on this feature originally - so you might want to take a look as well.

inez:
@Grunny Or maybe you want to take a look? :)

Grunny:
LGTM.

pauloslund:
:+1:

front/scripts/main/models/article-add-photo.js
```diff
@@ -145,30 +145,34 @@ App.ArticleAddPhotoModel.reopenClass(
...
-						if (resp && resp.addmediapermanent) {
-							resolve(resp.addmediapermanent);
-						} else if (resp && resp.error) {
-							reject(resp.error.code);
-						} else {
-							reject();
-						}
-					},
-					error: (err) => reject(err)
+			return this.getEditToken(title)
```
bognix: I'm not sure if it's the best idea to call instance method (`this.editToken`) from within static function. Even though it works right now it may cause some problems in the future (in case Ember changes the way how objects are created and stored in memory). I would consider creating plain ES6 module with getEditToken function exported and call it from within this model. I would suggest something like utils dir (check out http://ember-cli.com/user-guide/#module-directory-naming-structure)

inez: @bognix Are you sure that permanentUpload is a static method?

rogatty: It's not static per se but it's defined inside `reopenClass` which makes it a class method, not an instance one. `this.editToken` probably works only because we use a mixin (I'm not sure how exactly it works, but as you can see it's messy).
Moving `getEditToken` to a regular module inside /front/scripts/main/utils instead of a mixin would make this situation more straightforward.

inez: I agree. It will be nice to have a "general" util method for retrieving token.

inez:
Hey. This is ready for another round of code review.

front/scripts/main/utils/edit-token.js
```diff
@@ -1,10 +1,10 @@
-export default Ember.Mixin.create({
+export default {
 
 	/**
 	 * @param {string} title
 	 * @returns {Ember.RSVP.Promise}
 	 */
-	getEditToken(title) {
+	getToken(title) {
```
rogatty: This is the only method here so how about `export default function (title) {`? Then you would import it as
```
import getEditToken from '../utils/edit-token'
```

rogatty:
One suggestion, but it LGTM

PR #1783, opened 2015-12-16
Japan HP | Removed tslint task and references

perjg:


perjg:
Merging immediately as this is needed to test INT-206 fix

vforge:
LGTM. Sorry for not removing/noticing that when working on ES6 for homepage with you.

PR #1786, opened 2015-12-16
SOC- remove crumb

kvas-damian:
For better caching let's remove crumb CSRF protection.

We shouldn't pass CSRF tokens in cookies because:
* we break caching assets - ```Vary: cookie```
* users can't use application in many tabs at once

To solve that we should use hidden input in forms.


kvas-damian:
Ping @hakubo @vforge @rogatty

vforge:
LGTM

rafalkalinski:
LGTM

PR #1796, opened 2015-12-17
XW-121 | Used fixed version of dependencies

bognix:
PR to release: https://github.com/Wikia/mercury/pull/1800


tomasznapieralski:
LGTM :+1:

rogatty:
LGTM

hakubo:
LGTM - it would be cool to automate checking it, so we don't add back again ~ by accident

hakubo:
hmm that should work: `cat package.json | jq '.devDependencies[]' -Mr | grep '[x\^~]' > /dev/null`
:)

bognix:
@hakubo - not exactly, it would fail if we have something like Wikia/x#v2.0.0 (because you are testing against "x" presence

PR #1802, opened 2015-12-18
SOC-1763 | Prevented opening links in discussion posts from forum view

Bloodmoon24:
https://wikia-inc.atlassian.net/browse/SOC-1763
@kvas-damian @bkoval please CR

### DIFF
--- a/front/scripts/main/components/post-detail.js
+++ b/front/scripts/main/components/post-detail.js
@@ -45,7 +45,7 @@ export default Ember.Component.extend(
 		actions: {
 			/**
 			 * @param {number} postId
-			 * @param {object} event
+			 * @param {MouseEvent} event
 			 * @returns {void}
 			 */
 			goToPost(postId, event) {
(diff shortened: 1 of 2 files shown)


front/scripts/main/components/post-detail.js
```diff
@@ -49,6 +49,7 @@ export default Ember.Component.extend(
 			 * @returns {void}
 			 */
 			goToPost(postId, event = {}) {
+				event.preventDefault();
```
bkoval: This triggers "Uncaught SyntaxError: Unexpected token .(…)" when event is an empty object (as in default parameter).  Maybe check for if (typeof event.preventDefault === 'function')?

Bloodmoon24: After analysis and tests there is no more need of ``` event = {} ``` All using cases are passing event to action, so that if is no needed. Thanks for hint!

bkoval:
LGTM

front/scripts/main/components/post-detail.js
```diff
@@ -48,7 +48,8 @@ export default Ember.Component.extend(
 			 * @param {object} event
```
rogatty: It's `MouseEvent` type, right? https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

rogatty:
LGTM

PR #1806, opened 2015-12-21
INT-305 | New extra small screen and medium screen breakpoints for he…

perjg:
…ro slider

Various adjustments to make the hero slider look better on small (<380px screens) and medium (615-1000px screens)


mari-mcmurtrie:
LGTM

bkoval:
$small-range: (0px, 320px);
$medium-range: (321px, 767px);
$large-range: (768px, 1063px);
$xlarge-range: (1064px, 1574px); // Desktop
$xxlarge-range: (1575px, 99999px); // Desktop XL

These are the values that we use in the style-guide for breakpoints. I believe we should be as consistent between experiences as possible in terms of such values. Would it be possible to use these? I realize I might be wrong suggesting this if your specs explicitly mention values used in the pull request :)

perjg:
@bkoval Our design only covers desktop ($large-range and up) and mobile.

This specific set of breakpoints were added ad-hoc to make the top hero slider look good on any screen size. They are only used for layout of the hero slider, and should be kept with these particular pixel values.

For more general breakpoints, we will use the style guide.

PR #1807, opened 2015-12-21
DAT-3085 | do not collapse/uncollapse infobox when click target is an img

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3085

### DIFF
--- a/front/scripts/main/components/portable-infobox.js
+++ b/front/scripts/main/components/portable-infobox.js
@@ -70,8 +70,6 @@ export default Ember.Component.extend(
 			const collapsed = this.get('collapsed'),
 				$target = $(event.target);
 
-			console.log("target: ", $target)
-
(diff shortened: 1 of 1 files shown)


tomasznapieralski:
LGTM

front/scripts/main/components/portable-infobox.js
```diff
@@ -68,7 +70,7 @@ export default Ember.Component.extend(
 			const collapsed = this.get('collapsed'),
 				$target = $(event.target);
 
-			if ($target.is('a') || $target.is('button')) {
+			if ($target.is('a') || $target.is('button') || $target.is('img')) {
```
SebastianMarzjan: Would it make sense to define an 'excluded' tag array at this point and use it? This condition is becoming long :)

PR #1808, opened 2015-12-21
gitignore | Adding coverage/ to gitignore

bkoval:
@rogatty @hakubo @Warkot Seems like this directory is making our life harder :)


rogatty:
Where did it come from? Coverage should be in test/coverage dir. LGTM anyway as it doesn't break anything.

tomasznapieralski:
LGTM

PR #1811, opened 2015-12-22
DAT-3085 | add shouldHandleClick()

dianafa:
Fix for #1807 

https://wikia-inc.atlassian.net/browse/DAT-3085

### DIFF
--- a/front/scripts/main/components/portable-infobox.js
+++ b/front/scripts/main/components/portable-infobox.js
@@ -69,12 +69,9 @@ export default Ember.Component.extend(
 		 */
 		onInfoboxClick(event) {
 			const collapsed = this.get('collapsed'),
-				clickableElements = this.get('clickableElements'),
-				$target = $(event.target);
+				clickableElements = this.get('clickableElements');
 
-			if ($.inArray($target, clickableElements)) {
-				return;
-			}
+			if (!this.shouldHandleClick($(event.target))) return;
 
 			if (!collapsed) {
 				const body = window.document.body,
@@ -86,6 +83,20 @@ export default Ember.Component.extend(
 				this.set('collapsed', false);
 				this.$().height('auto');
 			}
+		},
+
+		shouldHandleClick($target) {
+			const clickableElements = this.get('clickableElements');
+			var shouldHandleClick = true;
+
+			clickableElements.forEach((element) => {
+				if ($target.is(element)) {
+					shouldHandleClick = false;
+					return;
+				}
+			});
+
+			return shouldHandleClick;
 		},
 
 		/**
--- a/front/scripts/main/components/portable-infobox.js
+++ b/front/scripts/main/components/portable-infobox.js
@@ -87,8 +87,8 @@ export default Ember.Component.extend(
 		},
 
 		/**
-		 * If element is a link, button or image,
-		 * should return false.
+		 * If element is one of clickableElements, collapsing of infobox should not be handled.
+		 * As this element has it's own action, not connected to collapsing/uncollapsing infobox.
 		 *
 		 * @param {JQuery} $target
 		 * @returns {bool}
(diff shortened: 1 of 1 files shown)


front/scripts/main/components/portable-infobox.js
```diff
@@ -69,12 +69,9 @@ export default Ember.Component.extend(
...
 		onInfoboxClick(event) {
 			const collapsed = this.get('collapsed'),
-				clickableElements = this.get('clickableElements'),
-				$target = $(event.target);
+				clickableElements = this.get('clickableElements');
 
-			if ($.inArray($target, clickableElements)) {
-				return;
-			}
+			if (!this.shouldHandleClick($(event.target))) return;
```
tomasznapieralski: please wrap `return` with curly brackets and move it to new line

rogatty: @tomasznapieralski ćśś, we have eslinter saying those things now :)

dianafa: yep, fixing :see_no_evil:

tomasznapieralski: :D

front/scripts/main/components/portable-infobox.js
```diff
@@ -88,6 +85,20 @@ export default Ember.Component.extend(
 			}
 		},
 
+		shouldHandleClick($target) {
+			const clickableElements = this.get('clickableElements');
+			var shouldHandleClick = true;
+
+			clickableElements.forEach((element) => {
+				if ($target.is(element)) {
+					shouldHandleClick = false;
```
rogatty: I don't get it. If the target is one of `clickableElements` then we return `false` from `shouldHandleClick`? Maybe we could change the naming to be more clear.

tomasznapieralski:
please make sure that all selenium-test in `MercuryPortableInfoboxTests` are passing

front/scripts/main/components/portable-infobox.js
```diff
@@ -88,6 +86,20 @@ export default Ember.Component.extend(
 			}
 		},
 
+		shouldHandleClick($target) {
+			const clickableElements = this.get('clickableElements');
+			let shouldHandleClick = true;
+
+			clickableElements.forEach((element) => {
+				if ($target.is(element)) {
+					shouldHandleClick = false;
```
rogatty: I don't get it. If the target is one of `clickableElements` then we return `false` from `shouldHandleClick`? Maybe we could change the naming to be more clear.

dianafa: They're `clickable` because when you click on them something happens - or lightbox appear, or we go to a link etc.
But gotcha, good point. What about `shouldHandleCollapsing()`?

rogatty: Yup, much better!

front/scripts/main/components/portable-infobox.js
```diff
@@ -88,6 +86,20 @@ export default Ember.Component.extend(
...
+					return;
```
rogatty: If you try to break from `forEach` loop here then it won't work, see http://stackoverflow.com/questions/2641347/how-to-short-circuit-array-foreach-like-calling-break. Probably `some` would fit here better.

tomasznapieralski:
LGTM

front/scripts/main/components/portable-infobox.js
```diff
@@ -89,6 +87,21 @@ export default Ember.Component.extend(
 		},
 
 		/**
+		 * If element is a link, button or image,
```
rogatty: Will everyone remember to update this comment when changing `this.clickableElements`? Probably not ;) IMO it would make more sense to explain WHY we need this function, not what it does - it's already visible in the code.

PR #1812, opened 2015-12-23
XW-879 | Introduce Ember CLI

hakubo:
https://wikia-inc.atlassian.net/browse/XW-879


front/.gitignore
```diff
@@ -0,0 +1,2 @@
+# compiled output
+vendor/
```
rogatty: There is no `front/vendor`.

front/common/gulpfile.js
```diff
@@ -0,0 +1,122 @@
...
+	svgSymbols = require('gulp-svg-symbols'),
+	compile = require('../../gulp/utils/compile-es6-modules'),
+	environment = require('../../gulp/utils/environment'),
+	log = require('../../gulp/utils/logger'),
+	piper = require('../../gulp/utils/piper'),
+	paths = require('../../gulp/paths'),
+	pathsCommon = paths.common,
+	Server = require('karma').Server;
+
+gulp.task('test-common', ['build-common-vendor'], function (done) {
```
rogatty: @tomasznapieralski why does it depend on `build-common-vendor` but not on `build-common`?

tomasznapieralski: it uses some dependencies like jquery only during tests.
you need to run tests on already builded app

hakubo:
yay!

PR #1814, opened 2015-12-29
Update handlebars to 4.0.5 and hapi to 11.1.4

paladox:
Reasons because of security problems described at https://david-dm.org/Wikia/mercury


rogatty:
Thanks, we will have to do some testing before merging this.

hakubo:
hapi won't just work with no changes, we need to do a bit more than that. Thanks anyway!

PR #1821, opened 2016-01-07
INT-277 | Rename logout handler to signout

perjg:
To work around the fact that the /logout route has special handling globally, change the logout for Japan HP to /signout instead.

Note that INT-299 depends on this fix, so right now if you log out it will redirect to / but will still show your status as logged in.

### DIFF
--- a/homepage/server/routes.js
+++ b/homepage/server/routes.js
@@ -29,8 +29,8 @@ exports.routes = [
 	},
 	{
 		method: 'GET',
-		path: '/logout',
-		handler: require('./facets/logout')
+		path: '/signout',
+		handler: require('./facets/signout')
 	},
 	{
 		method: 'GET',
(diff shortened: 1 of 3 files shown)


homepage/server/routes.js
```diff
@@ -29,8 +29,8 @@ exports.routes = [
 	},
 	{
 		method: 'GET',
-		path: '/logout',
```
rogatty: If the homepage is cached, users will still get the old URL in HTML but there will be no route to handle it. If that's the case you could leave this route for now and remove it later when you're sure that the cache is expired. You probably can check somewhere in Kibana if there were any requests to this endpoint to be sure.

perjg: Good point, I added the logout route back in to use for a transition period.

mari-mcmurtrie:
LGTM

PR #1827, opened 2016-01-08
INT-299 | Invalidate cache when signing out

perjg:
When logging out from the Japan HP, the logged in UI state would remain for a few minutes due to caching. This fixes the issue by forcefully refreshing the cookie when logging out.

### DIFF
--- a/homepage/server/facets/signout.js
+++ b/homepage/server/facets/signout.js
@@ -6,6 +6,11 @@
 
 function signout(request, reply) {
 	reply.unstate('access_token');
+
+	if (request.response) {
+		// Make sure to invalidate cache
+		request.response.vary('cookie');
+	}
 	reply.redirect('/');
 }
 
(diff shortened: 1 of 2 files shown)


kvas-damian:
Current code won't invalidate cache - it just don't let browser and varnish cache this one response.

```Vary: Cookie``` header should be applied to every HTML response - not only for logout.

Example scenario:
1. User visits any page
2. This page gets cached on varnish and in the browser
3. User logs in
4. User visits page that was visited before. Because ```Vary: cookie``` header wasn't present before, browser will check it's or varnish cache and serve cached anon version of a page. Because of lack of that header browser and Varnish don't know that they should fetch fresh content when cookie was changed.

perjg:
@kvas-damian Good point.

I'll set this up as middleware for the html routes, sending it every time. I guess this means that the html parts will not be cached, but I'll make sure that the static assets do not have this header.

perjg:
@kvas-damian I've moved response.vary to onPreResponse, and send it for all requests that are not 'file', just like in the main mercury app.

kvas-damian:
LGTM

PR #1829, opened 2016-01-08
SOC-1483 | Displaying message on empty forums

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-1483

@kvas-damian @rafalkalinski @Bloodmoon24 

This will go through design review in the meantime as I didn't add a "plus" icon yet. I'll find out if we need it.


front/templates/main/components/forum-wrapper.hbs
```diff
@@ -17,18 +19,23 @@
 		replyCount=item.postCount
 		undeletePost=(action this.attrs.undeletePost)
 		upvote='upvote'
-	}}
-{{/each}}
-{{#if hasMore}}
-	{{#if isLoading }}
-		{{loading-spinner
+		}}
+	{{/each}}
```
rogatty: You can use:
```
{{#each posts as |item|}}
   ...
{{else}}
   ...
{{/each}}
```
to handle no posts. See http://handlebarsjs.com/builtin_helpers.html#iteration

bkoval: Oh, that's awesome, didn't know this would work! It's just that in my case I don't want to render anything below the each block as well :)

rogatty: If there are no posts then `hasMore` should be `false`, right?

bkoval: right!

rafalkalinski:
LGTM

PR #1833, opened 2016-01-12
SOC-1576 | remove active highlighting on all links

katanka:
https://wikia-inc.atlassian.net/browse/SOC-1576

### DIFF
--- a/front/styles/main/base/_anchors.scss
+++ b/front/styles/main/base/_anchors.scss
@@ -1,3 +1,0 @@
-a {
-	@extend .touch-active;
-}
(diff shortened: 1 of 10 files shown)


bkoval:
@rogatty @hakubo This will remove highlighting from all the links everywhere on Mercury. We do have a green light from UX to do so on Discussions, but the question is if we can remove it everywhere. It might make sense, since these things should be consistent throughout the platform, but it appears to be more of a product decision.

If not, we'll make sure to just touch discussions with these changes for now.

katanka:
@bkoval I confirmed with Derek Hart that it should be removed for all of Mercury.

bkoval:
OK, perfect!

rafalkalinski:
LGTM

hakubo:
do we have plans to replace that with something?

PR #1834, opened 2016-01-12
INT-293 | Remove code for Google Custom Search integration for Mercur…

perjg:
…y app

We've stopped the Google Custom Search A/B test for Mercury, and the current conclusion is that there is no significant performance difference between google and oasis search for Search in a single wikia.
(instead we are running a separate test for global wikia search: See IMNT-285)

If we do decide to add Google Custom Search for Mercury in the future, it will be via a Pandora service, so the current code is in either case obsolete.


perjg:
@rogatty @hakubo It would be great if one of you have time to check this out.

perjg:
The second linter fix (1a3baba) is in code I didn't touch in this commit, so I think it is an error that was in the code already.

As far as I can see shouldBeVisibleObserver is not used anywhere, so maybe that whole function should be removed too?

front/scripts/main/components/side-nav.js
```diff
@@ -85,30 +84,13 @@ export default Ember.Component.extend({
...
-				// Hide SideNav
-				this.sendAction('toggleVisibility', false);
-				this.send('searchCancel');
-
-				this.sendAction('search', value);
-			} else {
-				// Use Wikia Search
-				window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
-			}
+			// Use Oasis Search
```
rogatty: It doesn't have anything to do with Oasis skin. Let's stay with `Wikia Search`.

rogatty:
The second linter fix reverts what you touched in the previous one ;) `shouldBeVisibleObserver` is called when `shouldBeVisible` changes and that's how we do the tracking.

rogatty:
One minor thing and it'll LGTM.

perjg:
@rogatty Ah, must have been a bit to quick on the editing, I didn't notice that 'gLabel' sneaking in.

I'll change the comment to // Use Wikia Search and merge.

front/scripts/main/components/side-nav.js
```diff
@@ -85,30 +84,13 @@ export default Ember.Component.extend({
...
+			window.location.assign('%@Special:Search?search=%@&fulltext=Search'.fmt(Mercury.wiki.articlePath, value));
```
hakubo: while at it, could you update it to use string template?

hakubo:
ditto, one minor thing otherwise lgtm

perjg:
@hakubo @rogatty I changed to use a string template.

Is something wrong with the build machine?

/var/lib/jenkins/workspace/Mercury-UPDATE-node-modules-old/node_modules/gulp-sass/node_modules/node-sass/lib/index.js:22
    throw new Error('`libsass` bindings not found. Try reinstalling `node-sass

rogatty:
@perjg it was caused by the fact that node version was upgraded over the last night but on Jenkins we used packages built on the old version to run tests.

It's green now. LGTM :+1:

hakubo:
lgtm

perjg:
Cool. What node version are you running now?

On Friday, 15 January 2016, Jakub Olek <notifications@github.com> wrote:

> lgtm
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Wikia/mercury/pull/1834#issuecomment-171981006>.
>


-- 
Sincerely,
 Per Johan Groland
 Akafugu Corporation

Blog: http://www.akafugu.jp/blog/
Twitter: akafuguJP
Facebook: http://www.facebook.com/akafugu

rogatty:
It's 0.12.9, see https://github.com/Wikia/chef-repo/blob/master/cookbooks/_nodejs/attributes/default.rb

PR #1836, opened 2016-01-12
Maintenance | Removing unused getLanguage method

bkoval:
@rogatty @kvas-damian 

To avoid confusion, let's delete this method (since it's not used anywhere)


rogatty:
@mari-mcmurtrie mentioned in an email that she wants to use it. On the other side it doesn't work in its current state so it makes sense to remove it and if needed it can be brought back, this time fixed and with some unit tests.

LGTM

kvas-damian:
LGTM

PR #1845, opened 2016-01-12
SOC-1549 - fixed loading spinner in firefox

katanka:
`transform-origin` is buggy in firefox currently, so I removed it and wrapped the circle in a svg group that is translated, so there is no need for `transform-origin` on the circle.

https://wikia-inc.atlassian.net/browse/SOC-1549


rafalkalinski:
LGTM

bkoval:
LGTM!

PR #1854, opened 2016-01-14
INT-311 | Final assets and text for category sliders

perjg:
All final assets for the Japan HP category slider, and updated json resource file with text and urls.


hakubo:
LGTM

PR #1805, opened 2015-12-20
XW | Fix phantomas errors

vforge:
Fixes for Phantomas errors:
- [x] changes inside SVGs - replace stuff like `id="a"` with unique versions
- [x] run through SVG optimizer again
- [x] inline brackets.svg

/cc @rogatty @tomasznapieralski

PR #1863, opened 2016-01-15
XW | Bump gulp-svg-symbols, install gulp-imagemin for minimizing SVGs and fix three SVGs for phantomas test

vforge:
leftover after: https://github.com/Wikia/mercury/pull/1805

/cc @rogatty @hakubo


vforge:
As we're moving towards Ember CLI, this PR is no longer valid.

:cry:

PR #1869, opened 2016-01-19
WIP SOC-1719 - delete all posts by a user, front end

katanka:
https://wikia-inc.atlassian.net/browse/SOC-1719


front/main/app/components/discussion-dialog.js
```diff
@@ -18,6 +20,17 @@ export default Ember.Component.extend(LoginLinkMixin, {
...
 			this.get('modalDialogService').close();
+			if (this.get('modalDialogService').isConfirm) {
+				this.get('modalDialogService').confirmCallback(false);
+			}
+		},
+		/**
+		 * @returns {void}
+		 */
+		confirm() {
+			this.set('isDialogVisible', false);
```
bkoval: Wondering if you meant "isVisible" here? If not, is there a need to have another flag like this? :)

front/main/app/mixins/discussion-delete-model.js
```diff
@@ -24,6 +24,28 @@ export default Ember.Mixin.create({
 	},
 
 	/**
+	 * Delete post in service
```
bkoval: Wondering if we should do AJAX calls from mixins - this seems like a model-only thing (I don't think there's any situation where 2 models are sharing the same ajax call).

Also, if the method delets all posts, why does it need a post as a param? Or maybe it's still a work in progress here on an actual method body? :)

front/main/app/mixins/discussion-delete-route.js
```diff
@@ -29,6 +30,24 @@ export default Ember.Mixin.create({
 		},
 
 		/**
+		 * Pass post deletion to model
+		 * @param {object} post
+		 * @returns {void}
+		 */
+		deleteAllPosts(post) {
+			const loadingSpinnerContainer = this.getLoadingSpinnerContainer(post);
+			var self = this;
```
bkoval: Just a side note - you probably know that we're not using vars anymore - we use let for variables that will likely change their value and const for variables that won't be entirely reasigned anymore.

bkoval:
I think overall it lookes great, just a couple of question from me - pinging @kvas-damian and @rafalkalinski to take a look as well

front/main/app/templates/components/discussion-delete.hbs
```diff
@@ -9,4 +9,8 @@
 		{{svg 'delete' viewBox='0 0 16 19' class='icon delete'}}
 		<span>{{deleteMessage}}</span>
 	</a>
+	<a class="delete-link" {{action this.attrs.deleteAll post}}>
```
kvas-damian: According to designs this button should not be placed in pop-over.

front/main/app/templates/components/discussion-dialog.hbs
```diff
@@ -0,0 +1,18 @@
+{{#if isVisible}}
+	{{#modal-dialog
+		isVisible = isVisible
+	}}
+		<div class="modal-dialog-content">{{i18n dialogMessage ns='discussion'}}</div>
+		<footer>
+			{{#if isConfirm}}
+			 	<button class="active-element-theme-color" {{action 'confirm'}}>{{i18n 'main.modal-dialog-confirm' ns='discussion'}}</button>
+				<button class="active-element-theme-color" {{action 'close'}}>{{i18n 'main.modal-dialog-cancel' ns='discussion'}}</button>
```
kvas-damian: This button can be moved out from if block.

kvas-damian: Ohh it uses different message key... Ignore my last comment

front/main/app/components/discussion-dialog.js
```diff
@@ -18,6 +20,17 @@ export default Ember.Component.extend(LoginLinkMixin, {
 		close() {
 			this.set('isDialogVisible', false);
 			this.get('modalDialogService').close();
+			if (this.get('modalDialogService').isConfirm) {
```
hakubo: I think you can do: `this.get('modalDialogService.isConfirm')`

front/main/app/components/discussion-dialog.js
```diff
@@ -18,6 +20,17 @@ export default Ember.Component.extend(LoginLinkMixin, {
...
+				this.get('modalDialogService').confirmCallback(false);
+			}
+		},
+		/**
+		 * @returns {void}
+		 */
+		confirm() {
+			this.set('isDialogVisible', false);
+			this.get('modalDialogService').close();
```
hakubo: minor: you could create a local variable for the service here

PR #1871, opened 2016-01-19
XW | Don't use eslint warnings

rogatty:
There is no point in using warnings as we don't display them anywhere but in Jenkins log. Let's use errors instead.

/cc @vforge


vforge:
LGTM

PR #1875, opened 2016-01-19
Run nodemon after the first change

rogatty:
There is no need to run it before Ember is built as it's confusing and might lead to the error we sometimes see.

/cc @hakubo


hakubo:
in npm run start should we pass this as false?

hakubo:
LGTM

PR #1880, opened 2016-01-21
INT-320 | Implement loading indicator

perjg:
Fixes visual glitches when slick slider is loading, and implements svg loading indicator for the page.


perjg:
Merging tomorrow unless there are specific comments (for JP Homepage, does not affect Mercury)

PR #1881, opened 2016-01-21
INT-310 | Final text and assets for hero slider

perjg:
* Cherry-picked changes for INT-308 as they are needed for this CR
* Added final text and assets
* Minor visual adjustments
* Fix incorrect alignment for slider buttons


perjg:
Merging tomorrow unless there are specific comments (for JP Homepage, does not affect Mercury)

PR #1892, opened 2016-01-22
INT-321 | Replace community central and Wikia University images

perjg:


perjg:
Merging directly as this is just an asset change.

PR #1901, opened 2016-01-26
Click on clear should not close keyboard

hakubo:
- focus on input after clear search is clicked    
- click area have 30px on side-nav clear now

### DIFF
--- a/front/main/app/styles/component/_side-nav.scss
+++ b/front/main/app/styles/component/_side-nav.scss
@@ -42,27 +42,6 @@
 			top: rem-calc(15);
 		}
 
-		input,
-		input:focus {
-			background-color: $color-white;
-			border: 1px solid $search-box-border-color;
-			border-radius: $border-radius-global;
-			height: rem-calc(28);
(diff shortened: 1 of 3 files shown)


vforge:
LGTM

bognix:
LGTM

front/main/app/styles/component/_side-nav.scss
```diff
@@ -159,6 +138,27 @@
 	}
 }
 
+.side-search__input,
+.side-search__input:focus {
+	background-color: $color-white;
+	border: 1px solid $search-box-border-color;
+	border-radius: $border-radius-global;
+	height: rem-calc(28);
+	margin: rem-calc(8) rem-calc(13);
```
bkoval: Just a side question - why do you use rems for defining the box model? I was pretty sure we had a signal from the UX team that we should be using pixels for defining paddings / margins.

hakubo: this signal did not propagate here yet. It sounds reasonable, but I won't be changing it now.

PR #1903, opened 2016-01-26
XW-918 | Update article.js to get new API response; unify UA dimensions set up

vforge:
https://wikia-inc.atlassian.net/browse/XW-918

https://github.com/Wikia/app/pull/9554

### DIFF
--- a/front/common/modules/Trackers/UniversalAnalytics.js
+++ b/front/common/modules/Trackers/UniversalAnalytics.js
@@ -36,7 +36,7 @@ class UniversalAnalytics {
 		this.accountPrimary = 'primary';
 		this.accountSpecial = 'special';
 		this.accountAds = 'ads';
-		this.dimensionsUpdated = true;
+		this.dimensionsUpdated = false;
 
 		if (!UniversalAnalytics.dimensions.length) {
 			throw new Error(
@@ -69,14 +69,14 @@ class UniversalAnalytics {
 	 * @returns {void}
 	 */
 	updateDimensions() {
-		if (UniversalAnalytics.dimensionsUpdated) {
+		if (!UniversalAnalytics.dimensionsUpdated) {
 			this.tracked.forEach((account) =>
 				UniversalAnalytics.dimensions.forEach((dimension, idx) =>
 					ga(`${UniversalAnalytics.getPrefix(account)}set`,
 						`dimension${idx}`,
 						UniversalAnalytics.getDimension(idx))));
 
-			UniversalAnalytics.dimensionsUpdated = false;
+			UniversalAnalytics.dimensionsUpdated = true;
 		}
 	}
 
@@ -98,7 +98,7 @@ class UniversalAnalytics {
 			$.extend(UniversalAnalytics.dimensions, dimensions);
 		}
 
-		UniversalAnalytics.dimensionsUpdated = true;
+		UniversalAnalytics.dimensionsUpdated = false;
 
 		return true;
 	}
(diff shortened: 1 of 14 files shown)


front/common/modules/Trackers/UniversalAnalytics.js
```diff
@@ -36,6 +36,7 @@ class UniversalAnalytics {
 		this.accountPrimary = 'primary';
 		this.accountSpecial = 'special';
 		this.accountAds = 'ads';
+		this.dimensionsUpdated = false;
```
hakubo: why do we need this flag?

front/main/app/models/article.js
```diff
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import MediaModel from './media';
 import {normalizeToWhitespace} from 'common/utils/string';
+import UniversalAnalytics from 'common/modules/Trackers/UniversalAnalytics';
```
hakubo: This is weird that model knows about UA...
maybe we could move it at least to something like article-content component?

front/common/modules/Trackers/UniversalAnalytics.js
vforge: Timeline:
(1) dimensions are being set up in Ember initializer
(2) trackers are being initialized
(3) **later** article is fetched from api - and one dimension is set there
(4) track for page view *OR* event is being sent to the server

My changes:
A) Before my changes dimensions were set in (2), but right now we're waiting for (4)
B) Before my changes dimensions were set up multiple times, right now it's only each time we've changed `UniversalAnalytics.dimensions`, before track event - that's why we need a flag answering the quesion: "is anything changed and we need to re-set dimensions?"

front/main/app/models/article.js
vforge: I'm thinking that this solution is simplest one - we're stopped polluting M.prop with useless property and we're setting dimension only and directly when we need it.

vforge: And as we know, we'll improve tracking in the future, I think it's enough for time being.

front/main/app/instance-initializers/setup-tracking.js
```diff
@@ -46,7 +46,7 @@ export function initialize() {
 	// TODO: Krux segmenting not implemented in Mercury https://wikia-inc.atlassian.net/browse/HG-456
 	// ga(prefix + 'set', 'dimension16', getKruxSegment());
 	dimensions[17] = Mercury.wiki.vertical;
-	dimensions[19] = M.prop('article.type');
+	// dimension[19] - articleType is being set in models/article.js
```
rogatty: `articleType is being updated in models/article.js on every article change`?

front/main/app/models/article.js
```diff
@@ -206,13 +207,12 @@ ArticleModel.reopenClass({
 				articleProperties.mainPageData = data.mainPageData;
 				articleProperties.isCuratedMainPage = true;
 			}
+
+			if (data.articleType) {
+				UniversalAnalytics.setDimension(19, articleProperties.articleType);
```
rogatty: Remember about the cache. Let's have a fallback for `articleProperties.type` with a ticket number to clean it up.

front/main/app/models/article.js
rogatty: I agree with @hakubo. This should be done in `components/article-wrapper.js#performArticleTransforms()` where the rest of tracking logic is. Also please use this opportunity to rename this function to something more appropriate ;)

No need to use `M.prop`.

front/main/app/instance-initializers/setup-tracking.js
vforge: Yes.

hakubo:
LGTM

vforge:
@nandy-andy can you LGTM this change: https://github.com/Wikia/mercury/commit/5a5fe03a5c04d7f709f866bb645df4cfaa3a1d69 ?

PR #1904, opened 2016-01-26
XW-1019 | scroll back to top while navigating through local nav

Warkot:
https://wikia-inc.atlassian.net/browse/XW-1019

/cc @Wikia/x-wing

### DIFF
--- a/front/main/app/components/side-nav-local-navigation-root.js
+++ b/front/main/app/components/side-nav-local-navigation-root.js
@@ -19,6 +19,7 @@ export default Ember.Component.extend({
 
 		updateContent(newLocalNavContent) {
 			this.set('localNavContent', newLocalNavContent);
+			this.scrollToTop();
 		},
 
 		goBack() {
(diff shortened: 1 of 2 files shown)


Warkot:
@bognix I made some changes. However, I don't think we need to bubble/handle the `scrollToTop()` action higher, as there we change the whole component, which makes it scroll by itself. If global nav would have more levels, then we could think of that.

hakubo:
lgtm

bognix:
it does look beautiful! LGTM! Although I'm a little bit worried about the future-proofing. When you pass the class to the component helper and then inside one of components you add something to http://emberjs.com/api/classes/Ember.Component.html#property_classNames what would happen? Would the class be overridden? Would it be appended? Could you check it?

hakubo:
It should be merged.

hakubo:
https://ember-twiddle.com/2e9fd327dcbf033e1798?openFiles=my-comp.component.js%2Cmy-comp.template.hbs

Funny thing is I don't know how to get that from component/template :)
but when you inspect it is merged

rogatty:
LGTM

PR #1905, opened 2016-01-27
SOC-1798 | Fixing bad behavior of ellipses by setting a different editor placeholder height

katanka:
When the post editor is stickied, the clamp class was not rendering properly, so I force it to re-render.

https://wikia-inc.atlassian.net/browse/SOC-1798


front/main/app/components/discussion-editor.js
```diff
@@ -91,6 +93,14 @@ export default Ember.Component.extend(ViewportMixin, {
 	},
 
 	/**
+	 * Re-render post content to make sure ellipses appear in the correct position.
+	 * @returns {void}
+	 */
+	updateEllipses() {
+		$('.post-details-link').hide().show(0);
```
rafalkalinski: So this is a  hack for ellipsis for Chrome only. 
We are using -webkit-line-clamp css property for this truncation - and if it's buggy, my call would be to use current IE/Firefox solution for all the browsers (it would be enough to remove the condition) 
https://github.com/Wikia/mercury/blob/dev/front/main/app/mixins/discussion-parsed-content.js#L15

bkoval:
@katanka That's a cool idea and it can definitely help. Unfortunately, it might generate a bit of performance issues: a handler (updateEllipses) will cause the whole render engine to run twice (after hiding and showing the post details link, which might even be two DOM elements).

Now, this method will run and repaint the view twice each time our throttled scroll event fires and the conditions inside are met.

It seems like this could cause a serious downgrade in performance of the post list page on scroll.

rafalkalinski:
LGTM

bkoval:
LGTM, thanks @katanka !

rogatty:
LGTM but PR title seems inaccurate.

PR #1907, opened 2016-01-27
Fix styles when running Ember tests in a browser

rogatty:
Those styles are not needed for tests and they break some colors.


vforge:
LGTM

PR #1908, opened 2016-01-27
XW-905 | Upgrade New Relic agent to v1.25.1

rogatty:
https://wikia-inc.atlassian.net/browse/XW-905
https://github.com/newrelic/node-newrelic/blob/master/NEWS.md


rogatty:
Based off wrong branch :facepunch:

PR #1909, opened 2016-01-27
XW-905 | Upgrade New Relic agent to v1.25.1

rogatty:
https://wikia-inc.atlassian.net/browse/XW-905
https://github.com/newrelic/node-newrelic/blob/master/NEWS.md

We do it as a hotfix to isolate it from other changes. If anything breaks we will be sure what was the reason.


bognix:
LGTM

PR #1912, opened 2016-01-28
XW-977 | Show native smart banner on ios

hakubo:


front/common/utils/browser.js
```diff
@@ -3,16 +3,17 @@
...
-	let system;
+const userAgent = navigator.userAgent;
 
-	if (ua.match(/iPad|iPhone|iPod/i) !== null) {
-		system = 'ios';
-	} else if (ua.match(/Android/i) !== null) {
-		system = 'android';
-	}
+export let system,
+	supportsNativeSmartBanner,
```
rogatty: It's not used, is it?

front/common/utils/browser.js
```diff
@@ -3,16 +3,17 @@
...
+	standalone;
 
-	return system;
+if (userAgent.match(/iPad|iPhone|iPod/i) !== null) {
```
rogatty: We can use `userAgent.test(...)` which returns boolean.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

front/main/app/components/smart-banner.js
```diff
@@ -18,59 +23,30 @@ export default Ember.Component.extend({
...
 	}),
 
-	labelInStore: Ember.computed('system', function () {
-		return i18n.t(`app.smartbanner-store-${this.get('system')}`);
-	}),
-
-	labelInstall: Ember.computed('system', function () {
-		return i18n.t(`app.smartbanner-install-${this.get('system')}`);
-	}),
+	isVisible: system === 'android',
```
rogatty: I think this shouldn't be here. We set `smartBannerVisible: false` in `components/application-wrapper.js` and `controllers/application.js` and here override it. It will be visible always when on Android and then hidden by `checkForHiding` according to cookie value.

Also, would be nice if you check why it's set to false twice and clean it up if possible.

front/main/app/components/smart-banner.js
```diff
@@ -126,19 +102,16 @@ export default Ember.Component.extend({
...
+		// Show custom smart banner only when a device is Android
+		// website isn't loaded in app and user did not dismiss it already
+		if (system === 'android' && !standalone &&
+			name && !disabled &&
 			Ember.$.cookie('sb-closed') !== '1'
 		) {
 			this.sendAction('toggleVisibility', true);
 			this.track(trackActions.impression);
-		} else {
-			this.set('isVisible', false);
```
rogatty: Sooo, how do we hide it currently?

Warkot: It's actually hidden by default → https://github.com/Wikia/mercury/blob/XW-997/front/main/app/components/application-wrapper.js#L41.

Then it's checked should it be displayed, if yes, then it's scheduled to be shown after render → https://github.com/Wikia/mercury/blob/XW-997/front/main/app/components/smart-banner.js#L98.

From what I've experimented, it's to display the smart banner above (in the y-axis) the top bar (the floating one with the Wikia logo).

Please, correct me if I'm wrong.

front/main/app/components/smart-banner.js
Warkot: Actually, this wasn't taken into account at all (changed it to a computed property with a `debugger;` inside — it didn't trigger). Mechanism for toggling the visibility of the smart banner is more… sophisticated – just as you say and I wrote here → https://github.com/Wikia/mercury/pull/1912#discussion_r51176200.

front/main/app/components/smart-banner.js
```diff
@@ -1,7 +1,12 @@
...
 import {track as mercuryTrack, trackActions} from 'common/utils/track';
-import {getSystem} from 'common/utils/browser';
+import {system, standalone} from 'common/utils/browser';
 
+/**
+ * Component for a custom Smart Banner
+ * it's visible only for Android devices
+ * iOS has its own native smart banner - no need to render it there
+ */
 export default Ember.Component.extend({
```
bognix: if this component is only used on android why not calling the component android-smart-banner to make it more obvious?

front/main/app/components/smart-banner.js
```diff
@@ -18,59 +23,26 @@ export default Ember.Component.extend({
...
-		return Ember.getWithDefault(Mercury, 'wiki.smartBanner', {});
-	}),
-
-	dbName: Ember.computed(() => {
-		return Ember.get(Mercury, 'wiki.dbName');
-	}),
-
+	appId: Ember.computed.oneWay(`config.appId.android`),
+	appScheme: Ember.computed.oneWay(`config.appScheme.android`),
+	config: Ember.getWithDefault(Mercury, 'wiki.smartBanner', {}),
```
bognix: what if wiki.smartBanner is null, not undefined? The thing I mentioned seems like an edge case but I guess it's worth pointing out at least.

front/main/app/components/smart-banner.js
```diff
@@ -18,59 +23,26 @@ export default Ember.Component.extend({
...
-		return i18n.t(`app.smartbanner-store-${this.get('system')}`);
-	}),
-
-	labelInstall: Ember.computed('system', function () {
-		return i18n.t(`app.smartbanner-install-${this.get('system')}`);
-	}),
-
-	link: Ember.computed('appId', 'dbName', 'system', function () {
+	link: Ember.computed('appId', 'dbName', function () {
 		const appId = this.get('appId');
```
bognix: appId is in variable, dbName is inserted in string directly from `this.get` - how about unification? My vote goes to  - remove the variable

front/main/app/routes/article.js
```diff
@@ -178,5 +179,22 @@ export default Ember.Route.extend({
...
+		const appId = Ember.get(Mercury, 'wiki.smartBanner.appId.ios'),
+			canonicalUrl = Ember.get(Mercury, 'wiki.basePath') + model.get('url');
+
+		let $descriptionMetaTag = Ember.$('head meta[name=apple-itunes-app]');
+
+		if (Ember.isEmpty($descriptionMetaTag)) {
+			$descriptionMetaTag = Ember.$('<meta name="apple-itunes-app">').appendTo('head');
+		}
+
+		$descriptionMetaTag.prop('content', `app-id=${appId}, app-argument=${canonicalUrl}`);
```
bognix: how about first setting the content prop and then appending to head? not sure if I'm not suggesting to-much-optimization

front/main/app/routes/article.js
```diff
@@ -178,5 +179,22 @@ export default Ember.Route.extend({
 		}
 
 		$descriptionMetaTag.prop('content', description);
+	},
+
+	/**
+	 * @param {ArticleModel} model
+	 * @returns {void}
+	 */
+	updateIOSSmartBannerMetaTag(model) {
```
bognix: The meta tag won't be updated when navigating from an article to a curated main page? Is it desired behavior?

front/main/app/components/smart-banner.js
rogatty: I like it!

front/main/app/routes/article.js
Warkot: Not really. Is there a solution you want to suggest?

front/main/app/routes/article.js
Warkot: Might be a good idea. I just mostly copy-pasted from @hakubo's code above. :stuck_out_tongue_winking_eye:

front/main/app/components/smart-banner.js
Warkot: Sounds like a plan. The code will be clearer for the coders new to our SPA. :+1:

front/main/app/routes/article.js
bognix: move the logic to mixin and add in both routes? Move to application route? (however I'm not so sure about the second solution)

rogatty:
We can also clean up `_smart-banner.scss` a little.

front/main/app/components/smart-banner.js
rogatty: It shouldn't be null:
```
$smartBannerConfig = $this->getSmartBannerConfig();
if ( !is_null( $smartBannerConfig ) ) {
	$wikiVariables[ 'smartBanner' ] = $smartBannerConfig;
}
```
https://github.com/Wikia/app/blob/dev/extensions/wikia/MercuryApi/MercuryApiController.class.php#L328

front/main/app/routes/article.js
rogatty: It won't be used for now as apps don't support it yet. Let's not waste time now. When it comes to fixing it we will probably have `ember-cli-meta-tags` already in place.

front/main/app/routes/article.js
```diff
@@ -178,5 +179,22 @@ export default Ember.Route.extend({
...
+		const appId = Ember.get(Mercury, 'wiki.smartBanner.appId.ios'),
+			canonicalUrl = Ember.get(Mercury, 'wiki.basePath') + model.get('url');
```
rogatty: Let's use backticks to concatenate strings.

front/main/app/routes/article.js
bognix: with this approach - what is the point of adding it right now? If we are going to have ember-cli-meta-tags by that time maybe we should implement it then and don't implement it now partially?

front/common/utils/browser.js
```diff
@@ -3,16 +3,16 @@
...
 
-	let system;
+const userAgent = navigator.userAgent;
 
-	if (ua.match(/iPad|iPhone|iPod/i) !== null) {
-		system = 'ios';
-	} else if (ua.match(/Android/i) !== null) {
-		system = 'android';
-	}
+export let system,
```
tomasznapieralski: export shouldn't be after assign operation on that variables? (after line 18)

bognix: you are exporting a reference so it should be fine

front/main/app/components/smart-banner.js
Warkot: So a question then: `smart-banner-android` vs `android-smart-banner`?

rogatty: I vote for no. 1

front/main/app/routes/article.js
Warkot: I'll drop it — this would just complicate code, add some duplication or helper variables, and yet a possibility that it happens is close to none, as both — app meta tag and canonical link are added to index.html.

Warkot:
I'm wondering if this is a good condition — we're generating this meta tag for both iOS and Android. https://github.com/Wikia/mercury/pull/1912/files#diff-bbc6e83f9f49313cd818aaa97bd0c91aR1
But the custom banner now is rendered for both (but hidden for iOS), too.

bognix:
LGTM

rogatty:
LGTM

PR #1915, opened 2016-01-28
SOC-1924 user posts/replies lists - reply card header text

rafalkalinski:
user posts/replies lists - reply card header text

https://wikia-inc.atlassian.net/browse/SOC-1924


kvas-damian:
LGTM

PR #1916, opened 2016-01-28
XW-1036 | Create fallback for localStorage

vforge:
https://wikia-inc.atlassian.net/browse/XW-1036

https://github.com/Wikia/app/pull/9608


front/common/utils/storage.js
```diff
@@ -0,0 +1,38 @@
+/* eslint no-undefined:0 */
```
hakubo: can we not modify eslint rules per file? :)
or change the rule if it ain't make sense

front/common/utils/storage.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Returns localStorage or compatible in-memory object
+ *
+ * @returns {Object}
+ */
+export default function () {
```
hakubo: I don't think it has to be a function - we could just export localStorage or the object
especially that calling the function will not share the object
meaning that for localStorage you would have 'global' data
and for a fallback not

front/common/utils/storage.js
```diff
@@ -0,0 +1,38 @@
...
+	return (('localStorage' in window) && window.localStorage) ?
+		window.localStorage :
+		{
```
hakubo: you can treat localStorage like an array localStorage['myKey']
If we want to make it look like it's a localStorage I think it would be good to mimick it

maybe there is something in Open Source?

vforge: This code was OpenSource'd from a gist found under 'best localStorage in-memory object alternative` ;)

front/common/utils/localStorageAdapter.js
```diff
@@ -0,0 +1,40 @@
...
+		},
+
+		clear() {
+			this._data = {};
+
+			return {};
+		}
+	};
+
+export default localStorageAdapter;
```
bognix: how about unit tests for this module? Seems like a perfect candidate

front/common/utils/localStorageAdapter.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Returns localStorage or compatible in-memory object
+ *
+ * @returns {Object}
```
bognix: you can add typedef for this object. The API is known, setItem, getItem, removeItem and clear

front/common/utils/localStorageAdapter.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+const localStorageAdapter = (('localStorage' in window) && window.localStorage) ?
```
bognix: how about `export default ((...`

rogatty: :laughing: 
https://github.com/Wikia/mercury/pull/1916#discussion_r51188503

front/main/app/components/wikia-in-your-lang.js
```diff
@@ -3,6 +3,7 @@ import AlertNotificationsMixin from '../mixins/alert-notifications';
 import LanguagesMixin from '../mixins/languages';
 import WikiaInYourLangModel from '../models/wikia-in-your-lang';
 import {track, trackActions} from 'common/utils/track';
+import localStorageAdapter from 'common/utils/localStorageAdapter';
```
rogatty: This should be imported from `../utils/local-storage-adapter`, let's not pollute common with things that won't be used by auth.

front/common/utils/localStorageAdapter.js
rogatty: Is the first check needed? How is it different without it?

front/common/utils/localStorageAdapter.js
```diff
@@ -0,0 +1,40 @@
...
+	window.localStorage :
+	{
+		_data: {},
+
+		setItem(id, value) {
```
rogatty: Shouldn't we have JSDocs for these functions?

front/common/utils/localStorageAdapter.js
bognix: I'm not suggesting to export default a function but an object

front/main/app/utils/local-storage-adapter.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {Function} removeItem
+ * @property {Function} clear
+ */
+
+/**
+ * Returns window.localStorage or compatible in-memory object
+ *
+ * @returns {LocalStorage}
+ */
+export default window.localStorage ?
```
bognix: `export default window.localStorage || { ...`

bognix:
Rename the FakeLocalStorage and you have LGTM from me :+1:

rogatty:
LGTM

PR #1921, opened 2016-01-29
Discussion refactors

hakubo:
Hey,

While working on something else I did couple of changes to discussion code.
Do they look good to you?
@bkoval @kvas-damian @rogatty

### DIFF
--- a/front/main/app/components/discussion-editor.js
+++ b/front/main/app/components/discussion-editor.js
@@ -176,7 +176,7 @@ export default Ember.Component.extend(ViewportMixin, {
 	 * @returns {void}
 	 */
 	didInsertElement() {
-		this._super();
+		this._super(...arguments);
 
 		this.handleIOSFocus();
 		this.initializeStickyState();
--- a/front/main/app/components/discussion-post-editor.js
+++ b/front/main/app/components/discussion-post-editor.js
@@ -20,7 +20,7 @@ export default DiscussionEditorComponent.extend({
 	 */
 	initializeStickyState() {
 		this.setProperties({
-			offsetTop: this.element.offset().top,
+			offsetTop: this.$().offset().top,
 			siteHeadHeight: Ember.$('.site-head').outerHeight(true)
 		});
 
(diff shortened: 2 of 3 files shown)


front/main/app/components/discussion-post-editor.js
```diff
@@ -9,20 +9,23 @@ export default DiscussionEditorComponent.extend({
...
+	didInsertElement() {
+		this._super(...arguments);
+		this.initializeStickyState();
+	},
+
 	/**
 	 * Initialize onScroll binding for sticky logic
 	 * @returns {void}
 	 */
-	initializeStickyState: Ember.on('didInsertElement', function () {
```
rogatty: Seems like this was a discouraged pattern http://notmessenger.com/proper-use-of-ember-on/

We probably should add information to the guidelines.

rogatty:
LGTM

bkoval:
LGTM

kvas-damian:
LGTM

front/main/app/components/discussion-editor.js
```diff
@@ -176,7 +176,7 @@ export default Ember.Component.extend(ViewportMixin, {
 	 * @returns {void}
 	 */
 	didInsertElement() {
-		this._super();
+		this._super(...arguments);
```
vforge: Can you use rest parameters explicitly instead of using `arguments`?

hakubo: From what I see this is encouraged way. 
We don't have guideline about that yet also

hakubo: especially when you look at the output of babel:
https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=true&code=function%20a%20()%20%7B%0A%20%20console.log(...arguments)%3B%0A%7D%0A%0Afunction%20b(...rest)%20%7B%0A%20%20console.log(rest)%3B%0A%7D

vforge: Thanks for the link! We just need to remember about not doing that with arrow functions.

PR #1923, opened 2016-01-29
XW-41 | Replace placeholders with components dynamically

rogatty:
https://wikia-inc.atlassian.net/browse/XW-41

Replace placeholders coming from MW API with components dynamically without losing Ember context. For example if this placeholder comes as a part of article:
```
<div
	data-component='image-media'
	data-attrs='{
		"width": 100,
		"height": 100,
		"media": {
			"type": "image",
			"url": "http://vignette2.wikia.nocookie.net/starwars/images/d/d6/Yoda_SWSB.png/revision/latest?cb=20150206140125",
			"fileUrl": "http://starwars.wikia.com/wiki/File:Yoda_SWSB.png",
			"title": "Yoda SWSB.png",
			"user": "Alexrd",
			"width": 1600,
			"height": 1920
		}
	}'>
	This is image placeholder!
</div>
```

The `image-media` component will be created and put in the place of placeholder. No need for any additional code, just create a component in Ember and pass its name and attributes from MediaWiki. As opposed to the current situation, it will have all the Ember goodness available, including bindings.

This is a prototype that combines code from https://github.com/mitchlloyd/ember-islands and https://github.com/yapplabs/ember-wormhole and adjusts it to our needs. Currently, only single images work. Related app PR: https://github.com/Wikia/app/pull/9597

/cc @Wikia/x-wing


front/main/app/components/article-content.js
```diff
@@ -152,6 +170,18 @@ export default Ember.Component.extend(
 			}
 		},
 
+		getAttributesForMedia({ name, attrs, element }) {
+			if (name === 'article-media-image' && attrs.ref) {
+				const media = this.get('media.media');
+
+				if (media && media[attrs.ref]) {
+					attrs = Object.assign(attrs, media[attrs.ref]);
```
hakubo: Are we sure this is supported in everybrowser that we support?
http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign

vforge:
[](http://i.imgur.com/0fXgf8E.jpg)

front/main/app/components/article-content.js
rogatty: Good call. I though it would be handled by Babel but it's not. Will use `Ember.$.extend` instead.

front/main/app/utils/render-component.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * @param {JQuery} $element
+ * @returns {Array}
+ */
+export function queryPlaceholders($element) {
+	let components = [];
+
+	$element.find('[data-component]').each(function () {
```
vforge: Wouldn't this be better?
```
$element.find('[data-component]').map(function (element) {
  const name = element.getAttribute('data-component'),
    attrs = componentAttributes(element);

  return {attrs, name, element};
});
```

tomasznapieralski:
LGTZespol

PR #1928, opened 2016-02-01
XW-821 | don't use deprecated functionality in article-comments

hakubo:


front/main/app/components/article-comments.js
```diff
@@ -67,14 +67,32 @@ export default Ember.Component.extend({
...
+
 		this.set('model', ArticleCommentsModel.create({
 			articleId: this.get('articleId')
 		}));
+	},
+
+	/**
+	 * @returns {void}
+	 */
+	didRender() {
```
rogatty: I think we should use `this._super(...arguments);` in all hooks.

front/main/app/components/article-comments.js
```diff
@@ -67,14 +67,32 @@ export default Ember.Component.extend({
 	}),
 
 	/**
+	 * Sets model when we get new articleId
+	 *
 	 * @returns {void}
 	 */
-	didInsertElement() {
+	didInitAttrs() {
```
rogatty: Are you sure this shouldn't be `didReceiveAttrs`?

https://guides.emberjs.com/v2.3.0/components/the-component-lifecycle/#toc_formatting-component-attributes-with-code-didreceiveattrs-code

hakubo: yes, the way article-comments are build now
creating model can happen only once
when articleId changes there is different hook that will change that

this whole component should be rewritten not to use observers

rogatty:
LGTM

PR #1930, opened 2016-02-01
SOC-1895 | Google Registration page

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-1895

@kvas-damian @rafalkalinski @katanka 

Now, this is a WiP but I want en early code review. There's a couple of things to address here:

1) This first does normal reg, then google connect with a new account. We need to add a logout in case the reg itself is fine but the connect raises an error.

2) Werid situtation with commits listed here that belong to the base branch - in the diff there are no changes from them.


kvas-damian:
Fix linter issues ;)

front/auth/app/google/GoogleLogin.js
```diff
@@ -39,12 +39,21 @@ export default class GoogleLogin extends ExternalLogin {
 		});
 	}
 
+	redirectToRegistrationPage() {
```
kvas-damian: We're talking about it. Please add link to ticket where we're going to refactor this code.

server/app/facets/auth/register.js
```diff
@@ -106,7 +106,8 @@ function getGoogleRegistrationPage(request, reply) {
 			{
 				formId: 'googleRegistrationForm',
 				headerText: 'auth:google-register.register-with-google',
-				heliosExternalRegistrationURL: authUtils.getHeliosUrl('/facebook/users'),
+				heliosExternalRegistrationURL: authUtils.getHeliosUrl('/users'),
+				additionalRegistrationURL: authUtils.getExternalAuthUrl('/google/users/'),
```
kvas-damian: what ```additional``` means here?

front/auth/app/google/GoogleRegistration.js
```diff
@@ -0,0 +1,158 @@
...
+
+	/**
+	 * @param {string} email from Google
+	 *
+	 * @returns {void}
+	 */
+	setUpEmailInput(email) {
+		const emailInput = this.form.elements.namedItem('email');
+
+		if (email && emailInput) {
```
kvas-damian: We already have similar code in ```FacebookRegistration```. I believe we can extract it and create base class.

package.json
```diff
@@ -9,9 +9,9 @@
     "setup": "npm run setup-root && npm run setup-common && npm run setup-auth && npm run setup-main && npm run setup-server",
     "setup-for-local": "npm install -g bower gulp@3.x.x ember-cli@2.2.0-beta.6",
     "setup-root": "npm install",
-    "setup-common": "cd front/common && bower install",
-    "setup-auth": "cd front/auth && bower install",
-    "setup-main": "cd front/main && npm install && bower install",
+    "setup-common": "cd front/common && bower install --allow-root",
```
rafalkalinski: I don't think you wanted to left this here.

front/auth/app/google/GoogleRegistration.js
```diff
@@ -0,0 +1,158 @@
...
+			emailInput.disabled = true;
+			emailInput.value = email;
+		}
+	}
+
+	/**
+	 * @param {string} userId
+	 * @returns {void}
+	 */
+	onSuccessfulRegistration(userId) {
```
rafalkalinski: the super method of this takes two parameters

front/auth/app/google/GoogleLogin.js
bkoval: https://wikia-inc.atlassian.net/browse/SOC-1977

server/app/facets/auth/register.js
bkoval: We're using the same template for both google and facebook reg, but google reg needs two urls (first is the url for setting up a regular account and the second one is for connecting it with google account).

Now the alternative seems to be removing additionalUrl from here and passing it through pageParams, which works in this case as we cannot fall back to non-js form, as the google reg needs to calls. The handlebars method seemed nicer to me, but that's just my preference.

What do you think?

kvas-damian: Current solution is ok ;)

front/auth/app/google/GoogleRegistration.js
```diff
@@ -0,0 +1,141 @@
...
+		this.pageName = 'googleRegistration';
+		this.urlHelper = new UrlHelper();
+		this.googleConnectUrl = form.getAttribute('data-registration-url');
+
+		params = this.urlHelper.urlDecode(window.location.search.substr(1));
+
+		this.email = params.email;
+		this.gToken = params.gToken;
+
+		new GoogleApi(this.init.bind(this));
```
hakubo: any chance to not construct class for side effects?

front/auth/app/main.js
```diff
@@ -53,6 +54,8 @@ export function init() {
 				new SignupForm(formElement).watch();
 			} else if (formElement.id === 'facebookRegistrationForm') {
 				new FacebookRegistration(formElement);
+			} else if (formElement.id === 'googleRegistrationForm') {
+				new GoogleRegistration(formElement).watch();
```
hakubo: here also - maybe that should be just a module?

front/auth/app/signup/SignupForm.js
```diff
@@ -126,7 +127,10 @@ export default class SignupForm {
 			label: VisitSourceWrapper.lifetimeVisitSource.get()
 		});
 
-		AuthUtils.authSuccessCallback(this.redirect);
+		if (shouldRunSuccessCallback) {
+			AuthUtils.authSuccessCallback(this.redirect);
+
```
hakubo: omg, omg! empty line

hakubo:
I would consider refactoring classes that are there for side effects into simple modules. other than that LGTM

PR #1931, opened 2016-02-01
XW-1076 | Implement prerender.io on 10 communities

vforge:
https://wikia-inc.atlassian.net/browse/XW-1076

Uses https://github.com/vforge/hapi-prerender (with patches)

### DIFF
--- a/server/config/localSettings.base.js
+++ b/server/config/localSettings.base.js
@@ -196,7 +196,18 @@ const localSettings = {
 	prerenderApiKey: process.env.PRERENDER_API,
 	// prerendered host lists
 	prerenderHost: [
-		'assassinscreed.wikia.com'
+		'assassinscreed.wikia.com',
+		'onepiece.wikia.com',
+		'vampirediaries.wikia.com',
+		'gameofthrones.wikia.com',
+		'dragonball.wikia.com',
+		'fairytail.wikia.com',
+		'ja.starwars.wikia.com',
+		'gta.wikia.com',
+		'animalcrossing.wikia.com',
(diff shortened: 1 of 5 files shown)


server/config/localSettings.base.js
```diff
@@ -191,6 +192,12 @@ const localSettings = {
 	mwPreviewSalt: process.env.MW_PREVIEW_SALT,
 	// Special salt for calculating GA userID
 	gaUserSalt: process.env.GA_USERID_SALT,
+	// Prerender.io API Key
+	prerenderApiKey: process.env.PRERENDER_API,
+	// prerendered host lists
+	prerenderHost: [
+		'assassinscreed.wikia.com'
```
vforge: This is temporary / need final list

server/package.json
```diff
@@ -22,6 +22,7 @@
     "handlebars": "4.0.5",
     "hapi": "9.5.1",
     "hapi-i18next": "2.2.0",
+    "hapi-prerender": "vforge/hapi-prerender",
```
rogatty: Won't it be downloaded every time you install modules? I think we should use commit hash or release tag here.

vforge: We'll add tag when we'll have final list of wikis.

hakubo: !

server/config/localSettings.base.js
hakubo: mkay

hakubo:
LGTM

tomasznapieralski:
LGTM

PR #1933, opened 2016-02-02
PLATFORM-1855 Image review does not show flagged images

RafalWilinski:
Simple if change, makes flagged images appear. 
/cc @rogatty @pchojnacki


rogatty:
LGTM

PR #1935, opened 2016-02-02
CE-3302 Add route to Recent Wiki Activity list

adamkarminski:
@Wikia/community-engineering @rogatty

### DIFF
--- a/front/main/app/routes/recent-wiki-activity.js
+++ b/front/main/app/routes/recent-wiki-activity.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+import RecentWikiActivityModel from '../models/recent-wiki-activity';
+
+export default Ember.Route.extend({
+
+	/**
+	 * @returns {*}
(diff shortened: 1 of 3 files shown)


front/main/app/routes/recent-wiki-activity.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+import RecentWikiActivityModel from '../models/recent-wiki-activity';
+
+export default Ember.Route.extend({
```
rogatty: If you want `meta` to work, this route should be extended by `MetaTagsMixin` like that:

```
import MetaTagsMixin from '../mixins/meta-tags';

export default Ember.Route.extend(MetaTagsMixin, {
```

Only in there is a logic that uses `this.get('meta')`.

adamkarminski: Thank you!

PR #1945, opened 2016-02-03
use const {...} = Ember;

hakubo:
This is a pattern that I have noticed core team Ember members are using.

For some reason you can't do 
`import {Component} from 'ember`;
just yet

but we can achieve that with
```
import Ember from 'ember';
const {Component} = Ember;
```

I'd like to propose that we use this pattern in our code to make code more succinct and readable.

@rogatty @kvas-damian @adamkarminski @dianafa @bkoval


rogatty:
LGTM

bkoval:
Do we have a place in the guidelines repo to add this rule?

gabrys:
What sort of magic is that?

hakubo:
Object destructuring
https://babeljs.io/docs/learn-es2015/#destructuring

vforge:
@gabrys JavaScript of the future ;)

LGTM

bognix:
@hakubo - can we merge it?

hakubo:
@bognix yes - next part would be to figure out if we want to do that for all files
or it would happen over time 
and 
put that in our guidelines (I'll do that)

thanks!

PR #1954, opened 2016-02-04
CE-3300 Add styling of a diff page

adamkarminski:
@kamilkoterba


front/main/app/templates/article-diff.hbs
```diff
@@ -1,12 +1,13 @@
...
-}}
-{{model.timestamp}}
-{{#link-to 'article' model.title}}{{model.title}}{{/link-to}}
+<div class="diff-page">
+{{recent-change model=model}}
 {{#each model.diffs as |diffs|}}
 	{{#each diffs as |diff|}}
-		{{diff.content}}
+		{{#if diff.isPrevious}}
+			<div class="diff-type">Previous</div>
```
Grunny: i18n needed on these two.

front/main/app/templates/components/article-diff.hbs
```diff
@@ -0,0 +1,10 @@
+{{recent-change model=model}}
+{{#each model.diffs as |diffs|}}
+	{{#each diffs as |diff|}}
+		<div class="{{diff.class}}{{if diff.allChanged ' all-changed'}}">{{diff.content}}</div>
+	{{/each}}
+{{/each}}
+
+{{#if currentUser.isAuthenticated}}
+<button {{action 'undo'}} class="">{{i18n 'main.undo-label' ns='recent-wiki-activity'}}</button>
```
bkoval: +1 tab

Grunny: Left over from testing something that I didn't remove properly. :)

front/main/app/components/recent-change.js
```diff
@@ -4,5 +4,8 @@ export default Ember.Component.extend({
 	classNames: ['recent-change'],
 	unixTimestamp: Ember.computed('model.timestamp', function () {
 		return new Date(this.get('model.timestamp')).getTime() / 1000;
+	}),
+	hasDiff: Ember.computed('model.old_revid', 'model.revid', function () {
```
hakubo: Ember.computed.and
http://emberjs.com/api/classes/Ember.computed.html#method_and

front/main/app/controllers/article-diff.js
```diff
@@ -0,0 +1,70 @@
+import Ember from 'ember';
+import ArticleDiffModel from '../models/article-diff';
+import {track, trackActions} from 'common/utils/track';
+
+export default Ember.Controller.extend({
```
hakubo: Controllers are kind of deprecated in Ember

We should try to have route that renders component and then have this logic in a main component for a route
they are working on something called routable components - and want to remove controllers from the framework

front/main/app/controllers/article-diff.js
```diff
@@ -0,0 +1,70 @@
...
+	},
+
+	actions: {
+		/**
+		 * @returns {void}
+		 */
+		undo() {
+			this.get('application').set('isLoading', true);
+
+			ArticleDiffModel.undo(this.get('model')).then(
```
hakubo: this API looks odd can we refactor this to have:
this.get('model').undo(...) ?

front/main/app/models/article-diff.js
```diff
@@ -9,7 +10,8 @@ const ArticleDiffModel = Ember.Object.extend({
 	pageid: null,
 	timestamp: null,
 	title: null,
-	user: null
+	user: null,
+	useravatar: null
 });
 
 ArticleDiffModel.reopenClass({
```
hakubo: We set this style in our models, but I am not sure if that is the best way to do this
it would be good if methods were part of the Object so you can work on that model directly
eg this.get('model').undo ...

front/main/app/controllers/article-diff.js
Grunny: Done in https://github.com/Wikia/mercury/commit/47348d483f3c8ab.

front/main/app/models/article-diff.js
Grunny: Done in https://github.com/Wikia/mercury/commit/47348d483f3c8ab.

front/main/app/controllers/article-diff.js
```diff
@@ -0,0 +1,70 @@
+import Ember from 'ember';
+import ArticleDiffModel from '../models/article-diff';
```
hakubo: you don't need this in this file anymore

front/main/app/controllers/article-diff.js
hakubo: or actually since these methods below work mainly with model, I think a route is the best home for it

front/main/app/models/article-diff.js
```diff
@@ -9,7 +10,39 @@ const ArticleDiffModel = Ember.Object.extend({
 	pageid: null,
 	timestamp: null,
 	title: null,
-	user: null
+	user: null,
+	useravatar: null,
+
+	undo() {
```
hakubo: JSDoc

front/main/app/models/article-diff.js
```diff
@@ -9,7 +10,39 @@ const ArticleDiffModel = Ember.Object.extend({
...
+						success: (resp) => {
+							if (resp && resp.edit && resp.edit.result === 'Success') {
+								resolve();
+							} else if (resp && resp.error) {
+								reject(resp.error.code);
+							} else {
+								reject();
+							}
+						},
+						error: (err) => reject(err)
```
hakubo: minor: 
```
error: reject
```
should be enough right?

front/main/app/styles/component/_article_diff.scss
```diff
@@ -0,0 +1,65 @@
+// TODO move $faint-gray to styleguide
+$faint-gray: #cccccc;
+
+.diff-page {
```
hakubo: we are experimenting a bit with BEM
you could try it too

front/main/app/styles/component/_article_diff.scss
```diff
@@ -0,0 +1,65 @@
...
+	font-size: $type-min;
+	line-height: $line-height-smallest;
+
+	.diff-lineno {
```
hakubo: .diff-page__lineno
and then you should not need to have this under .diff-page

same goes for next elements

front/main/app/styles/component/_article_diff.scss
```diff
@@ -0,0 +1,65 @@
...
+	.diff-addedline,
+	.diff-context,
+	.diff-deletedline {
+		& + .diff-lineno {
+			border-top: 1px dashed $faint-gray;
+			margin-top: 30px;
+		}
+	}
+
+	.diff-addedline,
```
hakubo: instead of doing this manually you could use %placeholder

front/main/app/templates/components/article-diff.hbs
```diff
@@ -0,0 +1,17 @@
+<header class="sub-head">
+	{{#link-to 'recent-wiki-activity' class="sub-head--cancel"}}
```
hakubo: this does not look like a modifier, more like a sub components
sub-head__cancel

front/main/app/components/article-diff.js
```diff
@@ -0,0 +1,16 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	classNames: ['diff-page'],
+	currentUser: Ember.inject.service(),
+	showDiffLink: false,
```
hakubo: this looks like a state of a component
these are better initialized in an init hook

reason being that if you render this component twice
they will share this state
so both will either showDiffLink or hide it
```
init() {
    this.set('showDiffLink', false);
}
```

hakubo: ok actually for literals it doesn't matter :)

but do check out this: 
https://poteto.github.io/component-best-practices/#/leaky

front/main/app/templates/components/article-diff.hbs
kamilkoterba: This is already used app wide.

front/main/app/components/article-diff.js
kamilkoterba: thanks

PR #1958, opened 2016-02-04
XW-1075 | use display title

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1075

### DIFF
--- a/front/main/app/models/article.js
+++ b/front/main/app/models/article.js
@@ -20,7 +20,7 @@ const ArticleModel = Ember.Object.extend({
 	content: null,
 	basePath: null,
 	categories: [],
-	cleanTitle: null,
+	displayTitle: null,
 	comments: 0,
 	description: null,
 	isMainPage: false,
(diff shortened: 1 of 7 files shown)


vforge:
LGTM

front/main/app/models/article.js
```diff
@@ -164,6 +164,7 @@ ArticleModel.reopenClass({
 
 				articleProperties = $.extend(articleProperties, {
 					content: article.content,
+					displayTitle: article.displayTitle,
```
rogatty: Won't it override `displayTitle: details.title` even if `article.displayTitle` is not set?

server/app/facets/operations/prepareArticleData.js
```diff
@@ -31,11 +31,12 @@ export default function prepareArticleData(request, data) {
 
 		if (articleData.details) {
 			articleDetails = articleData.details;
-			title = articleDetails.cleanTitle ? articleDetails.cleanTitle : articleDetails.title;
+			title = articleDetails.title;
 			result.canonicalUrl = wikiVariables.basePath + articleDetails.url;
 		}
 
 		if (articleData.article) {
+			title = articleData.article.displayTitle;
```
rogatty: There is always `articleData.details` and `articleData.article`. What's the point of setting it once and then overriding it?

PR #1962, opened 2016-02-04
(CE-3266) Add the ability to undo an edit from a diff page

Grunny:
Add the ability to undo an edit from a diff page. This also moves the diff page
to a proper component, and limits displaying the undo button to logged-in users.

/cc @Wikia/community-engineering


kamilkoterba:
LGTM

PR #1963, opened 2016-02-04
SOC-1376 | As a user I would like to see the timestamp of a post or a reply a certain way

Bloodmoon24:
https://wikia-inc.atlassian.net/browse/SOC-1376
@bkoval @rafalkalinski


front/main/bower.json
```diff
@@ -18,6 +18,7 @@
     "jquery": "2.1.4",
     "loader.js": "3.5.0",
     "numeral": "1.5.3",
+    "moment": "^2.8.0",
```
bognix: please use the fixed versions of packages

front/main/bower.json
```diff
@@ -27,6 +28,7 @@
     "wikia-style-guide": "Wikia/style-guide#v1.8.2",
     "Autolinker.js": "0.22.0",
     "sinonjs": "1.17.1",
+    "moment-timezone": "^0.5.0",
```
bognix: please use the fixed versions of packages

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,29 @@
...
-		return '';
+	if (now.diff(date, 'days') > 5) {
+		output = date.format('L');
+	} else if (now.diff(date, 'minutes') < 1) {
+		output = i18n.t('app.now-label');
+	} else {
+		output = date.fromNow(shouldHideAgoString);
 	}
+
+	return `<span class='timestamp' title='${date.format('LLL')}'>&bull; ${output} </span>`;
```
hakubo: can this be more generic?
return only date formattted
so it's up to a template to style it 

we will use it in more places I think

hakubo:
@kamilkoterba you might be interested about this PR

front/main/app/templates/components/post-reply.hbs
```diff
@@ -16,9 +16,8 @@
 			displayProfileName=true
 			username=author.name
 		}}
-		<span class="timestamp">
-			&bull; {{time-ago post.creationDate.epochSecond}}
-		</span>
+
+		{{{time-ago post.creationDate.epochSecond}}}
```
bognix: `{{{` is used on purpose?

bognix: `{{{` is used on purpose?

Bloodmoon24: Yeah, helper is returning <span> with content, so I need to parse it

front/main/app/helpers/time-ago.js
bognix: :+1:

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,29 @@
...
 
 /**
  * Helper to give textual representation of time interval between past date
  * and the current time/date in the form
- * {timeAgo unixTimestamp}
- * which returns something like '2 days ago'
+ * {timeAgo unixTimestamp shouldDisplayAgo}
+ * which returns something like '2 days ago' if interval is below 6 days or formated param date
  *
  * @param {Array} params
```
bognix: since we are obligated to add JSDocs how about explaining in it what params represent.
What I mean is something like
```
param[0] - time
param[1] - ...
```

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,29 @@
...
-		return i18n.t('app.years-ago-label', {count: interval.value});
-	default:
-		Ember.Logger.error('Unexpected date interval for timestamp', unixTimestamp);
-		return '';
+	if (now.diff(date, 'days') > 5) {
+		output = date.format('L');
+	} else if (now.diff(date, 'minutes') < 1) {
+		output = i18n.t('app.now-label');
+	} else {
+		output = date.fromNow(shouldHideAgoString);
```
bognix: could you rename the `shouldHideAgoString` to `shouldHideAgoPrefix` or something like that. It was hard for me to understand what this flag does without checking the documenation

Bloodmoon24: Makes sense

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,31 @@
...
+	if (now.diff(date, 'days') > 5) {
+		output = date.format('L');
+	} else if (now.diff(date, 'minutes') < 1) {
+		output = i18n.t('app.now-label');
+	} else {
+		output = date.fromNow(shouldHideAgoPrefix);
 	}
+
+	return `<span class='timestamp' title='${date.format('LLL')}'>&bull; ${output} </span>`;
+
```
tomasznapieralski: remove this line

front/main/app/helpers/time-ago.js
vforge: :+1:

front/main/config/environment.js
```diff
@@ -12,7 +12,9 @@ module.exports = function (environment) {
 				// e.g. 'with-controller': true
 			}
 		},
-
+		moment: {
```
tomasznapieralski: do you need to have it here? also can you provide mere meaningful name?

front/main/config/environment.js
```diff
@@ -12,7 +12,9 @@ module.exports = function (environment) {
...
+			localeOutputPath: 'assets/vendor/moment/locales'
```
tomasznapieralski: IMO you shouldn't modify vendor files, you can place your modification in different place

front/main/config/environment.js
Bloodmoon24: It's pattern from vendor's page, so I thought it will be ok. This is lib name, so I think there is no more meaningful name. If it's not clear I can add comment.

front/main/config/environment.js
Bloodmoon24: it's just moving moment locales to the specific path to be sure that there will be in dir with public access. Do you have better idea?

front/common/public/locales/moment/de.json
```diff
@@ -0,0 +1,14 @@
+{
+  "longDateFormat": {
```
bkoval: Are these autogenerated files that might be overwritten? If not, we could apply our convention of whitespacing to them, like tab-based indentation.

front/main/app/initializers/moment.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * @param {*} container
+ * @param {*} application
+ *
+ * @returns {void}
+ */
+export function initialize(container, application) {
+	const language = Ember.get(Mercury, 'wiki.language.content') || 'en',
+		/**
+		 * Defaults, other languages than english are overriding this
```
bkoval: Nah, just a line comment is enough, no reason for a block comment

front/main/app/initializers/moment.js
```diff
@@ -0,0 +1,61 @@
...
+				mm: '%d m',
+				h: '1 h',
+				hh: '%d h',
+				d: '1 d',
+				dd: '%d d'
+			}
+		};
+
+	if (language !== 'en') {
+		application.deferReadiness();
```
bkoval: Yeah, now that is the part which worries me :)

front/common/public/locales/moment/de.json
rogatty: We already use spaces for JSON files everywhere. I can't see it in guidelines though.

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,30 @@
...
  * and the current time/date in the form
- * {timeAgo unixTimestamp}
- * which returns something like '2 days ago'
+ * {timeAgo unixTimestamp shouldHideAgoPrefix}
+ * which returns something like '2 days ago' if interval is below 6 days or formated param date
  *
- * @param {Array} params
+ * @param {Array}
+ * 		param[0] date in unix format
+ * 		param[1] bool should hide ago prefix, true defaults
```
rogatty: How about using destructured assignment and impressing all your plebeian co-workers?
http://emberup.co/mastering-ember-js-helper-arguments/

front/main/app/initializers/moment.js
bkoval: From offline talk: Let's try to merge our configuration with the one coming from moment on the backend side. If there's no way to do this, let's use inlining of the language configuration code into the mercury's backend response right away, to prevent blocking the whole app while waiting for the ajax call to be completed.

front/main/app/helpers/time-ago.js
bkoval: Yes, let's do the impressing co-workers part!

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,30 @@
...
-		return '';
+	if (now.diff(date, 'days') > 5) {
+		output = date.format('L');
+	} else if (now.diff(date, 'minutes') < 1) {
+		output = i18n.t('app.now-label');
+	} else {
+		output = date.fromNow(shouldHideAgoPrefix);
 	}
+
+	return `<span class='timestamp' title='${date.format('LLL')}'>&bull; ${output} </span>`;
```
rogatty: Let's use double quotes for HTML attributes:
https://github.com/Wikia/guidelines/tree/master/Handlebars#quotes

rogatty: Is this space at the end intentional?

front/main/app/templates/components/article-comment.hbs
```diff
@@ -2,7 +2,7 @@
 	<img src='{{unbound user.avatar}}' width='50px'>
 </div>
 <div class='content'>
-	<span class='timestamp'>{{time-ago comment.created}}</span>
+	{{time-ago comment.created}}
```
rogatty: `{{{`?

front/main/app/helpers/time-ago.js
Bloodmoon24: Looks awesome, but what about IE support?

front/main/bower.json
```diff
@@ -27,6 +28,7 @@
     "wikia-style-guide": "Wikia/style-guide#v1.8.2",
     "Autolinker.js": "0.22.0",
     "sinonjs": "1.17.1",
+    "moment-timezone": "0.5.0",
```
rogatty: Can we use this opportunity to sort these libs alphabetically?

Bloodmoon24: Indeed, that was bothering me ;)

front/main/app/helpers/time-ago.js
rogatty: We already use ES6 a lot, which is compiled by Babel to ES5 in the build process.

front/common/public/locales/moment/de.json
Bloodmoon24: @bkoval Nope, these are file which are extending the basic moment translations. @rogatty So spaces are staying.

front/main/config/environment.js
rogatty: This is how you configure ember-moment:
https://github.com/stefanpenner/ember-moment#write-all-the-locales-to-a-folder-relative-to-dist

front/main/app/helpers/time-ago.js
Bloodmoon24: :+1:

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,30 @@
...
-		return '';
+	if (now.diff(date, 'days') > 5) {
+		output = date.format('L');
+	} else if (now.diff(date, 'minutes') < 1) {
+		output = i18n.t('app.now-label');
+	} else {
+		output = date.fromNow(shouldHideAgoPrefix);
 	}
+
+	return `<span class="timestamp" title="${date.format('LLL')}">&bull; ${output}</span>`;
```
kvas-damian: I believe `LLLL` format is more suitable here, see: https://wikia-inc.atlassian.net/browse/SOC-1376?focusedCommentId=200295&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-200295

server/app/views/_helpers/moment.js
```diff
@@ -0,0 +1,20 @@
+import Logger from '../../lib/Logger';
+
+module.exports = function (lang) {
+	if (lang !== 'en') {
+		const sourcePath = `../../../../front/common/locales/moment/${lang}.json`;
+		let translation;
+
+		try {
```
kvas-damian: Please fallback to `en` when language doesn't exist.

server/app/views/_helpers/moment.js
```diff
@@ -0,0 +1,20 @@
+import Logger from '../../lib/Logger';
+
+module.exports = function (lang) {
+	if (lang !== 'en') {
```
kvas-damian: I'd feel better if we have consistent way of loading translations. Can we import for en here and delete `config` from `front/main/app/initializers/moment.js`?

front/main/app/helpers/time-ago.js
kvas-damian: What do you think about escaping `output`?

front/main/app/helpers/time-ago.js
Bloodmoon24: @hakubo @bognix  Does this fulfills you?

front/main/app/templates/components/article-comment.hbs
```diff
@@ -2,7 +2,9 @@
 	<img src='{{unbound user.avatar}}' width='50px'>
 </div>
 <div class='content'>
-	<span class='timestamp'>{{time-ago comment.created}}</span>
+	<span class="timestamp" title="{{timestamp-to-date post.creationDate.epochSecond}}">
+		{{{time-ago post.creationDate.epochSecond}}}
```
rogatty: `{{time-ago comment.created}}`

vforge:
/var/lib/jenkins/workspace/Mercury-PR-Checks/front/main/app/helpers/time-ago.js
  26:2  error  Expected space or tab after "//" in comment  spaced-comment

✖ 1 problem (1 error, 0 warnings)

front/main/app/helpers/time-ago.js
Bloodmoon24: Yeah, I overlooked that comment.  Now, when helper is not returning HTML we don't need escaping, but thanks for pointing that.

server/app/views/_helpers/moment.js
Bloodmoon24: Do you really wanna load/extend translation each time when you are using helper?

kvas-damian: If I understand code correctly it's once per request. IMO, it functionally doesn't matter if load it from `config` object or a file.

front/main/app/templates/components/article-comment.hbs
Bloodmoon24: Thanks!

server/app/views/_helpers/moment.js
Bloodmoon24: Sorry I Iooked at wrong file. At this moment in json files are keyword that are different from en relative time. If we will do this that way we'll need to 
1 ) store all translations, not only different than en in json files
or
2 )  load en or move config in its current state to backend helper, extend it here by the wiki language 
Both of 'em are not a hudge deal. Which path I should take?

front/main/package.json
```diff
@@ -44,6 +45,7 @@
     "ember-derequire": "0.1.2",
     "ember-disable-proxy-controllers": "1.0.1",
     "ember-export-application-global": "1.0.4",
+    "ember-moment": "5.1.0",
```
hakubo: actually why do we need both?
ember-moment and ember-cli-moment-shim?

sounds like those two are two separate projects, what am I missing?

front/main/app/templates/components/article-comment.hbs
```diff
@@ -2,7 +2,9 @@
 	<img src='{{unbound user.avatar}}' width='50px'>
 </div>
 <div class='content'>
-	<span class='timestamp'>{{time-ago comment.created}}</span>
+	<span class="timestamp" title="{{timestamp-to-date comment.created}}">
+		{{{time-ago comment.created}}}
```
hakubo: from what I see ember-moment exposes some awesome helpers already:
https://github.com/stefanpenner/ember-moment

can we refrain from writing custom ones?

front/common/public/locales/moment/de.json
```diff
@@ -0,0 +1,10 @@
+{
```
hakubo: from what I can see you don't need all these files:
ember-moment can copy them to our dist folder: https://github.com/stefanpenner/ember-moment#write-all-the-locales-to-a-folder-relative-to-dist

front/main/bower.json
```diff
@@ -14,20 +16,20 @@
     "hammerjs": "2.0.4",
     "headroom.js": "0.7.0",
     "i18next": "1.7.4",
-    "jquery.cookie": "1.4.1",
     "jquery": "2.1.4",
+    "jquery.cookie": "1.4.1",
     "loader.js": "3.5.0",
     "numeral": "1.5.3",
+    "moment": "2.8.0",
```
hakubo: if you added ember-moment you don't need to add it here
ember addons can install their dependencies on their own

hakubo:
This gets to some details of what ember addon is capable of: https://gist.github.com/kristianmandrup/ae3174217f68a6a51ed5

front/main/app/templates/components/article-comment.hbs
Bloodmoon24: The problem is that they can't take unix timestamp as an input param, co I had to write custom ones. Moreover ticket requires short relative date format, so I had to override moment's defaults

front/common/public/locales/moment/de.json
Bloodmoon24: These files are for shortening moment's relative time format which I have to override.

front/main/bower.json
Bloodmoon24: before `ember-cli-moment-shim 1.0.0 `there was need for that ( https://github.com/stefanpenner/ember-moment/commit/6068a259a6b4161b9f8a449794220f820c5823da ). I'll upgrade ember-moment to `6.0.0` and this one will be solve.

front/main/package.json
Bloodmoon24: As I can see `ember-cli-moment-shim` is `ember-moment` dependency, but I can be wrong.

rogatty: Actually, when you use `ember install ember-moment` it adds:
```
"ember-cli-moment-shim": "1.0.0",
"ember-moment": "6.0.0",
"moment": "2.11.2",
"moment-timezone": "0.5.0"
```
to package.json

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,27 @@
 import Ember from 'ember';
-import {timeAgo, interval as dateTimeInterval} from 'common/utils/dateTime';
+import moment from 'moment';
 
 /**
  * Helper to give textual representation of time interval between past date
  * and the current time/date in the form
- * {timeAgo unixTimestamp}
- * which returns something like '2 days ago'
+ * {time-ago unixTimestamp shouldHideAgoPrefix}
```
hakubo: {{time-ago

front/main/app/templates/components/article-comment.hbs
hakubo: kk!

server/app/views/_partials/server-data.hbs
```diff
@@ -4,6 +4,7 @@
 	M.prop('queryParams', {{{json queryParams}}});
 	M.prop('tracking', {{{json tracking}}});
 	M.prop('translations', {{{json (translations wikiVariables.language.content ns=localSettings.translationFiles)}}});
+	M.prop('momentTranslation', {{{json (moment wikiVariables.language.content)}}});
```
hakubo: can we rename moment to something like momentTranslationFiles
as this looks like invoking moment lib here

front/common/public/locales/moment/de.js
```diff
@@ -0,0 +1,78 @@
+//! moment.js locale configuration
```
hakubo: these files look like copy and paste from somewhere. can we install this and use it from vendor?

front/common/public/locales/moment/de.js
```diff
@@ -0,0 +1,78 @@
...
+//! locale : german (de)
+//! author : lluchs : https://github.com/lluchs
+//! author: Menelion Elensúle: https://github.com/Oire
+//! author : Mikolaj Dadela : https://github.com/mik01aj
+
+;
+(function (global, factory) {
```
hakubo: if this would be compiled we would not have to have this 'low level' stuff here.

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,40 @@
...
-		return i18n.t('app.days-ago-label', {count: interval.value});
-	case dateTimeInterval.Month:
-		return i18n.t('app.months-ago-label', {count: interval.value});
-	case dateTimeInterval.Year:
-		return i18n.t('app.years-ago-label', {count: interval.value});
-	default:
-		Ember.Logger.error('Unexpected date interval for timestamp', unixTimestamp);
-		return '';
+		if (!momentTranslationsService.get('isLoaded')) {
+			return '<span class="datePlaceholder"/>';
```
hakubo: classes should be dasherized instead of camel case

hakubo: also is it valid html? I thought span always have open/close tag.

front/main/app/initializers/moment-translation.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @param {*} container
+ * @param {*} application
+ *
+ * @returns {void}
+ */
+export function initialize(container, application) {
+	application.inject('helper:time-ago', 'momentTranslation', 'service:moment-translation');
```
hakubo: I am preety sure that in Helper you can also do
`momentTranslation: Ember.service.inject()`
or that did not work?

front/main/app/services/moment-translation.js
```diff
@@ -0,0 +1,44 @@
...
+			}
+		});
+		Ember.run.next(() => {
+			this.changeLoadingStatus();
+		});
+	},
+	loadTranslation(lang = 'en') {
+		this.changeLoadingStatus(false);
+		if (lang !== 'en') {
+			Ember.$.getScript(M.buildUrl({path: `/front/common/locales/moment/${lang}.js`})).done(() => {
```
kvas-damian: Can we make sure those assets are fingerprinted to avoid problems with caching in future?

front/main/app/initializers/moment-translation.js
Bloodmoon24: Yeah, they can. But there is the case, when loading 'en' translation -> 1st helper is not recomputing, for all other languages it works fine, 'cuz ya need to load that script. It could be too fast for observer to realize that value has changed or something like that, so I decided to initialize it there and it works fine ;)

front/main/app/services/moment-translation.js
Bloodmoon24: Thanks for mentioning, I'll find the way to do that.

front/common/public/locales/moment/de.js
rogatty: Not really, because there are some custom modifications made here.

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,44 @@
...
+export default Ember.Helper.extend({
+	momentTranslation: Ember.inject.service(),
+	onTranslationChange: Ember.observer('momentTranslation.isLoaded', function () {
+		this.recompute();
+	}),
+
+	compute([unixTimestamp, shouldHideAgoPrefix = true]) {
+		const date = moment.unix(unixTimestamp),
+			now = moment(),
+			momentTranslationsService = this.get('momentTranslation');
```
rogatty: Decide on one form, translation or translations ;)

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,44 @@
...
-		return i18n.t('app.days-ago-label', {count: interval.value});
-	case dateTimeInterval.Month:
-		return i18n.t('app.months-ago-label', {count: interval.value});
-	case dateTimeInterval.Year:
-		return i18n.t('app.years-ago-label', {count: interval.value});
-	default:
-		Ember.Logger.error('Unexpected date interval for timestamp', unixTimestamp);
-		return '';
+		if (!momentTranslationsService.get('isLoaded')) {
+			if (!momentTranslationsService.get('isLoading')) {
```
rogatty: How about moving this condition to `momentTranslationsService.loadTranslation()`?

front/main/app/helpers/timestamp-to-date.js
```diff
@@ -0,0 +1,27 @@
+import Ember from 'ember';
+import moment from 'moment';
+
+/**
+ * Helper to convert unix timestamp to date in momentjs localized format
+ * {timestamp-to-date unixTimestamp dateFormat}
```
rogatty: Double braces.

front/main/app/services/moment-translation.js
```diff
@@ -0,0 +1,50 @@
+import Ember from 'ember';
+import moment from 'moment';
+
+export default Ember.Service.extend({
+	isLoaded: false,
+	isLoading: false,
+	onLanguageChange: Ember.observer('Mercury.wiki.language.content', function () {
```
rogatty: I don't think this will work. You can observe only the properties of Ember object and `Mercury` is a simple global.

front/main/app/services/moment-translation.js
```diff
@@ -0,0 +1,50 @@
...
+				this.changeLoadingStatus();
+			}).fail((jqxhr, settings, exception) => {
+				Ember.Logger.error(`Can't get moment translation for ${lang} | ${exception}`);
+				this.setEnTranslation();
+			});
+		} else {
+			this.setEnTranslation();
+		}
+	},
+	init() {
```
rogatty: Use `this._super(...arguments)` in Ember's hooks.

http://emberjs.com/api/classes/Ember.Service.html#method_init
>NOTE: If you do override init for a framework class like Ember.View, be sure to call this._super.apply(this, arguments) in your init declaration! If you don't, Ember may not have an opportunity to do important setup work, and you'll see strange behavior in your application.

front/main/app/services/moment-translation.js
```diff
@@ -0,0 +1,50 @@
...
+		Ember.run.next(() => {
+			this.changeLoadingStatus();
+		});
+	},
+	loadTranslation() {
+		const lang = Ember.get(Mercury, 'wiki.language.content') || 'en';
+
+		this.changeLoadingStatus(false);
+		if (lang !== 'en') {
+			Ember.$.getScript(M.buildUrl({path: `/front/common/locales/moment/${lang}.js`})).done(() => {
```
rogatty: Because of the string concatenation those locales won't be revved. If you want them to be revved you need a simple array of all possible paths.

front/main/app/services/moment-translation.js
```diff
@@ -0,0 +1,50 @@
...
+			}).fail((jqxhr, settings, exception) => {
+				Ember.Logger.error(`Can't get moment translation for ${lang} | ${exception}`);
+				this.setEnTranslation();
+			});
+		} else {
+			this.setEnTranslation();
+		}
+	},
+	init() {
+		moment.locale('en', {
```
rogatty: You call `moment.locale('en'...` here and in `setEnTranslation()`. It's not clear what does what, can you explain in function's JSDoc?

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,44 @@
...
+				momentTranslationsService.loadTranslation();
+			}
+			return '<span class="date-placeholder"/>';
```
rogatty: Doesn't it cause security warnings in the console? I think this should be wrapped in:
```
return new Ember.Handlebars.SafeString(...);
```
see https://guides.emberjs.com/v2.3.0/templates/writing-helpers/#toc_escaping-html-content

front/main/app/services/moment-translation.js
Bloodmoon24: This is for testing purposes only. I'm using it with `Ember.set(Mercury, 'wiki.language.content', 'ja')` and it works ;)

front/main/app/services/moment-translation.js
Bloodmoon24: I was wondering how rev that, thanks for hint!

front/main/app/helpers/time-ago.js
Bloodmoon24: Yeah, this is essential for xss security.

front/main/app/services/moment-translation.js
Bloodmoon24: @rogatty is JSDoc clear now?

front/main/app/initializers/moment-translation.js
Bloodmoon24: After moving extending en translation to service's init() it works fine, so I'm back with injecting it in helper instead of initializer.

front/main/app/services/moment-translation.js
Bloodmoon24: Done!

front/main/app/helpers/time-ago.js
Bloodmoon24: Does the current solution correct?

front/main/app/helpers/time-ago.js
```diff
@@ -1,37 +1,41 @@
...
-		return i18n.t('app.days-ago-label', {count: interval.value});
-	case dateTimeInterval.Month:
-		return i18n.t('app.months-ago-label', {count: interval.value});
-	case dateTimeInterval.Year:
-		return i18n.t('app.years-ago-label', {count: interval.value});
-	default:
-		Ember.Logger.error('Unexpected date interval for timestamp', unixTimestamp);
-		return '';
+		if (!momentLocaleService.isLocaleLoaded()) {
+			return new Ember.Handlebars.SafeString('<span class="date-placeholder"> </span>');
```
bkoval: Is there a need for a space inside?

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,91 @@
...
+	},
+	/**
+	 * Changes status of downloading moment's locale to trigger helper's observers
+	 *
+	 * @param {boolean} done
+	 * @return {void}
+	 */
+	changeLoadingStatus(done = true) {
+		this.setProperties({
+			isLoaded: done,
```
bkoval: Since isLoading is always a negation of isLoaded, is it really needed?

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,91 @@
...
+		});
+	},
+	/**
+	 * Changes moment locale to en. It's loaded by default, so we don't need to download it
+	 *
+	 * @return {void}
+	 */
+	setEnLocale() {
+		moment.locale('en');
+		Ember.run.next(() => {
```
bkoval: Please, add a comment why do we need to introduce additional async operation here ;)

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,91 @@
...
+	/**
+	 * Returns true when locale is set, so helpers will display date instead of placeholders
+	 *
+	 * @return {boolean}
+	 */
+	isLocaleLoaded() {
+		if (this.isLoaded) {
+			return true;
+		} else if (!this.isLoading) {
+			this.loadLocale();
```
bkoval: This code seems like a side effect of a method that should only check if the locale if needed.

front/main/bower.json
```diff
@@ -1,7 +1,9 @@
 {
   "name": "main",
   "dependencies": {
+    "Autolinker.js": "0.22.0",
     "cropper": "0.11.0",
+    "dom-ruler": "0.1.13",
```
bkoval: Thanks for alphabetizing these properties!

front/main/app/services/moment-locale.js
Bloodmoon24: Need something to call `loadLocale()` only once

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,81 @@
+import Ember from 'ember';
+import moment from 'moment';
+
+export default Ember.Service.extend({
+	isLoaded: false,
+	isLoading: false,
+	// Path to all supported locales, so they can be revved
```
rafalkalinski: 'revved'?

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,81 @@
...
+	},
+	/**
+	 * Downloads locale for moment if content language is not en, otherwise just changes to en
+	 *
+	 * @return {void}
+	 */
+	loadLocale() {
+		if (!this.isLoading) {
+			const contentLang = Ember.get(Mercury, 'wiki.language.content'),
+				lang = this.localePath.hasOwnProperty(contentLang) ? contentLang : 'en';
```
rafalkalinski: so if 'localePath" object will inherit "contentLang" property form some parent object, then it should be always 'en'?

front/main/app/services/moment-locale.js
Bloodmoon24: https://github.com/rickharrison/broccoli-asset-rev

Now, when I look at this maybe I should write fingerprinted

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,81 @@
...
+
+			this.changeLoadingStatus(false);
+			if (lang === 'en') {
```
rafalkalinski: I would put 'en' to a constant, as a default value.

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,81 @@
...
+			isLoaded: done,
+			isLoading: !done
+		});
+	},
+	/**
+	 * Changes moment locale to en. It's loaded by default, so we don't need to download it
+	 *
+	 * @return {void}
+	 */
+	setEnLocale() {
```
rafalkalinski: Why special method for setting 'en' locale? Better thing would be a method which sets any given locale, with 'en' as a default.

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,81 @@
...
+					this.setEnLocale();
+				});
+			}
+		}
+	},
+	// Extends default en translation by needed relative time on init
+	init() {
+		this._super();
+		moment.locale('en', {
+			relativeTime: {
```
rafalkalinski: it would be good to define this structure as the moment-locale object property - just like with localePath.

front/main/app/services/moment-locale.js
Bloodmoon24: The English locale is build-in moment, so there is no possibility to load it like others. Moreover this code is used in 2 places, so I decided to extract it. Was that wrong?

rafalkalinski: Okay, let it be :)

front/main/app/services/moment-locale.js
Bloodmoon24: There shall be any situation like that, all paths to locales should be in one place, if there is other requirement, then this check have to be refactored

rafalkalinski:
LGTM

front/main/app/services/moment-locale.js
```diff
@@ -0,0 +1,83 @@
...
+			this.changeLoadingStatus();
+		});
+	},
+	/**
+	 * Downloads locale for moment if content language is not en, otherwise just changes to en
+	 *
+	 * @return {void}
+	 */
+	loadLocale() {
+		if (!this.isLoading) {
```
rogatty: We should use getter here to be sure it's an up to date value (it's set by setter).

rogatty:
Just one thing and it'll LGTM.

front/main/app/services/moment-locale.js
Bloodmoon24: How could I overlook that, thanks!

PR #1965, opened 2016-02-05
XW-439 | add favicon to index.html

hakubo:
Unfortunatelly when browser get 500 it retries to get the favicon
and on devboxes this url is broken...
and on sandboxes too

since this is low impact I guess we'll have to test it on production


rogatty:
LGTM

PR #1966, opened 2016-02-05
XW-816 | Disable Ember performance sender

hakubo:
We do not use this data at all right now
and it's a hit on performance on our users

~8KB gzipped code less 
We don't hook into every View on every page load


rogatty:
LGTM

vforge:
LGTM

PR #1967, opened 2016-02-05
CE-3325 Handle revisions made by Anonymous users

adamkarminski:
@hakubo @rogatty @kamilkoterba

### DIFF
--- a/front/main/app/components/user-avatar.js
+++ b/front/main/app/components/user-avatar.js
@@ -18,5 +18,7 @@ export default Ember.Component.extend({
 			title: this.get('profileName'),
 		});
 	}),
-	shouldWrapInHref: true
+	shouldWrapInHref: Ember.computed('anonymous', function () {
+		return this.get('anonymous') !== true;
+	})
 });
--- a/front/main/app/components/user-avatar.js
+++ b/front/main/app/components/user-avatar.js
@@ -18,7 +18,5 @@ export default Ember.Component.extend({
 			title: this.get('profileName'),
 		});
 	}),
-	shouldWrapInHref: Ember.computed('anonymous', function () {
-		return this.get('anonymous') !== true;
-	})
+	shouldWrapInHref: Ember.computed.not('anonymous')
 });
--- a/front/main/app/models/article-diff.js
+++ b/front/main/app/models/article-diff.js
@@ -66,6 +66,7 @@ ArticleDiffModel.reopenClass({
 					content = $(revision.diff['*']),
 					diffs = this.prepareDiff(content),
 					model = ArticleDiffModel.create({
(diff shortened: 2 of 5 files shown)


front/main/app/components/user-avatar.js
```diff
@@ -18,5 +18,7 @@ export default Ember.Component.extend({
 			title: this.get('profileName'),
 		});
 	}),
-	shouldWrapInHref: true
+	shouldWrapInHref: Ember.computed('anonymous', function () {
```
rogatty:
```
shouldWrapInHref: Ember.computed.not('anonymous')
```
http://emberjs.com/api/classes/Ember.computed.html#method_not

front/main/app/models/article-diff.js
```diff
@@ -76,13 +74,26 @@ ArticleDiffModel.reopenClass({
...
+				 * if it is not an anonymous one (userId !== 0). Otherwise, just resolve */
+				if (userId === 0) {
+					model.set('anonymous', true);
+					resolve(model);
+				} else {
+					UserModel.find({userId}).then((user) => {
+						model.setProperties({
+							user: user.name,
+							useravatar: user.avatarPath
+						});
```
rogatty: Should we set `anonymous: false`?

front/main/app/models/article-diff.js
```diff
@@ -76,13 +74,26 @@ ArticleDiffModel.reopenClass({
...
 						timestamp: revision.timestamp,
 						title: page.title,
-						user: user.name,
-						useravatar: user.avatarPath
-					}));
-				}).catch((error) => {
-					Ember.Logger.error(error);
-					reject(error);
-				});
+						user: i18n.t('diff.anonymous', {ns: 'recent-wiki-activity'})
```
rogatty: It looks strange here. How about moving it inside `if (userId === 0) {`?

front/main/app/templates/components/user-avatar.hbs
```diff
@@ -12,5 +12,9 @@
 	{{/if}}
 </figure>
 {{#if displayProfileName}}
-	<a href="{{profileUrl}}" class="avatar-username">{{profileName}}</a>
+	{{#if shouldWrapInHref}}
+		<a href="{{profileUrl}}" class="avatar-username">{{profileName}}</a>
+	{{else}}
+		{{profileName}}
```
rogatty: Is it still "user-avatar" if it's for anonymous and we don't display the avatar? ;)

adamkarminski: It is an anonymous user, still a user. :)

rogatty:
LGTM

PR #1973, opened 2016-02-07
CE-3324: Use new revisions diff api

lukaszkonieczny:
Use new revisions diff api

### DIFF
--- a/front/main/app/models/article-diff.js
+++ b/front/main/app/models/article-diff.js
@@ -1,6 +1,5 @@
 import Ember from 'ember';
 import getEditToken from '../utils/edit-token';
-import UserModel from './user';
 
 const ArticleDiffModel = Ember.Object.extend({
 	diffs: null,
@@ -97,8 +96,8 @@ ArticleDiffModel.reopenClass({
 	},
 
 	/**
-	 * Transforms diff data received from API to match required format
-	 * @param {Array} content
+	 * Prepares diffs data received from API
+	 * @param {Array} diffs
 	 * @returns {Array}
      */
 	prepareDiffs(diffs) {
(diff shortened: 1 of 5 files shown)


front/main/app/models/article-diff.js
```diff
@@ -57,142 +56,57 @@ ArticleDiffModel.reopenClass({
...
-					rvprop: 'timestamp|parsedcomment|userid',
-					revids: oldid,
-					rvdiffto: newid
+					controller: 'RevisionApi',
+					method: 'getRevisionsDiff',
+					avatar: true,
+					newId: newid,
+					oldId: oldid
 				}
 			).done((response) => {
```
hakubo: .done({article, revision, diffs})
you can use object destruct here

front/main/app/models/article-diff.js
```diff
@@ -57,142 +56,57 @@ ArticleDiffModel.reopenClass({
...
-	getDiff(diff, diffClass, allChanged) {
-		if (diffClass === 'diff-deletedline' || diffClass === 'diff-addedline') {
-			return {
-				content: Ember.String.htmlSafe(diff.html()),
-				class: diffClass,
-				allChanged
-			};
-		}
+	prepareDiffs(diffs) {
+		diffs.forEach((diff) => {
```
hakubo: consider using map here and create new array instead of modifying diffs array

front/main/app/models/recent-wiki-activity.js
```diff
@@ -26,12 +26,22 @@ RecentWikiActivityModel.reopenClass({
...
+				model.set('recentChanges', recentChanges);
 
-				model.set('recentChanges', data.query.recentchanges);
 				resolve(model);
 			}).fail((err) => reject(err));
 		});
+	},
+
+	prepareTimestamps(recentChanges) {
+		recentChanges.forEach((rc) => {
```
hakubo: consider using Array.map here also
also can we thing of better var name other than rc?

hakubo:
LGTM

azarek:
LGTM

PR #1976, opened 2016-02-08
(CE-3331) Do not show undo button to blocked users

Grunny:
Add blocked status to current user service and make use of it when
deciding whether or not to show the undo button.

Requires https://github.com/Wikia/app/pull/9674

/cc @Wikia/community-engineering


rogatty:
Apart from linter it LGTM

Grunny:
@rogatty fixed, thanks. :)

owend:
I think this is being called for anon users, which is causing a huge server problem...

PR #1979, opened 2016-02-08
CE-3317 Enable on aga.wikia.com for MVP testing

kamilkoterba:
We want to test it on pure new wikia as well
@Wikia/community-engineering


hakubo:
lgtm

PR #1983, opened 2016-02-09
XW-608 clean up prepare* functions

bognix:


server/app/facets/operations/preparePageData.js
```diff
@@ -0,0 +1,81 @@
+import localSettings from '../../../config/localSettings';
+import deepExtend from 'deep-extend';
+
```
bognix: I'm not sure if name of this file is the best, although it was the best idea that came to my mind at that time.

server/app/facets/operations/preparePageData.js
```diff
@@ -0,0 +1,81 @@
...
+export function getQualarooScriptUrl(request) {
+	// all the third party scripts we don't want to load on noexternals
+	if (!request.query.noexternals) {
+		// qualaroo
+		if (localSettings.qualaroo.enabled) {
+			return localSettings.qualaroo.scriptUrl;
+		}
+	}
+
+	return false;
```
bognix: does someone have a better idea how to write this function so it has consistent return type?

server/app/facets/operations/prepareCuratedContentData.js
```diff
@@ -19,69 +38,25 @@ export default function prepareCuratedContentData(request, curatedContentPageDat
...
-			result.openGraph.image = mainPageDetails.thumbnail;
-		}
-	}
 
+	result.openGraph = getOpenGraphData('website', wikiVariables.siteName, result.canonicalUrl, result.mainPageData);
 	// clone object to avoid overriding real localSettings for future requests
-	result.localSettings = deepExtend({}, localSettings);
+	result.localSettings = getLocalSettings();
 
 	if (typeof request.query.buckySampling !== 'undefined') {
```
bognix: it turned out that this query param is accessed in two places in front - trackPerf and in common.js#getTracker

server/app/facets/operations/preparePageData.js
hakubo: well you could return here empty string - but that does not sound great either
you could throw - but we don't do that often in our code

I think false is ok here
at least you can decide easily what to do and fallback onto something when you don't have this url

server/app/facets/operations/preparePageData.js
hakubo: Page sounds good for now
we might consider renaming it to prepareMWPageData to be more explicit

PR #1985, opened 2016-02-09
XW-41 | Video thumbnails

hakubo:


front/main/app/templates/components/article-media-gallery.hbs
```diff
@@ -1,12 +1,12 @@
...
-			forceHeight=imageSize
-			cropMode=cropMode
-		}}
-	</span>
+	{{article-media-image
+		url=item.url
+		link=item.link
+		ref=ref
+		galleryRef=item.galleryRef
+		itemType=(concat 'gallery-' item.type)
```
rogatty: `itemType` is passed here as an attribute and then used as a computed property. I don't think we should do that.

hakubo: do you have an idea how to do that more cleanly?

hakubo: or I could add another param
itemContext='article|gallery'

and use that in the component

how does that sound?

rogatty: The first thing that comes to my mind is that we already know it's a gallery (`galleryRef` attribute was passed) so we can use it in the computed property.

rogatty:
LGTM

bognix:
LGTM

PR #1986, opened 2016-02-09
CE-3351 Add tracking for Recent wiki activity feature

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3351

Added tracking for impressions of recent-wiki-activity and diff page
and click tracks on opening and closing confirmation modals

@Wikia/community-engineering, @rogatty

### DIFF
--- a/front/main/app/initializers/link-to.js
+++ b/front/main/app/initializers/link-to.js
@@ -16,8 +16,8 @@ export function initialize() {
 		 * @returns {boolean}
 		 */
 		_invoke(event) {
-			const action = this.get('action');
-			let trackingCategory = this.get('tracking-category'),
+			const action = this.get('action'),
+				trackingCategory = this.get('tracking-category'),
 				trackingLabel = this.get('tracking-label');
 
 			if (action) {
--- a/front/main/app/initializers/link-to.js
+++ b/front/main/app/initializers/link-to.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import {track, trackActions} from 'common/utils/track';
 
 /**
  * @returns {void}
@@ -16,6 +17,8 @@ export function initialize() {
 		 */
 		_invoke(event) {
 			const action = this.get('action');
+			let trackingCategory = this.get('tracking-category'),
+				trackingLabel = this.get('tracking-label');
 
 			if (action) {
 				// There was an action specified (in handlebars) so take custom action
@@ -25,6 +28,14 @@ export function initialize() {
 
 				// trigger the action on the controller
 				this.get('parentView').get('context').send(action, this.get('actionParam'));
+			}
+
+			if (trackingCategory && trackingLabel) {
+				track({
+					action: trackActions.click,
+					category: trackingCategory,
+					label: trackingLabel
+				});
 			}
 
 			return this._super(event);
--- a/front/main/app/routes/article-diff.js
+++ b/front/main/app/routes/article-diff.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import {updateTrackedUrl, trackPageView} from 'common/utils/track';
 import ArticleDiffModel from '../models/article-diff';
 
 export default Ember.Route.extend({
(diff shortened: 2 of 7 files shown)


front/main/app/routes/article-diff.js
```diff
@@ -8,5 +9,16 @@ export default Ember.Route.extend({
...
 		return ArticleDiffModel.fetch(params.oldid, params.newid);
+	},
+
+	actions: {
+		/**
+		 * @returns {boolean}
+		 */
+		didTransition() {
+			updateTrackedUrl(window.location.href);
+			trackPageView(null);
```
lukaszkonieczny: are we sure we dont want to send any context?

vforge: I would also ping @nandy-andy if that's a valid solution.

kamilkoterba: For an impression URL is enough context. Clicks are collected separately.

front/main/app/initializers/link-to.js
```diff
@@ -27,6 +28,14 @@ export function initialize() {
 				this.get('parentView').get('context').send(action, this.get('actionParam'));
 			}
 
+			if (this.get('tracking-category') || this.get('tracking-label')) {
+				track({
+					action: trackActions.click,
+					category: this.get('tracking-category'),
```
rogatty: You use them twice so how about extracting to a variable? Also, what about a default?

front/main/app/templates/components/article-diff.hbs
```diff
@@ -1,8 +1,12 @@
 <header class="sub-head">
-	{{#link-to 'recent-wiki-activity' class="sub-head--cancel"}}
+	{{#link-to 'recent-wiki-activity'
+	class="sub-head--cancel"
+	tracking-category='recent-wiki-activity'
+	tracking-label='sub-head--cancel'
```
rogatty: Let's add more indentation.

rogatty: Are we sure that we want to use CSS syntax in tracking labels?

front/main/app/routes/article-diff.js
lukaszkonieczny: If it's url like wiki.wikia.com/recent-wiki-activity (ie muppet.wikia.com/recent-wiki-activity) then it can be hard to collect and filter all data. Also for diff page we will get thousands (or more ;)) of urls because it consists of newId and oldId.

kamilkoterba: true, I revisited it and actually it would be hard to filter that.
I'm changing it to contain label: 'recent-wiki-activity' or 'article-diff'

front/main/app/initializers/link-to.js
```diff
@@ -27,6 +28,14 @@ export function initialize() {
 				this.get('parentView').get('context').send(action, this.get('actionParam'));
 			}
 
+			if (this.get('tracking-category') || this.get('tracking-label')) {
```
vforge: Shouldn't it be `&&` instead? If not, better provide defaults, as @rogatty suggested in comment below.

front/main/app/templates/components/side-nav-local-navigation-menu.hbs
```diff
@@ -10,7 +10,12 @@
 
 {{#if shouldDisplayRecentWikiActivity}}
 	<li class="side-nav-menu__item mw-content">
-		{{#link-to 'recent-wiki-activity'}}{{i18n 'main.title' ns='recent-wiki-activity'}}{{/link-to}}
+		{{#link-to 'recent-wiki-activity'
+		tracking-category='recent-wiki-activity'
```
vforge: Missing `\t`

front/main/app/templates/components/recent-change.hbs
```diff
@@ -18,7 +26,12 @@
 }}
 <small class="recent-change__time-ago">{{time-ago unixTimestamp}}</small>
 <h3 class="recent-change__article-title">
-	{{#link-to 'article' model.title}}{{model.title}}{{/link-to}}
+	{{#link-to 'article' model.title
+	tracking-category='recent-wiki-activity'
```
vforge: Missing `\t`

front/main/app/routes/article-diff.js
nandy-andy: Yep, we've talked offline.

front/main/app/initializers/link-to.js
kamilkoterba: I checked that it's required by GA anyway, I'll require both of them in this case.

rogatty:
LGTM

PR #1989, opened 2016-02-10
SOC-1933 | Registration modal broken on pt wikis

Bloodmoon24:
https://wikia-inc.atlassian.net/browse/SOC-1933
@bkoval @katanka  CR please

### DIFF
--- a/server/app/facets/auth/BirthdateInput.js
+++ b/server/app/facets/auth/BirthdateInput.js
@@ -1,5 +1,5 @@
 import authLocaleSettings from '../../../config/authLocaleSettings';
-import {isLanguageDefined} from '../../lib/AuthUtils.js';
+import {getLanguageWithDefault} from '../../lib/AuthUtils.js';
 
 /**
  * @typedef {Object} DateEndian
@@ -45,7 +45,7 @@ export default class BirthdateInput {
 	 */
 	constructor(endian, lang) {
 		this.endian = endian;
-		this.lang = isLanguageDefined(lang);
+		this.lang = getLanguageWithDefault(lang);
 
 		const langSettings = authLocaleSettings[this.lang];
 
--- a/server/app/facets/auth/BirthdateInput.js
+++ b/server/app/facets/auth/BirthdateInput.js
@@ -44,7 +44,7 @@ export default class BirthdateInput {
 	 * @returns {void}
 	 */
 	constructor(endian, lang) {
-		const langSettings = authLocaleSettings[getLanguageWithDefault(lang)];
+		const langSettings = authLocaleSettings[lang];
 
 		this.endian = endian;
 		this.settings = langSettings ? langSettings.date : authLocaleSettings.en.date;
--- a/server/app/facets/auth/BirthdateInput.js
+++ b/server/app/facets/auth/BirthdateInput.js
@@ -1,5 +1,4 @@
 import authLocaleSettings from '../../../config/authLocaleSettings';
-import {getLanguageWithDefault} from '../../lib/AuthUtils.js';
 
 /**
  * @typedef {Object} DateEndian
--- a/server/app/facets/auth/BirthdateInput.js
+++ b/server/app/facets/auth/BirthdateInput.js
@@ -44,11 +44,9 @@ export default class BirthdateInput {
 	 * @returns {void}
 	 */
 	constructor(endian, lang) {
-		let langSettings;
+		const langSettings = authLocaleSettings[getLanguageWithDefault(lang)];
 
 		this.endian = endian;
-		this.lang = getLanguageWithDefault(lang);
-		langSettings = authLocaleSettings[this.lang];
 		this.settings = langSettings ? langSettings.date : authLocaleSettings.en.date;
 	}
 
--- a/server/app/facets/auth/BirthdateInput.js
+++ b/server/app/facets/auth/BirthdateInput.js
@@ -44,11 +44,11 @@ export default class BirthdateInput {
 	 * @returns {void}
 	 */
 	constructor(endian, lang) {
+		let langSettings;
+
 		this.endian = endian;
 		this.lang = getLanguageWithDefault(lang);
-
-		const langSettings = authLocaleSettings[this.lang];
-
+		langSettings = authLocaleSettings[this.lang];
 		this.settings = langSettings ? langSettings.date : authLocaleSettings.en.date;
 	}
 
--- a/server/app/facets/auth/BirthdateInput.js
+++ b/server/app/facets/auth/BirthdateInput.js
@@ -1,4 +1,5 @@
 import authLocaleSettings from '../../../config/authLocaleSettings';
+import {isLanguageDefined} from '../../lib/AuthUtils.js';
 
 /**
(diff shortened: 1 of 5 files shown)


rogatty:
This fixes the issue for pt-br but maybe we should fall back to english when there is no configuration for language?

kvas-damian:
agree with @rogatty

bkoval:
+1 to a graceful fallback :)

Bloodmoon24:
@rogatty @kvas-damian @bkoval Thanks for that hint!

server/app/lib/AuthUtils.js
```diff
@@ -92,3 +93,11 @@ export function getWhoAmIUrl() {
 		pathname: localSettings.whoAmIService.path
 	});
 }
+
+/**
+ * @param {string} lang
+ * @returns {string}
+ */
+export function isLanguageDefined(lang) {
```
Warkot: Name of this function is confusing. `is` suggests it returns an answer to a yes/no question → `boolean`. I would rename it to something like `getLanguageWithDefault`, so it explicitly says what happens inside and what expect to be returned.

bognix: :+1:

server/app/lib/AuthUtils.js
```diff
@@ -92,3 +93,11 @@ export function getWhoAmIUrl() {
 		pathname: localSettings.whoAmIService.path
 	});
 }
+
+/**
+ * @param {string} lang
+ * @returns {string}
+ */
+export function getLanguageWithDefault(lang) {
+	return (authLocaleSettings.hasOwnProperty(lang)) ? lang : 'en';
```
bkoval: No need for the additional parenthesis around hasOwnProperty check

server/app/lib/AuthUtils.js
```diff
@@ -92,3 +93,11 @@ export function getWhoAmIUrl() {
 		pathname: localSettings.whoAmIService.path
 	});
 }
+
+/**
+ * @param {string} lang
+ * @returns {string}
+ */
+export function getLanguageWithDefault(lang) {
```
bkoval: lang = 'en'

server/app/facets/auth/BirthdateInput.js
```diff
@@ -43,10 +44,11 @@ export default class BirthdateInput {
 	 * @returns {void}
 	 */
 	constructor(endian, lang) {
-		const langSettings = authLocaleSettings[lang];
-
 		this.endian = endian;
-		this.lang = lang;
+		this.lang = getLanguageWithDefault(lang);
+
+		const langSettings = authLocaleSettings[this.lang];
```
bkoval: Use const definition at the beginning of the method :)

server/app/lib/AuthUtils.js
Bloodmoon24: Nope, It will not work here. We have to check is there translation which you demand, if no, serve en

Bloodmoon24:
@tomasznapieralski Thanks for fixing linter!

server/app/facets/auth/BirthdateInput.js
```diff
@@ -43,10 +44,11 @@ export default class BirthdateInput {
 	 * @returns {void}
 	 */
 	constructor(endian, lang) {
-		const langSettings = authLocaleSettings[lang];
+		let langSettings;
 
 		this.endian = endian;
-		this.lang = lang;
+		this.lang = getLanguageWithDefault(lang);
```
bkoval: Does this class actually use this property? Or maybe it's just used to get language settings? I cannot find any references to this field in this class...

server/app/facets/auth/register.js
```diff
@@ -43,7 +43,7 @@ import deepExtend from 'deep-extend';
  * @returns {DefaultRegistrationContext}
  */
 function getDefaultRegistrationContext(request, i18n) {
-	const lang = i18n.lng();
+	const lang = authUtils.getLanguageWithDefault(i18n.lng());
```
bkoval: Seems like obtaining lng() is a repeatable action - maybe we could pass the whole i18n object to the method and call the lng() method inside it?

server/tests/unit/lib/AuthUtils.js
```diff
@@ -0,0 +1,20 @@
+/* globals getLanguageWithDefault:true */
+
+QUnit.module('lib/AuthUtils');
+
+QUnit.test('getLanguageWithDefault', function (assert) {
+	var testCases = [
```
bkoval: You could add more edge cases here, like passing a variable of a wrong type (any non-string, undefined, null etc.)

server/app/lib/AuthUtils.js
bkoval: We could add it anyway, to fall back to english when there's no language provided.

server/tests/unit/lib/AuthUtils.js
Warkot: IMO, you should mock `authLocaleSettings`, to check just if the `getLanguageWithDefault` functions correctly and not if a language (in the case below `ja`) is defined in the `authLocaleSettings`.

You could use `beforeEach` and `afterEach` hooks like [here](https://github.com/Wikia/mercury/blob/dev/front/main/tests/unit/routes/article-test.js#L7-L17) and then set `authLocaleSettings` in every test case like I propose below.

```
var testCases = [
		{
			lang: 'ja',
			expected: 'ja'
			authLocaleSettingsMock: {ja: {}}
		},
		{
			lang: 'Quenya',
			expected: 'en'
			authLocaleSettingsMock: {}
		}
	];

	testCases.forEach(function (testCase) {
		authLocaleSettings = testCase.authLocaleSettingsMock;
		assert.equal(getLanguageWithDefault(testCase.lang), testCase.expected);
	});
```

server/app/facets/auth/BirthdateInput.js
```diff
@@ -43,10 +44,11 @@ export default class BirthdateInput {
 	 * @returns {void}
 	 */
 	constructor(endian, lang) {
```
tomasznapieralski: Have you considered this:
```
constructor(endian, lang) {
		this.endian = endian;
		this.lang = getLanguageWithDefault(lang);

		this.setSettings();
}

setSettings() {
		const langSettings = authLocaleSettings[this.lang];

		this.settings = langSettings ? langSettings.date : authLocaleSettings.en.date;
}
```
and using `dateSetting` instead of just `settings`?

server/app/lib/AuthUtils.js
```diff
@@ -92,3 +93,13 @@ export function getWhoAmIUrl() {
...
 }
+
+/**
+ * @param {object} i18n
+ * @returns {string}
+ */
+export function getLanguageWithDefault(i18n = {}) {
+	const lang = i18n.lng() || 'en';
+
+	return authLocaleSettings.hasOwnProperty(lang) ? lang : 'en';
```
rafalkalinski: is it possible, that authLocaleSettings.hasOwnProperty('en') would be empty?

rafalkalinski:
One comment, except that: LGTM

bkoval:
LGTM

PR #1990, opened 2016-02-11
CE-3264 Limit width of username for recent change module...

kamilkoterba:
to not interfere with arrow

@Wikia/community-engineering

### DIFF
--- a/front/main/app/styles/component/_recent-change.scss
+++ b/front/main/app/styles/component/_recent-change.scss
@@ -12,6 +12,7 @@
 
 	.avatar-username {
 		font-size: $type-min;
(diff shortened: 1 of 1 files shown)


lukaszkonieczny:
LGTM

front/main/app/styles/component/_recent-change.scss
```diff
@@ -12,6 +12,7 @@
 
 	.avatar-username {
 		font-size: $type-min;
+		max-width: 120px;
```
rogatty: You can use `@include ellipsis;` to make it look better if it's a single line.

PR #1995, opened 2016-02-12
XW-802 | Disable prototype extensions

hakubo:
https://wikia-inc.atlassian.net/browse/XW-802

I did disable prototype extensions for Functions and Strings
As in this case being explicit is easier to reason about what is happening

A left Array extensions - as this does make it easier to write code
disabling that would mean we'd have to wrap any [] in Ember.A()
I tried it and it did not look good - and I think it would lead us to more bugs than it's worth it

### DIFF
--- a/front/main/app/components/gallery-media.js
+++ b/front/main/app/components/gallery-media.js
@@ -35,21 +35,18 @@ export default MediaComponent.extend(
 		/**
 		 * @returns {void}
 		 */
-		setUp() {
-			const mediaArray = Ember.A(),
-				emptyGif = this.get('emptyGif');
-
+		setup() {
 			/**
 			 * @property {ArticleMedia} media
 			 * @property {number} index
 			 * @return {void}
 			 */
-			this.get('media').forEach((media, index) => {
+			const mediaArray = this.get('media').map((media, index) => {
 				media.galleryRef = index;
-				media.thumbUrl = emptyGif;
+				media.thumbUrl = this.emptyGif;
 				media.captionClass = Ember.get(media, 'caption.length') > 0 ? ' has-caption' : '';
 
-				mediaArray.pushObject(Ember.Object.create(media));
+				return Ember.Object.create(media);
 			});
 
 			this.setProperties({
@@ -108,7 +105,7 @@ export default MediaComponent.extend(
 				thumbSize = this.get('thumbSize'),
 				maxImages = Math.ceil(galleryWidth / thumbSize);
 
-			this.setUp();
+			this.setup();
 			this.loadImages(0, maxImages);
 
 			$this.on('scroll', () => {
--- a/front/main/app/mixins/curated-content-editor-sortable-items.js
+++ b/front/main/app/mixins/curated-content-editor-sortable-items.js
@@ -6,11 +6,7 @@ export default Ember.Mixin.create(
 	TrackClickMixin,
 	{
 		persistentSort: false,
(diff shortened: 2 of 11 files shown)


front/main/app/mixins/curated-content-editor-sortable-items.js
```diff
@@ -6,11 +6,7 @@ export default Ember.Mixin.create(
 	TrackClickMixin,
 	{
 		persistentSort: false,
-		sortableItems: Ember.computed('model.items', function () {
-			const items = this.get('model.items') || [];
-
-			return Ember.A().pushObjects(items);
-		}),
+		sortableItems: Ember.computed.reads('model.items'),
```
rogatty: Does it work when there are no items? It no longer defaults to an empty array, right?

front/main/app/mixins/discussion-upvote-component.js
```diff
@@ -10,7 +10,7 @@ export default Ember.Mixin.create({
 	post: null,
 	hasUpvoted: Ember.computed('post._embedded.userData.@each.hasUpvoted', function () {
 		if (Ember.isArray(this.get('post._embedded.userData'))) {
-			return this.get('post._embedded.userData')[0].hasUpvoted;
+			return this.get('post._embedded.userData.0.hasUpvoted');
```
rogatty: Are you sure this works? I suggested this syntax once to the Social Team and IIRC they said it doesn't.

/cc @bkoval @kvas-damian

hakubo: ha, ok, will check it out.

front/main/app/mixins/curated-content-editor-sortable-items.js
hakubo: I tested empty main page
added 3 items 
and I were able to sort it

rogatty: Cool, it seems that we have this fallback somewhere else too `¯\_(ツ)_/¯`

front/main/app/mixins/discussion-upvote-component.js
hakubo: I can upvote and it is reflecte in UI
no errors in console - any idea how else can I check it?

bkoval: I think we already started using this proposed syntax somewhere on our feature branches. It should be fine, we have automated tests that cover upvoting so we could run them against this branch or just wait for the run on preview since the manual check was OK.

front/main/app/components/gallery-media.js
```diff
@@ -108,7 +105,7 @@ export default MediaComponent.extend(
 				thumbSize = this.get('thumbSize'),
 				maxImages = Math.ceil(galleryWidth / thumbSize);
 
-			this.setUp();
+			this.setup();
```
bkoval: Why? ;>

bkoval:
Thanks, that makes sense!

front/main/app/components/gallery-media.js
hakubo: one word: http://dictionary.cambridge.org/dictionary/english/setup

rogatty:
LGTM

front/main/app/components/gallery-media.js
bkoval: :+1:

PR #1996, opened 2016-02-12
XW-1103 | Prepare support for non-article page types

vforge:
https://wikia-inc.atlassian.net/browse/XW-1103


server/app/lib/MediaWikiPage.js
```diff
@@ -5,14 +5,14 @@ import logger from './Logger';
...
  */
 
 /**
- * @class ArticleRequestError
+ * @class MediaWikiPageRequestError
  */
-export class ArticleRequestError {
+export class MediaWikiPageRequestError {
 	/**
 	 * @param {ArticlePageData} data
```
rogatty: `MediaWikiPageData`

server/app/lib/MediaWikiPage.js
```diff
@@ -100,16 +100,16 @@ export class ArticleRequestHelper {
 				}
 
 				data = {
-					article,
+					article: mediaWikiPage,
```
rogatty: `page`?

vforge: Can you explain a little more?

server/app/lib/MediaWikiPage.js
```diff
@@ -100,16 +100,16 @@ export class ArticleRequestHelper {
 				}
 
 				data = {
-					article,
+					page: mediaWikiPage,
```
rogatty: This is `MediaWikiPage.js` so I think using `page` would be better. Less to read / write and no need to rename it here :)

server/app/views/_partials/server-data.hbs
```diff
@@ -12,6 +12,9 @@
 	M.prop('servicesDomain', {{{json localSettings.servicesDomain}}});
 	M.prop('discussionBaseRoute', {{{json localSettings.discuss.baseAPIPath}}});
 	M.prop('imageReviewBaseRoute', {{{json localSettings.imageReview.baseAPIPath}}});
+	{{#if mediaWikiNamespace}}
+		M.prop('mediaWikiNamespace', '{{{mediaWikiNamespace}}}');
```
rogatty: How about moving it next to other MW stuff like `article`?

server/app/views/_partials/server-data.hbs
```diff
@@ -24,9 +27,9 @@
 	{{#if asyncArticle}}
 		M.prop('asyncArticle', {{asyncArticle}});
 	{{/if}}
-	{{#if article}}
+	{{#if page}}
```
rogatty: We want to do this only if it's a content namespace, right?

server/app/facets/api/article.js
```diff
@@ -1,4 +1,4 @@
-import * as Article from '../../lib/Article';
+import * as MediaWikiPage from '../../lib/MediaWikiPage';
```
hakubo: while at it 
can we refactor it so we don't have to do `* as `?
It would be good to either
return default
or import only used functions

server/app/facets/api/article.js
```diff
@@ -60,10 +60,10 @@ export default function get(request, reply) {
 		allowCache = false;
 	}
 
-	article = new Article.ArticleRequestHelper(params);
+	mediaWikiPageHelper = new MediaWikiPage.MediaWikiPageRequestHelper(params);
```
hakubo: Butter.Butter :)
MediaWikiPage.requestHelper ?

server/app/facets/mediaWikiPage.js
```diff
@@ -50,21 +56,31 @@ function redirectToMainPage(reply, article) {
...
-
-	// mainPageData is set only on curated main pages - only then we should do some special preparation for data
-	if (data.article.data && data.article.data.isMainPage && data.article.data.mainPageData) {
-		result = deepExtend(result, prepareMainPageData(data));
-		delete result.adsContext;
+function handleResponse(request, reply, data, allowCache = true, code = 200) {
+	let result, response;
+
+	switch (data.ns) {
+	case MediaWikiNamespace.MAIN:
```
hakubo: missing tab here

server/app/lib/MediaWiki.js
```diff
@@ -18,6 +18,35 @@ import Url from 'url';
  */
 
 /**
+ * This list is taken from MediaWiki:app/includes/Defines.php
+ * @type {{name: number}}
+ */
+export const MediaWikiNamespace = {
```
hakubo: it's already in MediaWiki module
how about name this namespaces?

server/app/lib/MediaWikiPage.js
```diff
@@ -22,15 +22,15 @@ export class ArticleRequestError {
 		this.data = data;
 	}
 }
-ArticleRequestError.prototype = Object.create(Error.prototype);
+MediaWikiPageRequestError.prototype = Object.create(Error.prototype);
```
hakubo: we use ES6 now - it think you can do:
```
export class MediaWikiPageRequestError extend Error {}
```

server/app/views/_partials/server-data.hbs
```diff
@@ -24,6 +24,9 @@
 	{{#if asyncArticle}}
 		M.prop('asyncArticle', {{asyncArticle}});
 	{{/if}}
+	{{#if mediaWikiNamespace}}
```
hakubo: I think we already have that in articleDetails, no?

server/app/facets/mediaWikiPage.js
vforge: Not according to https://github.com/Wikia/guidelines/blob/master/JavaScript/CodingConventions.md#switch-statements and `.eslint`

server/app/lib/MediaWikiPage.js
vforge: During our moving to ES6 for some reason this didn't work.

server/app/views/_partials/server-data.hbs
rogatty: We need the namespace even if it's not an article.

server/app/lib/MediaWiki.js
```diff
@@ -18,6 +18,35 @@ import Url from 'url';
...
+	// virtual namespaces
+	MEDIA: -2,
+	SPECIAL: -1,
+	// real namespaces
+	MAIN: 0,
```
rogatty: What if it's a content page but in a custom namespace? It's not really a part of this ticket but it would be good to think about the next step right now to not create a solution that will block the future (close one) work.

vforge: I think custom namespace can only be for content, right? So we can return `0` for non-default ones - because we're using this only for decide what component to use.

rogatty: Overriding it doesn't sound clean and surely can be confusing. I think we should pass two values to front end:
- `namespace`
- `isContentNamespace`

There won't be a lot of overhead and it will be very clear from the dev perspective.

front/main/app/components/mediawiki-category.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+import TrackClickMixin from '../mixins/track-click';
+
+export default Ember.Component.extend(
```
hakubo: const {Component} = Ember;

:D

front/main/app/routes/wiki-page.js
```diff
@@ -74,6 +84,16 @@ export default Ember.Route.extend({
 	},
 
 	/**
+	 * @param {Ember.controller} controller
+	 * @param {Ember.model} model
+	 * @returns {void}
+     */
+	setupController(controller, model) {
+		controller.set('model', model);
```
hakubo: setProperties

front/main/app/templates/wiki-page.hbs
```diff
@@ -1,6 +1,6 @@
 {{ad-slot name='INVISIBLE_HIGH_IMPACT' noAds=noAds}}
 {{ad-slot name='MOBILE_TOP_LEADERBOARD' noAds=noAds}}
-{{article-wrapper
+{{component wikiPageComponentName
 	commentsPage=commentsPage
```
hakubo: some of those properties won't have sense for some components. I wonder how could we improve it

front/main/app/utils/mediawiki-handlers/article.js
```diff
@@ -0,0 +1,115 @@
...
+
+		$descriptionMetaTag.prop('content', content);
+	}
+}
+
+
+/**
+ * @returns {string}
+ */
+function getComponentName() {
```
hakubo: does it have to be 'getter' maybe we could just export string?

front/main/app/utils/mediawiki-namespace.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * All supported namespaces
+ */
+const supportedNamespaces = [0, 14],
```
hakubo: can we add todo to add here all content namespaces?

server/app/facets/operations/prepareArticleData.js
```diff
@@ -27,15 +27,15 @@ export function getTitle(request, articleData) {
  * Prepares article data to be rendered
  *
  * @param {Hapi.Request} request
- * @param {ArticlePageData} data
+ * @param {MediaWikiPageData} data
  * @returns {object}
  */
 export default function prepareArticleData(request, data) {
```
hakubo: I would actually use destructuring here:
(request, {page})

vforge: I've just rename some stuff here - I've also noticed some common elements between `prepare*` functions and I've created https://wikia-inc.atlassian.net/browse/XW-1145 to fix all that mess. Is that solution acceptable for now instead of refactoring this one file/line?

front/main/app/utils/mediawiki-namespace.js
vforge: I've added Articles for now (because we're supporting only those ATM), plus Category namespace for this ticket - I thought it would be better to support all content namespaces in https://wikia-inc.atlassian.net/browse/XW-1069 .

front/main/app/templates/wiki-page.hbs
bognix: I think it would be cleaner to call renderTemplate hook inside wiki-page route with different components depending on which handler is used

front/main/app/routes/wiki-page.js
```diff
@@ -74,6 +84,18 @@ export default Ember.Route.extend({
 	},
 
 	/**
+	 * @param {Ember.controller} controller
+	 * @param {Ember.model} model
+	 * @returns {void}
+	 */
+	renderTemplate(controller, model) {
+		this.render(this.getHandler().viewName, {
```
hakubo: really minor: I would call getHandler once

front/main/app/utils/mediawiki-namespace.js
hakubo: :100:

server/app/facets/operations/prepareArticleData.js
hakubo: :100: x2

hakubo:
only minor things + I've noticed you have not used const {} = Ember in other files, but this can be done later

thanks!

front/main/app/utils/mediawiki-namespace.js
```diff
@@ -0,0 +1,55 @@
...
+		IMAGE_TALK: 7
+	};
+
+/**
+ * Returns current MW namespace if we're supporting them, false otherwise
+ *
+ * @returns {*}
+ */
+function getCurrentNamespace() {
+	const ns = parseInt(M.prop('mediaWikiNamespace'), 10);
```
hakubo: any chance to not use global?

front/main/app/utils/mediawiki-namespace.js
```diff
@@ -0,0 +1,55 @@
...
+
+	if (supportedNamespaces.indexOf(ns) > -1) {
+		return ns;
+	}
+
+	return false;
+}
+
+function isContentNamespace() {
```
hakubo: if above was not global - this could accept namespace as a param

front/main/app/utils/mediawiki-handlers/curated-main-page.js
```diff
@@ -0,0 +1,38 @@
...
+ * @returns {Object}
+ */
+function getCuratedContentModel(mainPageModel) {
+	if (mainPageModel.mainPageData.curatedContent) {
+		return CuratedContentModel.create({
+			type: 'section',
+			items: CuratedContentModel.sanitizeItems(mainPageModel.mainPageData.curatedContent)
+		});
+	}
+	return {};
```
hakubo: I think this should be Ember.Object.create - so it has the same api as CuratedContentModel

vforge:
All tests passed, merging

front/main/app/utils/mediawiki-namespace.js
vforge: Talked offline - but no.

PR #1998, opened 2016-02-12
CE-3362 Enable RWA on all english communities

kamilkoterba:
https://wikia-inc.atlassian.net/browse/CE-3362
@Wikia/community-engineering @hakubo


hakubo:
LGTM

PR #2000, opened 2016-02-14
Infobox builder - editing infobox elements

RafLeszczynski:
https://wikia-inc.atlassian.net/browse/DAT-3714
https://wikia-inc.atlassian.net/browse/DAT-3715
https://wikia-inc.atlassian.net/browse/DAT-3716

DO NOT MERGE - requires changes on app (PR to be prepared)


rogatty:
`infobox-builder-image-placeholder.png` seems to have 2px of empty space that shouldn't be there.

front/main/app/components/infobox-builder-edit-item.js
```diff
@@ -0,0 +1,43 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	editPanelOption: Ember.computed('item.type', function () {
+		return `infobox-builder-edit-item-${this.get('item.type')}`;
+	}),
+
+	header: Ember.computed('item.{data.label,type,value,data.default,infoboxBuilderData.index}', function () {
```
rogatty: Isn't this syntax awesome?

front/main/app/models/infobox-builder.js
```diff
@@ -1,29 +1,36 @@
...
+	 * @returns {void}
+	 */
+	init() {
+		this._super(...arguments);
+		this._itemIndex = {
+			row: 0,
+			image: 0,
+			title: 0
+		};
+		this.infoboxState = Ember.A([]);
```
rogatty: Do we need to do this explicitly? @hakubo did some investigation recently and he claims that it's done anyway by Ember's array prototype extension.

front/main/app/components/infobox-builder-edit-item.js
RafLeszczynski: we could work on improving readability :-)

front/main/app/models/infobox-builder.js
```diff
@@ -38,12 +45,12 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
 			},
 			source: `${itemType}${index}`,
-			type: xmlTag
+			type: itemType
 		});
 	},
 
 	/**
-	 * @returns {void}
+	 * @returns {ImageItem} added item
```
rogatty: Currently nothing is returned from this function.

front/main/app/models/infobox-builder.js
```diff
@@ -1,29 +1,36 @@
...
 	 */
 	addToState(object) {
 		this.get('infoboxState').pushObject(object);
+
+		return object;
 	},
 
 	/**
-	 * @returns {void}
+	 * @returns {RowItem} added item
```
rogatty: Currently nothing is returned from this function.

front/main/app/models/infobox-builder.js
```diff
@@ -126,8 +136,41 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
+	 * @param {RowItem} item
+	 * @param {string} value
+	 * @returns {void}
+	 */
+	editRowItem(item, value) {
+		const index = this.get('infoboxState').indexOf(item);
+
+		this.set(`infoboxState.${index}.data.label`, value);
+
+		if (value.trim().length) {
```
rogatty: Would it make sense to test if this is a string? Can there be an edge case where null or undefined is sent here?

RafLeszczynski: no, I could make it bullet proof and add the check but on other hand I'd probably prefer to throw error rather than silently skip setting source value

rogatty:
Minor things and it'll LGTM.

PR #2003, opened 2016-02-15
CE-3368: Fix RWA when preserved diff is not at the list

lukaszkonieczny:
Fix RWA when preserved diff is not at the list


tomasznapieralski:
LGTM

azarek:
LGTM

PR #2005, opened 2016-02-15
use postinstall hook instead of run setup

hakubo:
We can leverage post install hook
instead of our custom `npm run setup`
so we can just run `npm install` in our root folder.

What do you think? :)

Jenkins fill fail - but I want to get your opinion first.


kvas-damian:
Why didn't you show us it before? :)
:+1:

rogatty:
Cool, but I wouldn't deprecate `npm run setup` yet to not complicate things too much. Let's add this hook, merge it to dev, wait until most of branches have it merged, update deploy tools plugin and only then deprecate it. Slightly more work for us, less annoyance created.

tomasznapieralski:
For us this will be still using one command to install all, but this will complicate the way how we use cached dependencies. Especially in situation when in project root package.json change, because if it change, it will trigger all setups even if they shouldn't be triggered

hakubo:
@tomasznapieralski if nothing changes in all sub folders running npm/bower install is pretty quick and I think we are safer that way.

bkoval:
Cool!

hakubo:
not much of an improvement
but would cause our caching  system on jankins fail

let's just remember we can use pre/post hooks here :)

PR #2009, opened 2016-02-16
SOC-1906 reported indicator on posts

rafalkalinski:
the reported indicator on posts in the post list and on replies on the post details page

https://wikia-inc.atlassian.net/browse/SOC-1906

### DIFF
--- a/front/main/app/components/post-reply.js
+++ b/front/main/app/components/post-reply.js
@@ -7,8 +7,9 @@ export default Ember.Component.extend(
 	DiscussionMoreOptionsMixin,
 	{
 		classNames: ['post-reply'],
-		classNameBindings: ['isNew', 'isDeleted', 'isParentDeleted'],
+		classNameBindings: ['isNew', 'isDeleted', 'isReported', 'isParentDeleted'],
 
-		isDeleted: Ember.computed.alias('post.isDeleted')
+		isDeleted: Ember.computed.alias('post.isDeleted'),
+		isReported: Ember.computed.alias('post.isReported')
 	}
 );
--- a/front/main/app/styles/module/_discussion.scss
+++ b/front/main/app/styles/module/_discussion.scss
@@ -9,18 +9,6 @@ $post-details-reply-upvote-margin-left: 10px;
 $post-details-content-line-height: 1.4rem;
 $post-mark-icon-size: 12px;
 
-@mixin discussion-post-mark-indicator {
-	border-color: transparent;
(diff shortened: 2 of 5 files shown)


Bloodmoon24:
LGTM

front/main/app/components/post-reply.js
```diff
@@ -7,8 +7,9 @@ export default Ember.Component.extend(
 	DiscussionMoreOptionsMixin,
 	{
 		classNames: ['post-reply'],
-		classNameBindings: ['isNew', 'isDeleted', 'isParentDeleted'],
+		classNameBindings: ['isNew', 'isDeleted', 'isReported', 'isParentDeleted'],
```
bkoval: We could move the ones that are used for both posts and replies to one of the mixins, what do you think?

front/main/app/components/post-reply.js
```diff
@@ -7,8 +7,9 @@ export default Ember.Component.extend(
...
 
-		isDeleted: Ember.computed.alias('post.isDeleted')
+		isDeleted: Ember.computed.alias('post.isDeleted'),
```
bkoval: Looks movable to a mixin as well :)

front/main/app/components/post-reply.js
rafalkalinski: I would say no, since it's like two lines of codes for separate mixin.

front/main/app/styles/module/_discussion.scss
```diff
@@ -161,15 +173,15 @@ body.discussions {
...
-				top: 0;
-				width: $deleted-mark-size;
+				@include discussion-post-mark-indicator;
+				border-top-color: $discussion-deleted-mark-color;
+			}
+		}
+
+		&.is-reported {
+			&::before {
+				@include discussion-post-mark-indicator;
```
kvas-damian: Instead of including this mixin two times, maybe it's possible to have additional class name and applies styles only once?
That approach could limit size of output CSS file.

front/main/app/components/post-reply.js
rafalkalinski: Yeah, it would be the second line of code I mentioned in previous comment :)

front/main/app/styles/module/_discussion.scss
rafalkalinski: Yes, you're right.

bkoval:
L------G------T-------M

PR #2010, opened 2016-02-16
SOC-957 | Reporting posts and replies

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-957

@rafalkalinski @rogatty @Bloodmoon24 

Also, @adamkarminski and @kamilkoterba as I know you guys are planning similar features


front/main/app/components/discussion-more-options.js
```diff
@@ -14,4 +14,8 @@ export default Ember.Component.extend({
 	}),
 
 	canDeleteOrUndelete: Ember.computed.or('canDelete', 'canUndelete'),
+
+	canReport: Ember.computed('post.isReported', function () {
```
hakubo: you can use Ember.computed.not

front/main/app/mixins/discussion-moderation-controller.js
```diff
@@ -45,6 +45,24 @@ export default Ember.Mixin.create({
...
 			this.get('target').send('undeleteReply', reply);
-		}
+		},
+
+		/**
+		 * Bubbles up to Route
+		 * @param {object} post
+		 * @returns {void}
+		 */
+		reportPost(post) {
```
hakubo: in controller you have access to model
so you could do this.get('model')...

this would simplify code here a lot I think

front/main/app/mixins/discussion-moderation-route.js
```diff
@@ -95,6 +95,32 @@ export default Ember.Mixin.create({
...
+
+		/**
+		 * Pass post reporting to model
+		 * @param {object} post
+		 * @returns {void}
+		 */
+		reportPost(post) {
+			const loadingSpinnerContainer = this.getLoadingSpinnerContainer(post);
+
+			Ember.set(loadingSpinnerContainer, 'isLoading', true);
```
hakubo: loadingSpinnerContainer.set('isLoading', true)
Ember.set/get should not actually be used - many times it can be convenient, but we should try not to use.

front/main/app/components/discussion-more-options.js
bkoval: I used a full one as we might be adding more conditions here, so it's easier to refactor :)

front/main/app/mixins/discussion-moderation-controller.js
```diff
@@ -45,6 +45,24 @@ export default Ember.Mixin.create({
 		 */
 		undeleteReply(reply) {
 			this.get('target').send('undeleteReply', reply);
-		}
+		},
+
+		/**
+		 * Bubbles up to Route
```
tomasznapieralski: Description is not required by `eslint` do you think we should keep this `Bubbles up to Route` isn't that obvious?

front/main/app/mixins/discussion-moderation-controller.js
bkoval: You mean bypassing the route? We specifically move as much logic as possible to route to avoid building up controllers. This code is the slimmest one for the controller we can get until we finally need to remove it all once Ember gets rid of them ;)

front/main/app/mixins/discussion-moderation-route.js
bkoval: OK, I can see many places we use it for loading spinner. Created https://wikia-inc.atlassian.net/browse/SOC-2005 to refactor it.

front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -105,6 +105,26 @@ export default Ember.Mixin.create({
 		}
 	},
 
+	/**
+	 * Report post in service
+	 * @param {object} post
+	 * @returns {Ember.RSVP.Promise|void}
```
tomasznapieralski: i don't see situation when void is returned, do you?

front/main/app/templates/components/discussion-forum.hbs
```diff
@@ -44,17 +44,19 @@
...
 					pageNum=model.pageNum
 					posts=model.posts
 					postsDisplayed=model.posts.length
-					totalPosts=model.totalPosts
+					reportPost=(action this.attrs.reportPost)
 					siteHeadPinned=siteHeadPinned
+					totalPosts=model.totalPosts
+					undeletePost=(action this.attrs.undeletePost)
+					upvote=(action this.attrs.upvote)
+
```
tomasznapieralski: remove this empty line

front/main/app/mixins/discussion-moderation-route.js
bkoval: Changed.

BTW. Why is it wrong to use it?

tomasznapieralski:
LGTM

front/main/app/components/discussion-more-options.js
kvas-damian: It sounds like premature optimization ;)
Getting back to condition, we're aiming to allow post to be reported by multiple users. This condition doesn't allow moderators to report already reported item. Don't know if it's expected behavior or not. For more details about `isReported` flag see: https://wikia-inc.atlassian.net/wiki/pages/viewpage.action?pageId=38109866

front/main/app/mixins/discussion-moderation-route.js
```diff
@@ -91,10 +91,36 @@ export default Ember.Mixin.create({
...
+
+		/**
+		 * Pass post reporting to model
+		 * @param {object} post
+		 * @returns {void}
+		 */
+		reportPost(post) {
+			const loadingSpinnerContainer = this.getLoadingSpinnerContainer(post);
+
+			loadingSpinnerContainer.set('isLoading', true);
```
kvas-damian: On post details page, do we need to put overlay on entire page? This operation doesn't modify replies, IMO it should be enough to put loading overlay on post content only.

front/main/app/templates/components/post-reply.hbs
```diff
@@ -28,7 +28,9 @@
 			{{discussion-more-options
 				delete=(action 'delete')
 				deleteMessage=(i18n 'main.delete-reply' ns='discussion')
+				isReply=true
```
kvas-damian: Maybe instead of passing additional flag we could pass `reportMessage` to component?

front/common/public/locales/en/discussion.json
```diff
@@ -44,7 +44,9 @@
     "user-replied-to": "A reply to __userName__",
     "delete-all": "Delete All",
     "user-post-list-header": "Posts by __userName__",
-    "empty-forum-text": "No posts yet. Get the discussion started, create the first post now!"
+    "empty-forum-text": "No posts yet. Get the discussion started, create the first post now!",
+    "report-post": "Report Post",
+    "report-reply": "Report Reply"
```
rafalkalinski: Is there any localization ticket for that already created? :)

front/main/app/components/discussion-more-options.js
rafalkalinski: Actually there is a need currently to add here the second condition : 
post._embedded.userData.@each.hasUpvoted !== true :)

front/main/app/mixins/discussion-moderation-route.js
rafalkalinski: +1

front/main/app/mixins/discussion-more-options.js
```diff
@@ -27,6 +27,18 @@ export default Ember.Mixin.create({
...
 			this.attrs.undelete(item);
+		},
+
+		/**
+		 * Report an item and close pop-over
+		 * @param {object} item - post or reply
+		 * @returns {void}
+		 */
+		report(item) {
+			// There's no API for closing Ember Pop-Up, so that's
```
rafalkalinski: Actually it is possible to close it using pop-over api. It is not documented, but you can check how I did it in discussion-filters component:
https://github.com/Wikia/mercury/blob/9deb92fbc7ba1c61311f494bca0ae4f08c250298/front/main/app/components/discussion-filters.js#L18

front/main/app/components/discussion-more-options.js
bkoval: Seems like isReported should be set to false if a current user didn't report the post, at least to my understanding :)

front/main/app/templates/components/post-reply.hbs
bkoval: I'm not a fan of doing so. The less passing of the view-specific items (like i18n keys, messages etc.) we do throughout the logic part of app the better. Plus this variable might become handy at some point.

front/common/public/locales/en/discussion.json
bkoval: It's not posted to crowdin yet. Wait's for CR to finish as the shape of i18n might get changed here.

rafalkalinski:
LGTM

PR #2011, opened 2016-02-16
SOC-1242 update the status of a reported post by taking an action on the content

rafalkalinski:
When a post or reply is reported, the moderator will need to be able to mark the report as having been reviewed. They can do this by deleting the content or clearing the report in some other way. Clearing the report should essentially clear it from the post or reply but leave said post or reply visible to all users.

https://wikia-inc.atlassian.net/browse/SOC-1242

### DIFF
--- a/front/main/app/styles/module/_discussion.scss
+++ b/front/main/app/styles/module/_discussion.scss
@@ -7,7 +7,19 @@ $post-details-post-vertical-padding: 12px;
 $post-details-reply-upvote-width: 12px;
 $post-details-reply-upvote-margin-left: 10px;
 $post-details-content-line-height: 1.4rem;
-$deleted-mark-size: 12px;
+$post-mark-icon-size: 12px;
+
+@mixin discussion-post-mark-indicator {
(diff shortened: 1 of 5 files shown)


front/main/app/styles/module/_discussion.scss
```diff
@@ -7,7 +7,19 @@ $post-details-post-vertical-padding: 12px;
 $post-details-reply-upvote-width: 12px;
 $post-details-reply-upvote-margin-left: 10px;
 $post-details-content-line-height: 1.4rem;
-$deleted-mark-size: 12px;
+$post-mark-icon-size: 12px;
+
+@mixin discussion-post-mark-indicator {
```
hakubo: consider using placeholder here

bkoval: Per offline discussion, we'll drop the mixin and use a semicolon notation. Thanks!

rafalkalinski: It is going to be removed :)

PR #2013, opened 2016-02-16
XW-1133 | Don't collapse infobox when user clicks on image/video caption

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1133


Warkot:
As a P2 fix — LGTM. :+1:

PR #2017, opened 2016-02-17
DAT-3744 Infobox-builder: edit already existing template

dianafa:
https://wikia-inc.atlassian.net/browse/DAT-3744

### DIFF
--- a/front/main/app/models/infobox-builder.js
+++ b/front/main/app/models/infobox-builder.js
@@ -49,6 +49,7 @@ const InfoboxBuilderModel = Ember.Object.extend({
 				item = this.createImageItem();
 				break;
 			default:
+				Ember.Logger.warn(`Unsupported infobox builder type encountered: '${type}'`);
 				break;
 		}
 
--- a/front/main/app/models/infobox-builder.js
+++ b/front/main/app/models/infobox-builder.js
@@ -39,18 +39,18 @@ const InfoboxBuilderModel = Ember.Object.extend({
 		let item = {};
 
 		switch (type) {
-			case 'title':
-				item = this.createTitleItem();
-				break;
-			case 'row':
-				item = this.createRowItem();
-				break;
-			case 'image':
-				item = this.createImageItem();
-				break;
-			default:
-				Ember.Logger.warn(`Unsupported infobox builder type encountered: '${type}'`);
-				break;
+		case 'title':
+			item = this.createTitleItem();
+			break;
+		case 'row':
+			item = this.createRowItem();
+			break;
+		case 'image':
+			item = this.createImageItem();
+			break;
+		default:
+			Ember.Logger.warn(`Unsupported infobox builder type encountered: '${type}'`);
+			break;
 		}
 
 		item = InfoboxBuilderModel.extendItemData(item, elementData);
@@ -242,10 +242,7 @@ const InfoboxBuilderModel = Ember.Object.extend({
 	},
 
 	setupExistingState(state) {
-		state.forEach((element) => {
-			console.log('element: ', element);
-			this.addItem(element.type, element);
-		});
+		state.forEach((element) => this.addItem(element.type, element));
 	},
 
 	/**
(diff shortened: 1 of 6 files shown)


front/main/app/models/infobox-builder.js
```diff
@@ -27,13 +28,47 @@ const InfoboxBuilderModel = Ember.Object.extend({
 	},
 
 	/**
+	 * @desc add item to infobox state. Extend it with already existing
+	 * data if present.
+	 * @param {string} type type of element we want to add
+	 * @param {Object} elementData optional data if we already have some
```
vforge: If you're providing default for a param please not this in JSDoc:
`@param {Object} [elementData=null]`
It's optional - hence brackets and it defaults to `null`.

Warkot:
LGTM

vforge:
LGTM :shipit:

front/main/app/models/infobox-builder.js
```diff
@@ -220,10 +235,14 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
-		this.addImageItem();
-		this.addRowItem();
-		this.addRowItem();
+		this.addItem('title');
+		this.addItem('image');
+		this.addItem('row');
+		this.addItem('row');
+	},
+
+	setupExistingState(state) {
```
RafLeszczynski: missing JS doc

front/main/app/routes/infobox-builder.js
```diff
@@ -128,6 +139,28 @@ export default Ember.Route.extend({
...
+	 * @returns {Array} not modified promiseResponseArray
+	 */
+	setupInfoboxState(promiseResponseArray) {
+		const assets = promiseResponseArray[0].data;
+		let templateData = {};
+
+		if (assets) {
+			templateData = JSON.parse(assets);
+			if (templateData.data) {
+				this.controllerFor('infobox-builder').set('templateData', templateData.data);
```
RafLeszczynski: could you not use controller here for storing parsed template data and instead add it to `promiseResponseArray`?

front/main/app/routes/infobox-builder.js
```diff
@@ -30,7 +33,13 @@ export default Ember.Route.extend({
 	},
 
 	afterModel(model) {
-		model.setupInitialState();
+		const templateData = this.controllerFor('infobox-builder').get('templateData');
```
RafLeszczynski: Could you use promise passed between route hooks instead of model for passing `templateData` between route hooks?

front/main/tests/unit/models/infobox-builder-test.js
```diff
@@ -134,3 +202,50 @@ test('sanitize custom row source', (assert) => {
 		infoboxBuilderModelClass.sanitizeCustomRowSource(testCase.input), testCase.output
 	));
 });
+
+test('extend row data', (assert) => {
```
RafLeszczynski: please consider adding tests for extending `title` and `image` data

front/main/app/routes/infobox-builder.js
dianafa: when we add it to the `promiseResponseArray` we still'll have no access to it where we need it, that is in `afterModel` hook. Model neither seems to be a good place for it as it's at this stage only a bunch of raw data received from the server. Thus, using controller here seems reasonable.

hakubo:
minor things from me only. LGTM!

front/main/app/models/infobox-builder.js
```diff
@@ -268,6 +290,89 @@ InfoboxBuilderModel.reopenClass({
...
+	 * @returns {Object}
+	 */
+	extendImageData(item, itemData) {
+		if (itemData) {
+			item.source = itemData.source;
+			item.data.caption.source =
+				itemData.data &&
+				itemData.data.caption &&
+				itemData.data.caption.source ?
+					itemData.data.caption.source :
```
dianafa: maybe can we use object destructuring here?

RafLeszczynski:
LGTM

PR #2021, opened 2016-02-18
CE-3347: Banner to notify user about RWA

lukaszkonieczny:
We added banner with recent edit on given wiki. It appears on the bottom of screen for logged in users on article page. User can close it (banner will appear after 24h), go to RWA or dismiss displaying banner. For last two banner will never appear again.


lukaszkonieczny:
ping @hakubo @vforge @rogatty @bognix @Wikia/community-engineering

vforge:
@lukaszkonieczny You can use @Wikia/x-wing :)

front/main/app/models/recent-wiki-activity.js
```diff
@@ -1,18 +1,18 @@
...
+	RecentWikiActivityModel = Ember.Object.extend({
+		init() {
+			this._super(...arguments);
+			this.recentChanges = {};
+		}
+	});
 
 RecentWikiActivityModel.reopenClass({
 	/**
 	 * Gets the last 50 changes on a given wiki.
```
vforge: Please update JSDoc with missing
```
@param {number} [limit=50]
@param {string} [props=defaultProps]
```

front/main/app/styles/component/_recent-edit-banner.scss
```diff
@@ -0,0 +1,76 @@
...
+	width: 100%;
+
+	&.loaded {
+		animation-duration: 1s;
+		animation-iteration-count: 1;
+		animation-name: displayBanner;
+		bottom: 0;
+
+		&.dismissed {
+			animation-duration: 1s;
```
vforge: As `.dismissed` and `.loaded` are nested anyway I think you don't have to declare `animation-duration` and `animation-iteration-count`

lukaszkonieczny:
@vforge thanks I will use @Wikia/x-wing next time :)

front/main/app/components/article-wrapper.js
```diff
@@ -143,6 +143,10 @@ export default Ember.Component.extend(
 
 		curatedContentToolButtonVisible: Ember.computed.and('model.isMainPage', 'currentUser.rights.curatedcontent'),
 
+		displayRecentEdit: Ember.computed(function () {
```
bognix:
```
Ember.computed('currentUser.isAuthenticated', function() {
...
}
```

front/main/app/components/recent-edit-banner.js
```diff
@@ -0,0 +1,50 @@
...
+	classNameBindings: ['loaded', 'dismissed'],
+	dismissed: false,
+	loaded: false,
+	recentEdit: null,
+	recentWikiActivityLink: '/recent-wiki-activity',
+	init() {
+		this._super(...arguments);
+		RecentWikiActivityModel.getRecentActivityList().then((recentEdit) => {
+			this.set('loaded', true);
+			this.set('recentEdit', recentEdit.recentChanges.get('firstObject'));
```
bognix: minor note
consider using `this.setProperties` instead of `this.set` twice

front/main/app/styles/component/_recent-edit-banner.scss
```diff
@@ -0,0 +1,76 @@
...
+	h3 {
+		background-color: $color-blue-light;
+		color: $color-white;
+		font-size: $type-smallest;
+		margin: 0;
+		padding: 7px 16px;
+		text-transform: uppercase;
+	}
+
+	.recent-edit-close {
```
hakubo: do these have to be nested? can we flatten this file a bit?

azarek:
Testing done LGTM!

front/main/app/components/recent-edit-banner.js
```diff
@@ -0,0 +1,52 @@
...
+			action: trackActions.click,
+			category: 'recent-edit-banner',
+			label
+		});
+
+		this.set('dismissed', true);
+	},
+	actions: {
+		postpone(label) {
+			this.setCookie(1, label);
```
bognix: I would consider calling track inside actions not inside setCookie - this way setCookie method would be more generic and could be easily re-use in the future

front/main/app/models/recent-wiki-activity.js
```diff
@@ -1,18 +1,22 @@
...
-	 * Gets the last 50 changes on a given wiki.
+	 * Gets the last changes on a given wiki.
+	 *
+	 * @param {number} [limit=50] number of changes to fetch
+	 * @param {string} [props=defaultProps]
+	 *
 	 * @returns {Ember.RSVP.Promise}
 	 */
-	getRecentActivityList() {
+	getRecentActivityList(limit = 50, props = defaultProps) {
```
bognix: :+1:

front/main/app/components/recent-edit-banner.js
```diff
@@ -0,0 +1,52 @@
...
+		track({
+			action: trackActions.click,
+			category: 'recent-edit-banner',
+			label
+		});
+
+		this.set('dismissed', true);
+	},
+	actions: {
+		postpone(label) {
```
bognix: is it worth to send label along with action?

Consider hardcoding it here instead of passing it from template. This way it would be easier to debug this value in the future (no need to dig through files to determine from where it came from)

bognix:
Two minor notes - LGTM

front/main/app/components/recent-edit-banner.js
lukaszkonieczny: We are passing two different labels to dismiss action so I think it's worth to send it from template.

lukaszkonieczny:
Thank you :)

bognix:
LGTM

PR #2023, opened 2016-02-18
SOC-957 | SOC-1906 | SOC-1241 | SOC-1242 | SOC-1908 Reporting tools for Discussions

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-957
https://wikia-inc.atlassian.net/browse/SOC-1906
https://wikia-inc.atlassian.net/browse/SOC-1242
https://wikia-inc.atlassian.net/browse/SOC-1241
https://wikia-inc.atlassian.net/browse/SOC-1908


front/main/app/components/discussion-post-card-base.js
```diff
@@ -0,0 +1,21 @@
+import Ember from 'ember';
+import DiscussionParsedContentMixin from '../mixins/discussion-parsed-content';
+import DiscussionMoreOptionsMixin from '../mixins/discussion-more-options';
+
+/**
+ * Basic methods/properties for discussion-post-card-detail and discussion-post-card-reply.
+ */
+export default Ember.Component.extend(
```
hakubo: you can use destructuring here to make the class even more succinct 

const {Component, computed} = Ember;

front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -0,0 +1,81 @@
+import Ember from 'ember';
+import {checkPermissions} from 'common/utils/discussionPermissions';
+
+export default Ember.Component.extend({
+	classNames: ['top-note'],
+
+	canDelete: Ember.computed(function () {
+		return checkPermissions(this.get('post'), 'canModerate') && checkPermissions(this.get('post'), 'canDelete');
```
hakubo: I would consider using local variable for this.get('post')

front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -0,0 +1,81 @@
...
+	actions: {
+		/**
+		 * Delete item - shows modal dialog first
+		 * @param {object} item - post or reply
+		 * @param {boolean} isReply - if this is a reply
+		 * @returns {void}
+		 */
+		delete(item, isReply) {
+			this.get('modalDialogService').display(
+				i18n.t(`main.modal-dialog-delete${isReply ? '-reply' : ''}-text`, {ns: 'discussion'}),
```
hakubo: generating keys for translations is a root of all evil :)
can you please at list enumerate what keys you use here in JSDoc above

front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -14,7 +14,10 @@ export default Ember.Mixin.create({
 				method: 'PUT',
 				url: M.getDiscussionServiceUrl(`/${this.wikiId}/threads/${post.threadId}/delete`),
 				success: () => {
-					Ember.set(post, 'isDeleted', true);
+					Ember.setProperties(post, {
```
hakubo: is post an Ember object? 
can you use post.setProperties?

front/main/app/styles/module/_discussion.scss
```diff
@@ -439,6 +503,10 @@ body.discussions {
 				}
 				width: $post-details-outer-width;
 
+				&.show-top-note {
```
hakubo: getting deep here:
please consider BEM here to simplify css here

front/main/app/components/discussion-post-card-top-note.js
bkoval: Switched to classic if/else statement, good catch!

front/main/app/styles/module/_discussion.scss
bkoval: Spoke to @vforge about that, we'll be introducing BEM as we go with changing our stylesheets in the future.

front/main/app/mixins/discussion-moderation-model.js
bkoval: No, post is not an Ember object. It will be, after we're done with the mdoels refactor in https://wikia-inc.atlassian.net/browse/SOC-2084

front/common/public/locales/en/discussion.json
```diff
@@ -41,10 +41,31 @@
     "modal-dialog-delete-all-header": "Delete All Posts?",
     "modal-dialog-delete-all-message": "Are you sure you want to delete all posts by __userName__ on __wikiName__? Note that only individual posts can be restored and that you cannot undelete posts in bulk.",
     "modal-dialog-log-in": "Sign in",
+    "modal-dialog-delete": "Delete",
```
kvas-damian: Do we have translations somewhere?

front/main/app/services/discussion-sort.js
```diff
@@ -1,4 +1,5 @@
 export default Ember.Service.extend({
+	onlyReported: false,
```
kvas-damian: I believe sort service shouldn't be responsible for filtering. Maybe we should rename it?

front/main/app/components/discussion-filters.js
```diff
@@ -3,24 +3,64 @@ import nearestParent from 'ember-pop-over/computed/nearest-parent';
...
+
+			/**
+			 * Sets onlyReported flag in sync with onlyReported checkbox
+			 *
+			 * @param {event} event
+			 *
+			 * @returns {void}
+			 */
+			toggleOnlyReported(event) {
+				const isCheckboxChecked = event.target.checked;
```
kvas-damian: Maybe we could use Ember input helper and passed checked value?
See examples here: https://guides.emberjs.com/v2.4.0/templates/input-helpers/

front/main/app/routes/discussion/reported-posts.js
```diff
@@ -0,0 +1,85 @@
...
+				});
+			},
+
+			/**
+			 * @param {string} sortBy
+			 * @returns {void}
+			 */
+			setSortBy(sortBy) {
+				this.setSortBy(sortBy);
+			},
```
kvas-damian: I see a lot of C&P between routes. Can we make it less visible?

front/main/app/components/discussion-filters.js
```diff
@@ -3,24 +3,72 @@ import nearestParent from 'ember-pop-over/computed/nearest-parent';
...
 		sortBy: Ember.computed.oneWay('discussionSort.sortBy'),
 
-		popover: nearestParent('pop-over'),
+		onlyReportedClassName: Ember.computed('onlyReported', function () {
+			return this.get('onlyReported') === true ? 'active-element-background-color' : null;
+		}),
+
+		trendingDisabled: Ember.computed('onlyReported', function () {
+			return this.get('onlyReported') === true ? 'disabled' : undefined;
+		}),
```
kvas-damian: It's template logic, can we move it to hbs file?

front/common/public/locales/en/discussion.json
rafalkalinski: https://wikia-inc.atlassian.net/browse/LOC-278

front/main/app/services/discussion-sort.js
rafalkalinski: On some point the sorting becomes a part of filtering. And that's a good catch - it's probably should be renamed.

front/main/app/components/discussion-filters.js
rafalkalinski: I think Bart tried to do that this way and it didn't work. Input helper is used, but this is one way binding.

kvas-damian: Did you try this solution: http://stackoverflow.com/a/24155149 ?

bkoval: We use an action instead of observer which we consulted with X-Wing. We also try to avoid observers as they onfuscate the flow of the application.

front/main/app/routes/discussion/reported-posts.js
bkoval: Good catch, moved it to a mixin :)

front/main/app/services/discussion-sort.js
bkoval: Agreed. However, it's too risky to rename dependencies in that many files while the ticket is being QA-ed, and that's why I created https://wikia-inc.atlassian.net/browse/SOC-2127 for renaming.

front/main/app/components/discussion-filters.js
kvas-damian: I agree, observers makes application flow harder to follow, however it's key concept of Ember framework. Ember uses observers everywhere (directly - Ember.observer, under computed properties, in templates, ...), Ember is optimized for using observers and it's designed to for heavy usage of observers. I believe we agreed on using observers when we decided to use Ember and it's conventions over configuration rule. 

Avoiding observers mean that we have to "invent a wheel" and probably spend more resources on maintaining/migrating code in future.

BTW We had the same story with MW Hooks - we hate them but we have to use them.

/cc @hakubo @rogatty

front/main/app/routes/discussion/reported-posts.js
```diff
@@ -0,0 +1,95 @@
...
+			discussionSort.setOnlyReported(true);
+
+			return DiscussionReportedPostsModel.find(Mercury.wiki.id, params.forumId, this.get('discussionSort.sortBy'));
+		},
+
+		/**
+		 * @param {string} sortBy
+		 * @returns {EmberStates.Transition}
+		 */
+		setSortBy(sortBy) {
```
kvas-damian: I see you moved some code to mixin, although there is still a lot of code that is almost identical. The only difference is route name. What do you think about extracting all code that is in common by using: `this.get('routeName')`?

PR #2028, opened 2016-02-19
XW-1120 | send namespace & poweruser types to GA

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1120

/cc @Wikia/x-wing


tomasznapieralski:
Depends on https://github.com/Wikia/app/pull/9755

front/main/app/components/article-wrapper.js
```diff
@@ -248,7 +249,29 @@ export default Ember.Component.extend(
 				});
 
 				updateTrackedUrl(window.location.href);
-				trackPageView(model.get('adsContext.targeting'));
+
+				this.get('currentUser.powerUserTypes').then((powerUserTypes) => {
```
hakubo: you can use object destructuring here
`{poweruser_frequent, poweruser_lifetime}`

front/main/app/components/article-wrapper.js
```diff
@@ -248,7 +249,29 @@ export default Ember.Component.extend(
...
+					if (powerUserTypes.poweruser_frequent) {
+						UniversalAnalytics.setDimension(24, 'yes');
```
hakubo: to avoid repetion consider
`UniversalAnalytics.setDimension(24, (poweruser_frequent ? 'Yes' : 'No')`

hakubo: I am not entirely sure - but I think we send Yes and No capitalised

front/main/app/components/article-wrapper.js
```diff
@@ -248,7 +249,29 @@ export default Ember.Component.extend(
...
+
+					if (powerUserTypes.poweruser_lifetime) {
+						UniversalAnalytics.setDimension(23, 'yes');
+					} else {
+						UniversalAnalytics.setDimension(23, 'no');
+					}
+
+					trackPageView(model.get('adsContext.targeting'));
+				}).catch(() => {
+					if (this.get('currentUser.userId') !== null) {
```
hakubo: why do we have to check it here? what happens if we don't?

front/main/app/routes/article.js
```diff
@@ -48,12 +48,17 @@ export default Ember.Route.extend({
...
-			articleType = model.articleType;
+			articleType = model.articleType,
+			namespace = model.get('ns');
 
 		if (!Ember.isEmpty(exception)) {
 			Ember.Logger.warn('Article model error:', exception);
 		}
 
+		if (typeof namespace !== 'undefined') {
+			UniversalAnalytics.setDimension(25, `${namespace}`);
```
hakubo: I would use namespace directly here :)

front/main/app/routes/main-page.js
```diff
@@ -26,6 +27,10 @@ export default Ember.Route.extend(MainPageRouteMixin, {
 			title: Ember.getWithDefault(Mercury, 'wiki.siteName', 'Wikia')
 		});
 
+		if (typeof namespace !== 'undefined') {
+			UniversalAnalytics.setDimension(25, `${namespace}`);
```
hakubo: here as well

front/main/app/services/current-user.js
```diff
@@ -32,20 +32,45 @@ export default Ember.Service.extend({
 		return cookieUserId > 0 ? cookieUserId : null;
 	}),
 
+	userModel: Ember.computed(function () {
```
hakubo: I see how this could become more succinct
how about returning UserModel promise and in case we don't have userId
returning Ember.RSVP.Reject()?

front/main/app/services/current-user.js
```diff
@@ -32,20 +32,45 @@ export default Ember.Service.extend({
...
+					});
+			} else {
+				reject();
+			}
+		});
+	}),
+
+	powerUserTypes: Ember.computed('userModel', function () {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			this.get('userModel').then((userModel) => {
```
hakubo: here as well - I would think how not to create layers of promises
after all they are chainable

hakubo:
couple of notes left,
thanks Tomasz!

front/main/app/components/article-wrapper.js
rogatty: Can we keep it DRY and outside of article-wrapper / main-page? This logic doesn't really belong here. If extracted we could also write a test or two.

vforge: I have to agree with @rogatty on that - please don't WET :)

front/main/app/components/article-wrapper.js
vforge: Yea, and even the brackets are not needed

front/main/app/components/article-wrapper.js
hakubo: yes, from what I know - Tomek did not do it now as we wait for XW-1103 to land and move it to wiki-page so it has single parent, that is what we talked about yesterday.

front/main/app/components/article-wrapper.js
tomasznapieralski: we won't set dimension 23 & 24 is user is anon

front/main/app/routes/article.js
tomasznapieralski: GA says that string is expected

front/main/app/routes/main-page.js
tomasznapieralski: here as well :D

front/main/app/routes/article.js
vforge: I think he thought about `UniversalAnalytics.setDimension(25, namespace);` ;)

front/main/app/routes/main-page.js
vforge: here as well :>

front/main/app/routes/article.js
hakubo: or `String(namespace)`

front/main/app/services/current-user.js
```diff
@@ -32,20 +32,31 @@ export default Ember.Service.extend({
...
-
-		if (userId !== null) {
-			UserModel.find({userId})
-				.then((result) => {
-					this.setProperties(result);
-				})
-				.catch((err) => {
-					Ember.Logger.warn('Couldn\'t load current user model', err);
-				});
+		this.get('userModel').then((userModel) => {
```
hakubo: Ember.RSVP.allSettled([this.get('userModel'), this.loadUserInfo()]).then( ([userModel, userInfo]) => {
this.setProperties(userModel);
this.loadUserLanguage(userInfo);
...
})

front/main/app/routes/article.js
```diff
@@ -48,12 +48,17 @@ export default Ember.Route.extend({
 	 */
 	afterModel(model) {
 		const exception = model.exception,
-			articleType = model.articleType;
+			articleType = model.articleType,
+			namespace = model.get('ns');
```
rogatty: Why do we get `ns` with getter and `articleType` with dot notation? What's the difference? Can it be unified?

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
-				.then(this.loadBlockedStatus.bind(this))
-				.then(this.loadUserRights.bind(this))
-				.catch((err) => {
-					this.setUserLanguage();
-					Ember.Logger.warn('Couldn\'t load current user info', err);
-				});
-		} else {
-			this.setUserLanguage();
+			return UserModel.find({userId})
+					.catch((err) => {
```
rogatty: Too many tabs, I think.

front/main/app/components/main-page.js
```diff
@@ -44,7 +45,27 @@ export default Ember.Component.extend(
 				});
 
 				updateTrackedUrl(window.location.href);
-				trackPageView(this.get('adsContext.targeting'));
+
+				this.get('currentUser.userModel').then(({powerUserTypes}) => {
```
rogatty: `currentUser.userModel` returns a promise so this will make a call to UserApi on every article transition, right?

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
-	 * @param {string|null} [userLang=null]
 	 * @returns {void}
 	 */
-	setUserLanguage(userLang = null) {
-		const contentLanguage = Ember.getWithDefault(Mercury, 'wiki.language.content', 'en'),
-			userLanguage = userLang || contentLanguage;
+	init() {
+		Ember.RSVP.all([this.get('userModel'), this.loadUserInfo()]).then(([userModel, userInfo]) => {
+			if (userModel) {
+				this.setProperties(userModel);
```
rogatty: Something weird is going on here. `userModel` is an instance of `models/user` class and here we set its properties on `services/current-user` instance and yet in `components/article-wrapper` we use `currentUser.userModel.powerUserTypes`.

This is hard to read and prone to errors, would be nice to clean it up.

/cc @hakubo as I see that you moved a lot of things here.

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
+			if (userInfo) {
+				this.setUserLanguage(userInfo);
+				this.setBlockedStatus(userInfo);
+				this.setUserRights(userInfo);
+			}
+		});
 
-		this.set('language', userLanguage);
-		M.prop('userLanguage', userLanguage);
+		this._super(...arguments);
```
rogatty: This probably should be the first thing we do in this function.

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
+	 * @param {string|null} [userLang=null]
+	 * @returns {void}
 	 */
-	loadUserLanguage(result) {
-		return new Ember.RSVP.Promise((resolve) => {
-			const userLanguage = Ember.get(result, 'query.userinfo.options.language');
-
-			this.setUserLanguage(userLanguage);
+	setUserLanguage({query}) {
+		const userLanguage = query.userinfo.options.language;
```
rogatty: Seems pretty fragile, how about sticking with `Ember.get(query, 'userinfo.options.language')`?

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
 
-			resolve(result);
-		});
+		if (userLanguage) {
+			this.set('language', userLanguage);
+			M.prop('userLanguage', userLanguage);
```
rogatty: This isn't used anywhere but in this file, I believe it can be removed safely.

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
 
-		this.set('language', userLanguage);
-		M.prop('userLanguage', userLanguage);
+		this._super(...arguments);
 	},
 
 	/**
-	 * @param {QueryUserInfoResponse} result
-	 * @returns {Ember.RSVP.Promise<QueryUserInfoResponse>}
+	 * @param {string|null} [userLang=null]
```
rogatty: We pass an object here.

front/main/app/services/current-user.js
```diff
@@ -32,102 +32,84 @@ export default Ember.Service.extend({
...
-
-			if (!Ember.isArray(rightsArray)) {
-				reject(result);
-			}
+	setUserRights({query}) {
+		const rightsArray = query.userinfo.rights,
+			rights = {};
 
+		if (Ember.isArray(rightsArray)) {
+			// TODO - we could use contains instead of making an object out of an array
```
rogatty: We already use it in few places so it's probably tested.

front/main/app/components/article-wrapper.js
```diff
@@ -252,7 +253,27 @@ export default Ember.Component.extend(
...
+							24,
+							powerUserTypes.contains('poweruser_frequent') ? 'yes' : 'no'
+						);
+					} else {
+						UniversalAnalytics.setDimension(23, 'no');
+						UniversalAnalytics.setDimension(24, 'no');
+					}
+
+					trackPageView(model.get('adsContext.targeting'));
+				}).catch(() => {
```
hakubo: you can use finally so you don't have to repeat trackPageView

front/main/app/components/main-page.js
hakubo: no, you always work on the same promise

front/main/app/components/main-page.js
```diff
@@ -44,7 +45,27 @@ export default Ember.Component.extend(
...
+							24,
+							powerUserTypes.contains('poweruser_frequent') ? 'yes' : 'no'
+						);
+					} else {
+						UniversalAnalytics.setDimension(23, 'no');
+						UniversalAnalytics.setDimension(24, 'no');
+					}
+
+					trackPageView(this.get('adsContext.targeting'));
+				}).catch(() => {
```
hakubo: use finally here as well

front/main/app/services/current-user.js
hakubo: that is a good catch.
so I guess this is not needed here. we just need to be sure that we get that so request is done

front/main/app/services/current-user.js
hakubo: on the other hand Ember.get will hide some errors for us. 
This will brake and we'll know what is going on (api changed for example)

rogatty: I don't think we should test that on our users ;) API tests should catch that and we should fallback gracefully (and this would throw an error and probably stop Ember).

tomasznapieralski: For now, let't leave it as it is.

front/main/app/services/current-user.js
tomasznapieralski: It was checked before, don't change too many things

Warkot:
LGTM, tested, merging.

PR #2029, opened 2016-02-19
use link-to instead of a tag

hakubo:
todo: 
- [x] url should be in form of 'Namespace:Title' instead of '/wiki/Namespace:Title'


front/main/app/models/mediawiki/category.js
```diff
@@ -27,8 +29,42 @@ function getUrlBatchContent(categoryName, index, batch) {
...
 }
+/**
+ * add title to collectionItem based on url eg. /wiki/Namespace:Title -> Namespace:Title
+ *
+ * @param  {Array.<{url: string, name: string}>} collectionItems - array of items
+ * @returns {Array.<{url: string, name: string, title: string}>}
+ */
+function addTitlesToCollection(collectionItems) {
+	return collectionItems.map((item) => {
+		item.title = item.url.replace('/wiki/', '');
```
rogatty: Does it make sense to be safer and use `new RegExp('^/wiki/')`?

front/main/app/models/mediawiki/category.js
```diff
@@ -28,8 +30,42 @@ function getUrlBatchContent(categoryName, index, batch) {
...
+
+		return item;
+	});
+}
+
+/**
+ * Adds titles to collection
```
rogatty: Can you also explain when `collections` param is an object and when it's an array? This doc isn't really helpful, especially that there is another function called `addTitlesToCollection`.

front/main/app/models/mediawiki/category.js
hakubo: string replace replaces only first occurrence

rogatty: I guess we can safely assume that there is always `/wiki/` at the beginning and not worry about later occurrences, right?

rogatty:
LGTM

front/main/app/models/mediawiki/category.js
hakubo: we always send local path to client and all of our wikis have /wiki/ as article path

PS. this probably should be moved to api - now when I think about that. 
but would go with that for now

front/main/app/models/mediawiki/category.js
```diff
@@ -51,32 +87,21 @@ const CategoryModel = Ember.Object.extend({
...
 		const url = getUrlBatchContent(this.get('name'), index, batchToLoad);
 
-		return Ember.$.ajax({
+		return $.ajax({
 			url,
 			dataType: 'json',
 			method: 'get',
 		}).then((pageData) => {
-			this.set(
-				`collections.${index}.items`,
```
bognix: ticket for that already created you can annotate this method with @TODO - https://wikia-inc.atlassian.net/browse/XW-1165

front/main/app/models/mediawiki/category.js
bognix: ticket for that already created you can annotate this method with @TODO - https://wikia-inc.atlassian.net/browse/XW-1165

PR #2032, opened 2016-02-19
XW-1147 create dedicated handler for main pages

bognix:


front/main/app/models/mediawiki/wiki-page.js
```diff
@@ -0,0 +1,62 @@
...
+			Ember.$.ajax({
+				url: this.url(params),
+				dataType: 'json',
+				success: (data) => {
+					resolve(WikiPageModel.create(data));
+				},
+				error: (err) => {
+					// Temporary solution until we can make error states work - ideally we should reject on errors
+					if (err.status === 404 && isContentNamespace()) {
+						model = ArticleModel.create(params);
```
vforge: Please don't reuse variable from outside of the scope :)

PR #2035, opened 2016-02-21
DO NOT MERGE/DELETE | Optimize all SVGs in repo

vforge:
TESTING BRANCH, DO NOT MERGE/DELETE


bkoval:
Optimize -> not so much, more like prettify?

vforge:
@bkoval Optimize too - but if we want to see if svg has class, fill, color or it's better to ready prettyfied output - we're saving only 1byte per line (so it's on avg under 10bytes per file), but it leaves file in a readable state.

rogatty:
There are few optimized files and we should verify case by case if they still look the same.

As for new lines - I believe that having them in source files is good - for readability. But I don't think that we should serve them in that form to users. The less, the better. How about we minify them during the build process, only for production build?

bkoval:
@rogatty These svgs are concatenated into batch svg files, right? Removing them before building the batch file might cause some problems with concatenation. I'm not saying it will, just might. And since concatenation script handles them well, does it make sense to remove them during the build step?

bkoval:
Talked it through offline :)

vforge:
Closing this PR, as there's not much to add here :)

PR #2037, opened 2016-02-22
DAT-3770 infobox builder reordering of elements

dianafa:
#https://wikia-inc.atlassian.net/browse/DAT-3770

This PR allows for reordering elements in infobox builder.
It's introducing a fork of the https://github.com/jgwhite/ember-sortable because the original version didn't have the functionality we needed. Once `jgwhite` accept the PR and will introduce changes needed by us,  we need to change package.json to use that version of code.

@hakubo @rogatty can you please take a look?


front/main/app/components/infobox-builder.js
```diff
@@ -0,0 +1,38 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	classNameBindings: ['isPreviewItemDragged'],
+
+	init() {
+		this._super(...arguments);
+		this.toolTipPosX = null;
```
hakubo: I don't think it's needed, these are primitives and won't be shared between components
if we to do this everywhere - I think it'll create much overhead on development just to save us from couple of places where state can be leaked.

Maybe instead of that we should learn how this works and teach others?

front/main/app/components/infobox-builder.js
```diff
@@ -0,0 +1,38 @@
...
+		this.isPreviewItemHovered = false;
+		this.isPreviewItemDragged = false;
+		this.isReorderTooltipVisible = Ember.computed('isPreviewItemHovered', 'isPreviewItemDragged', function () {
+			return this.get('isPreviewItemHovered') && !this.get('isPreviewItemDragged');
+		});
+	},
+
+	actions: {
+		showReorderTooltip(posX, posY) {
+			this.set('toolTipPosX', posX + this.get('toolTipDistanceFromCursor'));
```
hakubo: you can usesetProperties

front/main/app/components/infobox-builder.js
```diff
@@ -0,0 +1,38 @@
...
+	classNameBindings: ['isPreviewItemDragged'],
+
+	init() {
+		this._super(...arguments);
+		this.toolTipPosX = null;
+		this.toolTipPosY = null;
+		this.toolTipDistanceFromCursor = 20;
+		this.isPreviewItemHovered = false;
+		this.isPreviewItemDragged = false;
+		this.isReorderTooltipVisible = Ember.computed('isPreviewItemHovered', 'isPreviewItemDragged', function () {
```
hakubo: This looks even weirdier - before we change how we write apps in Ember - maybe RFC would do where we can have a chat about that?

front/main/app/components/wikia-ui-components/on-hover-tooltip.js
```diff
@@ -0,0 +1,31 @@
...
+ *      message='lorem ipsum dolor'
+ *      fixed=true
+ * }}
+ *
+ * fixed attribute adds position fixed to the tooltip otherwise it will be absolutely positioned
+ */
+
+import Ember from 'ember';
+
+export default Ember.Component.extend({
```
hakubo: wait! we have just added tooltip addon - we should reuse that

front/main/app/routes/infobox-builder.js
```diff
@@ -223,93 +198,28 @@ export default Ember.Route.extend({
...
-				return this.redirectToTemplatePage(title);
+		redirectToTemplatePage(title) {
+			return new Ember.RSVP.Promise((resolve, reject) => {
+				const ponto = window.Ponto;
+
+				ponto.invoke(
+					'wikia.infoboxBuilder.ponto',
+					'redirectToTemplatePage',
+					title,
+					(data) => {
```
hakubo: you can just pass function as a reference

front/main/app/styles/component/_infobox-builder.scss
```diff
@@ -38,39 +43,59 @@ $border-width: 2px;
...
-			content: '';
-			display: block;
-			height: 100%;
-			left: 0;
-			position: absolute;
-			top: 0;
-			width: 100%;
+		// transitions during reordering
+		&.sortable-item {
+			transition: all .125s;
```
hakubo: can you list what is transitioned?
all sometimes leads to transition properties that user won't even see.

front/main/package.json
```diff
@@ -48,6 +48,7 @@
     "ember-in-viewport": "2.0.4",
     "ember-pop-over": "kvas-damian/ember-pop-over#5aa7c38d437618364e0e5090b57bd49098e3e39d",
     "ember-resolver": "2.0.3",
-    "ember-sinon": "0.3.0"
+    "ember-sinon": "0.3.0",
+    "ember-sortable": "dianafa/ember-sortable"
```
hakubo: did you submit PR from you fork?
can you add todo to use original repo?

rogatty: There is no PR opened and it seems unifinished (no tests, for example). We can't add comments to json files.

dianafa: https://github.com/jgwhite/ember-sortable/pull/87

front/main/app/components/wikia-ui-components/on-hover-tooltip.js
dianafa: From what I see here: https://github.com/kvas-damian/ember-pop-over this addon doesn't provide an option to reuse it in a way we need it here, that is a tooltip which follows the mouse cursor.

front/main/app/components/infobox-builder.js
dianafa: good suggestion and perfect topic for a blog post or a voice talk.

I think, by introducing this pattern, we wanted to avoid a surprise with sharing the same non-primitive objects by all component instances. However, if all developers knew about this issue, we could avoid writing code `in case of someone doesn't know this`. I'll remove this unneeded init() from here and commit myself to prepare some knowledge sharing in this topic. Thanks!

front/main/app/components/infobox-builder.js
dianafa: yes. as you can see it's a fresh idea - `classNameBindings:` are outside of init here, and inside of init() in `on-hover-tooltip.js`, so let's keep this idea of improvement here: https://wikia-inc.atlassian.net/wiki/display/MER/Mercury+TODOs

front/main/app/components/infobox-builder-item-row.js
```diff
@@ -1,10 +1,14 @@
...
+import SortableItem from 'ember-sortable/mixins/sortable-item';
 
 export default Ember.Component.extend(
+	SortableItem,
 	InfoboxBuilderItemMixin, {
 		classNames: ['pi-item pi-data', 'pi-item-spacing', 'pi-border-color'],
-		label: Ember.computed.oneWay('item.data.label'),
+		// spacing around the element when dragged, used in SortableItem
+		spacing: 8,
+		label: Ember.computed.oneWay('model.data.label'),
```
bognix: minor note: http://emberjs.com/api/classes/Ember.computed.html#method_readOnly

front/main/app/components/infobox-builder.js
```diff
@@ -0,0 +1,38 @@
...
+	isPreviewItemHovered: false,
+	isPreviewItemDragged: false,
+	isReorderTooltipVisible: Ember.computed('isPreviewItemHovered', 'isPreviewItemDragged', function () {
+		return this.get('isPreviewItemHovered') && !this.get('isPreviewItemDragged');
+	}),
+
+	actions: {
+		showReorderTooltip(posX, posY) {
+			this.setProperties({
+				toolTipPosX: posX + this.get('toolTipDistanceFromCursor'),
```
bognix: minor: sometimes it's tooltip and sometimes it's toolTip

front/main/app/styles/component/_infobox-builder.scss
```diff
@@ -242,5 +268,16 @@ $border-width: 2px;
...
+		content: "";
+		display: block;
+		position: absolute;
+		top: 0;
+		left: 0;
+		height: 100%;
+		width: 100%;
+		z-index: 2;
+		background-color: $grayout-color;
+	}
```
bognix: minor: alphabetical order

bognix:
minor notes - LGTM

front/main/app/components/wikia-ui-components/on-hover-tooltip.js
```diff
@@ -0,0 +1,27 @@
...
+	tagName: 'span',
```
vforge: `span` will break if we pass HTML as message. `div` would be better.

PR #2041, opened 2016-02-23
PLATFORM-1859 Add information about number of images still in the queue to be reviewed

RafalWilinski:
## Links

* https://wikia-inc.atlassian.net/browse/PLATFORM-1859

## Description

Add the ability to fetch amount of remaining images to review from `image-review` service.

## Reviewers
@tomasznapieralski  @bognix


front/main/app/models/image-review.js
```diff
@@ -124,6 +124,21 @@ ImageReviewModel.reopenClass({
 				reject(data);
 			});
 		});
+	},
+
+	getImagesToReviewCount(images, contractId) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: M.getImageReviewServiceUrl(`/monitoring?status=UNREVIEWED`),
```
bognix: instead of adding `?...` to url you can use `query`

front/main/app/models/image-review.js
```diff
@@ -124,6 +124,21 @@ ImageReviewModel.reopenClass({
...
+				xhrFields: {
+					withCredentials: true
+				},
+				dataType: 'json',
+				method: 'GET',
+				success: (data) => resolve(ImageReviewModel.sanitize(images, contractId, data.countByStatus)),
```
bognix: how about using http://emberjs.com/api/classes/RSVP.html#method_all instead of promise inside of promise?

RafalWilinski:
@bognix I replaced chained promises with extra `getImagesAndCount` method which uses `RSVP.allSettled`

bognix:
Looks much better, fix linter and LGTM

front/main/app/models/image-review.js
```diff
@@ -124,6 +124,35 @@ ImageReviewModel.reopenClass({
...
+
+	getImagesAndCount(contractId) {
+		const promises = [
+			ImageReviewModel.getImages(contractId),
+			ImageReviewModel.getImagesToReviewCount()
+		];
+
+		return Ember.RSVP.allSettled(promises).then((promisesArray) => {
+			return ImageReviewModel.sanitize(promisesArray[0].value.data, promisesArray[0].value.contractId,
+				promisesArray[1].value.countByStatus);
```
rogatty: Minor, cosmetic thing - you could use destructuring here (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) so it would look like this:
```
Ember.RSVP.allSettled(promises).then(([getImagesPromise, getImagesToReviewCountPromise]) => { ... });
```
and you wouldn't need this bad looking `promisesArray[1]`.

front/common/public/locales/en/image-review.json
```diff
@@ -7,6 +7,7 @@
     "error-invalid-data": "Invalid data was returned from Image Review API",
     "review-and-get-more": "Review & get next batch",
     "get-flagged": "Get flagged",
+    "images-to-review-count": "Remaining images to review: ",
```
rogatty: I think we shouldn't add spaces at the end of translation strings - translators can miss them and we wouldn't want it to break UX. Also, you already have a space in the template.

RafalWilinski:
Fixed, cc @rogatty

rogatty:
LGTM

PR #2044, opened 2016-02-24
(MAIN-6507) Only load user info if the user is not an anon

Grunny:
## Links

* https://wikia-inc.atlassian.net/browse/MAIN-6507

## Description

The check for userId being null was moved out of the load method in
this release, leading to this API call being made for anons as well.
This has caused large performance problems, so putting the check
back in as a quick fix.

## Reviewers

@owend @hakubo


owend:
LGTM, thanks G

rogatty:
LGTM, thanks for the fix!

PR #2046, opened 2016-02-24
Release 212

Grunny:
Merging changes from Release 212 after hotfix. Looks like this wasn't done for the normal release either?

And we have duplicated commits because we merged https://github.com/Wikia/mercury/pull/2044 before realising mercury merges release to dev for hotfixes, so sorry for the mess on that front. :)

/cc @Wikia/x-wing


rogatty:
@Grunny thanks for fixing P2 and sorry for the failing build. It was tooling issue and was fixed in https://wikia-inc.atlassian.net/browse/XW-1170. I've cherry picked it to this branch so we don't merge red stuff.

PR #2047, opened 2016-02-24
DAT-3786 | Infobox Builder: add dialog on onbeforeunload

rybmat:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-3786

## Description

Handling onbeforeunload event to prevent user from accidentally loosing changes.

## Reviewers
@Wikia/x-wing 
@dianafa


front/main/app/routes/infobox-builder.js
```diff
@@ -294,6 +306,7 @@ export default Ember.Route.extend({
 		save() {
 			const model = this.modelFor('infoboxBuilder');
 
+			window.onbeforeunload = null;
```
vforge: It would be better to revert to previous value of `window.onbeforeunload` instead of setting null.

rogatty:
How about using https://github.com/jasonmit/ember-onbeforeunload?
It provides a mixin that has more sophisticated logic for doing that: https://github.com/jasonmit/ember-onbeforeunload/blob/master/addon/mixins/confirmation.js

front/main/app/routes/infobox-builder.js
```diff
@@ -161,6 +162,17 @@ export default Ember.Route.extend({
 	},
 
 	/**
+	 * @desc setups handlers for events
+	 *
+	 * @returns {void}
+	 */
+	setupEventHandlers() {
+		window.onbeforeunload = function () {
```
Warkot: Wouldn't it be possible to use the fat arrow shorthand here like this?
```
window.onbeforeunload = () => i18n.t('infobox-builder:main.leave-confirmation');
```

rybmat: it should be possible, but I'm going to change it to what Igor proposed

Warkot: :+1:

front/main/app/controllers/infobox-builder.js
```diff
@@ -24,6 +27,9 @@ export default Ember.Controller.extend({
 		save() {
 			const model = this.get('model');
 
+			// prevents showing confirmation dialog on save
+			this.isDirty = false;
```
rogatty: ember-onbeforeunload uses getter (https://github.com/jasonmit/ember-onbeforeunload/blob/master/addon/mixins/confirmation.js#L23) so please use setter everywhere (`this.set('isDirty', false)`).

front/main/app/controllers/infobox-builder.js
```diff
@@ -1,6 +1,9 @@
 import Ember from 'ember';
 
 export default Ember.Controller.extend({
+
+	isDirty: false,
```
rogatty: Please add a comment that this is used by ember-onbeforeunload.

front/main/app/components/infobox-builder-wrapper.js
```diff
@@ -3,3 +3,4 @@ import Ember from 'ember';
 export default Ember.Component.extend({
 	classNames: ['infobox-builder-preview']
 });
+
```
rogatty: No need for this empty line.

rogatty:
LGTM

PR #2049, opened 2016-02-25
XW-1173 | ASAP send an event to UA that user opened a page

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1173

## Description

We want to verify how many page views are we missing in analytics because of the fact that we track them only after Ember is fully loaded. This experiment will send
```
category: 'mercury-app',
action: 'impression',
label: 'analytics-loaded'
```
event to UA as soon as possible. Then we'll compare its number with the one that we already send as soon as Ember is loaded.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/server/app/views/_partials/tracking-experiment.hbs
+++ b/server/app/views/_partials/tracking-experiment.hbs
@@ -30,7 +30,7 @@
 			hitType: 'event',
 			eventCategory: 'mercury-app',
 			eventAction: 'impression',
-			eventLabel: 'article-visible',
+			eventLabel: 'analytics-loaded',
 			eventValue: 0,
 			nonInteraction: true
 		});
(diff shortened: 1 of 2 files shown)


server/app/views/_partials/tracking-experiment.hbs
```diff
@@ -0,0 +1,38 @@
...
+			autoLinkDomain = [
+				'wikia.com', 'ffxiclopedia.org', 'jedipedia.de',
+				'marveldatabase.com', 'memory-alpha.org', 'uncyclopedia.org',
+				'websitewiki.de', 'wowwiki.com', 'yoyowiki.org'
+			].filter(function (domain) {
+				return document.location.hostname.indexOf(domain) > -1;
+			})[0];
+
+		ga('create', id, 'auto', options);
+		ga(`require`, 'linker');
```
hakubo: `` - this might not work

wladekb:
I'm curious to know the result. In Oasis we measured the number to be ~2%.

server/app/views/_partials/tracking-experiment.hbs
rogatty: It works in my browser without issues but yeah, I replaced them with single quotes.

vforge:
LGTM

PR #2050, opened 2016-02-25
XW-1086 | Mercury should still work even with cookies disabled

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1086
* https://github.com/Modernizr/Modernizr/blob/master/feature-detects/storage/localstorage.js

## Description

You can't actually call window.localStorage when cookies are disabled. It'll throw. So you have to do a check in try/catch. 

## Reviewers
@Wikia/x-wing

### DIFF
--- a/front/main/app/routes/application.js
+++ b/front/main/app/routes/application.js
@@ -6,8 +6,16 @@ import {normalizeToUnderscore} from 'common/utils/string';
 import Ads from 'common/modules/Ads';
 import getLinkInfo from '../utils/article-link';
 
-export default Ember.Route.extend(
-	Ember.TargetActionSupport,
+const {
+	$,
+	getWithDefault,
+	Logger,
+	Route,
+	TargetActionSupport,
+} = Ember;
+
+export default Route.extend(
+	TargetActionSupport,
 	TrackClickMixin,
 	{
 		queryParams: {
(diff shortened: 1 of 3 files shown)


front/main/app/routes/application.js
```diff
@@ -112,7 +120,11 @@ export default Ember.Route.extend(
 				 */
 				if (target.className.indexOf('external') > -1) {
 					if (target.href.indexOf(`useskin=${defaultSkin}`) > -1) {
-						document.cookie = `useskin=${defaultSkin}${cookieDomain}; path=/`;
+						$.cookie('useskin', defaultSkin, {
+							domain: cookieDomain,
+							expires: date,
```
rogatty: `date` is undefined here.

rogatty:
LGTM

hakubo:
I saw you walking a minute ago, are you doing CR through phone ? :)

front/main/app/utils/local-storage-connector.js
```diff
@@ -55,9 +55,22 @@ export const localStorageAdapter = {
 	}
 };
 
+// Check if we have local storage available
+let localStorageAvailable = false;
```
bognix: minor: i would scope it inside method

hakubo: taken offline

variables inside module are scoped only to this module anyway

hakubo:
Tested manually.

PR #2055, opened 2016-02-26
DAT 3732 Infobox builder header section editing

RafLeszczynski:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-3732
* https://github.com/Wikia/app/pull/9832

## Description

Add ability to edit section header element using infobox builder UI

## Reviewers
@dianafa


front/main/app/components/infobox-builder-edit-item-section-header.js
```diff
@@ -0,0 +1,35 @@
...
+		set(key, value) {
+			this.get('editSectionHeaderItem')(this.get('item'), {collapsible: value});
+			return value;
+		}
+	}),
+
+	init() {
+		this._super(...arguments);
+		this.isHelpVisible = false;
+		this.classNames = ['sidebar-content-padding'];
```
vforge: You don't need to assign values in `init()` function in this case - see https://github.com/Wikia/mercury/pull/2055/files#diff-eefbfc8c2642ba8b1f8be514c7cdb243R11 below.

hakubo: https://poteto.github.io/component-best-practices/#/leaky

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -1,25 +1,17 @@
...
+		labelValue: Ember.computed('item.data.label', {
+			get() {
+				return this.get('item.data.label');
+			},
+			set(key, value) {
+				this.get('editRowItem')(this.get('item'), value);
+				return value;
+			}
+		})
 	}
```
dianafa: don't we need anymore `showHelp` here?

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -1,25 +1,17 @@
...
-
-	actions: {
-		showHelp() {
-			this.set('isHelpVisible', true);
-		}
+export default Ember.Component.extend(
+	InfoboxBuilderEditItemMixin,
+	{
+		labelValue: Ember.computed('item.data.label', {
+			get() {
```
dianafa: shouldn't we add here also `editRowItem` and `item` ?

front/main/tests/unit/models/infobox-builder-test.js
```diff
@@ -542,3 +602,76 @@ test('extend image data', (assert) => {
...
+				},
+				expected: {
+					data: '',
+					collapsible: true
+				}
+			},
+			{
+				newValues: null,
+				expected: {
+					data: defaultName,
```
dianafa: if would be nice to add here case when data is `null`, data is `{}` and data is `''` and newValues is `{}`. what do you think?

front/main/app/components/infobox-builder-edit-item-row.js
dianafa: we shouldn't :) :+1:

front/main/tests/unit/models/infobox-builder-test.js
RafLeszczynski: done

dianafa:
LGTM

PR #2063, opened 2016-02-26
Jenkins improve

tomasznapieralski:
## Description 
- bump karma version (fix socket hanging)
- ember test will use existing build

/cc @Wikia/x-wing

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "start": "nodemon --delay 2",
     "start-on-change": "nodemon --delay 2 --on-change-only",
     "test": "npm run build-dev && npm run test-front && npm run test-server",
-    "test-front": "gulp test-common && gulp test-auth && cd front/main && ember test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "start": "nodemon --delay 2",
     "start-on-change": "nodemon --delay 2 --on-change-only",
     "test": "npm run build-dev && npm run test-front && npm run test-server",
-    "test-front": "gulp test-common && gulp test-auth && cd front/main && ember test",
+    "test-front": "gulp test-common && gulp test-auth && cd front/main && ember test --path ../../www/front/main",
```
hakubo: What if I haven't build main before?

rogatty:
LGTM

package.json
rogatty: It won't work anyway, notice that `npm run test` does the build first and then runs `test-front`.

Currently `test-front` isn't independent.

hakubo: gotcha

hakubo:
lgtm

PR #2064, opened 2016-02-26
XW-957 | Clean up server logs

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-957

## Description

This PR does two things:
1. Stops logging 404s that MediaWiki API responds with. There is ~750000 of them per day.
2. Removes redundant logging which is covered in `server/app/lib/MediaWiki.js`.

## Reviewers

@Wikia/x-wing


tomasznapieralski:
LGTM

server/app/lib/MediaWiki.js
```diff
@@ -100,21 +100,25 @@ function requestCallback(params) {
...
-		const info = {
-			exception: {
-				message: 'Invalid response',
-				code: response.statusCode,
-				details: payload ? payload.toString('utf-8') : null
-			}
-		};
+		const rejectData = payload || {};
+
+		// Don't flood logs with 404s (there is a LOT)
```
hakubo: minor: (there is a LOT) - it's subjective and not that useful, I'd remove that :)

hakubo:
LGTM

Warkot:
:+1:

server/app/lib/MediaWiki.js
```diff
@@ -100,21 +100,25 @@ function requestCallback(params) {
...
+		}
 
-		Logger.error({
-			url,
-			headers: response.headers,
-			statusCode: response.statusCode
-		}, 'Bad HTTP response');
+		// Make sure that we have exception code as we rely on it later
+		if (!rejectData.exception || !rejectData.exception.code) {
+			rejectData.exception = rejectData.exception || {};
```
bognix: it may get problematic if an exception is not an Object but "string" for instance - the code would be set incorrectly to undefined.

Same thing if rejectData.exception would be null or undefined - the TypeError would be thrown with message `Cannot read property exception of null`

rogatty: Few lines above is
```
const rejectData = payload || {};
```
so it's never `undefined`. If exception is a string then it will work without issues:
```
> y = {exception: ''};
  if(!y.exception || !y.exception.code) {
      y.exception = y.exception || {};
      y.exception.code = 500;
  }

<- 500
```

rogatty: For the record - this should be fixed as it will break for `y = {exception: 'non empty string'};`

PR #2065, opened 2016-02-26
XW-1104 : handle exception

tomasznapieralski:
Remaining:
- [ ] navigate from 404 page


front/main/app/routes/wiki-page-error.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+
+	/**
+	 * @param {Ember.Controller} controller
+	 * @param {Ember.Model} model
+	 * @returns {void}
+	 */
+	renderTemplate(controller, model) {
```
bognix: I would do another check - if model has code or if model is an object to be sure this code won't throw in case of model being `undefined`

front/main/app/utils/wiki-handlers/wiki-page.js
```diff
@@ -60,6 +60,8 @@ export default function getPageModel(params) {
 				ArticleModel.setArticle(model);
 				return resolve(model);
 			}
+		} else if (Mercury.exception) {
+			return reject(Mercury.exception);
```
bognix: you should unset this value after you check it to be sure the code won't reach this `if` on proceeding transitions

front/main/app/utils/wiki-handlers/wiki-page.js
```diff
@@ -70,22 +72,7 @@ export default function getPageModel(params) {
...
-				if (err.status === 404) {
-					const model = ArticleModel.create(params);
-
-					ArticleModel.setArticle(model, err.responseJSON);
-					resolve(model);
-				} else {
-					reject(err);
-				}
-			});
+			.fail(reject());
```
bognix: reject with `err`

bognix:
@tomasznapieralski great work with cleaning up! Thank you!

PR #2069, opened 2016-02-29
Add version to Ember inspector

hakubo:
## Links

* https://github.com/embersherpa/ember-cli-app-version

## Description

By removing version from package.json I enabled ember-cli-app-version to do that for us
It already uses git latest commit hash and that is enough for us.

## Reviewers

@Wikia/x-wing 

PS. We actually talked with the team about that recently - when I found out what it takes to make that work - I lol'ed...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "name": "Mercury",
-  "version": "0.0.0",
   "description": "Wikia Web Application",
   "main": "www/server/app/server.js",
   "scripts": {


tomasznapieralski:
lol, that was simple :D

tomasznapieralski:
LGTM

hakubo:
For some reason - does not work when deployed via DT. Will have to investigate that.

hakubo:
will merge anyway - as we don't use version anyway.

PR #2070, opened 2016-03-01
SOC-1908 As a mod+ in desktop web, I want to see a filter for showing only reported content

rafalkalinski:
We need to add a new filter option to the right column in desktop web to let mods filter to view all reported content.

https://wikia-inc.atlassian.net/browse/SOC-1908


front/main/app/components/discussion-filters.js
```diff
@@ -13,6 +13,10 @@ export default Ember.Component.extend(
 			return this.get('onlyReported') === true ? 'active-element-background-color' : null;
 		}),
 
+		showApplyButton: false,
+		showSortSection: false,
+		canModerate: false,
```
bkoval: Alphabetical order

front/main/app/components/discussion-filters.js
```diff
@@ -54,6 +58,10 @@ export default Ember.Component.extend(
 				if (isCheckboxChecked !== this.get('onlyReported')) {
 					this.set('onlyReported', isCheckboxChecked);
 				}
+
+				if (!this.get('showApplyButton')) {
```
bkoval: I'm not a fan of this. In terms of UX of course :) But I can see we need to use it like this for now. Why not use the component's "applyFilters" method here?

front/main/app/models/discussion-forum.js
```diff
@@ -7,6 +7,15 @@ const DiscussionForumModel = DiscussionBaseModel.extend(
 	DiscussionModerationModelMixin,
 	DiscussionForumActionsModelMixin,
 	{
+
+		postDataNormalizer(post) {
```
bkoval: I would still stick to using verbs in method names - how about normalizePostData? Also, since it might not be clear what we're doing here, let's add JSDoc to this method.

front/main/app/styles/component/_discussion-app-promotion.scss
```diff
@@ -5,12 +5,10 @@ $discussion-app-join-text-line-height: 125%;
 
 $discussion-app-join-text-spacing-top: (
 	details-view: 85px,
-	list-view: 82px
+	list-view: 42px
```
bkoval: Why do we change it? Ie there a change for it in the specs?

front/main/app/styles/component/_discussion-filters.scss
```diff
@@ -146,6 +144,25 @@ $legend-font-size: 16px;
 	}
 }
 
+.rail {
+	.discussion-filters {
+		padding-top: 85px;
+		width: auto;
+
+		legend {
+			background: transparent;
```
bkoval: Does it normally have a background set somewhere in style-guide? If it's just in our SCSS, we can remove it as we always show it with no special background, right?

front/main/app/templates/components/discussion-forum.hbs
```diff
@@ -1,7 +1,8 @@
 {{discussion-header
+	canModerate=model.canModerate
 	hasNewPostButton=true
 	hasSiteName=true
-	hasSortMenu=true
+	hasFilters=true
```
bkoval: Alphabetical order

front/main/app/templates/components/discussion-user.hbs
```diff
@@ -63,7 +63,7 @@
 				}}
 			</div>
 		{{/if}}
-		<div class="rail large-3 columns">
+		<div class="rail rail large-3 columns">
```
bkoval: No need for setting the same classname twice :)

bkoval:
LGTM

PR #2073, opened 2016-03-01
XW-1221 fix path to model in article-preview route

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1221

## Description

Fix path to wiki-page file in article-preview route
 
## Reviewers
@Wikia/x-wing


hakubo:
lgtm

PR #2067, opened 2016-02-29
CE-3429 + CE-3430 Experiment: editing highlighed text

kamilkoterba:
This is a mobile contribution **experiment** and therefore is not intended to be built in a stable and future-looking way. It is expected that this code be entirely thrown away after the experiment has ended. The experiment will be disabled on March 29.

## Links
https://wikia-inc.atlassian.net/browse/CE-3429
https://wikia-inc.atlassian.net/browse/CE-3430

## Description
This feature notifies user about possibility to edit text that he/she selects and allows to perform the edit.

## Reviewers
@Wikia/community-engineering @Wikia/x-wing

PR #2074, opened 2016-03-01
CE-3445 Catch an event of scrolling to the first <p> and select a word

adamkarminski:
**Work in progress. Don't merge!**

## Links

* https://wikia-inc.atlassian.net/browse/CE-3430
* Visual context: https://drive.google.com/a/wikia-inc.com/file/d/0B_mnLiH2mWTwOWYtM09FbHdCS3c/view?usp=sharing

## Description

**This is a part of an experiment that will go live on Tue, Mar 8 and will be retired after 3 weeks on Tue, Mar 29.**

The code will be served only to 10% of traffic on EN wikis. Given the experimental nature of the code we did not invest much in performance micro-optimisation (like treeWalkers - thanks @bkoval).

## Reviewers

@bkoval @lukaszkonieczny @kamilkoterba @Grunny


front/main/app/components/article-content.js
```diff
@@ -540,6 +541,33 @@ export default Ember.Component.extend(
...
 				});
+		},
+
+		insertHighlightedTextEditorDemo() {
+			const highlightedId = 'highlighted-text__demo',
+				selection = window.getSelection(),
+				range = document.createRange(),
+				$paragraphs = this.$('.article-content').children('p');
+
+			let $paragraph, plain, paragraphHtml, word, $highlightedElement;
```
Grunny: The scope of const and let should be as small as possible, so most of these should be moved within the `for` loop (and some could become const's as a result).

front/main/app/components/article-content.js
```diff
@@ -540,6 +541,33 @@ export default Ember.Component.extend(
...
+
+			for (let i = 1; i <= 3; i++) {
+				$paragraph = $($paragraphs[i]);
```
Grunny: This should avoid the `$` global as you do below with `Ember.$` but in this case could use `const $paragraph = $paragraphs.eq(i);`

Grunny:
Two minor comments, otherwise LGTM.

front/main/app/components/article-content.js
```diff
@@ -540,6 +541,33 @@ export default Ember.Component.extend(
 
 					$element.wrap(wrapper);
 				});
+		},
+
+		insertHighlightedTextEditorDemo() {
+			const highlightedId = 'highlighted-text__demo',
```
bkoval: No I can see it's sort of a BEM-alike selector, but why it is using dashes if it's an ID? :)

front/main/app/components/article-content.js
bkoval: +1, block scopes work well with "lets".

front/main/app/components/article-content.js
```diff
@@ -540,6 +541,33 @@ export default Ember.Component.extend(
...
+				selection = window.getSelection(),
+				range = document.createRange(),
+				$paragraphs = this.$('.article-content').children('p');
+
+			let $paragraph, plain, paragraphHtml, word, $highlightedElement;
+
+			for (let i = 1; i <= 3; i++) {
```
bkoval: I'd consider moving "3" to a const with a name, this way code is self-explanatory (name of the const will tell you why it's exactly 3)

front/main/app/components/article-content.js
bkoval: Does it mean you don't care about $paragraphs[0]?

front/main/app/components/article-content.js
```diff
@@ -540,6 +541,33 @@ export default Ember.Component.extend(
...
+					word = plain.split(' ')[1];
+					$paragraph.html(paragraphHtml.replace(word, `<span id="${highlightedId}">${word}</span>`));
+					$highlightedElement = $paragraph.find(`#${highlightedId}`);
+
+					Ember.$(window).scrollTop($highlightedElement.offset().top - 150);
+
+					range.selectNodeContents($highlightedElement[0]);
+					selection.removeAllRanges();
+					selection.addRange(range);
+					break;
```
bkoval: I know this is an experiment, but using break is just bad :) I would consider $(selector).slice(0,numberOfResults) to get your desired number of results. Then, use Array.some() to iterate over it and return false when the "if" condition is met. How about that? :)

bkoval:
LGTM :hamburger: :hammer:

kamilkoterba:
Will be tested on feature branch #2067

PR #2076, opened 2016-03-01
CE-3434 Update translations for Recent Wiki Activity and enable it for all langs

adamkarminski:
## Links
* https://wikia-inc.atlassian.net/browse/CE-3434

## Description
Since we're done with the Localization QA process - let's enable the RWA for all languages according to the release plan.

## Reviewers
@Wikia/community-engineering


tomasznapieralski:
LGTM

PR #2078, opened 2016-03-01
XW-1203 | send Comscore tracking on server side

tomasznapieralski:
- [x] smoke tests
- [x] functional tests
- [x] ads tests

## Links

* https://wikia-inc.atlassian.net/browse/XW-1203

## Reviewers

@Wikia/x-wing


server/app/views/_partials/tracking-comscore.hbs
```diff
@@ -0,0 +1,29 @@
+<script>
+	if (!M.prop('queryParams.noexternals')) {
```
rogatty: Why not `{{#unless queryParams.noexternals}}` at the top of the file? What's the point of including this code if not used?

Aaand, I figured it out by myself. We're using `window.trackComscorePageView();` in `front/common/utils/track.js` and it would break without this code. Would be nice to have it in a comment above this line so no one tries to "optimize" it.

server/app/views/_partials/tracking-comscore.hbs
```diff
@@ -0,0 +1,29 @@
...
+		window.trackComscorePageView = function() {
+			var prefix = (document.location.protocol === 'https:' ? 'https://sb' : 'http://b'),
```
hakubo: we do have loadScripts.js loaded in baseline,
maybe we could use that here?

vforge: And while checking that, can we use it in other trackers too?

server/app/views/_partials/tracking-comscore.hbs
vforge: :+1:

Warkot:
Ads tests http://jenkins:8080/view/ads_synthetic_tmp_248/ — two tests are failing, but same happened for prod. I mentioned the issue on slack#adeng-anomalies.

PR #2081, opened 2016-03-02
RECIRC 49: ABTest Fandom content in footer

pauloslund:
## Links

* https://wikia-inc.atlassian.net/browse/RECIRC-49
* Design: https://wikia-inc.atlassian.net/browse/RECIRC-50

## Description

Running an AB Test with Fandom content in the footer instead of the standard related pages. This is very similar to a test being run in Oasis with displaying Fandom content in the Right Rail. Given this is an AB Test, the code is considered temporary and will likely change in the near future

## Reviewers

@Wikia/x-wing


front/common/modules/AbTest.js
```diff
@@ -0,0 +1,3 @@
+const AbTest = window.Wikia && window.Wikia.AbTest || {};
```
Warkot: You could probably just write `export default window…` as it's not a named export.

front/main/app/components/content-recommendations.js
```diff
@@ -0,0 +1,33 @@
...
+			category: 'recirculation',
+			label: 'footer'
+		});
+	},
+
+	didRender() {
+		const pages = this.get('pages');
+
+		if (pages) {
+			Ember.run.throttle(this, 'trackImpression', 200);
```
Warkot: Why do you need the throttle here? Isn't this hook run just once per page transition and rendering? It doesn't run for every recommended page, does it?

front/main/app/components/content-recommendations.js
```diff
@@ -0,0 +1,33 @@
...
+		}
+	},
+
+	actions: {
+		trackClick() {
```
Warkot: You could consider naming it a little bit different—like e.g. `trackExperimentClick`—to prevent any confusion with this mixin → https://github.com/Wikia/mercury/blob/RECIRC-49/front/main/app/mixins/track-click.js#L27.

front/main/app/templates/components/fandom-posts.hbs
```diff
@@ -0,0 +1,15 @@
+<section class="fandom-posts content-recommendations">
+	<h2>{{ title }}</h2>
```
Warkot: A minor thing—I don't think the spaces after `{{` and before `}}` are recommended.

Warkot:
IMO, it looks pretty neat—especially for an experimental code. :+1:

front/main/app/components/content-recommendations.js
pauloslund: `didRender` was being fired twice in rapid succession when transitioning between articles. Not entirely sure why it was happening, my initial guess due to the article model potentially clearing the `relatedPages` property in between rendering the page but I could be entirely off base. I'll do some more research to figure out the reason.

vforge:
@pauloslund I have more of a general question: Is there any reason not ot use optimizely for this? It seems a bit redundant to implement our in-house AbTesting when we have working solution already. 
Also I think we should consider sending proper dimensions (that descibes our in-house AbTesting tests and current groups) to the UA as we currently are doing in Oasis/Desktop.

pauloslund:
@vforge The primary reasons for not using Optimizely for these experiments are because of Optimizely's usage limits. We're limited on the number of impressions as well as the number of experiments we can run that report to GA (custom dimensions 48-50 are reserved for Optimizely with 48, and 49 already being in use for other experiments). This is only the initial set of experiments we plan to run for recirculation on Mercury so we felt it made sense to leverage the AbTesting tool.

Very good point about sending proper dimensions to the UA.

vforge:
@pauloslund Thanks for clarification! :+1:
Do you have plan to remove all testing code from Mercury and AbTesting with it? Or we want to leave it in our codebase? (I'm just curious)

front/common/modules/AbTest.js
```diff
@@ -0,0 +1 @@
+export default window.Wikia && window.Wikia.AbTest || {};
```
rogatty: If you return an empty object then `AbTest.getGroup(experiment)` will fail with `getGroup is not a function` or something similar.

front/main/app/components/ab-test.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+import AbTest from 'common/modules/AbTest';
+
+export default Ember.Component.extend({
+	willRender() {
```
rogatty: What we do in this hook isn't really related to rendering stuff, right? How about using `didReceiveAttrs()`?
https://guides.emberjs.com/v2.4.0/components/the-component-lifecycle

Also, please call `this._super(...arguments)` when overriding Ember hooks.

rogatty: What we do in this hook isn't really related to rendering stuff, right? How about using `didReceiveAttrs()`?
https://guides.emberjs.com/v2.4.0/components/the-component-lifecycle

Also, please call `this._super(...arguments)` when overriding Ember hooks.

front/main/app/components/content-recommendations.js
rogatty: You can use `didReceiveAttrs()` instead of `didRender()`. It's called only once per article view.

front/main/app/components/fandom-posts.js
```diff
@@ -0,0 +1,48 @@
...
+			}, (error) => {
+				// At some point this error should possibly be handled
+			});
+	},
+
+	didRender() {
+		const posts = this.get('posts');
+
+		if (posts) {
+			Ember.run.throttle(this, 'trackImpression', 200);
```
rogatty: The same as https://github.com/Wikia/mercury/pull/2081/files?diff=split#r54878608

front/main/app/models/fandom-posts.js
```diff
@@ -0,0 +1,33 @@
+import Ember from 'ember';
+
+const FandomPostsModel = Ember.Object.extend({
+	title: null,
+	type: null,
+	posts: [],
```
rogatty: It's risky, see https://wikia-inc.atlassian.net/wiki/pages/viewpage.action?pageId=44007507

front/common/modules/AbTest.js
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	const abList = AbTest.getExperiments(true);
+
+	for (let abIndex = 0; abIndex < abList.length; abIndex++) {
+		const abExp = abList[abIndex],
+			abSlot = AbTest.getGASlot(abExp.name);
+
+		if (abExp && abExp.flags && abExp.flags.ga_tracking) {
+			if (abSlot >= 40 && abSlot <= 49) {
```
hakubo: what is 40 and 49?
can you add a comment about that?

front/main/app/components/fandom-posts.js
```diff
@@ -0,0 +1,45 @@
...
+
+	init() {
+		this._super(...arguments);
+
+		FandomPostsModel
+			.create({
+				type: this.get('type')
+			})
+			.load()
+			.then((model) => {
```
hakubo: hmm this looks odd, 
what do you thing about setting
this.set('model', Fandom...)

and then access the model data through `model.`

this would be more clear where the data is coming from

front/main/app/templates/components/article-wrapper.hbs
```diff
@@ -111,7 +111,15 @@
 		{{/collapsible-menu}}
 	{{/if}}
 </section>
-{{content-recommendations pages=model.relatedPages}}
+{{#ab-test experiment='RECIRCULATION_MERCURY_FOOTER' as |group| }}
```
hakubo: neat :)

front/main/app/templates/components/fandom-posts.hbs
```diff
@@ -0,0 +1,17 @@
+{{#if posts.length}}
+<section class="fandom-posts content-recommendations">
+	<h2>{{title}}</h2>
+	<ul class="thumbnails">
+		{{#each posts as |post|}}
+			<li class="title-thumbnail" {{action 'trackExperimentClick' preventDefault=false}}>
+				<a href="{{post.url}}" class="image image-thumbnail">
+					<img src="{{post.image_url}}" alt="{{post.title}}">
+				</a>
+				<p class="title" title="{{post.title}}">
```
hakubo: never saw someone setting title for p, what is it for? tooltip?

also you have 2 links next to each other that point to the same place
could this a wrap the whole content - so the whole thing is clickable?

front/main/app/components/fandom-posts.js
```diff
@@ -0,0 +1,41 @@
...
+
+	init() {
+		this._super(...arguments);
+
+		const model = FandomPostsModel.create({
+			type: this.get('type')
+		});
+
+		model.load();
+		this.set('model', model);
```
hakubo: there are no listeners set up when in init hook
so you could actually do
this.model = model;

pauloslund:
Got a final LGTM from @hakubo over Slack so I'm going to merge this branch

front/common/modules/AbTest.js
```diff
@@ -0,0 +1,45 @@
...
+	const abList = AbTest.getExperiments(true);
+
+	for (let abIndex = 0; abIndex < abList.length; abIndex++) {
+		const abExp = abList[abIndex],
+			abSlot = AbTest.getGASlot(abExp.name);
+
+		if (abExp && abExp.flags && abExp.flags.ga_tracking) {
+			// GA Slots 40-49 are reserved for our AB Testing tool. Anything outside that
+			// range could potentially overwrite something that we don't want to
+			if (abSlot >= 40 && abSlot <= 49) {
```
rogatty: Are you sure about these numbers? According to https://one.wikia-inc.com/wiki/Engineering/ABTesting/Optimizely#Google_Analytics_Integrations dimensions 48-50 are used for Optimizely.
/cc @Warkot @vforge

front/common/modules/AbTest.js
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+/**
+ * Integrates AbTest with Universal Analytics
+ *
+ * @param {Array} dimensions
+ * @returns {Array}
+ */
+export function integrateAbTestWithUA(dimensions) {
```
rogatty: This method could really use some unit test :) I'm fine with merging this PR without it and adding them later, though.

front/common/modules/AbTest.js
pauloslund: Hmm, interesting. I should have mentioned this in the code but this is pulled almost directly from the code in Oasis. There is actually a test being run in Oasis currently that is tracking on slot 48. Not entirely sure what the best way to handle this would be. We can limit it down to 40-47 on Mercury however that puts it out of sync with the main site

front/common/modules/AbTest.js
pauloslund: Unit tests will be following shortly in another PR :)

front/common/modules/AbTest.js
Warkot: There's a Slack channel #optimizely, you raise this topic there.

rogatty: If it's the same issue in Oasis then IMO this can be merged and resolved later, globally.

rogatty:
LGTM

front/common/modules/AbTest.js
pauloslund: Sounds like a plan

vforge: @pauloslund @rogatty You're both right - I remember that we've selected dimensions for Optimizely not to collide with in-house AbTesting - I couldn't find original document about it, though.
I suggest editing this page on WikiaOne and put information that we can use dimension >= 50 in Optimizely - followed by an e-mail.
Alternatively you can ask Zach Hobbs, as it's he who added this information to One - https://one.wikia-inc.com/index.php?title=Engineering%2FABTesting%2FOptimizely&diff=248985&oldid=248696

PR #2087, opened 2016-03-02
XW-1229 | Create tracking function, always

Warkot:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1229

## Description

Quantserve P2 fix.

## Reviewers

@Wikia/x-wing


bognix:
LGTM

PR #2089, opened 2016-03-02
DAT-3789 | Infobox Builder - source attribute immutable after first save

rybmat:
**Work in progress. Don't merge!** State it here and use labels.

## Links

* https://wikia-inc.atlassian.net/browse/DAT-3789

## Description

Prevent infobox builder from changing source attribute of row elements in existing infoboxes

## Reviewers
@RafLeszczynski @dianafa


front/main/tests/unit/models/infobox-builder-test.js
```diff
@@ -63,13 +63,36 @@ test('add items by type', (assert) => {
...
-					type: 'row'
+					type: 'row',
+					sourceFreezed: false
 				},
 				message: 'add row item'
 			},
 			{
 				dataMock: {
 					data: {
+						label: messageMock
```
RafLeszczynski: I wander if we need this extending test here as we already have test for extending rows in a separate test case below. What do you think?

RafLeszczynski:
one small concern, except it LGTM and please merge newest dev, run linter and wait for build to pass before moving to QA

front/main/app/models/infobox-builder.js
```diff
@@ -81,7 +81,8 @@ const InfoboxBuilderModel = Ember.Object.extend({
 				component: InfoboxBuilderModel.createComponentName(itemType)
 			},
 			source: `${itemType}${index}`,
-			type: itemType
+			type: itemType,
+			sourceFreezed: false
```
rogatty: Minor - there is no such word as "freezed", it should be "frozen".
http://english.stackexchange.com/questions/23991/whats-the-proper-past-tense-of-to-freeze-freezed-or-frozen

RafLeszczynski: your're absolutely right :-)

PR #2090, opened 2016-03-02
Authentication | Do not override userId from Helios

bkoval:
## Description

We should always rely on userID from Helios, not from Mediawiki. Additionally, this is causing isAuthenticated check to fail on polish devboxes, when using servicesDomain in san jose.

## Reviewers

@rogatty @vforge @rafalkalinski @Warkot


rogatty:
It sucks that `UserModel` has `userId` property and that it'll always be null. This service / model hybrid needs to be cleaned anyway but maybe for now we should just set `userId` to the one that is passed to `UserModel.find()`? Or remove it from `UserModel` entirely, whatever you like better.

hakubo:
ping @bkoval

Warkot:
Hey, this PR is two weeks old—this slacker should be merged or closed and, after one of the aforementioned, the branch deleted to keep the repo clean. @bkoval @rafalkalinski @kvas-damian

bkoval:
@Warkot I addressed a comment @rogatty - can you guys CR this? :)

bkoval:
Also, should we create a ticket about cleanups in the service-model chain here?

Are there any tickets for starting to use User Attribute service instead of monolith for fetching data like username and avatar?

rogatty:
Great question! There is no ticket that I would be aware of. IMHO it kinda looks like Social one ;)

rogatty:
LGTM

bkoval:
@rogatty Yep, it sounds like our job to do, just wanted to make sure we won't create a duplicate :)

bkoval:
https://wikia-inc.atlassian.net/browse/SOC-2160

PR #2092, opened 2016-03-02
DAT-3862 | Scroll to the bottom of infobox builder preview after new item is added

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-3862

## Description

After new item is added and rendered, scroll to the bottom of `.infobox-builder-preview` so the item is visible. Animation is there for better experience.

## Reviewers

@RafLeszczynski @dianafa


front/main/app/components/infobox-builder.js
```diff
@@ -15,6 +15,14 @@ export default Ember.Component.extend({
 	}),
 
 	actions: {
+		addItem() {
+			this.get('addItem')(...arguments);
+
+			Ember.run.scheduleOnce('afterRender', this, () => {
+				Ember.run.debounce(this, this.scrollPreviewToBottom, 200);
```
RafLeszczynski: it a small thing but what do you think about moving this magic number to local const or make it a component property?

front/main/tests/unit/components/infobox-builder-test.js
```diff
@@ -0,0 +1,49 @@
...
+
+test('calls scrollPreviewToBottom with debounce after new item is added', function (assert) {
+	const component = this.subject(),
+		/** stub Ember.run.debounce so we can run it synchronously */
+		debounceStub = sinon.stub(Ember.run, 'debounce', (target, func) => {
+			func.call(target);
+		});
+
+	component.set('addItem', sinon.spy());
+	component.scrollPreviewToBottom = sinon.spy();
```
RafLeszczynski: shouldn't we use `component.set('scrollPreviewToBottom', sinon.spy()')`?

front/main/tests/unit/components/infobox-builder-test.js
```diff
@@ -0,0 +1,49 @@
...
+
+	Ember.run(() => {
+		component.send('addItem', 'row');
+	});
+
+	assert.ok(component.scrollPreviewToBottom.calledOnce);
```
RafLeszczynski: you could also check if debounceStub was called with the expected milliseconds number as last argument

front/main/app/components/infobox-builder.js
```diff
@@ -49,5 +57,24 @@ export default Ember.Component.extend({
...
+	scrollPreviewToBottom() {
+		const $preview = this.$('.infobox-builder-preview'),
+			scrollHeight = $preview.prop('scrollHeight'),
+			scrollTop = $preview.prop('scrollTop'),
+			height = $preview.height();
+
+		if (scrollTop + height < scrollHeight) {
+			$preview.animate({
+				scrollTop: scrollHeight - height
+			}, 200);
```
RafLeszczynski: please check the comment about debounce magic number above

front/main/tests/unit/components/infobox-builder-test.js
```diff
@@ -0,0 +1,49 @@
...
+					scrollTop: 0
+				}[propName];
+			}
+		};
+	});
+
+	component.scrollPreviewToBottom();
+
+	assert.ok(animateSpy.calledWith({
+		scrollTop: 100
```
RafLeszczynski: what do you think about moving all magic numbers to local consts and change the assert to:

```JS
assert.ok(animateSpy.calledWith({
    scrollTop: scrollHeight - height
});
```

RafLeszczynski: please also consider adding second assertion for the animation time

front/main/tests/unit/components/infobox-builder-test.js
```diff
@@ -0,0 +1,49 @@
...
+	const component = this.subject(),
+		/** stub Ember.run.debounce so we can run it synchronously */
+		debounceStub = sinon.stub(Ember.run, 'debounce', (target, func) => {
+			func.call(target);
+		});
+
+	component.set('addItem', sinon.spy());
+	component.scrollPreviewToBottom = sinon.spy();
+
+	Ember.run(() => {
```
RafLeszczynski: What is the reason for using `Ember.run(()` instead of just `component.send('addItem', 'row');`

If we need `Ember.run(()` you could make it shorted: `Ember.run(() => component.send('addItem', 'row')); unless our linter will complain because of its configuration :-)

rogatty: If I didn't use it there would be this error:
```
	
Error: Assertion Failed: You have turned on testing mode, which disabled the run-loop's autorun.
You will need to wrap any code with asynchronous side-effects in a run
```

PR #2094, opened 2016-03-03
DAT-3769 add tracking to infobox builder & element change recognition

dianafa:
## Links

https://wikia-inc.atlassian.net/browse/DAT-3769

## Reviewers

@RafLeszczynski @rogatty @hakubo


front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -1,15 +1,20 @@
...
 	InfoboxBuilderEditItemMixin,
+	TrackClickMixin,
 	{
 		labelValue: Ember.computed('item.data.label', {
 			get() {
 				return this.get('item.data.label');
 			},
 			set(key, value) {
-				this.get('editRowItem')(this.get('item'), value);
+				const item = this.get('item');
```
RafLeszczynski: what do you think about creating a track function in `InfoboxBuilderEditItemMixin` and just call in each edit item component? you will avoid copy-pasting `this.trackClick('infobox-builder', `edit-item-${item.type}-label`) in each component

front/main/app/components/infobox-builder-button.js
```diff
@@ -1,12 +1,19 @@
...
+		label: Ember.computed('name', function () {
+			return i18n.t(`infobox-builder:main.add-${this.get('name')}`);
+		}),
 
-	click() {
-		this.get('onButtonClick')(this.get('name'));
+		click() {
+			const buttonName = this.get('name');
+
+			this.trackClick('infobox-builder', `add-item-${buttonName}`);
```
RafLeszczynski: what do you think about defining tracking category `infobox-builder` as a const or a property of `infobox-builder` and pass it to its children to avoid copy pasting the category name?

front/main/app/components/infobox-builder-edit-item-section-header.js
```diff
@@ -1,24 +1,33 @@
...
 	{
 		value: Ember.computed('item.data', {
 			get() {
 				return this.get('item.data');
 			},
 			set(key, value) {
-				this.get('editSectionHeaderItem')(this.get('item'), {data: value});
+				const item = this.get('item');
+
+				this.trackClick('infobox-builder', `edit-item-${item.type}-value`);
```
RafLeszczynski: please consider using `the.get('item.type')` instead of storing `item` in a local const

front/main/app/components/infobox-builder-edit-item.js
```diff
@@ -1,46 +1,56 @@
...
+			back() {
+				const item = this.get('item');
 
-	actions: {
-		removeItem() {
-			this.get('onDeleteItem')(this.get('item'));
-		},
-		back() {
-			this.get('onBackArrowClick');
+				this.trackClick('infobox-builder', `navigate-back-from-edit-panel-${item.type}`);
```
RafLeszczynski: what do you think about using `ember.computed` for creating tracking labels. This will allow you to write unit test in order to check if we are creating correct tracking labels

front/main/app/models/infobox-builder.js
```diff
@@ -161,11 +161,26 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
+			let itemData = item.data;
+
+			if (item.type === 'section-header') {
+				itemData = {
+					value: item.data,
+					collapsible: item.collapsible
+				};
+			}
+			// we need a copy of itemData, not a reference to it
+			item.infoboxBuilderData.oldData = jQuery.extend({}, itemData);
```
RafLeszczynski: is there any Ember jQuery wrapper or this is the correct way of using jQuery in ember app?

front/main/app/models/infobox-builder.js
```diff
@@ -161,11 +161,26 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
 	/**
-	 * @desc sets item to the edit mode
+	 * @desc sets item to the edit mode and saves its current
+	 * data in the moment of beginning editing
+	 *
 	 * @param {Object} item
 	 * @returns {void}
 	 */
 	setEditItem(item) {
+		if (item && !item.infoboxBuilderData.oldData) {
```
RafLeszczynski: please consider moving this logic to separate function

RafLeszczynski:
do we need such detailed diffs for tracking ?

front/main/app/components/infobox-builder-edit-item.js
```diff
@@ -1,46 +1,56 @@
...
+export default Ember.Component.extend(
+	TrackClickMixin,
+	{
+		editPanelOption: Ember.computed('item.type', function () {
+			return `infobox-builder-edit-item-${this.get('item.type')}`;
+		}),
 
-	header: Ember.computed('item.{data,data.label,type,value,data.defaultValue,infoboxBuilderData.index}', function () {
-		let header;
+		header: Ember.computed('item.{data,data.label,type,value,data.defaultValue,infoboxBuilderData.index}', function () {
```
rogatty: Missing `\t`

front/main/app/components/infobox-builder-edit-item-row.js
rogatty: +1

front/main/app/components/infobox-builder.js
```diff
@@ -1,53 +1,75 @@
...
-			});
-		},
+		trackChangedItems() {
+			const diffArray = this.get('getDiffArray')();
 
-		onPreviewItemDrag(actionTrigger) {
-			this.set('isPreviewItemDragged', true);
-			this.get('handleItemInEditMode')(actionTrigger);
+			diffArray.forEach((element) =>
+				this.trackClick('infobox-builder', `changed-element-${element.type}-${element.changedField}`)
```
rogatty: This isn't really a click event, right?

front/main/app/components/infobox-builder.js
```diff
@@ -1,53 +1,75 @@
...
+				this.set('isPreviewItemDragged', false);
+			},
+
+			save() {
+				this.set('isLoading', true);
+
+				this.trackClick('infobox-builder', 'saving-attempt');
+				this.trackChangedItems();
+				this.get('saveAction')().then(() => {
+					this.trackClick('infobox-builder', 'saving-successful');
```
rogatty: This isn't a click event, right?

front/main/app/mixins/infobox-builder-item.js
```diff
@@ -1,18 +1,25 @@
...
 
-	click() {
-		this.get('setEditItem')(this.get('model'));
-	},
-	mouseMove(event) {
-		this.get('onMouseEnter')(event.clientX, event.clientY);
-	},
-	mouseLeave() {
-		this.get('onMouseLeave')();
+		click() {
```
rogatty: This is removed in https://github.com/Wikia/mercury/pull/2095/files#diff-24d56f341a8bec24cbdb436ad04951d8L9 so after merging, tracking should be moved, probably to `setEditItemAndStopPropagation()` (https://github.com/Wikia/mercury/pull/2095/files#diff-edc3dc94c910eeccd54b08a647379879R60)

front/main/app/models/infobox-builder.js
rogatty: Use `Ember.$.extend()`

rogatty: Also, can we call it `originalData`? It's the word we use in the rest of the app.

front/main/app/models/infobox-builder.js
```diff
@@ -264,6 +279,50 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
+						break;
+					case 'section-header':
+						diff = InfoboxBuilderModel.createSectionHeaderDiff(
+							item.infoboxBuilderData.oldData,
+							item.data,
+							item.collapsible
+						);
+						break;
+					default:
+						break;
```
rogatty: Can we just skip the `default`?

front/main/app/components/infobox-builder.js
```diff
@@ -1,112 +1,146 @@
...
+				});
+			},
+
+			/**
+			 * @param {Object} actionTrigger - dragged item
+			 * @returns {void}
+			 */
+			onPreviewItemDrag(actionTrigger) {
+				this.set('isPreviewItemDragged', true);
+				this.trackClick('infobox-builder', `drag-element-${actionTrigger.type}`);
```
RafLeszczynski: do we want to track this as click or should we add drag event?

front/main/app/components/infobox-builder.js
```diff
@@ -1,112 +1,146 @@
...
+
+				if (actionTrigger !== this.get('activeItem')) {
+					this.get('setEditItem')(null);
+				}
+			},
+
+			onPreviewItemDrop() {
+				this.set('isPreviewItemDragged', false);
```
RafLeszczynski: we should probably also track drops to see if user successfully reorder item???

front/main/app/components/infobox-builder.js
```diff
@@ -1,112 +1,146 @@
...
+				this.get('setEditItem')(targetItem);
+			},
+
+			save() {
+				this.set('isLoading', true);
+
+				this.trackClick('infobox-builder', 'save-attempt');
+				this.trackChangedItems();
+				this.get('saveAction')().then(() => {
+					this.trackSuccess('infobox-builder', 'save-successful');
```
RafLeszczynski: this method doesn't exist in trackClick Mixin

front/common/utils/track.js
```diff
@@ -89,7 +91,9 @@ const trackers = {
 		// Action to take a survey
 		takeSurvey: 'take-survey',
 		// View
-		view: 'view'
+		view: 'view',
+		// Input focus
+		focus: 'focus'
```
SebastianMarzjan: Can we keep the keys in alphabetic order?

front/main/app/components/infobox-builder.js
```diff
@@ -1,112 +1,146 @@
...
-
-		save() {
-			this.set('isLoading', true);
-			this.get('saveAction')().then(() =>
-				this.setProperties({
-					isLoading: false,
-					showSuccess: true
-				})
+			diffArray.forEach((element) =>
+				this.trackChange('infobox-builder', `changed-element-${element.type}-${element.changedField}`)
```
RafLeszczynski: this method doesn't exist in trackClick mixin

front/main/app/components/infobox-builder.js
```diff
@@ -1,112 +1,146 @@
...
+				this.trackClick('infobox-builder', `item-${targetItem.type}`);
+
+				this.get('setEditItem')(targetItem);
+			},
+
+			save() {
+				this.set('isLoading', true);
+
+				this.trackClick('infobox-builder', 'save-attempt');
+				this.trackChangedItems();
```
RafLeszczynski: do we want to track this change as successful changes? if yes this code should be probably invoked on after successful save

front/main/app/models/infobox-builder.js
```diff
@@ -162,11 +162,26 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
+	 * data in the moment of beginning editing
+	 *
 	 * @param {Object} item
 	 * @returns {void}
 	 */
 	setEditItem(item) {
+		if (item && !item.infoboxBuilderData.originalData) {
+			let itemData = item.data;
+
+			if (item.type === 'section-header') {
```
RafLeszczynski: could we move this logic to separate method?

RafLeszczynski: we could do something like:
```JS
item.infoboxBuilderData.originalData = Ember.$.extend({}, item.type === 'section-header' ?  serilizeSectionHeaderData(itemData) : itemData);

```

rogatty:
As hooks are currently broken, here's log from failing linter:
http://jenkins.wikia-prod:8080/job/Mercury-PR-Checks/3964/artifact/jenkins/linter.log

front/main/app/utils/infobox-builder-diff.js
```diff
@@ -0,0 +1,133 @@
...
+ * @desc Provides information about whether and which elements'
+ * values have been changed, basing on previously saved original data.
+ *
+ * @returns {Array} of Objects with element type and changed property
+ */
+export default function (currentState) {
+	const diffs = [];
+	let diff = [];
+
+	currentState
```
RafLeszczynski: I know what's going on in this function but I have mixed feelings about the code. Seams a bit complicated. I'll try to figure out how to simplify it. Maybe @SebastianMarzjan or @rogatty have any suggestions?

front/main/app/utils/infobox-builder-diff.js
```diff
@@ -0,0 +1,133 @@
...
+			return createTitleDiff;
+			break;
+		case 'row':
+			return createRowDiff;
+			break;
+		case 'section-header':
+			return createSectionHeaderDiff;
+			break;
+		default:
+			return null;
```
RafLeszczynski: what do you think about adding some logging here?

front/main/app/utils/infobox-builder-diff.js
```diff
@@ -0,0 +1,133 @@
...
+ */
+export function createRowDiff(originalData, data) {
+	const changes = [];
+
+	if (typeof originalData.label === 'undefined') {
+		return changes;
+	}
+
+	if (data.label !== originalData.label) {
+		changes.push({
```
RafLeszczynski: this pattern is repeated in every Diff function. Looks like a good candidate for some inheritance - default class and extending it for each item type. Of course we don't won't to invest more time into diffs in this sprint so it something for future improvements

front/main/app/utils/infobox-builder-diff.js
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * @param {Object} originalData data in form as it was when opening infobox builder
+ * @param {string} value section header value
+ * @param {boolean} collapsible
+ * @returns {Object}
+ */
+export function createSectionHeaderDiff(originalData, value, collapsible) {
+	const changes = [];
+
+	if (typeof originalData.collapsible === 'undefined' || typeof originalData.value === 'undefined') {
```
RafLeszczynski: shouldn't it be `&&`?

RafLeszczynski:
Did first round of CR. @SebastianMarzjan please take a look too

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -4,14 +4,47 @@ import InfoboxBuilderEditItemMixin from '../mixins/infobox-builder-edit-item';
...
+
 		labelValue: Ember.computed('item.data.label', {
 			get() {
 				return this.get('item.data.label');
 			},
 			set(key, value) {
-				this.get('editRowItem')(this.get('item'), value);
+				const item = this.get('item');
+
+				// check if user interacted with label input
```
SebastianMarzjan: Is it really a check?

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -4,14 +4,47 @@ import InfoboxBuilderEditItemMixin from '../mixins/infobox-builder-edit-item';
...
 			}
-		})
+		}),
+
+		actions: {
+			onLabelInputFocus() {
+				// save current label value for tracking change on blur
+				this.set('labelValueOnFocus', this.get('labelValue'));
+
+				//tack focus on label input
```
SebastianMarzjan: lack of space and typo in 'track'

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -4,14 +4,47 @@ import InfoboxBuilderEditItemMixin from '../mixins/infobox-builder-edit-item';
...
+			},
+
+			onLabelInputBlur() {
+				// track interaction with label input
+				if (this.get('wasLabelAltered')) {
+					this.trackEditItemOption('keypress', this.get('labelFocusTrackingKey'));
+					this.set('wasLabelAltered', false);
+				}
+
+				// tack change of label value
```
SebastianMarzjan: s/tack/track/

front/main/app/components/infobox-builder-edit-item-section-header.js
```diff
@@ -4,23 +4,56 @@ import InfoboxBuilderEditItemMixin from '../mixins/infobox-builder-edit-item';
 export default Ember.Component.extend(
 	InfoboxBuilderEditItemMixin,
 	{
+		// params required for tracking edit actions
+		headerValueFocusTrackingKey: 'header-value',
+		headerValueOnFocus: null,
+		wasHeaderValueAltered: false,
```
SebastianMarzjan: I'm wondering if that needs to be specific to the item type or could be abstracted to the InfoboxBuilderEditItemMixin

front/main/app/components/infobox-builder-edit-item-section-header.js
```diff
@@ -4,23 +4,56 @@ import InfoboxBuilderEditItemMixin from '../mixins/infobox-builder-edit-item';
...
 			}
-		})
+		}),
+
+		actions: {
+			onHeaderValueInputFocus() {
+				// save current label value for tracking change on blur
+				this.set('headerValueOnFocus', this.get('labelValue'));
+
+				//tack focus on label input
```
SebastianMarzjan: space and typo

front/main/app/components/infobox-builder-edit-item-section-header.js
```diff
@@ -4,23 +4,56 @@ import InfoboxBuilderEditItemMixin from '../mixins/infobox-builder-edit-item';
...
+			},
+
+			onHeaderValueInputBlur() {
+				// track interaction with label input
+				if (this.get('wasHeaderValueAltered')) {
+					this.trackEditItemOption('keypress', this.get('headerValueFocusTrackingKey'));
+					this.set('wasHeaderValueAltered', false);
+				}
+
+				// tack change of label value
```
SebastianMarzjan: s/tack/track/

copypaste I sense...

front/main/app/components/infobox-builder.js
rogatty: User clicks and then holds the button, IMO it's good enough for now and we can adjust later if needed.

front/main/app/utils/infobox-builder-diff.js
SebastianMarzjan: It has already been simplified a LOT :)

front/main/app/components/infobox-builder.js
rogatty: But this will always fire so we'll have exactly the same number as on drag, right? If we care about successful reorder then maybe we should track it as a part of diff?

SebastianMarzjan:
I left my two sense. There are also linter issues.

front/main/app/components/infobox-builder.js
rogatty: I think we want to know if user made some changes and clicked the save button. Even if there is network error, lack of authentication or whatever, we still want to know that user managed to use the tool, right? For counting failures we have `save-successful` event.

front/main/app/components/infobox-builder-edit-item-section-header.js
rogatty: Right now we could abstract all logic for focus / blur and it would make sense... but!
If we will need to add another input to header / row or some other type (more than one per type) then it won't work. I think it's [The Wrong Abstraction](http://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction) to make.

Altough, to keep it more DRY I've extracted some logic in https://github.com/Wikia/mercury/commit/dce47552039051a0dec29f14fab747497a77779e

front/main/app/components/infobox-builder.js
RafLeszczynski: @rogatty Agree, let's add it to Mercury TODOs

front/main/app/components/infobox-builder.js
RafLeszczynski: we could also add event to reorder method on model

front/main/app/components/infobox-builder.js
RafLeszczynski: true

front/main/app/utils/infobox-builder-diff.js
RafLeszczynski: OK let's not invest more time into this at the moment as this logic is covered with unit tests. We could rethink this approach next time we will be working on this functionality

front/main/app/components/infobox-builder.js
RafLeszczynski: Added additional tracking to reorder https://github.com/Wikia/mercury/commit/63edc2ec9638fa5ccabc59147ae7eb97e2574eb1

PR #2096, opened 2016-03-03
CE-3446 Add popover

kamilkoterba:
**This is for ~two week experiment**

## Links
* https://wikia-inc.atlassian.net/browse/CE-3446 Popover
* https://wikia-inc.atlassian.net/browse/CE-3448 Tracking

## Description
* This is ugly! But we don't aim for quality here ;)
* Text is hard coded as it's not going to be translated - test is only on EN wikis.
* Popover is reused from discussions (ember-pop-over) necessary custom styles added. Full width is forced. Properly done it should introduce new flow in flows.js, but for this purpose it's probably enough.
* As we're pinning popover to element within article - again we can't do it ember way and jquery is used to force pop-over opening and `run.later` to wait for pop-over to get aware of element to track click to open

## Reviewers
@Wikia/community-engineering  @Wikia/x-wing

### DIFF
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -568,6 +568,9 @@ export default Ember.Component.extend(
 					break;
 				}
 			}
+			Ember.run.later(function () {
+				$highlightedElement.trigger('mousedown');
+			}, 500);
 		}
 	}
 );
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -14,6 +14,7 @@ import WidgetVKComponent from '../components/widget-vk';
 import WidgetPolldaddyComponent from '../components/widget-polldaddy';
 import WidgetFliteComponent from '../components/widget-flite';
 import {getRenderComponentFor, queryPlaceholders} from '../utils/render-component';
+import {getExperimentVariationNumber} from 'common/utils/variantTesting';
 import {track, trackActions} from 'common/utils/track';
 
 /**
@@ -36,6 +37,9 @@ export default Ember.Component.extend(
 		uploadFeatureEnabled: null,
 		displayTitle: null,
 		headers: null,
+		highlightedEditorDemoEnabled: Ember.computed(() => {
+			return getExperimentVariationNumber({dev: '5170910064', prod: '5164060600'}) === 1;
+		}),
 
 		newFromMedia(media) {
 			if (media.context === 'infobox' || media.context === 'infobox-hero-image') {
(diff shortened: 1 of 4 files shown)


front/main/app/components/article-content.js
```diff
@@ -35,6 +37,9 @@ export default Ember.Component.extend(
 		uploadFeatureEnabled: null,
 		displayTitle: null,
 		headers: null,
+		highlightedEditorDemoEnabled: Ember.computed(() => {
+			return getExperimentVariationNumber({dev: '5170910064', prod: '5164060600'}) === 2;
```
lukaszkonieczny: Are you sure it should be separate variation?

front/main/app/styles/component/_article-wrapper.scss
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	p {
+		font-size: $type-smallest;
+		line-height: $line-height-smallest;
+	}
+
+	.pop-over-compass {
+		left: 20px!important;
+		width: calc(100% - 40px)!important;
```
lukaszkonieczny: Can we avoid `!important` ?

front/main/app/components/article-content.js
```diff
@@ -572,6 +579,16 @@ export default Ember.Component.extend(
 					range.selectNodeContents($highlightedElement[0]);
 					selection.removeAllRanges();
 					selection.addRange(range);
+
+					Ember.run.later(() => {
+						$highlightedElement.trigger('mousedown');
```
lukaszkonieczny: Why we are triggering `mousedown` ?

front/main/app/styles/component/_article-wrapper.scss
kamilkoterba: We're fighting inline style counted by ember-pop-over internally. We would need to provide this behavior as a variant of popover by extending it's functionality. flows.js is probably good way to start, but do we want it for experiment purposes? or maybe you have another idea?

front/main/app/components/article-content.js
kamilkoterba: Unfortunately there's no other way available right now https://github.com/Wikia/mercury/blob/9560edbe6ad69899b1b6c217069d1ac4134b41da/front/main/app/mixins/discussion-more-options.js#L28-L28 ;(

front/main/app/styles/component/_article-wrapper.scss
vforge: :sob:

front/main/app/components/article-content.js
rogatty: Maybe we should use this way around? https://github.com/tim-evans/ember-pop-over/issues/24

rogatty:
LGTM

front/main/app/styles/component/_article-wrapper.scss
kamilkoterba: @vforge :sob: :sob:

kamilkoterba:
Will be tested on feature branch #2067

PR #2098, opened 2016-03-03
ADEN-3043 Add timestamp to Nielsen url

drets:
## Links

* https://wikia-inc.atlassian.net/browse/ADEN-3043

## Description

Add timestamp to Nielsen request URL for preventing library from browser caching.


jacekjursza:
LGTM

PR #2099, opened 2016-03-03
XW | stop calling quantserve twice on auth

bognix:
## Description
Quantserve was called twice on auth pages because the initialPageView was not set
## Reviewers
@Wikia/x-wing


vforge:
LGTM

PR #2109, opened 2016-03-03
Set up geo in baseline so ivw3 can use that for tracking

hakubo:
## Description

IVW3 need geo to work as it is being tracked only for some countries
This therefore have to happen much earlier - where we don't have jQuery yet

Subject to refactor - I guess...
## Reviewers

@Wikia/x-wing


bognix:
LGTM

PR #2112, opened 2016-03-03
XW-1211 | track specific article click events

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1211

## Description

A specific click tracking.

## Reviewers

@Wikia/x-wing


front/main/app/components/article-content.js
```diff
@@ -117,6 +119,14 @@ export default Ember.Component.extend(
 			this.destroyChildComponents();
 		},
 
+		click() {
+			const label = this.getTrackingEventLabel(event);
+
```
hakubo: Missed parameter? Also you could use destructuring here {target}

front/main/app/components/article-content.js
```diff
@@ -176,6 +186,26 @@ export default Ember.Component.extend(
 			}
 		},
 
+		/**
+		 * @param event
+		 * @returns {string}
+		 */
+		getTrackingEventLabel(event) {
```
hakubo: This only cares about Dom element, you could pass here only that

front/main/app/components/article-content.js
```diff
@@ -176,6 +187,24 @@ export default Ember.Component.extend(
 			}
 		},
 
+		/**
+		 * @param {jQuery} $element — element of which context is to be checked
+		 * @returns {string}
+		 */
+		getTrackingEventLabel($element) {
+			if ($element.length) {
```
tomasznapieralski: what if `$element` is `undefined`?

front/main/app/components/article-content.js
```diff
@@ -176,6 +187,24 @@ export default Ember.Component.extend(
...
+				if ($element.closest('.portable-infobox').length) {
+					return 'portable-infobox-link';
+				} else if ($element.closest('.context-link').length) {
+					return 'context-link';
+				} else if ($element.closest('blockquote').length) {
+					return 'blockquote-link';
+				}
+				return 'regular-link';
```
tomasznapieralski: please add line above `return` statement

front/main/app/components/article-content.js
```diff
@@ -176,6 +187,24 @@ export default Ember.Component.extend(
...
+			}
+			return '';
```
tomasznapieralski: please add line above `return` statement

front/main/app/components/article-content.js
```diff
@@ -176,6 +187,24 @@ export default Ember.Component.extend(
 			}
 		},
 
+		/**
+		 * @param {jQuery} $element — element of which context is to be checked
+		 * @returns {string}
+		 */
+		getTrackingEventLabel($element) {
+			if ($element.length) {
+				if ($element.closest('.portable-infobox').length) {
```
tomasznapieralski: you don't need to check length, closest can return element or null

front/main/app/components/article-content.js
```diff
@@ -176,6 +187,24 @@ export default Ember.Component.extend(
...
+					return 'portable-infobox-link';
+				} else if ($element.closest('.context-link').length) {
```
tomasznapieralski: you don't need to check length, closest can return element or null

front/main/app/components/article-content.js
```diff
@@ -176,6 +187,24 @@ export default Ember.Component.extend(
...
+					return 'context-link';
+				} else if ($element.closest('blockquote').length) {
```
tomasznapieralski: you don't need to check length, closest can return element or null

front/main/app/components/article-content.js
Warkot: I must disagree, `closest` always returns list of elements—if there are no elements found, it's just an empty list. And in JS `Boolean([]) === true`.

front/main/app/components/article-content.js
Warkot: https://github.com/Wikia/mercury/pull/2112#discussion_r55197687

front/main/app/components/article-content.js
Warkot: https://github.com/Wikia/mercury/pull/2112#discussion_r55197687

front/main/app/components/article-content.js
vforge: @Warkot can you jsfiddle your solution?

tomasznapieralski:
LGTM

vforge:
LGTM

PR #2115, opened 2016-03-04
(CE-3464,CE-3466) Improve arrow and popover

kamilkoterba:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3464 Arrow position
* https://wikia-inc.atlassian.net/browse/CE-3466 Popover permanent dismiss

## Description

Introduced new popover that extends built one. Overridden documentClick to additionally disable not just hide and also made positionPointer to set positioning of arrow to point middle of targeted element.

## Reviewers

@Wikia/community-engineering, @Wikia/x-wing


front/main/app/components/highlighted-editor-pop-over.js
```diff
@@ -0,0 +1,18 @@
+import PopOver from 'ember-pop-over/components/pop-over';
+import Ember from 'ember';
+
+export default PopOver.extend({
+	positionPointer($compass, pointerRect) {
+		const $target = Ember.$(Ember.get(this, 'targets')[0].element),
```
Grunny: Should be `this.get('targets')`.

front/main/app/components/highlighted-editor-pop-over.js
```diff
@@ -0,0 +1,18 @@
...
+			// shift is padding-left and half of pointer width
+			// TODO calculate shift properly
+			shift = 26;
+
+		pointerRect.left = $target.position().left + $target.width() / 2 - shift;
+		this._super($compass, pointerRect);
+	},
+	documentClick(evt) {
+		this._super(evt);
+		Ember.set(this, 'disabled', true);
```
Grunny: Likewise `this.set('disabled', true);`

Grunny:
Rather than duplicating the template pop-over template to `highlighted-editor-pop-over.hbs` without changes, I think it would be better just to reuse the original template (i.e. by setting `layoutName: 'components/pop-over'` on the extended component), then we reuse the existing template, and avoid sending the same template to the client twice.

kamilkoterba:
Perfect, thanks!

Grunny:
LGTM, will be tested on feature branch.

PR #2116, opened 2016-03-04
CE-3445: Highlight text on ios

lukaszkonieczny:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3445

## Description

Highlight text by setting background (we cannot highlight text in ios from javascript)

## Reviewers

@Wikia/community-engineering

### DIFF
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -576,16 +576,16 @@ export default Ember.Component.extend(
 					$paragraph.html(paragraphHtml.replace(word, `<span id="${highlightedId}">${word}</span>`));
 					$highlightedElement = $paragraph.find(`#${highlightedId}`);
 
+					range.selectNodeContents($highlightedElement[0]);
+					selection.removeAllRanges();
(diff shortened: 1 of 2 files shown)


front/main/app/components/article-content.js
```diff
@@ -576,7 +576,12 @@ export default Ember.Component.extend(
 					$paragraph.html(paragraphHtml.replace(word, `<span id="${highlightedId}">${word}</span>`));
 					$highlightedElement = $paragraph.find(`#${highlightedId}`);
 
-					Ember.$(window).scrollTop($highlightedElement.offset().top - 150);
+					Ember.$('body').animate({scrollTop: $highlightedElement.offset().top - 150}, () => {
+						Ember.$(document).one('selectionchange', () => {
```
kamilkoterba: I see a risk here that this event listener will be added after selection changed and will never happen?

kamilkoterba:
LGTM

azarek:
LGTM

PR #2118, opened 2016-03-04
CE-3460 Initialize the HTE demo when user scrolls to content

adamkarminski:
## Links
* http://wikia-inc.atlassian.net/browse/CE-3460

## Description
The demo will now be launched when a user scrolls to an article's content.

## Reviewers

### DIFF
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -647,7 +647,6 @@ export default Ember.Component.extend(
 		 * @returns {void}
 		 */
 		onScroll() {
-			console.log(window.scrollY , this.get('targetParagraphOffset'));
 			if (window.scrollY > this.get('targetParagraphOffset')) {
 				this.launchHighlightedTextEditorDemo();
 			}
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -570,28 +570,20 @@ export default Ember.Component.extend(
 			$paragraphs.toArray().some((paragraph) => {
 				const $paragraph = Ember.$(paragraph),
 					paragraphHtml = $paragraph.html(),
-					plain = Ember.$('<div>').html(paragraphHtml).children().remove().end().html();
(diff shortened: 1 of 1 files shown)


front/main/app/components/article-content.js
```diff
@@ -569,33 +572,84 @@ export default Ember.Component.extend(
...
-					range.selectNodeContents($highlightedElement[0]);
-					selection.removeAllRanges();
-					selection.addRange(range);
+					this.set('targetParagraphOffset', $paragraph.offset().top);
 
-					Ember.run.later(() => {
-						$highlightedElement.trigger('mousedown');
-					}, 500);
+					Ember.$(document).on('touchmove', this, this.debouncedScroll);
+					Ember.$(window).on('scroll', this, this.debouncedScroll);
```
lukaszkonieczny: I think these events should be bind if there is a word longer than `minLettersLimit`

lukaszkonieczny: Also you can consider using `one` method which will trigger just once.

front/main/app/components/article-content.js
```diff
@@ -569,33 +572,84 @@ export default Ember.Component.extend(
...
+
+			Ember.$(document).unbind('touchmove', this.debouncedScroll);
+			Ember.$(window).unbind('scroll', this.debouncedScroll);
+
+			if ($highlightedElement) {
+				range.selectNodeContents($highlightedElement[0]);
+				selection.removeAllRanges();
+				selection.addRange(range);
+
+				Ember.$('html, body').animate({scrollTop: `${$highlightedElement.offset().top - 150}px`});
```
lukaszkonieczny: Is it some kind of workaround for something or we want to animate both elements?

kamilkoterba: http://stackoverflow.com/questions/19303405/difference-between-html-body-animate-and-body-animate
Both sounds safer but possibly this browser differences are not longer valid. We can test this if we have spare time before shipping the experiment.

adamkarminski:
LGTM!

PR #2122, opened 2016-03-07
XW-842 | Make it run under node 5.X

hakubo:
## Links
- https://nodejs.org/en/download/releases/
- https://github.com/kenkouot/hapi-i18next
## Description

We'd like to upgrade node to latest version. Currently 4.X is LTS I think we'd gain a lot if we jumped straight to 5.x. Then when 6 lands (next LTS) we could stay longer on it.

I ran it with:
`node v5.7.1
npm 3.8.1`

This branch just go as far as - make it run. There are still missing pieces that have to be solved:
- [ ] Fork hapi-18next under Wikia and install dev dependencies as dev dependencies
- [ ] test if all works as expected
- [ ] load test it

Branch made so when we have a way to deploy mercury with latest version - we have something to deploy.
## Reviewers

@rogatty @RafLeszczynski 
@rmarchei - when we have a way to deploy mercury with latest node - we can use this branch to test it
## Update

We didn't make it work with with node 5.7 but we made it work with node 5.10.1


package.json
```diff
@@ -78,6 +78,7 @@
     "multipipe": "0.1.2",
     "nodemon": "1.8.1",
     "npm-run-all": "1.4.0",
+    "phantomjs-prebuilt": "^2.1.4",
```
tomasznapieralski: use fixed version

server/gulpfile.js
```diff
@@ -43,7 +42,7 @@ gulp.task('build-server-scripts', ['build-server-init-config'], function (done)
  * Copy node dependencies to www/server/
  */
 gulp.task('build-server-node-modules', function () {
-	var deps = '/{' + nodeDeps.join('/**/*,') + '/**/*}';
```
hakubo: npm 3 installs all as flat as possible so this won't work
unless we are more clever here - IMHO not worth it
node_modules in server take 37Megs

server/gulpfile.js
```diff
@@ -43,7 +42,7 @@ gulp.task('build-server-scripts', ['build-server-init-config'], function (done)
...
+	var deps = '/**/*';
```
tomasznapieralski: everything everywhere, are you sure?

hakubo:
latest node is v5.11.0 can we use that already?

tomasznapieralski:
sure, we had problems with `apt-get` on Friday, but it seems that it is availabe now

tomasznapieralski:
Blocked by https://wikia-inc.atlassian.net/browse/XW-1441

PR #2134, opened 2016-03-08
DAT-3856 | Use infobox theme property to set additional class

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-3856

## Description

In https://github.com/Wikia/app/pull/9907 we started to return additional property from the Infobox Builder API - the theme. If it's present then let's add `` `theme-${theme}` `` class to the `infobox-builder-preview` element.

I also used this occasion to clean up `infobox-builder` route and model. If you want to review just the theme part then look at the second commit.

## Reviewers

@RafLeszczynski @SebastianMarzjan


front/main/app/templates/components/infobox-builder.hbs
```diff
@@ -7,7 +7,7 @@
 }}
 
 <div class="infobox-builder">
-	<div class="infobox-builder-preview" {{action 'onPreviewBackgroundClick'}}>
+		<div class="infobox-builder-preview{{themeClassName}}" {{action 'onPreviewBackgroundClick'}}>
```
RafLeszczynski: we need this class on sortable group

front/main/app/components/infobox-builder.js
```diff
@@ -23,6 +23,16 @@ export default Ember.Component.extend(
 			return this.get('isPreviewItemHovered') && !this.get('isPreviewItemDragged');
 		}),
 
+		themeClassName: Ember.computed('theme', function () {
+			const theme = this.get('theme');
+
+			if (theme) {
+				return ` theme-${theme}`;
```
RafLeszczynski: because of the way themes are serialised to class names this class should be `pi-theme-${theme}`

front/main/app/models/infobox-builder.js
```diff
@@ -247,7 +247,26 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
+	 * @returns {void}
+	 */
+	setupInfoboxData(data) {
+		if (data.data) {
+			this.setupExistingState(data.data);
+		} else {
+			this.setupInitialState();
+		}
+
+		if (!Ember.isEmpty(data.theme)) {
```
RafLeszczynski: what about new infobox templates should have europa theme by default?

front/main/app/components/infobox-builder.js
SebastianMarzjan: Do we have a helper to avoid having a whitespace here?

front/main/app/models/infobox-builder.js
rogatty: @SebastianMarzjan said that no, they should not. And if one day we decide that they should then it would ideally be handled by the API side and returned explicitly.

RafLeszczynski:
LGTM

do we want to test it separately or have all changes for https://wikia-inc.atlassian.net/browse/DAT-3856 in a single branch?

tomasznapieralski:
LGTM

front/main/app/models/infobox-builder.js
```diff
@@ -247,7 +249,28 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
+	 * @param {Object} data
+	 * @param {Boolean} isNew
+	 * @returns {void}
+	 */
+	setupInfoboxData(data, isNew) {
+		if (isNew) {
+			this.setupInitialState();
+			this.set('theme', this.get('defaultTheme'));
+		} else {
+			this.setupExistingState(data.data);
```
hakubo: what is data?
maybe a better var name here would help?

front/main/app/components/infobox-builder.js
```diff
@@ -23,6 +23,17 @@ export default Ember.Component.extend(
 			return this.get('isPreviewItemHovered') && !this.get('isPreviewItemDragged');
 		}),
 
+		sortableGroupClassNames: Ember.computed('theme', function () {
```
RafLeszczynski: please consider adding test for this computed property

front/main/app/models/infobox-builder.js
RafLeszczynski: +1. In addition I'd recommend avoid `data` as  argument or variable name. What do you think about adding this to guidelines?

front/main/app/models/infobox-builder.js
```diff
@@ -324,24 +351,32 @@ InfoboxBuilderModel.reopenClass({
...
-	 * @param {Em.Array} state
+	 * @param {Ember.Object} model
 	 * @returns {String} stringified object
 	 */
-	prepareStateForSaving(state) {
-		const plainState = state.map((item) => {
-			delete item.infoboxBuilderData;
-			return item;
-		}).toArray();
+	prepareDataForSaving(model) {
```
RafLeszczynski: what do you think about adding test for this method ?

front/main/app/routes/infobox-builder.js
```diff
@@ -7,56 +7,119 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+					dataAndAssets: this.loadInfoboxDataAndAssets(templateName),
+					ponto: this.loadPonto()
+				};
+
+				Ember.RSVP.hash(promises)
+					.then((responses) => {
+						this.setupStyles(responses.dataAndAssets);
+						this.setupInfoboxData(responses.dataAndAssets);
+					})
+					.then(this.isWikiaContext)
```
RafLeszczynski: shouldn't we check `isWikiaContext` before setup? if this check fails there is no need for setup

front/main/app/routes/infobox-builder.js
```diff
@@ -7,56 +7,119 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+					.then(resolve)
+					.catch(reject);
 			} else {
-				reject();
+				reject('Infobox builder has to be loaded in an iframe');
```
RafLeszczynski: do we want to translate this message?

front/main/app/routes/infobox-builder.js
```diff
@@ -126,111 +191,68 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
-
-		if (assets) {
-			templateData = JSON.parse(assets);
-			if (templateData.data) {
-				this.controllerFor('infobox-builder').set('templateData', templateData.data);
-			}
+	setupInfoboxData(serverResponse) {
+		const infoboxData = serverResponse.data,
+			controller = this.controllerFor('infobox-builder');
+		let infoboxDataParsed = {};
```
RafLeszczynski: do we want to set `infoboxData` on controller only if `serverResponse.data` exist? if yes what about:

```JS
if (infoboxData) {
    controller.set('infoboxData', JSON.parse(infoboxData));
}
```
in this case we don't need extra `infoboxDataParsed` variable

the other option is to always set `infoboxData` on controller and if `serverResponse.data` doesn't exist set `{}`

```JS
    controller.set('infoboxData', infoboxData ? JSON.parse(infoboxData) : {});
```

RafLeszczynski: another thing is doing `JSON.parse` without `try catch`. Do we want to fail hard or have some graceful error fallback?

front/main/app/routes/infobox-builder.js
```diff
@@ -126,111 +191,68 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+
+		if (infoboxData) {
+			infoboxDataParsed = JSON.parse(infoboxData);
+			controller.set('infoboxData', infoboxDataParsed);
 		}
 
-		return promiseResponseArray;
+		controller.set('isNew', serverResponse.isNew);
```
RafLeszczynski: shouldn't we set `isNew` to false if `serverResponse.isNew` is `undefined`?

front/main/app/routes/infobox-builder.js
rogatty: It will be visible only in the console and we don't translate console messages anywhere else. 
TL;DR: no, not really.

front/main/app/routes/infobox-builder.js
rogatty: Resolved offline - it would require adding complexity as we would have to pass `responses` into `isWikiaContext` and then resolve promise which is there with the same `responses`. It's simpler as it is now and isn't that much of performance hit.

front/main/app/models/infobox-builder.js
rogatty: I've changed the variable name to `infoboxData` but `data` property is trickier and I've opened a ticket to fix it properly:
https://wikia-inc.atlassian.net/browse/DAT-3932

front/main/app/routes/infobox-builder.js
```diff
@@ -126,111 +191,73 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+	 * @param {Object} serverResponse
 	 * @returns {Ember.RSVP.Promise}
 	 */
-	setupStyles(promiseResponseArray) {
+	setupStyles(serverResponse) {
 		Ember.$('body').addClass('infobox-builder-body-wrapper');
 
-		return new Ember.RSVP.Promise((resolve) => {
+		if (serverResponse.css) {
 			let html = '';
```
RafLeszczynski: you could also do:

```JS
const html = serverResponse.css.map((url) => {return `<link type="text/css" rel="stylesheet" href="${url}">`}).join('');
```

front/main/app/routes/infobox-builder.js
```diff
@@ -126,111 +191,73 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+		const infoboxData = serverResponse.data,
+			controller = this.controllerFor('infobox-builder');
+
+		let infoboxDataParsed = null;
+
+		if (infoboxData) {
+			try {
+				infoboxDataParsed = JSON.parse(infoboxData);
+			} catch (e) {
+				Ember.Logger.error('Could not parse infobox data as JSON', infoboxData);
```
RafLeszczynski: it would be better for the user to get general IB error here instead of default infobox state for new template

front/main/app/components/infobox-builder.js
rogatty: Done in https://github.com/Wikia/mercury/commit/2bb894bbddc74c7841399de27e0dc93198ccfd42

RafLeszczynski:
LGTM

PR #2136, opened 2016-03-09
CE-3475 Add shift on Android due to Tap to Search

kamilkoterba:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3475

## Description

Moving edit button away from bottom edge on Android phones to avoid being covered by [Google Tap to Search](https://developers.google.com/web/updates/2015/10/tap-to-search)

## Reviewers

@Wikia/community-engineering


front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -5,8 +5,7 @@ import LanguagesMixin from '../mixins/languages';
 export default Ember.Component.extend(
 	LanguagesMixin,
 	{
-		classNames: ['article-edit-highlighted-entrypoint'],
-		classNameBindings: ['displayEdit'],
+		classNameBindings: ['displayEdit', 'entrypoinClass'],
```
adamkarminski: A typo - should be entrypoin**t**Class

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -18,6 +17,12 @@ export default Ember.Component.extend(
 
 			return '';
 		}),
+		entrypoinClass: Ember.computed(() => {
+			// CE-3475 Shift for Android due to Tap to Search
+			return navigator.userAgent.toLowerCase().indexOf('android') > -1 ?
```
adamkarminski: shouldn't we check for a more specific version, not all "android" devices?

rogatty: Please use https://github.com/Wikia/mercury/blob/dev/front/common/utils/browser.js
If you need a specific version then add it to this util too. Let's keep it in a single place.

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -19,6 +19,27 @@ export default Ember.Component.extend(
 
 			return '';
 		}),
+		shouldShift: Ember.computed(() => {
+			// CE-3475 Shift for Android due to Tap to Search
+			// Only for Chrome v43+ on Android
+			return system === 'android' && isChromeMinVer(43);
```
hakubo: since this does not need anything from the component I would consider setting this as a 'normal' property

`shouldShift: system === 'android' && isChromeMinVer(43),`

and then you can call this like this;
`this.shouldShift`

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -19,6 +19,27 @@ export default Ember.Component.extend(
...
+		}),
+		scrollObserver: Ember.observer('showEdit', function () {
+			if (this.get('showEdit')) {
+				this.set('entrypointClass', this.get('shouldShift') ?
+					'article-edit-highlighted-entrypoint--shifted' :
+					'article-edit-highlighted-entrypoint'
+				);
+				Ember.$(document).on('touchmove.shiftEditButton', this, this.unShift.bind(this));
+				Ember.$(window).on('scroll.shiftEditButton', this, this.unShift.bind(this));
```
Grunny: Rather than calling off below, you can just use `Ember.$(document).one(...);` and `Ember.$(window).one(...);`.

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -19,6 +19,27 @@ export default Ember.Component.extend(
...
+			}
+		}),
+		unShift() {
+			this.set('entrypointClass', 'article-edit-highlighted-entrypoint');
+			Ember.$(document).off('touchmove.shiftEditButton', this.debouncedScroll);
+			Ember.$(window).off('scroll.shiftEditButton', this.debouncedScroll);
```
Grunny: If you use `.one()` this is irrelevant, but `this.debouncedScroll` is a copy/paste error as it isn't defined here and it's not the handler that was attached.

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -19,6 +19,27 @@ export default Ember.Component.extend(
 
 			return '';
 		}),
+		shouldShift: Ember.computed(() => {
+			// CE-3475 Shift for Android due to Tap to Search
+			// Only for Chrome v43+ on Android
+			return system === 'android' && isChromeMinVer(43);
+		}),
+		scrollObserver: Ember.observer('showEdit', function () {
```
hakubo: observers are discouraged to be used in apps
https://emberway.io/ember-js-functional-programming-and-the-observer-effect-48901c3b84d7#.wm5vza6pp

entrypointClass should be a computed property

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -1,13 +1,16 @@
 import Ember from 'ember';
 import TrackClickMixin from '../mixins/track-click';
 import LanguagesMixin from '../mixins/languages';
+import {system, isChromeMinVer} from 'common/utils/browser';
+
+const entrypointClassDefault = 'article-edit-highlighted-entrypoint',
```
hakubo: I think it would be easier to reason about if you declared this inside computed property

front/main/app/components/article-edit-highlighted-entrypoint.js
```diff
@@ -1,13 +1,16 @@
...
+	entrypointClassShifted = 'article-edit-highlighted-entrypoint--shifted';
```
hakubo: `${entrypointClassDefault}--shifted` - look less error prone

hakubo:
two minor comments
LGTM

azarek:
LGTM!

PR #2138, opened 2016-03-09
XW-1236 | remove $wgEnableCategoryPagesInMercury

hakubo:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1236

## Description

Checking if we have Category: in a link should be enough to enable that on EN comunities

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

vforge:
LGTM

vforge:
Remember about removing that in MW too.

PR #2141, opened 2016-03-09
DAT-3720, add hyphens to infobox labels

RafLeszczynski:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-3720
* https://github.com/Wikia/app/pull/9937/files

## Description

Add hyphens to infobox labels

## Reviewers

@SebastianMarzjan @rogatty @idradm


SebastianMarzjan:
Looks good; I'm thinking about the lang property on the articles (we set it to userlang, not content-lang)

rogatty:
@SebastianMarzjan are you sure about that?

In console:
```
> $('article').closest('[lang]')
< [<div class=​"page-wrapper row" lang=​"en" dir=​"ltr">​…​</div>​]
```

In the template that is rendered on server side:
```
<div class="page-wrapper" lang="{{wikiVariables.language.content}}" dir="{{wikiVariables.language.contentDir}}">
```

In the template that is used by Ember:
```
<div class="page-wrapper row" lang={{language.content}} dir={{language.contentDir}}>
```

front/main/app/styles/module/article/_portable-infobox.scss
```diff
@@ -191,6 +191,7 @@ $hero-caption-line-height: 18px;
 
 		.pi-data-label {
 			flex-basis: 40%;
+			hyphens: auto;
```
vforge: Shouldn't we re-use this solution? https://github.com/Wikia/app/pull/9937

SebastianMarzjan:
@rogatty Mental shortcut; I was referring to the MW app (Oasis skin) as part of this ticket. Good that we have that checked for Mercury too now, thanks!

front/main/app/styles/module/article/_portable-infobox.scss
hakubo: in mercury we have autoprefixer - I think this should handle that so we don't have to write mixin for that.

hakubo:
lgtm

PR #2146, opened 2016-03-10
XW-1206 | Track GA traffic before ember stands up

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1206
* https://github.com/Wikia/app/pull/9940

## Description

We want to send initial PV to UA as soon as it's possible. This PR is:
* [x] Querying new MercuryAPI endpoint which sets all required dimensions in `window.trackingDimensions`
* [x] Uses `$script.js` to loads optimizely, UA and MercuryApi asynchronously and then calls the initial PV with all the data
*  **Note**: This should make loading of Ember much faster, because before this PR we've started loading Optimizely when ember was starting and we've waited for it to end. (initial comparison shows between -200ms up to 2s gain)
* [x] Moves UA tracker to the baseline
* [x] Moves jQuery load to the head and disables loading it in Ember's `vendor.js`
* [x] Removes some code leftovers after moving rest of the trackers to the server-side
* [x] Is doing some additional optimizations, like removing need for separate `updateTrackedUrl`

## Reviewers

@Wikia/x-wing


server/app/views/_partials/tracking-ua.hbs
```diff
@@ -1,25 +1,22 @@
...
+		window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
 
-			// Creates an initial ga() function.  The queued commands will be executed once analytics.js loads.
-			i[r] = i[r] || function () {
-				(i[r].q = i[r].q || []).push(arguments);
-			},
+		// Load UA
+		$script(['{{server.gaUrl}}'], 'ua');
+		// load window.trackingDimensions
+		$script(['/wikia.php?controller=MercuryApi&method=getTrackingDimensions&title=' + encodeURIComponent(Mercury.article.data.article.displayTitle)], 'dimensions');
```
bognix: window.Mercury

front/common/baseline/trackerUniversalAnalytics.js
```diff
@@ -0,0 +1,374 @@
...
+			// eslint-disable-next-line no-console
+			console.log('Cannot initialize UA; please provide dimensions');
+		} else {
+			dimensions = integrateOptimizely(dimensions);
+			setDimensions(dimensions);
+
+			// All domains that host content for Wikia
+			// Use one of the domains below. If none matches, the tag will fall back to
+			// the default which is 'auto', probably good enough in edge cases.
+			const domain = [
```
hakubo: we moved all our wikis under wikia.com - I think we can simplify it now to always use wikia.com

front/common/baseline/trackerUniversalAnalytics.js
```diff
@@ -0,0 +1,374 @@
...
+	 * @param {UniversalAnalyticsDimensions} dimensions
+	 * @returns {void}
+	 */
+	function initialize(dimensions) {
+		if (typeof dimensions === 'undefined') {
+
+			// eslint-disable-next-line no-console
+			console.log('Cannot initialize UA; please provide dimensions');
+		} else {
+			dimensions = integrateOptimizely(dimensions);
```
bognix: could you rename the method to something more descriptive - for instance `setDimensionsForOptimizelyExepriments` ?

front/common/baseline/trackerUniversalAnalytics.js
```diff
@@ -0,0 +1,374 @@
...
+				'marveldatabase.com', 'memory-alpha.org', 'uncyclopedia.org',
+				'websitewiki.de', 'wowwiki.com', 'yoyowiki.org'
+			].filter((domain) => document.location.hostname.indexOf(domain) > -1)[0];
+
+			accounts = M.prop('tracking.ua');
+
+			initAccount(accountPrimary, domain);
+			initAccount(accountAds, domain);
+
+			if (Boolean(M.prop('isGASpecialWiki') || Mercury.wiki.isGASpecialWiki)) {
```
bognix: are you sure you don't have to use `window.Mercury` here?

server/app/facets/operations/prepareCategoryData.js
```diff
@@ -1,7 +1,8 @@
 import * as Utils from '../../lib/Utils';
 import {gaUserIdHash} from '../../lib/Hashing';
 import localSettings from '../../../config/localSettings';
-import {isRtl, getUserId, getQualarooScriptUrl, getOpenGraphData, getLocalSettings} from './preparePageData';
+// eslint-disable-next-line max-len
+import {isRtl, getUserId, getQualarooScriptUrl, getOptimizelyScriptUrl, getOpenGraphData, getLocalSettings} from './preparePageData';
```
hakubo: we can break this line!

server/app/facets/operations/prepareCuratedContentData.js
```diff
@@ -1,6 +1,7 @@
 import {parseQueryParams} from '../../lib/Utils';
 import {gaUserIdHash} from '../../lib/Hashing';
-import {isRtl, getUserId, getQualarooScriptUrl, getOpenGraphData, getLocalSettings} from './preparePageData';
+// eslint-disable-next-line max-len
+import {isRtl, getUserId, getQualarooScriptUrl, getOptimizelyScriptUrl, getOpenGraphData, getLocalSettings} from './preparePageData';
```
hakubo: break all the lines!

server/app/views/_partials/optimizely.hbs
```diff
@@ -1,3 +1,5 @@
 {{#if optimizelyScript}}
-	<script src="{{optimizelyScript}}"></script>
+	<script>$script(['{{optimizelyScript}}'], 'optimizely');</script>
```
hakubo: this looks odd - I would add a comment why we load it this way

server/app/views/_partials/tracking-ua.hbs
hakubo: since we load jquery here anyway
I think we should use $.getJson and not some custom script

server/app/views/_partials/tracking-ua.hbs
```diff
@@ -1,25 +1,22 @@
...
 
-			// Sets the time (as an integer) this tag was executed.  Used for timing hits.
-			i[r].l = 1 * new Date ();
+		// when everything loads, initialize trackers and track initial pageview
+		$script.ready(['dimensions', 'optimizely'], function() {
+			var ua = M.tracker.UniversalAnalytics;
```
bognix: I guess it should be `window.M`

front/common/utils/track.js
```diff
@@ -189,10 +166,10 @@ export function trackPageView(adsContext) {
 		const Tracker = trackers[tracker];
 
 		if (typeof Tracker.prototype.trackPageView === 'function') {
```
bognix: since the Internal is the only one that left can we simply call `Internal.prototype.trackPageView` without any foreach etc?

server/app/views/_partials/tracking-ua.hbs
vforge: I'll add the title to `M.prop` - now it looks strange - and on non-article pages we don't have `article`

front/common/baseline/trackerUniversalAnalytics.js
vforge: Nope, we don't  - but I may add it just to be sure. Do you want it?

server/app/facets/operations/prepareArticleData.js
```diff
@@ -38,6 +39,7 @@ export default function prepareArticleData(request, data) {
 			articlePage: data.page,
 			server: data.server,
 			wikiVariables: data.wikiVariables,
+			displayTitle: request.params.title.replace(/_/g, ' '),
```
bognix: consider calling getTitle here and inside getTitle add check if articleData is set

front/common/utils/track.js
```diff
@@ -160,42 +140,29 @@ export function track(params) {
...
 	if (M.prop('queryParams.noexternals')) {
 		return;
 	}
 
-	Object.keys(trackers).forEach((tracker) => {
-		const Tracker = trackers[tracker];
-
-		if (typeof Tracker.prototype.trackPageView === 'function') {
-			const instance = new Tracker(isSpecialWiki());
+	const instance = new Internal();
```
rogatty: Can this name be more explicit? E.g. `internalTrackerInstance`

front/common/utils/track.js
```diff
@@ -204,12 +171,26 @@ export function trackPageView(adsContext) {
...
 	}
 
 	Ads.getInstance().trackKruxPageView();
 }
 
 /**
+ * @param {string} experimentName
+ * @returns {*}
+ */
+export function getAbTestGroup(experimentName) {
```
rogatty: This really shouldn't be in `utils/track.js`. It's not related to tracking at all.

server/app/facets/mediawiki-page.js
```diff
@@ -79,6 +79,8 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
 		ns = pageData.ns;
 		result.mediaWikiNamespace = ns;
 	}
+	// pass page title to front
+	result.pageTitle = request.params.title;
```
rogatty: Why don't we use `displayTitle`? If we need the title from params and nothing else then maybe this name should be explicit about it? E.g. `result.urlTitleParam`

front/common/utils/track.js
vforge: It's cleaned in https://github.com/Wikia/mercury/tree/XW-1209
Also: talked offline

server/app/views/_partials/tracking-ua.hbs
```diff
@@ -1,25 +1,29 @@
...
-			// Creates an initial ga() function.  The queued commands will be executed once analytics.js loads.
-			i[r] = i[r] || function () {
-				(i[r].q = i[r].q || []).push(arguments);
-			},
+    // when optimizely loads, initialize trackers and track initial pageview
+    $script.ready(['optimizely'], function() {
+			$.getJSON('/wikia.php', {
+				controller: 'MercuryApi',
+				method: 'getTrackingDimensions',
+				title: M.prop('urlTitleParam'),
```
rogatty: It isn't really clear that this will be undefined for everything other than wiki page. Should we even send it when undefined?

server/app/facets/operations/prepare-page-data.js
```diff
@@ -46,6 +46,22 @@ export function getQualarooScriptUrl(request) {
 }
 
 /**
+ * @param {Hapi.Request} request
+ * @returns {boolean|String}
+ */
+export function getOptimizelyScriptUrl(request) {
+	// all the third party scripts we don't want to load on noexternals
+	if (!request.query.noexternals) {
+		// optimizely
```
rogatty: Those two comments suggest that there should be more scripts here. Is this still relevant?
If not then how about a single `if`?

server/app/views/_partials/optimizely.hbs
```diff
@@ -1,3 +1,12 @@
 {{#if optimizelyScript}}
-	<script src="{{optimizelyScript}}"></script>
+{{!--
+	We need to load optimizely via $script, because $script will tell us
+	when it loads - see tracking-ua.hbs for the usage
+--}}
+	<script>$script(['{{optimizelyScript}}'], 'optimizely');</script>
+{{else}}
+{{!--
+	Mark 'optimizely' as already loaded
```
rogatty: Why?

server/app/views/_partials/tracking-ua.hbs
```diff
@@ -1,25 +1,29 @@
...
-		(function (i, s, o, g, r, a, m) {
-			i['GoogleAnalyticsObject'] = r; // Acts as a pointer to support renaming.
+		// Load UA async and flag $script when ready
+		$script(['{{server.gaUrl}}'], 'ua');
 
-			// Creates an initial ga() function.  The queued commands will be executed once analytics.js loads.
-			i[r] = i[r] || function () {
-				(i[r].q = i[r].q || []).push(arguments);
-			},
+    // when optimizely loads, initialize trackers and track initial pageview
```
rogatty: Indentation is broken here (spaces instead of tab).

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -148,17 +149,46 @@ if (typeof window.M.tracker === 'undefined') {
 	 * @returns {boolean} true if dimensions were successfully set
 	 */
 	function setDimensions(dimensionsToSet, overwrite) {
-		if (!dimensionsToSet.length) {
+		if (!Object.keys(dimensionsToSet).length) {
```
rogatty: Please update the JSDoc as it says this is an array.

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -29,12 +29,13 @@ if (typeof window.M.tracker === 'undefined') {
...
-	let dimensions = [],
+	let tracked = [],
+		createdAccounts = [],
+		dimensions = {},
 		dimensionsSynced = false,
 		accounts;
 
-	const createdAccounts = [],
-		tracked = [],
+	const pageDimensions = [3, 14, 19, 25],
```
rogatty: Please add a comment explaining what is it for.

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -327,11 +369,10 @@ if (typeof window.M.tracker === 'undefined') {
 						variationName = optimizely.variationNamesMap[experimentId];
 
 					dimensions[dimension] = `Optimizely ${experimentName} (${experimentId}): ${variationName}`;
+					dimensionsSynced = false;
```
rogatty: Can we do it by using `setDimension()` instead of manually?

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -342,36 +383,30 @@ if (typeof window.M.tracker === 'undefined') {
...
-					const noGroup = abList.nouuid ? 'NOBEACON' : 'NOT_IN_ANY_GROUP';
+				if (abExp && abExp.flags && abExp.flags.ga_tracking) {
+					// GA Slots 40-49 are reserved for our AB Testing tool. Anything outside that
+					// range could potentially overwrite something that we don't want to
+					if (abSlot >= 40 && abSlot <= 49) {
+						const noGroup = abList.nouuid ? 'NOBEACON' : 'NOT_IN_ANY_GROUP';
 
-					dimensions[abSlot] = abExp.group ? abExp.group.name : noGroup;
+						dimensions[abSlot] = abExp.group ? abExp.group.name : noGroup;
+						dimensionsSynced = false;
```
rogatty: Can we do it by using `setDimension()` instead of manually?

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -0,0 +1,465 @@
...
+				 * Which, ultimately, tells us if new dimensions are
+				 * different from the old ones and if we need a re-syncing.
+				 */
+				$.each(dimensions, (index, value) => {
+					dimensionsSynced = dimensionsSynced && (oldDimensions[index] === value);
+				});
+			}
+		}
+
+		return true;
```
rogatty: Does it make any sense to always return `true`?

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -0,0 +1,465 @@
...
+	 */
+	function trackPageView(uaDimensions, overrideUrl) {
+		// reset per-page dimensions
+		$.each(pageDimensions, function () {
+			setDimension(this, '');
+		});
+
+		// set per-page dimensions if they were passed
+		if (typeof uaDimensions === 'object') {
+			$.each(uaDimensions, (key, value) => {
```
hakubo: Minor: ```$.each(uaDimensions, setDimension)```

front/common/baseline/tracker-universal-analytics.js
hakubo: begining of function:
if (!Object.keys(dimensionsToSet).length) {
 +			return false;
 +		}

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -406,12 +443,23 @@ if (typeof window.M.tracker === 'undefined') {
...
 	// API
 	M.tracker.UniversalAnalytics = {
 		initialize,
+		destroy,
 		setDimension,
 		track,
 		trackAds,
-		trackPageView
+		trackPageView,
+		// expose internals for unit test
```
rogatty: `destroy` is for unit tests only too, right?

front/common/baseline/tracker-universal-analytics.js
rogatty: Oh, you're right. Let me ask another question. Does it make any sense to return boolean from this function? I don't see it being used anywhere. Or did I miss something?

server/app/views/_partials/tracking-ua.hbs
```diff
@@ -18,15 +18,15 @@
 				cb: Date.now
 			}, function(data) {
 					if (data && data.dimensions) {
```
rogatty: Double `\t`

front/common/baseline/tracker-universal-analytics.js
vforge: Yes.

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	/**
+	 * Synchronously sets the UA dimensional context
+	 *
+	 * @param {UniversalAnalyticsDimensions} dimensionsToSet - array of dimensions to set, may be strings or functions
+	 * @param {boolean} [overwrite=true] - set to true to overwrite all preexisting dimensions and unset ones not declared
+	 * @returns {boolean} true if dimensions were successfully set
+	 */
+	function setDimensions(dimensionsToSet, overwrite) {
```
rogatty: Let's write a unit test for this method that would check all the logic that is here. This is not a blocker for merge.

bognix:
LGTM

rogatty:
LGTM

PR #2151, opened 2016-03-10
CE-3494 Enable the HTE on all EN communities without surfacing it

adamkarminski:
## Links

- https://wikia-inc.atlassian.net/browse/CE-3494

## Description

Since the Optimizely experiment had to be stopped - we want the users we already enabled it for to see the editor when they want to edit. Assuming that the editor is hard to find (low percentage of people selecting text on websites) - we want to enable it for EN communities.

## Reviewers

@Wikia/community-engineering .

### DIFF
--- a/front/main/app/components/article-wrapper.js
+++ b/front/main/app/components/article-wrapper.js
@@ -4,8 +4,6 @@ import TextHighlightMixin from '../mixins/text-highlight';
 import TrackClickMixin from '../mixins/track-click';
 import ViewportMixin from '../mixins/viewport';
 import {track, trackActions} from 'common/utils/track';
-import {getExperimentVariationNumber} from 'common/utils/variantTesting';
-
 
 /**
  * @typedef {Object} ArticleSectionHeader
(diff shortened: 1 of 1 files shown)


front/main/app/components/article-wrapper.js
```diff
@@ -152,9 +152,7 @@ export default Ember.Component.extend(
 				!this.get('highlightedEditorEnabled');
 		}),
 
-		highlightedEditorEnabled: Ember.computed(() => {
-			return getExperimentVariationNumber({dev: '5170910064', prod: '5164060600'}) === 1;
-		}),
+		highlightedEditorEnabled: Ember.computed(() => Mercury.wiki.language.content === 'en'),
```
hakubo: This will enabled it on all EN communities - is that what you want? PR title says ALL communities.

lukaszkonieczny:
Why don't we want to use optimizely bucket cookies?

rogatty:
LGTM

azarek:
LGTM

PR #2152, opened 2016-03-10
SOC-1844 & SOC-1522 | Locking and unlocking Discussion posts

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/JIRA-1844
* http://wikia-inc.atlassian.net/browse/JIRA-1522

## Description

We're adding a possibility of locking posts for moderators. After a post is locked users cannot add any more replies to it. Moderators can unlock the post.

## Reviewers

@rafalkalinski @kvas-damian


front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -164,6 +164,54 @@ export default Ember.Mixin.create({
...
+	 * @returns {Ember.RSVP.Promise|void}
+	 */
+	lockPost(post) {
+		if (checkPermissions(post, 'canDelete')) {
+			return ajaxCall({
+				method: 'PUT',
+				dataType: 'text',
+				url: M.getDiscussionServiceUrl(`/${this.wikiId}/threads/${post.threadId}/lock`),
+				success: () => {
+					Ember.setProperties(post, {
```
hakubo: why can't we use here
post.set('isLocked', true);?

What is the benefit of Ember.set/setProperties?

front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -164,6 +164,54 @@ export default Ember.Mixin.create({
...
 	},
 
+
+	/**
+	 * Locks a post in the service
+	 * @param {object} post
+	 * @returns {Ember.RSVP.Promise|void}
+	 */
+	lockPost(post) {
+		if (checkPermissions(post, 'canDelete')) {
```
hakubo: do you plan to do something if that is false?
otherwise it'll silently ignore it

jQuery does that - and I don't think we are super fans of that. What do you think?

front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -164,6 +164,54 @@ export default Ember.Mixin.create({
...
+			});
+		}
+	},
+
+	/**
+	 * Unlocks a post in the service
+	 * @param {object} post
+	 * @returns {Ember.RSVP.Promise|void}
+	 */
+	unlockPost(post) {
```
hakubo: this won't return promise if checkPermissions returns false. 
That is a suboptimal api. I think this should always return promise

front/main/app/mixins/discussion-moderation-route.js
```diff
@@ -120,5 +120,33 @@ export default Ember.Mixin.create({
 				Ember.set(item, 'isLoading', false);
 			});
 		},
+
+		/**
+		 * Pass post locking to the model
+		 * @param {object} post
+		 * @returns {void}
+		 */
+		lock(post) {
```
hakubo: even if that won't be called when checkPermissions(post, 'canDelete') returns false
I have no way of knowing that without reading whole discussions code.

another point that {un}lockPost should always return promise

front/main/app/components/discussion-more-options.js
```diff
@@ -20,4 +20,14 @@ export default Ember.Component.extend({
 		return this.get('post._embedded.userData.0.hasReported') !== true &&
 			this.get('currentUser.isAuthenticated') === true;
 	}),
+
+	canLock: Ember.computed('post.isEditable', 'post.canDelete', function () {
+		// 'canDelete' until we have 'canLock' coming from the service
```
kvas-damian: Ticket number?

front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -14,6 +14,10 @@ export default Ember.Component.extend({
 		return checkPermissions(this.get('post'), 'canModerate');
 	}),
 
+	showButtons: Ember.computed('isReported', function () {
+		return this.get('canShowModButtons') && this.get('isReported');
```
kvas-damian: Shouldn't we handle a case when `canShowModButtons` changes value? Probably it's not the case now, but maybe in future we'll add a periodical check of user permissions.

front/main/app/mixins/discussion-moderation-model.js
kvas-damian: post is not Ember object here

front/main/app/mixins/discussion-more-options.js
```diff
@@ -40,5 +40,29 @@ export default Ember.Mixin.create({
...
 			this.attrs.report(item);
 		},
+
+		/**
+		 * Locks an item and closes pop-over
+		 * @param {object} item - post
+		 * @returns {void}
+		 */
+		lock(item) {
+			// There's no API for closing Ember Pop-Up, so that's
```
kvas-damian: I heard there is an API for closing `pop-over`. @rafalkalinski can you confirm that?

front/main/app/models/discussion-forum.js
```diff
@@ -11,6 +11,7 @@ const DiscussionForumModel = DiscussionBaseModel.extend(
 		normalizePostData(post) {
 			post.firstPost = post._embedded.firstPost[0];
 			post.firstPost.isReported = post.isReported;
+			post.firstPost.isLocked = !post.isEditable;
```
kvas-damian: This is something that is about to be changed - https://wikia-inc.atlassian.net/browse/SOC-2079. Can we make sure which code will be released earlier?

front/main/app/components/discussion-more-options.js
```diff
@@ -20,4 +20,14 @@ export default Ember.Component.extend({
...
+		return this.get('isLockable') && !this.get('post.isLocked') && checkPermissions(this.get('post'), 'canDelete');
```
kvas-damian: Is it possible to lock deleted item?

front/main/app/mixins/discussion-moderation-model.js
bkoval: We're not planning on doing anything if it's false. We hide UI controls that allow you for such an action if you don't have permissions to do it anyway. This is just a safety precaution and we use it for all authenticated / mod actions in discussions.

front/main/app/mixins/discussion-moderation-model.js
bkoval: Doesn't make sense to fix it here as we use this scheme for all models. However, we have 2 technical debt tickets that we'll tackle during next 2 sprints that will help avoid confusion: 
https://wikia-inc.atlassian.net/browse/SOC-2084
https://wikia-inc.atlassian.net/browse/SOC-2093

front/main/app/mixins/discussion-moderation-route.js
bkoval: https://github.com/Wikia/mercury/pull/2152#discussion_r55864224

front/main/app/components/discussion-post-card-top-note.js
bkoval: Yeah, we will be cleaning up these things in https://wikia-inc.atlassian.net/browse/SOC-2084 :)

front/main/app/models/discussion-forum.js
bkoval: Yes, we agreed with @armonr that we'll adjust front-end flag as a part of a backend ticket https://wikia-inc.atlassian.net/browse/SOC-2079 (currently in progress).

front/main/app/components/discussion-more-options.js
bkoval: It is, Discussion Service allows us to do that :)

front/main/app/models/discussion-forum.js
armonr: :+1: Please use "isEditable" for now.
SOC-2079 will add "isLocked" in addition to existing "isEditable". Once you have time to change to "isLocked", the old field will be removed: https://wikia-inc.atlassian.net/browse/SOC-2143

front/main/app/components/discussion-more-options.js
armonr: The FE ticket to be added for this, should be marked as blocked by SOC-2079

front/main/app/mixins/discussion-moderation-model.js
rafalkalinski: Since we are setting only one property, Ember.set would be a little bit better here.

front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -164,6 +164,54 @@ export default Ember.Mixin.create({
...
+		if (checkPermissions(post, 'canDelete')) {
+			return ajaxCall({
+				method: 'DELETE',
+				dataType: 'text',
+				url: M.getDiscussionServiceUrl(`/${this.wikiId}/threads/${post.threadId}/lock`),
+				success: () => {
+					Ember.setProperties(post, {
```
rafalkalinski: Ember.set

front/main/app/mixins/discussion-more-options.js
rafalkalinski: Yes, there is.

front/main/app/models/discussion-reported-posts.js
```diff
@@ -40,12 +40,16 @@ const DiscussionForumModel = DiscussionBaseModel.extend(
...
 
-					let allPosts;
+					newPosts.forEach((post) => {
+						if (post.hasOwnProperty('createdBy')) {
+							post.createdBy.profileUrl = this.get('userProfileUrl');
+						}
 
-					newPosts.forEach(this.normalizeThreadData);
+						this.normalizeThreadData(post);
+						post.isLocked = !post.isReply && !post._embedded.thread.isEditable;
```
rafalkalinski: can we put this into the 'normalizeThreadData' method?

front/main/app/models/discussion-reported-posts.js
```diff
@@ -111,6 +115,8 @@ DiscussionForumModel.reopenClass({
 							namespace: 'User',
 							title: post.createdBy.name
 						});
+
+						post.isLocked = !post.isReply && !post._embedded.thread.isEditable;
```
rafalkalinski: I believe it belongs to 'normalizeThreadData' method and for sure (!) it should exists outside of that condition.

rafalkalinski:
LGTM

front/main/app/mixins/discussion-more-options.js
```diff
@@ -6,39 +6,48 @@ import Ember from 'ember';
 export default Ember.Mixin.create({
```
kvas-damian: I believe we don't need this mixin anymore.

kvas-damian:
LGTM

PR #2154, opened 2016-03-11
XW-849 | server/app files renamed to dashed-case

Warkot:
CHANGE ALL THE NAMES

## Links

* http://wikia-inc.atlassian.net/browse/JIRA-849

## Description

Dashed-case filenames are cool. For example they're system agnostic (e.g. for Linux file `FOO.JS` is different than `foo.js`, but for Windows it's the same file).

## Reviewers

@Wikia/x-wing


front/auth/tests/unit/common/cookie.js
```diff
@@ -1,4 +1,4 @@
-QUnit.module('auth/common/Cookie', function (hooks) {
+QUnit.module('auth/app/common/cookie', function (hooks) {
```
vforge: `app/`?

bognix: yes, this path is correct

front/main/app/instance-initializers/setup-tracking.js
```diff
@@ -1,7 +1,7 @@
-import Ads from 'common/modules/Ads';
-import UniversalAnalytics from 'common/modules/Trackers/UniversalAnalytics';
+import Ads from 'common/modules/ads';
+import UniversalAnalytics from 'common/modules/trackers/universal-analytics';
 import {integrateOptimizelyWithUA} from 'common/utils/variantTesting';
```
rogatty: `variant-testing`?

hakubo:
Guideline: https://github.com/Wikia/guidelines/pull/109

PR #2160, opened 2016-03-11
SOC-2105 | hardcoded string for the Discussions feature name in the desktop header

rafalkalinski:
## Links

https://wikia-inc.atlassian.net/browse/SOC-2105

## Description

We're currently using the forum name on the post list in desktop. However, we should just use the hardcoded word "Discussions" since the forum names (categories) will be displayed behind the filter menu.
This should be updated on the Post List and the Reported Posts List.

## Reviewers

@bkoval


bkoval:
LGTM

PR #2162, opened 2016-03-12
SOC-2119 Reporting: Delete reply option does not work on filtered view

rafalkalinski:
## Links

https://wikia-inc.atlassian.net/browse/SOC-2119

## Description

Like in the summary: 
"Reporting: Delete reply option does not work on filtered view"
and I changed one component name, to be more consistent with the rest .

## Reviewers

@bkoval @kvas-damian


bkoval:
LGTM :kimono:

kvas-damian:
LGTM

PR #2163, opened 2016-03-14
add host to logs so we can see if requests are made with correct host

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-90

## Description

After small investigation I can tell that 
* majority of errors are 502
* I can reproduce that constantly when wrong host is sent to icache (e.g. host.wikia.com/ or http://host.wikia.com)
* I need to see if we send correct host to icache

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

PR #2168, opened 2016-03-14
CE-3486 Save user feedback using UserFeedbackStorage API

adamkarminski:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3486
* https://github.com/Wikia/app/pull/9972

## Description

Saves users feedback using our UserFeedbackStorage API. I had to implement a different method to retrieve a token to protect against CSRF ([app PR](https://github.com/Wikia/app/pull/9972)). The existing one checks permissions to edit a given title and fails if a user does not have them. This is irrelevant to us since we gather feedback without making any changes to articles.

## Reviewers


front/main/app/components/feedback-form.js
```diff
@@ -100,6 +101,32 @@ export default BottomBanner.extend({
 		}, timeout);
 
 	},
+	incrementCookieCounter(cookieName) {
+		const cookieValue = parseInt(Ember.$.cookie(cookieName), 10) || 0;
```
Grunny: Any reason not to reuse `getCookieCounter` here?

front/main/app/components/feedback-form.js
```diff
@@ -100,6 +101,32 @@ export default BottomBanner.extend({
...
+
+		Ember.$.cookie(cookieName, cookieValue + 1);
```
Grunny: Since this cookie is per wikia anyway, it's better to use localStorage.

front/main/app/components/feedback-form.js
```diff
@@ -116,15 +143,50 @@ export default BottomBanner.extend({
...
+			if (userFeedback !== '') {
+				this.setProperties({
+					displayInput: false,
+					displayThanks: true,
+					message: completed.no.message
+				});
+				this.dismissBanner(completed.no.timeout);
 
-			this.dismissBanner(completed.no.timeout);
+				this.getCSRFToken()
```
lukaszkonieczny: Maybe we can extract this code to use it also in infobox experiment

front/main/app/components/feedback-form.js
```diff
@@ -116,15 +143,50 @@ export default BottomBanner.extend({
...
+									controller: 'UserFeedbackStorageApi',
+									method: 'saveUserFeedback'
+								}
+							}),
+							dataType: 'json',
+							data: {
+								token,
+								experimentId,
+								variationId: this.get('variationId'),
+								wikiId: Ember.get(Mercury, 'wiki.id'),
```
lukaszkonieczny: May this wiki id change when user goes to different wiki?

front/main/app/components/feedback-form.js
rogatty: If you decide to use localStorage then please do it with `utils/local-storage-connector`.

front/main/app/components/feedback-form.js
```diff
@@ -116,15 +143,50 @@ export default BottomBanner.extend({
...
-			this.setProperties({
-				displayInput: false,
-				displayThanks: true,
-				message: completed.no.message
-			});
+			const userFeedback = this.get('userFeedback') || '',
+				userFeedbackImpressions = this.getCookieCounter('userFeedbackImpressions'),
+				userFeedbackCount = this.getCookieCounter('userFeedbackCount');
+
+			if (userFeedback !== '') {
```
rogatty: If you used `!Ember.isEmpty(userFeedback)` there would be no need for `|| ''`
http://emberjs.com/api/#method_isEmpty

adamkarminski: Thanks :)

front/main/app/components/feedback-form.js
adamkarminski: We want to track users' exposure to the the feedback question across all wikis, that's why I used a cookie here. I'll update the code to save it under the main domain. Thanks!

front/main/app/components/feedback-form.js
adamkarminski: Nope. :)

front/main/app/components/feedback-form.js
adamkarminski: We want the wikiId to change. The reason it's retrieved in a different way than the `articleId` is that `wikiId` is available globally via the `Mercury` object and the `articleId` comes with the `article-wrapper` model.

front/main/app/components/feedback-form.js
adamkarminski: Yes!

front/main/app/components/feedback-form.js
rogatty: When user goes to different wiki the page is reloaded, so we don't need to worry about it being updated.

lukaszkonieczny: Thanks for clarification!

adamkarminski:
Changes made, UserFeedbackStorage related code abstracted - @Grunny @lukaszkonieczny @rogatty

lukaszkonieczny:
LGTM

PR #2170, opened 2016-03-15
RECIRC-76: Update Fandom tracking code

pauloslund:
## Links
https://wikia-inc.atlassian.net/browse/RECIRC-76

## Description
We've been seeing inconsistent tracking numbers around our Fandom content. This is to clear up those issues.

## Reviewers
@Wikia/x-wing


front/main/app/components/fandom-posts.js
```diff
@@ -17,24 +17,25 @@ export default Ember.Component.extend({
...
-		trackExperimentClick() {
+		trackExperimentClick(url) {
 			trackExperiment(this.get('experimentName'), {
 				action: trackActions.click,
 				category: 'recirculation',
 				label: 'footer'
 			});
+
+			setTimeout(() => {
+				window.location.assign(url);
```
vforge: One small question here: Isn't `href` in anchor element enough to change the page? do we need `window.location.assign` additionally?

pauloslund: The idea is to delay the page change by a small amount to ensure that we send the tracking event before navigating away from the page.

pauloslund:
Merging after a check with @vforge on Slack

PR #2174, opened 2016-03-15
CE-3471: Users satisfaction/emotional feedback

lukaszkonieczny:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3471

## Description

Experiment to gather users satisfaction/emotional feedback about articles.

## Reviewers


front/main/app/components/feedback-form.js
```diff
@@ -0,0 +1,190 @@
...
+		}
+	},
+	helpImproveMessage = 'Your input helps to improve this wiki. Every bit of feedback is highly valued. ' +
+		'What information was missing?',
+	offsetLimit = 0.2,
+	cookieName = 'feedback-form',
+	experimentId = 'USER_SATISFACTION_FEEDBACK';
+
+
+export default BottomBanner.extend(
```
bognix: extending components isn't a good pattern. We had it in our media components and we learned why it's bad pattern in a hard way. I understand that this code experimental but if (or when) implementing this try to shape components in a way that one component won't be dependent on another one.

front/main/app/mixins/user-feedback-storage.js
```diff
@@ -0,0 +1,73 @@
...
+					method: 'getEditToken'
+				},
+				(response) => {
+					if (response.token) {
+						resolve(response.token);
+					} else {
+						reject();
+					}
+				}
+			).fail(reject);
```
bognix: are you sure you want to reject without any error? How about adding some log here for your information (although only if it would be useful to you)

front/main/app/components/feedback-form.js
```diff
@@ -0,0 +1,190 @@
...
+					Ember.$(window).on('scroll.feedbackForm', () => this.checkOffsetPosition());
+				});
+			}
+		},
+		willDestroyElement() {
+			Ember.$(window).off('scroll.feedbackForm');
+		},
+		checkOffsetPosition() {
+			const scrollY = window.scrollY,
+				direction = scrollY >= this.get('lastOffset') ? 0 : 1;
```
Grunny: This is only being used as a boolean, so could just be equal to `scrollY < this.get('lastOffset')` rather than assigning an int value to it.

front/main/app/components/feedback-form.js
bkoval: @bognix Could your team make sure to add this information to guidelines repo? Probably there's a lot more tips like this one that could be added to https://github.com/Wikia/guidelines/tree/master/Ember

front/main/app/mixins/user-feedback-storage.js
adamkarminski: Yup, we'll monitor the logs of API exceptions for this case. And since we do not want to disturb a user's flow with any error, even if we're unable to save the feedback, it will be fine. Good catch though!

PR #2179, opened 2016-03-15
CE-3433 Condition for displaying the New badge and hide on child views

adamkarminski:
## Links

* http://wikia-inc.atlassian.net/browse/CE-3433

## Description

When a user goes through the whole path to get to the Recent Wiki Activity he should not see the new badge (blue dot) anymore at any step. Also, on child views of the local nav the RWA link should not be visible.

## Reviewers

@Wikia/community-engineering


kamilkoterba:
LGTM

azarek:
LGMT

PR #2180, opened 2016-03-16
RECIRC 29: Test collapsed sections on articles

pauloslund:
## Links
https://wikia-inc.atlassian.net/browse/RECIRC-29

## Description
AB Test will be run against 1% of traffic.

## Reviewers
@Wikia/x-wing


front/common/modules/abtest.js
```diff
@@ -13,6 +13,19 @@ export function getGroup(experimentName) {
 }
 
 /**
+ * Check if a user is in a group for an experiment
+ *
+ * @param {String} experimentName
+ * @param {String} groupName
+ * @returns {Boolean}
```
rogatty: Let's be consistent with the return type and return `false` instead of `void`.

front/main/app/models/wiki/article.js
```diff
@@ -130,7 +131,7 @@ ArticleModel.reopenClass({
 				article = data.article;
 
 				articleProperties = $.extend(articleProperties, {
-					content: article.content,
+					content: this.injectSections(article.content),
```
rogatty: Can we do this in `components/article-content.js`? It's where we do all those _bad things_ with article content. This is all about modifying how the article looks and models aren't supposed to be about presentation.

rogatty:
Few things:

1. This breaks lazy loading of images on wikis with `$wgEnableSeoFriendlyImagesForMobile === true`. After opening article, all images are loaded at once. I found that `didEnterViewport()` hook is called on all instances of `components/article-media-gallery` after article is rendered. On dev branch it waits until gallery actually entered the viewport. I currently don't have time to investigate why exactly this happens but it should be fixed before merging this PR. This hook is a part of https://github.com/DockYard/ember-in-viewport and you can test it on http://muppet.paulo.wikia-dev.com/wiki/Kermit if using devbox or http://sandbox-whatever.fallout.wikia.com/wiki/Fallout_4 if using sandbox.

2. This looks like a quick hack to verify some ideas. Please mark this PR with _Experiment_ label (read https://wikia.slack.com/archives/mercury/p1457448486000321) and state it explicitly in the code so this fact isn't lost.

front/main/app/styles/module/article/_content.scss
```diff
@@ -20,4 +20,32 @@
 			margin-bottom: 0;
 		}
 	}
+
+	.collapsible-section-header {
+		border-bottom: 1px solid $gainsboro;
+		margin-bottom: 0.7rem;
+		padding-bottom: 0.7rem;
```
Warkot: https://github.com/Wikia/guidelines/blob/master/SCSS/CodingConventions.md#leadingzero

Warkot:
This PR is over a month old—shouldn't this be closed (or merged)?

pauloslund:
@rogatty I fixed the issue with it loading all images on page load. Seemed to be a problem with how the visibility state manager was handling items that were not visible on screen due to them having a `display: none` applied to them a some point.

@Warkot Some other work ended up becoming a higher priority than this, but it should be good to go now

If it looks good I'll rebase it off of dev before merging

rogatty:
@pauloslund unfortunately this branch doesn't build because of error:
```
File: /usr/wikia/mercury/front/main/tmp/sass_compiler-input_base_path-SOCXWMv7.tmp/0/app/styles/module/_site-footer.scss (13)
The Broccoli Plugin: [SassCompiler] failed with:
Error: ".wikia-footer ul" failed to @extend ".type-smallest".
```
which was fixed in https://github.com/Wikia/mercury/pull/2386.

Please rebase it off of the latest dev, especially that there are some conflicts that might have impact on the logic you introduce (might be something minor, too).

front/main/app/mixins/is-visible.js
```diff
@@ -0,0 +1,42 @@
...
+	{
+		/**
+		 * Update _triggerDidAccessViewport to only trigger an event if the element
+		 * is displayed on the page. Original found here:
+		 * https://github.com/DockYard/ember-in-viewport/blob/2.0.4/addon/mixins/in-viewport.js#L123-L142
+		 *
+		 * @param {bool} hasEnteredViewport
+		 * @returns {void}
+		 */
+		_triggerDidAccessViewport(hasEnteredViewport = false) {
```
rogatty: Let's not do that as it would be PITA to maintain. Please create a fork of ember-in-viewport, use it in our code and then open PR with your change to the original repo. Then it'll be just a matter of watching the PR status and going back to mainstream when it's merged and released.

front/main/package.json
```diff
@@ -47,7 +47,7 @@
     "ember-disable-proxy-controllers": "1.0.1",
     "ember-export-application-global": "1.0.5",
     "ember-getowner-polyfill": "1.0.0",
-    "ember-in-viewport": "2.0.4",
+    "ember-in-viewport": "pauloslund/ember-in-viewport.git#visibility",
```
rogatty: Please use the commit hash instead of the branch name because right now on every time we use `bower update` it will download the code again instead of using existing files.

rogatty:
LGTM

PR #2182, opened 2016-03-16
CE-3508: Hide infobox field about which we ask

lukaszkonieczny:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3508
* https://wikia-inc.atlassian.net/browse/CE-3473

## Description

Hide infobox field about which is question about

## Reviewers


front/main/app/components/portable-infobox-question.js
```diff
@@ -21,66 +21,114 @@ export default Ember.Component.extend(
...
+				});
+			}
+
+			return experimentWikiPage;
 		}),
+		hideInfoboxField(field, value) {
+			Ember.$('.portable-infobox').find('h3').filter(function () {
+				if (this.textContent === field) {
+					if (value) {
+						this.hideFieldValue($(this).next('.pi-data-value'), value);
```
Grunny: `this.hideFieldValue` will not be defined in this scope as `this` will refer to the element.

Grunny:
LGTM.

PR #2186, opened 2016-03-16
CE-3513 A solution for iOS+Safari position: fixed bug 

adamkarminski:
## Links

* http://wikia-inc.atlassian.net/browse/CE-3513

## Description
This is bad. This is ugly. But this is also necessary to work around the [position: fixed Safari bug](http://stackoverflow.com/questions/15199072/ios-input-focused-inside-fixed-parent-stops-position-update-of-fixed-elements). For Safari + iOS combination we will make the fixed element an absolute one for the time an input is visible.

Forgive me.

## Reviewers
@Wikia/x-wing @Wikia/community-engineering


front/main/app/components/feedback-form.js
```diff
@@ -63,6 +65,13 @@ export default BottomBanner.extend(
 		displayQuestion: true,
 		displayInput: false,
 		displayThanks: false,
+		isiOS: Ember.computed(() => {
```
kamilkoterba: You don't need computed this should work as simple assignment. Plus consider moving to browser.js util

front/main/app/components/feedback-form.js
```diff
@@ -82,7 +91,19 @@ export default BottomBanner.extend(
 			}
 		},
 		willDestroyElement() {
-			Ember.$(window).off('scroll.feedbackForm');
+			Ember.$(window).off('scroll.feedbackForm').off('scroll.absoluteFeedbackForm');
```
kamilkoterba: I think we might have to cover touchmove event on document as well. Did you test that?

kamilkoterba: or not if it's iOS only solution...

kamilkoterba:
LGTM for an experiment purpose.

PR #2196, opened 2016-03-17
XW-1252 | Fix article related UA events click target context check

Warkot:
## Description

All contexts that were checked later than `.article-image` were treated as `.article-image` (because no `.length` check → and in JS `Boolean([]) === true`).

`.article-image` moved after `figcaption` as the `figcaption` is contained in the `figure` element which has the `.article-image` class. Without this change, all clicks in caption links would return `.article-image`.

## Reviewers

@Wikia/x-wing


vforge:
LGTM

bognix:
LGTM

PR #2199, opened 2016-03-17
Build Step | Quick way to restart the dev app

bkoval:
## Description

Do you need a quick way to restart Mercury when:
* The app crashes
* Your ssh session expires and you don't want to rebuild things, just run the app again
* You stopped a server yourself and you want to rerun it quickly

? :)

This task doesn't do any building. It just starts the app and runs all watchers.

## Reason

It takes my devbox around 65 - 75 seconds to run *npm run dev*, this task runs in 10-15 seconds. 

## RFC

Well, I couldn't come up with a more suitable task name, so feel free to suggest one :) Also, feel free to suggest any other quick restart solutions, whether they already exist in the repo (and I missed to spot them) or they're not there yet.

## Ping

@Wikia/x-wing @rafalkalinski @kvas-damian


rogatty:
Neat! Just add a description to README and will LGTM.

Also, this isn't 100% true:
>This task doesn't do any building. It just starts the app and runs all watchers.

`restart` runs `watch-main` which then runs `ember build --watch`. When creating new build pipeline I wasn't able to find any way to build Ember separately from watching it. The perfect solution would be to have something like:
```
ember build-without-watching
ember watch-without-building
```
which you would be able to reuse here, but it's currently not possible.

bognix:
Great! Thank you Bart!

bkoval:
@rogatty LGTY?

rogatty:
LGTM, thanks.
I suggest pinging `#mercury @channel` to let people know :)

PR #2200, opened 2016-03-17
fix abtest module import by lowercasing its path

Warkot:
## Description

JS console was throwing this error:
```
Uncaught Error: Could not find module `common/modules/AbTest` imported from `main/components/feedback-form`
```

All filenames, paths and the imports have to be lowercase. Linux machines, on which our app works, are case sensitive.

## Reviewers

@Wikia/x-wing @adamkarminski @lukaszkonieczny @kamilkoterba

### DIFF
--- a/front/main/app/components/feedback-form.js
+++ b/front/main/app/components/feedback-form.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import UserFeedbackStorageMixin from '../mixins/user-feedback-storage';
 import BottomBannerMixin from '../mixins/bottom-banner';
-import {getGroup} from 'common/modules/AbTest';
+import {getGroup} from 'common/modules/abtest';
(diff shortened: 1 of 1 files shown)


rogatty:
LGTM

kamilkoterba:
shouldn't we rename a file as well?

Warkot:
@kamilkoterba The file was already renamed to lower-dashed-case [here](https://github.com/Wikia/mercury/commit/8f28181def9171a47af0e60ef715932572523fb0#diff-1297dbce62d932408cafcd59bbf94e39). The import path was new, it was added in parallel to that case-changing PR. This divergence caused a difference between the import path and the path of the actual file caused the issue.

kamilkoterba:
Thanks, must have had outdated repo. That makes sense.

PR #2205, opened 2016-03-18
CE upvotes feature

kamilkoterba:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3295
* https://wikia-inc.atlassian.net/browse/CE-3265

## Description

This is addition to Recent Wiki Activity page - ability to upvote and downvote page edits.


front/main/app/styles/component/_article_diff.scss
```diff
@@ -107,10 +109,20 @@ $legend-line-height: $type-smallest;
...
 	position: fixed;
-	width: 100%;
+	width: 50%;
 	z-index: $z-site-head;
 }
 
+.diff-page__upvote,
+.diff-page__upvote--upvoted {
+	right: 0;
+	border-left: 1px solid $color-blue-gray;
```
bognix: minor: alphabetical order

front/main/app/components/recent-change.js
```diff
@@ -3,9 +3,17 @@ import Ember from 'ember';
 export default Ember.Component.extend({
 	classNames: ['recent-change'],
 	classNameBindings: ['active'],
+	currentUser: Ember.inject.service(),
 	active: Ember.computed('id', 'rc', function () {
 		return this.get('id') === this.get('rc');
 	}),
+	upvotesEnabled: Ember.get(Mercury, 'wiki.language.content') === 'en',
```
hakubo: if you have time I would advise to add to our toolset this:
https://github.com/kategengler/ember-feature-flags

front/main/app/templates/components/article-diff.hbs
```diff
@@ -26,6 +26,17 @@
 	<div class="{{diff.classes}}">{{diff.content}}</div>
 {{/each}}
 
-{{#if showUndoButton}}
-	<button {{action attrs.showConfirmation}} class="diff-page__undo">{{i18n 'main.undo-label' ns='recent-wiki-activity'}}</button>
+{{#if showButtons}}
+	<div class="diff-page__buttons">
+		<button {{action attrs.showConfirmation}} class="{{buttonUndoClass}}">
```
hakubo: more common way of changing class is to use inline if
{{if upvotesEnabled 'diff-page__undo' 'diff-page__undo--alone'}}

front/main/app/controllers/article-diff.js
```diff
@@ -1,101 +1,169 @@
...
+		 * Sends impression tracking for recent-wiki-activity category
+		 * @param {string} label
+		 * @returns {void}
+		 */
+		trackImpression(label) {
 			track({
-				action: trackActions.close,
-				category: 'recent-wiki-activity',
-				label: 'undo-confirmation-close'
+				action: trackActions.open,
```
rogatty: You call `this.trackImpression('upvote-success');` which then tracks action `open` :)
How about cleaning up the naming? We have `success` action so maybe `this.trackSuccession('upvote')`?

front/main/app/controllers/article-diff.js
```diff
@@ -1,101 +1,169 @@
...
+				label
 			});
 		},
+
+		actions: {
+			/**
+			 * Shows confirmation modal
+			 * @returns {void}
+			 */
+			showConfirmation() {
```
rogatty: Why do we do it in the controller? This is strictly UI thing so IMO it should live in the component. This whole file has a lot of logic which would fit better in a component.

Only the things that use:
- `this.get('application')`
- `this.transitionToRoute`
- `this.get('model')`

should be here. Any UI related logic (not the alert, though (we need to clean it up finally)) should be moved from them too. For example:

```
controller action:

removeUpvote(upvoteId) {
	return this.get('model').removeUpvote(upvoteId);
},


component method:

removeUpvote(upvoteId) {
	// This is the name passed in the template, e.g. removeUpvoteFromModel=(action 'removeUpvote')
	// This one sucks but it can't be removeUpvote as it would clash with the component method
	this.get('removeUpvoteFromModel')(upvoteId).then(
		this.handleRemoveUpvoteSuccess.bind(this),
		this.handleRemoveUpvoteError.bind(this)
	);
	this.trackClick(trackCategory, 'remove-upvote');
}

component action:

handleVote() {
	if (this.get('upvoted')) {
		this.removeUpvote(this.get('currentUserUpvoteId'));
	} else {
		this.upvote();
	}
}
```
In short - keep the controller as simple as possible.

front/main/app/models/article-diff.js
```diff
@@ -56,6 +59,83 @@ const ArticleDiffModel = Ember.Object.extend({
...
+	},
+
+	/**
+	 * Send request to server to remove previously added upvote for a revision
+	 * @param {int} upvoteId ID of upvote record to remove
+	 * @param {int} userId user ID who made an edit
+	 * @returns {Ember.RSVP.Promise}
+	 */
+	removeUpvote(upvoteId) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
```
rogatty: Both methods look almost identical. How about extracting common logic to
```
sendUpvoteRequest(data, successCallback)
```
?

front/main/app/components/recent-change.js
lukaszkonieczny: Looks nice :) We are going to remove this condition next week so this time we leave it but for sure we can try it in the future :)

front/main/app/components/article-diff.js
```diff
@@ -1,9 +1,152 @@
...
-});
+const trackCategory = 'recent-wiki-activity';
+
+export default Ember.Component.extend(
+	TrackClickMixin,
+	{
+		classNames: ['diff-page'],
+		currentUser: Ember.inject.service(),
+		revisionUpvotes: Ember.inject.service(),
+		currentUserUpvoteId: Ember.computed('revisionUpvotes.upvotes.@each.count', 'currentUser.userId', function () {
```
rogatty: `currentUser.userId` isn't used in this property so why do we observe it here?

front/main/app/components/article-diff.js
```diff
@@ -1,9 +1,152 @@
...
+			).then(
+				() => this.trackSuccess('upvote-success'),
+				() => this.handleError('main.error', 'upvote-error')
+			);
+			this.trackClick(trackCategory, 'upvote');
+		},
+
+		/**
+		 * Send request to server to remove previously added upvote for a revision
+		 * @param {int} upvoteId ID of upvote record to remove
```
rogatty: `@param {number}`

front/main/app/components/recent-change.js
```diff
@@ -1,11 +1,87 @@
...
+		classNames: ['recent-change'],
+		classNameBindings: ['active'],
+		currentUser: Ember.inject.service(),
+		revisionUpvotes: Ember.inject.service(),
+		active: Ember.computed('id', 'rc', function () {
+			return this.get('id') === this.get('rc');
+		}),
+		currentUpvotes: null,
+		upvotesCount: Ember.computed('revisionUpvotes.upvotes.@each.count', function () {
+			return this.get('currentUpvotes').count || 0;
```
rogatty: `this.get('currentUpvotes.count')` won't fail if `currentUpvotes` isn't an array.

front/main/app/components/recent-change.js
```diff
@@ -1,11 +1,87 @@
...
+		}),
+		upvotesEnabled: Ember.get(Mercury, 'wiki.language.content') === 'en',
+		currentUserUpvoteId: Ember.computed('upvotesCount', 'currentUser.userId', function () {
+			return this.get('currentUpvotes').userUpvoteId || 0;
```
rogatty: `this.get('currentUpvotes.userUpvoteId')` won't fail if `currentUpvotes` isn't an array.

adamkarminski:
LGTM

PR #2208, opened 2016-03-18
CE-3522: Share state between components

lukaszkonieczny:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3522

## Description

The blue dot didn't disappear from site navigation after going to RWA.
We added service to store the current blue dot state.

## Reviewers


rogatty:
>If we want to share state of mixin property between components, this property must be an array or an object.

This sounds like a hack :) And yes, I'm aware that we do this in other places too, but this this might be a good opportunity to introduce a better solution - service. Services are designed exactly for that - sharing state between multiple components.

bognix:
Exactly what Igor said.
Mixins are good for abstracting the common part between two components not for sharing state

lukaszkonieczny:
ready for CR @Wikia/x-wing

azarek:
Testing LGTM

front/main/app/services/new-features-badges.js
```diff
@@ -0,0 +1,50 @@
...
+		const badgesCookie = Ember.$.cookie('seenNewBadgeFor');
+
+		let features = [];
+
+		if (badgesCookie) {
+			features = JSON.parse(badgesCookie);
+		}
+		this.set('features', features);
+	},
+	/**
```
rogatty: How about using a single `\n` above JSDocs? IMHO it improves readibility and is used in the most of other files.

rogatty:
LGTM but what do you think about adding unit tests for the service methods?

front/main/app/services/new-features-badges.js
```diff
@@ -0,0 +1,50 @@
+import Ember from 'ember';
+import {getDomain} from '../utils/domain';
+
+export default Ember.Service.extend({
+	currentUser: Ember.inject.service(),
+	features: null,
+	init() {
+		const badgesCookie = Ember.$.cookie('seenNewBadgeFor');
```
kamilkoterba: The name of the cookie is same as it used to be, but you changed format from string to array, so if a user has it set to previous string 'recent-wiki-activity' you get error `Uncaught SyntaxError: Unexpected token r` trying `JSON.parse` on it, causing whole app stop further working.

rogatty: Perfect case to test against :)

front/main/tests/unit/services/new-features-badges-test.js
```diff
@@ -0,0 +1,117 @@
...
+
+moduleFor('service:new-features-badges', {
+	beforeEach: function () {
+		Ember.$.removeCookie('seenNewBadgeFor');
+	}
+});
+
+function setUpService(service, cookies) {
+	Ember.$.cookie('seenNewBadgeFor', JSON.stringify(cookies));
+	service.init();
```
rogatty: Isn't `init` run on `this.subject()`? If it is, we probably shouldn't run it twice.

front/main/tests/unit/services/new-features-badges-test.js
```diff
@@ -0,0 +1,117 @@
...
+	});
+
+	setUpService(service, ['feature-one', 'featureTwo']);
+
+	assert.equal(service.shouldDisplay('feature-one'), false);
+	assert.equal(service.shouldDisplay('featureTwo'), false);
+	assert.equal(service.shouldDisplay('feature3'), true);
+});
+
+test('should display badge only for logged in users and user is not', function (assert) {
```
rogatty: Instructions not clear ;)

front/main/tests/unit/services/new-features-badges-test.js
```diff
@@ -0,0 +1,117 @@
...
+	const service = this.subject();
+
+	service.set('currentUser', {
```
rogatty: You set a property and then run `init` in `setUpService`. Seems a little bit messy.

rogatty:
LGTM

front/main/app/services/new-features-badges.js
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		return !this.get('features').contains(featureName);
+	},
+
+	/**
+	 * Store information that user has seen information about new feature
+	 *
+	 * @param {string} featureName
+	 * @param {int} expires (default 10 years - 10 * 365)
```
vforge: `@param {number} [expires=10*365]`

front/main/app/services/new-features-badges.js
```diff
@@ -0,0 +1,61 @@
...
+	 * @param {string} path
```
vforge: `@param {string} [path='/']`

front/main/tests/unit/services/new-features-badges-test.js
```diff
@@ -0,0 +1,112 @@
...
+import {test, moduleFor} from 'ember-qunit';
+
+moduleFor('service:new-features-badges', {
+	beforeEach() {
+		Ember.$.removeCookie('seenNewBadgeFor');
+	}
+});
+
+/**
+ * Set cookies for the test
```
vforge: Please do not add not meaningful descriptions - here and in other places.

If function's name describes pretty clear what it does, we don't need description.

PR #2210, opened 2016-03-18
CE-3527 Reset iOS related properties also

adamkarminski:
## Links

* http://wikia-inc.atlassian.net/browse/CE-3527

## Description

The fix for banner not being reset on iOS when a user went to another route.

## Reviewers

@Wikia/community-engineering


lukaszkonieczny:
LGTM

PR #2213, opened 2016-03-21
CE-3492 Add ability to downvote

kamilkoterba:
## Links

* http://wikia-inc.atlassian.net/browse/JIRA-3492

## Description
Adding ability to downvote

## Reviewers

@Wikia/community-engineering, @Wikia/x-wing

### DIFF
--- a/front/main/app/components/article-diff.js
+++ b/front/main/app/components/article-diff.js
@@ -5,10 +5,10 @@ export default Ember.Component.extend({
 	currentUser: Ember.inject.service(),
 	currentUserUpvoteId: Ember.computed('model.upvotes', 'currentUser.userId', function () {
 		const upvotes = this.get('model.upvotes'),
-			currentUserUpvote = upvotes ? upvotes.filterBy(
+			currentUserUpvote = upvotes ? upvotes.findBy(
 				'from_user',
 				String(this.get('currentUser.userId'))
-			)[0] : null;
+			) : null;
 
 		return currentUserUpvote ? currentUserUpvote.id : null;
 	}),
--- a/front/main/app/components/article-diff.js
+++ b/front/main/app/components/article-diff.js
@@ -15,9 +15,5 @@ export default Ember.Component.extend({
 	userNotBlocked: Ember.computed.not('currentUser.isBlocked'),
 	showButtons: Ember.computed.and('currentUser.isAuthenticated', 'userNotBlocked'),
 	showDiffLink: false,
-	upvoted: Ember.computed.bool('currentUserUpvoteId'),
-	classUpvote: Ember.computed('upvoted', function () {
-		return this.get('upvoted') ? 'diff-page__upvote--upvoted' : 'diff-page__upvote';
-	})
-
+	upvoted: Ember.computed.bool('currentUserUpvoteId')
 });
--- a/front/main/app/components/article-diff.js
+++ b/front/main/app/components/article-diff.js
@@ -3,6 +3,16 @@ import Ember from 'ember';
 export default Ember.Component.extend({
 	classNames: ['diff-page'],
 	currentUser: Ember.inject.service(),
+	currentUserUpvoteId: Ember.computed(function () {
+		const upvotes = this.get('model.upvotes');
+
+		for (const upvote of upvotes) {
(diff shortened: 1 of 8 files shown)


front/main/app/components/article-diff.js
```diff
@@ -3,6 +3,16 @@ import Ember from 'ember';
 export default Ember.Component.extend({
 	classNames: ['diff-page'],
 	currentUser: Ember.inject.service(),
+	currentUserUpvoteId: Ember.computed(function () {
```
lukaszkonieczny: Actually this is what `upvoted` method is doing as well so you could use just one property

front/main/app/components/article-diff.js
```diff
@@ -3,6 +3,16 @@ import Ember from 'ember';
...
+		const upvotes = this.get('model.upvotes');
+
+		for (const upvote of upvotes) {
```
lukaszkonieczny: Could we use `filterBy` Ember method?

front/main/app/components/article-diff.js
```diff
@@ -16,4 +26,10 @@ export default Ember.Component.extend({
 	classUpvote: Ember.computed('upvoted', function () {
 		return this.get('upvoted') ? 'diff-page__upvote--upvoted' : 'diff-page__upvote';
 	}),
+
+	actions: {
```
lukaszkonieczny: Do we need to catch it here and resend to controller?

front/main/app/models/article-diff.js
```diff
@@ -87,6 +88,40 @@ const ArticleDiffModel = Ember.Object.extend({
...
+	 * @param {int} userId user ID who made an edit
+	 * @returns {Ember.RSVP.Promise}
+	 */
+	downvote(upvoteId, userId) {
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			getEditToken(this.title)
+				.then((token) => {
+					Ember.$.ajax({
+						url: M.buildUrl({
+							path: '/wikia.php?controller=RevisionUpvotesApiController&method=removeUpvote',
```
lukaszkonieczny: You should define `controller` and `method` inside `data object

kamilkoterba: nope, it won't work this way.
Wikia's MediaWiki is choosing WikiaApiController based on GET parameters and this is POST request.

front/main/app/components/article-diff.js
```diff
@@ -3,17 +3,17 @@ import Ember from 'ember';
 export default Ember.Component.extend({
 	classNames: ['diff-page'],
 	currentUser: Ember.inject.service(),
+	currentUserUpvoteId: Ember.computed('model.upvotes', 'currentUser.userId', function () {
+		const upvotes = this.get('model.upvotes'),
+			currentUserUpvote = upvotes ? upvotes.filterBy(
```
Grunny: `.findBy` will do the same as `.filterBy` but only return the first match, which is what you're doing here.

front/main/app/controllers/article-diff.js
```diff
@@ -125,6 +132,35 @@ export default Ember.Controller.extend(
 			},
 
 			/**
+			 * Send info to server that user downvoted a revision
+			 * @param {int} upvoteId ID of upvote record to remove
+			 * @param {int} userId user ID who made an edit
+			 * @returns {void}
+			 */
+			downvote(upvoteId, userId) {
```
Grunny: As a general comment, I think we should rename the `downvote` methods to `removeUpvote` or something similar so the code is clear about what that methods actually do. "downvote" suggests a separate action like "dislike" rather than simply undoing a previous upvote.

front/main/app/models/article-diff.js
lukaszkonieczny: I cannot say i'm not surprised ;p Thanks Kamil!
In this case I think you could use `query` property of `M.buildUrl`

kamilkoterba: Thanks. Addressed here b2580916a2b81f13a5322e2963302ed5e70e9bd5

front/main/app/controllers/article-diff.js
kamilkoterba: Thanks. Addressed here d6bdd28907d58b01d4e095ecd858bbf99a436ab3

PR #2215, opened 2016-03-21
PLATFORM-1860 Context for image review rework

RafalWilinski:
## Links

* http://wikia-inc.atlassian.net/browse/PLATFORM-1860
* https://github.com/Wikia/pandora/pull/863

## Description

Context is now fetched async from `static-asset` service on-demand when user clicks thumbnail to get details about picture. 

## Reviewers

Ping @hakubo @rogatty


front/main/app/components/image-review.js
```diff
@@ -10,8 +11,17 @@ export default Ember.Component.extend({
...
 	actions: {
 		showModal(popupModel) {
-			this.set('thumbnailModel', popupModel);
-			this.set('isModalVisible', true);
+			imageReviewModel.getImageContext(popupModel.imageId).then((data) => {
+				popupModel.context = data.context || '#';
+				popupModel.isLink = new RegExp('(http|https)?:\/\/[^\s]+').test(data.context);
+				this.set('thumbnailModel', popupModel);
+				this.set('isModalVisible', true);
+			}).catch(() => {
```
rogatty: How about keeping it DRY?
```
showModal(popupModel) {
	imageReviewModel.getImageContext(popupModel.imageId).then((data) => {
		popupModel.context = data.context;
	}).finally(() => {
		if (popupModel.context) {
			popupModel.isLink = new RegExp('(http|https)?:\/\/[^\s]+').test(data.context);
		} else {
			popupModel.context = '#';
			popupModel.isLink = false;
		}
		
		this.setProperties({
			thumbnailModel: popupModel,
			isModalVisible: true
		});
	});
}
```

front/main/app/models/image-review.js
```diff
@@ -160,6 +160,21 @@ ImageReviewModel.reopenClass({
...
+		return new Ember.RSVP.Promise((resolve, reject) => {
+			Ember.$.ajax({
+				url: M.getStaticAssetsServiceUrl(`/image/context/${imageId}`),
+				xhrFields: {
+					withCredentials: true
+				},
+				dataType: 'json',
+				method: 'GET',
+				success: (data) => resolve(data),
+				error: () => reject(i18n.t('app.image-review-error-invalid-data'))
```
rogatty: Are you sure this is the reject reason you want to use? How about moving some logic here from the component? You can set `context` and `isLink` here and the component would just consume it.

E.g.
```
reject({
    context: '#',
    isLink: false
})
```

front/main/app/components/image-review.js
rogatty: Or even better - do it in the model instead.

RafalWilinski: Clever!

RafalWilinski:
@rogatty A bit simplified, what do you think?

front/main/app/components/image-review.js
```diff
@@ -1,17 +1,25 @@
 import Ember from 'ember';
+import imageReviewItemModel from '../models/image-review-item';
 
 export default Ember.Component.extend({
 	classNames: ['image-review'],
 	isModalVisible: false,
 
 	isContextProvided: Ember.computed('thumbnailModel.context', function () {
-		return this.get('thumbnailModel.context') !== '#';
+		return this.get('thumbnailModel.context') || false;
```
rogatty: IMHO `Boolean(this.get('thumbnailModel.context'))` would be simpler and more consistent, function would always return the same type.

rogatty:
One minor thing and it'll LGTM

rogatty:
LGTM

PR #2219, opened 2016-03-22
DAT-3921 add focusing on inputs in Infobox Builder

dianafa:
## Links

https://wikia-inc.atlassian.net/browse/DAT-3921

## Description
Add focusing on input fields when displaying sidebar.

## Reviewers

@RafLeszczynski

### DIFF
--- a/front/main/app/mixins/infobox-builder-edit-item.js
+++ b/front/main/app/mixins/infobox-builder-edit-item.js
@@ -20,14 +20,16 @@ export default Ember.Mixin.create(
 		 * @returns {void}
 		 */
 		focusFirstInput() {
-			const input = this.$('.text-field-input');
-			let firstInput;
+			const inputs = this.$('.text-field-input');
+			let firstInput,
+				inputProperties;
 
-			if (input) {
-				firstInput = input.first().get(0);
-				input.first().focus();
+			if (inputs) {
+				firstInput = inputs.first();
+				inputProperties = firstInput.get(0);
+				firstInput.focus();
 				// required for moving cursor to the end of input on FF
-				firstInput.selectionStart = firstInput.selectionEnd = firstInput.value.length;
+				inputProperties.selectionStart = inputProperties.selectionEnd = inputProperties.value.length;
 			}
 		},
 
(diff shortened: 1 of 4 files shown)


front/main/app/mixins/infobox-builder-edit-item.js
```diff
@@ -6,7 +6,33 @@ export default Ember.Mixin.create(
...
+		 * Focuses on the end of the first text input element of this component
+		 * @returns {void}
+		 */
+		focusFirstInput() {
+			const input = this.$('.text-field-input');
+			let firstInput;
+
+			if (input) {
+				firstInput = input.first().get(0);
+				input.first().focus();
```
RafLeszczynski: Please consider avoiding duplication DOM traversal. Consider caching `input.first()` in local variable.

RafLeszczynski:
One small optimisation proposal. Except that LGTM

front/main/app/mixins/infobox-builder-edit-item.js
```diff
@@ -6,7 +6,35 @@ export default Ember.Mixin.create(
...
+		/**
+		 * Focuses on the end of the first text input element of this component
+		 * @returns {void}
+		 */
+		focusFirstInput() {
+			const inputs = this.$('.text-field-input');
+			let firstInput,
+				inputProperties;
+
+			if (inputs) {
```
bognix: It will always be truthy - check in console: `!!$('not-existing-tag')`

bognix:
two minor things - when fixed - LGTM

front/main/app/mixins/infobox-builder-edit-item.js
RafLeszczynski: that's true cause you will always receive jQuery collection object. If there won't be any elemenets with class name `.text-field-input` this collection will have `length = 0`

RafLeszczynski: we did have an issue with ember liquid fire (used here before, now removed) that `didRender` hook was fired twice and in second call `this.$('.text-field-input');` returns `undefined`

RafLeszczynski:
LGTM if build passes

PR #2224, opened 2016-03-22
CE-3334 Display a byte change of a given change

adamkarminski:
## Links
* http://wikia-inc.atlassian.net/browse/CE-3334
* https://github.com/Wikia/app/pull/10044

## Description
Display byte change for a given diff.

## Reviewers
@Wikia/community-engineering


Grunny:
Code LGTM, but it would be missing the acceptance criteria "Please use proper thousands separators (123 | 1,234 | 1,234,567)", right? So, we may want to consider how best to ensure the number is formatted correctly for the user's language.

front/main/app/services/numeral-locale.js
```diff
@@ -0,0 +1,69 @@
...
+		it: '/front/vendor/assets/numeral/it.js',
+		ja: '/front/vendor/assets/numeral/ja.js',
+		pl: '/front/vendor/assets/numeral/pl.js',
+		'pt-br': '/front/vendor/assets/numeral/pt-br.js',
+		ru: '/front/vendor/assets/numeral/ru.js',
+		'zh-cn': '/front/vendor/assets/numeral/zh-cn.js',
+		'zh-tw': '/front/vendor/assets/numeral/zh-tw.js'
+	},
+	/**
+	 * Changes status of downloading moment's locale to trigger helper's observers
```
Grunny: Copy and paste from moment-locale? ;)

adamkarminski: Noooo dad... <not looking you in the eye...>

adamkarminski:
@Grunny - Updated :)

front/main/app/helpers/localize-number.js
```diff
@@ -0,0 +1,34 @@
+import Ember from 'ember';
+
+/**
+ * A helper allowing you to localize numbers in your templates.
+ * Usage: {{number (true)}}. If you pass true as the second parameter, the number will be signed
```
rogatty: Something is wrong with this example. I see that you use it in form
```
{{localize-number model.lengthChange true}}
```
so let's put something similar here.

front/main/app/services/numeral-locale.js
```diff
@@ -0,0 +1,69 @@
...
+	},
+	/**
+	 * Changes numeral locale to en. It's loaded by default, so we don't need to download it
+	 *
+	 * @return {void}
+	 */
+	setEnLocale() {
+		numeral.language('en');
+		// Change status when moment finished changing locale
+		Ember.run.next(() => {
```
rogatty: Why do we use `Ember.run.next` here while in `loadLocale()` we just call `this.changeLoadingStatus()` immediately after setting the language? What's the difference?

front/main/app/helpers/localize-number.js
adamkarminski: True, sorry.

rogatty:
LGTM

azarek:
LGTM

PR #2227, opened 2016-03-23
Fix PhantomJS issues

rogatty:
## Links

* https://github.com/Medium/phantomjs/issues/501
* https://github.com/Medium/phantomjs#deciding-where-to-get-phantomjs
* https://github.com/karma-runner/karma-phantomjs-launcher/blob/master/CHANGELOG.md#100-2016-01-28

## Description

I've bumped into an error preventing me from installing node dependencies. It's described in the linked PhantomJS issue and the suggested fix is to use a mirror instead of BitBucket to download the binary.

I've also upgraded `karma-phantomjs-launcher` which has some bug fixes including:
>use phantomjs-prebuilt instead of deprecated phantomjs

## Reviewers

@Wikia/x-wing

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 save-exact=true
+phantomjs_cdnurl=http://cnpmjs.org/downloads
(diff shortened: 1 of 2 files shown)


hakubo:
LGTM

PR #2231, opened 2016-03-23
XW-1283 | Enable links with : (colon)

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1283

## Description

Some initial work. Has to be cleaned up.
Clicking on a link to any page in supported namespace transitions to it—even ones with colon.

## Reviewers

@rogatty


front/main/app/utils/mediawiki-namespace.js
```diff
@@ -1,12 +1,8 @@
...
+ * This list is taken from MediaWiki:app/includes/Defines.php
+ * @type {{name: number}}
  */
-const supportedNamespaces = [0, 14],
-	/**
-	 * This list is taken from MediaWiki:app/includes/Defines.php
-	 * @type {{name: number}}
-	 */
-	namespace = {
+const namespace = {
```
hakubo: do we still need that list?
can we reuse whatever comes from wikivariables?

server/app/views/_partials/server-data.hbs
```diff
@@ -29,7 +29,8 @@
 	{{#if asyncArticle}}
 		M.prop('asyncArticle', {{asyncArticle}});
 	{{/if}}
-	M.prop('mediaWikiNamespace', '{{{mediaWikiNamespace}}}', true);
+	M.prop('mediaWikiNamespace', {{{mediaWikiNamespace}}}, true);
```
hakubo: can we get this data from article model?

using more globals will make it harder for us to use fastboot

front/main/app/utils/article-link.js
```diff
@@ -72,29 +58,6 @@ export default function getLinkInfo(basePath, title, hash, uri) {
 			};
 		}
 
-		if (Ember.isArray(article) && article[3] && !isMercuryNamespaceHandlingOverridden(article[3])) {
```
bognix: sooo goood you removed it!

front/main/app/utils/mediawiki-namespace.js
rogatty: This list is here only for better readibility. We could use numbers when checking the namespace but it's easier to read
```
currentNamespace === MediawikiNamespace.CATEGORY
```
than
```
currentNamespace === 14
```

This is a config hardcoded in MediaWiki, not a variable per wiki.

front/main/app/utils/mediawiki-namespace.js
```diff
@@ -32,27 +28,21 @@ const supportedNamespaces = [0, 14],
...
+ * @param {number} ns
+ *
  * @returns {boolean}
  */
-function isContentNamespace() {
-	// In the future this method would be more sophisticated
-	// because each wiki can have its own content namespaces defined
-	return parseInt(M.prop('mediaWikiNamespace'), 10) === 0;
+function isContentNamespace(ns) {
+	return ns === namespace.MAIN || Mercury.wiki.contentNamespaces.indexOf(ns) >= 0;
```
rogatty: Let's remember to release Mercury after App or this will break.

front/main/app/utils/wiki-handlers/wiki-page.js
```diff
@@ -25,22 +25,21 @@ function getURL(params) {
  * @returns {Object}
  */
 function getModelForNamespace(data, params) {
+	const currentNamespace = getCurrentNamespace();
```
rogatty: It's already available in `data.data.ns`

front/main/app/utils/wiki-handlers/wiki-page.js
```diff
@@ -53,17 +52,16 @@ export default function getPageModel(params) {
 	let model;
 
 	return new Ember.RSVP.Promise((resolve, reject) => {
-		if (isContentNamespace()) {
-			if (M.prop('articleContentPreloadedInDOM') && !M.prop('asyncArticle')) {
-				model = ArticleModel.create(params);
+		if (M.prop('isContentNamespace') && M.prop('articleContentPreloadedInDOM') && !M.prop('asyncArticle')) {
```
rogatty: Do we still need `M.prop('isContentNamespace')`? Can we just use `isContentNamespace()`?

server/app/views/_partials/server-data.hbs
```diff
@@ -29,7 +29,8 @@
...
+	M.prop('isContentNamespace', {{{isContentNamespace}}});
```
bognix: don't do that please - you can set this value on model

front/main/app/utils/mediawiki-namespace.js
Warkot: The list that comes from `WikiVariables` is different and it looks like this:
```
namespaces: {
0: "",
1: "Talk",
2: "User",
3: "User_talk",
4: "MediaWiki",
5: "MediaWiki_talk",
…
2002: "Topic",
-2: "Media",
-1: "Special"
},
```

server/app/views/_partials/server-data.hbs
bognix: sometimes yes and sometimes no.
This data is needed in case of categories without articles - there is no article model.
I tried to removed this M.prop but it's used in `tracking-internal` partial which is included inside `index.html`.

front/main/app/routes/wiki-page.js
```diff
@@ -87,6 +88,14 @@ export default Ember.Route.extend({
 				this.set('wikiHandler', handler);
 
 				handler.afterModel(this, model);
+			} else {
+				transition.abort();
+				window.location.assign(M.buildUrl({
+					wikiPage: Ember.get(transition, 'params.wiki-page.title'),
+					query: {
+						useskin: 'oasis'
```
rogatty: Let's extend the original query with this param.

front/main/tests/unit/utils/wiki-handlers/wiki-page-test.js
```diff
@@ -0,0 +1,120 @@
...
+		articleSetArticleStub = sinon.stub(),
+		categoryCreateStub = sinon.stub(),
+		categorySetCategoryStub = sinon.stub(),
+		isContentNamespaceStub = sinon.stub();
+
+	hooks.beforeEach(() => {
+		require('main/models/wiki/article').default.create = articleCreateStub;
+		require('main/models/wiki/article').default.setArticle = articleSetArticleStub;
+		require('main/models/wiki/category').default.create = categoryCreateStub;
+		require('main/models/wiki/category').default.setCategory = categorySetCategoryStub;
```
rogatty: Can we keep it DRYer?

server/tests/unit/lib/mediawiki-namespace.js
```diff
@@ -0,0 +1,27 @@
...
+			expected: false
+		}, {
+			contentNamespaces: [0, '112'],
+			namespace: 112,
+			expected: true
+		}, {
+			contentNamespaces: [0, 112],
+			namespace: 0,
+			expected: true
+		}
```
rogatty: Let's check what happens when `undefined` is passed as param(s).

server/app/facets/mediawiki-page.js
```diff
@@ -88,43 +91,45 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
...
-			result.subtitle = i18n.t('app.category-page-subtitle');
-			break;
-
-		default:
-			Logger.warn(`Unsupported namespace: ${ns}`);
-			result = prepareMediaWikiData(request, data);
+		result = deepExtend(result, prepareCategoryData(request, data));
+		// Hide TOC on category pages
+		result.hasToC = false;
+		result.subtitle = i18n.t('app.category-page-subtitle');
```
rogatty: Can we move these two lines to `prepareCategoryData()`?

server/app/facets/mediawiki-page.js
```diff
@@ -88,43 +91,45 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
...
+		Logger.warn(`Unsupported namespace: ${ns}`);
+		result = prepareMediaWikiData(request, data);
 	}
 
 	// mainPageData is set only on curated main pages - only then we should do some special preparation for data
-	if (pageData.isMainPage && pageData.mainPageData) {
+	if (isMainPage && pageData.mainPageData) {
 		result = deepExtend(result, prepareMainPageData(data));
 		result.hasToC = false;
 		delete result.adsContext;
```
rogatty: Can we move these two lines to `prepareMainPageData()`?

server/app/lib/mediawiki-namespace.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * This list is taken from MediaWiki:app/includes/Defines.php
+ * @type {{name: number}}
+ */
```
rogatty: Let's add a comment to keep it in sync with `front/main/app/utils/mediawiki-namespace.js` and vice versa.

server/app/facets/mediawiki-page.js
```diff
@@ -88,43 +91,45 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
...
+		response;
 
 	if (data.page && data.page.data) {
 		pageData = data.page.data;
 		ns = pageData.ns;
-		result.mediaWikiNamespace = ns;
+		isMainPage = pageData.isMainPage;
 	}
+
+	result.mediaWikiNamespace = ns;
```
tomasznapieralski: if we always set namespace to result, maybe we should set default value? what do you think?

server/app/facets/mediawiki-page.js
```diff
@@ -88,43 +91,45 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
...
+
+	isContentNamespace = MediaWikiIsContentNamespace(ns, data.wikiVariables.contentNamespaces);
```
tomasznapieralski: what if `data` doesn't have `wikiVariables`?

front/main/app/routes/wiki-page.js
```diff
@@ -17,18 +17,19 @@ export default Ember.Route.extend({
...
 
-		switch (getCurrentNamespace()) {
-			case MediawikiNamespace.MAIN:
-				return ArticleHandler;
-			case MediawikiNamespace.CATEGORY:
-				return CategoryHandler;
-			default:
-				Ember.Logger.debug(`Unsupported NS passed to getHandler - ${getCurrentNamespace()}`);
-				return null;
+		if (isContentNamespace(currentNamespace)) {
```
Warkot: I wondering if would it be nicer and more cohesive if we would use `else if` here while connecting it with the `if` block above?

server/app/facets/mediawiki-page.js
bognix: yes

server/app/facets/mediawiki-page.js
Warkot: If yes then which one? `0`? Then we might get caught trying to display a non-content namespace as a such. If we have no data, we can throw an error on line 126.

bognix: it's hard to say what should be the default? an article? content? main page? some custom one?
It's better to set it as it comes from API and have proper error handling to gracefully handle situation when ns is empty because of what API return

server/app/facets/mediawiki-page.js
bognix: we set it explicitly in lib/mediawiki-page.js inside getFull method.
If request for Wiki Variables fails then an exception is thrown and this code is not exectued

server/tests/acceptance/wiki-page.js
```diff
@@ -197,4 +195,47 @@ describe('wiki-page', function () {
 			done();
 		});
 	});
+
+	it('renders article with custom namespace', function (done) {
```
Warkot: Other case I can think about is something like "redirects if content namespace is not supported" — for a custom and for a non-custom namespaces which are not the content namespaces.

rogatty: This will be handled in XW-1284.

rogatty:
LGTM

server/tests/acceptance/wiki-page.js
Warkot: OK, let's do it when it needed/possible. :wink:

Warkot:
LGTM :shipit:

server/app/facets/mediawiki-page.js
tomasznapieralski: :+1:

server/app/facets/mediawiki-page.js
tomasznapieralski: :+1:

tomasznapieralski:
LGTM

hakubo:
LGTM :thumbsup:

PR #2232, opened 2016-03-24
move wikia-ui to in-repo-addon

hakubo:
## Description

I moved wikia-ui to in-repo-addon. This is a proposal to enforce better code organisation in our repository.


front/main/app/styles/component/_infobox-builder.scss
```diff
@@ -53,11 +53,11 @@ $sidebar-z-index: 1;
 	box-sizing: border-box;
 	display: flex;
 	height: 100%;
-	padding-top: $wikia-ui-components-sub-head-height;
+	padding-top: 48px;
```
rogatty: So here's the downside of this solution as we have to keep these values in sync. Not sure what would be the best solution here.
/cc @RafLeszczynski

hakubo: so here is the upside - this should teach us not to reuse some 'internal' vars to components :)

for now I did hardcoded it - but we can create a 'global' settings to share those values

front/main/app/styles/app.scss
```diff
@@ -15,6 +15,9 @@
 @import 'component/index';
 @import 'module/index';
 
+// lib
+@import 'wikia-ui';
```
hakubo: I am sure this can be automated - but for now I'd say this is required if we want to have one css file at the end of build

PR #2237, opened 2016-03-24
CE-3556: Move most logic from controller to component

lukaszkonieczny:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3556

## Description

Move most logic from controller to component

## Reviewers
@Wikia/community-engineering @Wikia/x-wing @rogatty ;)

### DIFF
--- a/front/main/app/components/article-diff.js
+++ b/front/main/app/components/article-diff.js
@@ -34,7 +34,7 @@ export default Ember.Component.extend(
 		 */
 		handleError(messageKey, label) {
 			this.trackError(label);
-			this.get('error')(messageKey);
+			this.get('showError')(messageKey);
 		},
 
 		/**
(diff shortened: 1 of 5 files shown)


front/main/app/templates/components/article-diff.hbs
```diff
@@ -28,15 +28,26 @@
...
 			{{svg 'kudos' viewBox='0 4 18 18' class='diff-page__button-icon'}}
 			<span class="diff-page__button-text">{{i18n 'main.nice-work-button-label' ns='recent-wiki-activity'}}</span>
 		</button>
 	{{/if}}
 	</div>
 {{/if}}
+
+{{#if shouldShowUndoConfirmation}}
+	{{confirmation-modal
+	close=(action 'closeConfirmation')
```
rogatty: This could use some `\t`

front/main/app/components/article-diff.js
```diff
@@ -1,36 +1,153 @@
...
+			this.get('error')(messageKey);
+		},
+
+		/**
+		 * Send request to server to remove previously added upvote for a revision
+		 * @param {int} upvoteId ID of upvote record to remove
+		 * @returns {void}
+		 */
+		removeUpvote(upvoteId) {
+			this.get('model').removeUpvote(upvoteId).then(
```
rogatty: This breaks the DDAU rule. You pass `model` from the controller and then modify it here. It works but is harder to reason about. It's your decision what to do you with it but I would personally move `this.get('model').removeUpvote(upvoteId)` part to the controller action and leave the rest here :)

rogatty:
LGTM

PR #2240, opened 2016-03-25
Dat 4009

RafLeszczynski:
DO NOT merge without APP part!!! https://github.com/Wikia/app/pull/10070

## Links

* https://wikia-inc.atlassian.net/browse/DAT-4009

## Description

fix issue with go to source page redirects

## Reviewers

@SebastianMarzjan @rogatty

### DIFF
--- a/front/main/app/models/infobox-builder.js
+++ b/front/main/app/models/infobox-builder.js
@@ -322,7 +322,7 @@ const InfoboxBuilderModel = Ember.Object.extend({
 						method: 'POST',
 						success: (data) => {
 							if (data && data.success) {
-								resolve(this.get('title'));
+								resolve(data.urls);
 							} else {
 								reject(data.errors);
 							}
--- a/front/main/app/routes/infobox-builder.js
+++ b/front/main/app/routes/infobox-builder.js
@@ -128,10 +128,10 @@ export default Ember.Route.extend(ConfirmationMixin, {
 		goToSourceEditor(title) {
 			this.getRedirectUrls(title)
 				.then((urls) => {
-					this.redirectToPage(urls.sourceEditorUrl);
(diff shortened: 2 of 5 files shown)


front/main/app/models/infobox-builder.js
```diff
@@ -322,7 +322,7 @@ const InfoboxBuilderModel = Ember.Object.extend({
 						method: 'POST',
 						success: (data) => {
 							if (data && data.success) {
```
SebastianMarzjan: "This shouldn't happen" (TM)
But what if urls are not in data?

dianafa: nothing happens:

`			return model.saveStateToTemplate().then((urls = {}) => {
				if (shouldRedirectToPage) {
					this.get('target').send('redirectToPage', urls.templatePageUrl);
				}
			});`

that means we get redirected to the previous page:

`redirectToPage(url) {
			const action = url ?
				'redirectToPage' :
				'redirectToPreviousPage';`

front/main/app/routes/infobox-builder.js
```diff
@@ -282,5 +279,35 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+				url: M.buildUrl({
+					path: '/wikia.php'
+				}),
+				data: {
+					controller: 'PortableInfoboxBuilderController',
+					method: 'getRedirectUrls',
+					title
+				},
+				dataType: 'json',
+				method: 'POST',
```
SebastianMarzjan: POSTing to a get.* method?

front/main/app/models/infobox-builder.js
SebastianMarzjan: Cool, didn't notice the default value. Thanks!

SebastianMarzjan:
LG(TM)

PR #2244, opened 2016-03-28
RECIRC-86: Add utm tracking code to Fandom links in footer

pauloslund:
## Links
https://wikia-inc.atlassian.net/browse/RECIRC-86

## Description
We want to add UTM tracking information to Fandom links so we can better track from the Fandom side which traffic is coming from our Recirculation efforts


gabrys:
LGTM

PR #2254, opened 2016-03-29
XW-1284 | Redirect to Oasis on unsupported namespace

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1284

## Reviewers

@Wikia/x-wing


server/app/facets/mediawiki-page.js
```diff
@@ -118,9 +123,14 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
...
 		result = deepExtend(result, prepareCategoryData(request, data));
-	} else {
-		Logger.warn(`Unsupported namespace: ${ns}`);
+	} else if (code !== 200) {
+		// In case of status code different than 200 we want Ember to display an error page
+		// This method sets all the data required to start the app
 		result = prepareMediaWikiData(request, data);
+	} else {
+		Logger.info(`Unsupported namespace: ${ns}`);
+		redirectToSkin(request, reply, 'oasis');
```
bognix: I would call it redirect to desktop and avoid the unnecessary param.
Making the method "parametrized" is better in case of reuse in the future but right now we are not reusing it anywhere and passing more params decreases the readability because my mind has more variables to follow

bognix:
minor note about the naming - LGTM

server/app/lib/utils.js
```diff
@@ -340,3 +340,18 @@ export function getHtmlTitle(wikiVariables, displayTitle = '') {
...
 }
+
+/**
+ * @param {Hapi.Request} request
+ * @param {Hapi.Response} reply
+ * @returns {void}
+ */
+export function redirectToOasis(request, reply) {
+	const queryParams = QueryString.stringify(
+		Hoek.applyToDefaults(request.query, {
```
tomasznapieralski: could you destruct `QueryString` and `Hoek` and use their methods directly?

tomasznapieralski:
1 pierdola, LGTM

PR #2256, opened 2016-03-30
ADEN-3122 Fix GA ad/timing events

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3122

## Description

Fix GA `ad/timing` events by reseting `window.wgNow` on all pages where we use AdEngine. Previously this value was set on first pageview only:
https://github.com/Wikia/mercury/blob/dev/server/app/views/_partials/nirvana.hbs#L6


front/main/app/routes/main-page-category.js
```diff
@@ -2,8 +2,9 @@ import Ember from 'ember';
 import MetaTagsMixin from '../mixins/main-page-route';
 import MainPageRouteMixin from '../mixins/meta-tags';
 import CuratedContentModel from '../models/curated-content';
+import RouteWithAds from '../mixins/route-with-ads';
```
rogatty: The convention is to use `Mixin` suffix, just look above.

front/main/tests/unit/routes/main-page-category-test.js
```diff
@@ -72,3 +72,12 @@ test('sets controller properties', function (assert) {
...
 	routeMock.afterModel(modelWithDoubleEncodedTitle);
 });
+
+test('reset ads variables on before model', function (assert) {
+	const mock = this.subject();
+
+	window.wgNow = null;
+	mock.beforeModel();
+
+	assert.notEqual(window.wgNow, null);
```
rogatty: By doing that you rely on `M.prop('initialPageView')` being set to false by another test:

```
if (!M.prop('initialPageView')) {
	window.wgNow = new Date();
}
```
It would probably fail if you ran only this test.

How about just checking the fact that `beforeModel` from the mixin was called? It would be a proper unit test, then.

front/main/app/mixins/route-with-ads.js
```diff
@@ -0,0 +1,17 @@
...
+	 *
+	 * @returns {void}
+	 */
+	beforeModel() {
+		this._super();
+		if (!M.prop('initialPageView')) {
+			window.wgNow = new Date();
+		}
+	}
+
```
tomasznapieralski: remove this extra line, please

front/main/tests/unit/routes/main-page-section-test.js
```diff
@@ -71,3 +71,13 @@ test('sets controller properties', function (assert) {
 	};
 	routeMock.afterModel(modelWithDoubleEncodedTitle);
 });
+
+test('reset ads variables on before model', function (assert) {
+	const mock = this.subject();
+
+	M.prop('initialPageView', false);
+	window.wgNow = null;
```
tomasznapieralski: if you set sth on globals, it will be nice to store it's previous value in `beforeEach()` hook and restore it in `afterEach()` hook

front/main/tests/unit/routes/wiki-page-test.js
```diff
@@ -163,3 +163,25 @@ test('get correct handler based on model isMainPage flag and exception', functio
...
+test('reset ads variables on before model', function (assert) {
+	const mock = this.subject();
+
+	M.prop('initialPageView', false);
+	mock.controllerFor = () => {
+		return {
+			send: () => {}
+		};
+	};
+	window.wgNow = null;
```
tomasznapieralski: if you set sth on globals, it will be nice to store it's previous value in `beforeEach()` hook and restore it in `afterEach()` hook

tomasznapieralski:
3 minors, LGTM

nandy-andy:
LGTM :+1:

PR #2257, opened 2016-03-30
XW-1278 | send flag if user is logged in to cache responses for anon users

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1278
* https://github.com/Wikia/app/pull/10106

## Description

Base on the isanon flag we either cache responses from server or not

## Reviewers
@Wikia/x-wing


Warkot:
LGTM

rogatty:
LGTM

PR #2264, opened 2016-03-31
DAT-4016 add basic keyboard actions support to Infobox Builder

RafLeszczynski:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-4016

## Description

add basic keyboard actions support to Infobox Builder

## Reviewers

@SebastianMarzjan @idradm @dianafa


front/main/app/components/wikia-ui-components/text-field.js
```diff
@@ -71,6 +73,19 @@ export default Ember.Component.extend({
...
+
+		/**
+		 * @param {String} value
+		 * @param {Event} event
+		 * @returns {void}
+		 */
+		onKeyUp(value, event) {
+			const onKeyUpHandler = this.get('onKeyUpHandler');
+
+			if (typeof onKeyUpHandler === 'function') {
```
idradm: Maybe we could change this logic to only check if handler was defined, in other case let it fail. That way silent failure, when wrong handler type provided, won't happen.

idradm:
LGTM

dianafa:
LGTM

PR #2268, opened 2016-04-01
DAT-3991 Animation of sidebar in Infobox builder

dianafa:
# Links

* https://wikia-inc.atlassian.net/browse/DAT-3991

## Reviewers

@RafLeszczynski @bogna @vforge


front/main/app/components/infobox-builder-edit-item-section-header.js
```diff
@@ -26,6 +26,14 @@ export default Ember.Component.extend(
 			}
 		}),
 
+		textInputId: Ember.computed(function () {
```
RafLeszczynski: please add comment why do we need this unique ids for each input element

front/main/app/components/infobox-builder-sidebar-header.js
```diff
@@ -5,23 +5,17 @@ export default Ember.Component.extend(
...
 
 				return type ?
 					// possible message keys: add-title, add-image, add-row, add-section-header
-					i18n.t(`main.add-${this.get('item.type')}`, {
-						ns: 'infobox-builder'
-					}) :
-					i18n.t('main.sidebar-header', {
-						ns: 'infobox-builder'
-					});
+					i18n.t(`main.add-${type}`, { ns: 'infobox-builder' }) :
```
RafLeszczynski: is a valid single line object declaration when it comes to our ESLint config?

front/main/app/components/infobox-builder-sidebar-header.js
```diff
@@ -5,23 +5,17 @@ export default Ember.Component.extend(
...
+					i18n.t('main.sidebar-header', { ns: 'infobox-builder' });
 			}
 		),
 
-		showActionButtons: Ember.computed.bool('item'),
+		showActionButtons: Ember.computed.bool('type'),
```
RafLeszczynski: why do we change `item` to `type`?

front/main/app/components/infobox-builder.js
```diff
@@ -58,9 +58,18 @@ export default Ember.Component.extend(
...
+				id = 0,
+				type = null;
+
+			if (activeItem) {
+				name = `infobox-builder-edit-item-${activeItem.type}`;
+				id = activeItem.type + activeItem.infoboxBuilderData.id;
+				type = activeItem.type;
+			}
+
+			return {name, id, type};
```
RafLeszczynski: please consider changing this computed property name so it corresponds with return value + add some context on why we are returning this data?

front/main/app/transitions.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * File required for liquid-fire addon to define the animations in.
+ * @returns {void}
+ */
+export default function () {
+	const liquidFireSlideToLeftContainer = '.liquid-fire-slide-to-left-container',
+		liquidFireSlideToLeftItem = '.liquid-fire-slide-to-left-item',
+		durationOut = 1250,
```
RafLeszczynski: could we cache this magic number in single const and reuse it in both places?

front/main/app/transitions.js
```diff
@@ -0,0 +1,21 @@
...
+		durationIn = 1250;
+
+	this.transition(
+		this.childOf(liquidFireSlideToLeftContainer),
+		this.use('explode', {
+			pickOld: liquidFireSlideToLeftItem,
+			use: ['wait', durationOut]
```
RafLeszczynski: does this syntax (single line array) match our ESLint guidelines?

front/main/app/transitions.js
```diff
@@ -0,0 +1,21 @@
...
+		}, {
+			pickNew: liquidFireSlideToLeftItem,
+			use: ['toLeft', {duration: durationIn, easing: 'easeInSine'}]
```
RafLeszczynski: does this syntax (single line array and object) match our ESLint guidelines?

RafLeszczynski:
Please fix linter issues and tests + add tests for new logic

RafLeszczynski:
LGTM after build passes

front/main/app/transitions.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * File required for liquid-fire addon to define the animations in.
+ * @returns {void}
+ */
+export default function () {
```
vforge: As this file seems to be global - can we name this transition and do not export it as default one? (especially that file is named `transitions`)

front/main/app/utils/generate-guid.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * create unique string with prefix as a base of the string if passed
+ *
+ * @param {string} [prefix='']
+ * @returns {string} unique string
+ */
+export default function generateGuid(prefix = '') {
+	return `${prefix}${Date.now()}`;
```
vforge: Just a thought:
Have you considered using something beside timestamp as a factor here?
```
Math.floor((1 + Math.random()) * 0x10000).toString(16).substring(1);
```
This snippet is also returning a random 4-byte hex string.

vforge:
LGTM, just minor questions :)

PR #2269, opened 2016-04-01
XW-1272 redirect to main page instead of /

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1272

## Description

When we were navigating to "" instead of main page - main page route was redirecting to correct main page title. Instead of doing extra redirect we can go to main page straight from curated content editor

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

PR #2272, opened 2016-04-01
SOC-2093 | Create layouts and use error substates on Discussions

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-2093

## Description
Let's use the power of error substates to render discussion content errors. Let's also use this pull request to extract discussions layouts and make them reusable:

* [x] Make reusable layouts for discussions using block components
* [x] Ensure all the errors are properly styled (currently they're broken on desktop)
* [x] Make a generic error page to show by default on discussions
* [x] Prepare error pages for all routes
* [x] Hide everything we should when there's no data from the service (e.g. top contributors section)

##Changes
* We have 3 block components for discussion layouts living under /components/discussion/layouts
* All content errors are being rendered using error substates (with their own routes and controllers)
* To retry loading a right route from within connection error substate we preserve the failed transition and we attempt to retry it
* Default error state uses layout for entities list wrapper

## Reviewers
@kvas-damian @rafalkalinski


front/main/app/templates/components/discussion/layouts/user-entities-list.hbs
```diff
@@ -29,42 +29,14 @@
 	</div>
 	<div class="discussion-page row">
 		<div class="rail large-3 columns">
-			{{#link-to 'discussion.forum' model.forumId discussionSort.sortBy class="side-button back-button active-element-theme-color"}}
+			{{#link-to 'discussion' class="side-button back-button active-element-theme-color"}}
```
kvas-damian: Can we make sure that `sortBy` won't be lost after this transition.

front/main/app/components/discussions/layouts/user-entities-list.js
```diff
@@ -0,0 +1,11 @@
+import DiscussionModalDialogMixin from '../../../mixins/discussion-modal-dialog';
+
+export default Ember.Component.extend(DiscussionModalDialogMixin, {
+	discussionSort: Ember.inject.service(),
+
+	actions: {
+		deleteAllPosts() {
+			this.attrs.deleteAllPosts(this.get('model.posts'));
```
kvas-damian: While we're only bubbling up, maybe we could call this method from template?

front/main/app/controllers/discussion/user-error.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend(
+	{
+		application: Ember.inject.controller(),
+
+		smartBannerVisible: Ember.computed.oneWay('application.smartBannerVisible'),
+		siteHeadPinned: Ember.computed.oneWay('application.siteHeadPinned'),
```
kvas-damian: Please extract common part of error controller into base class

front/main/app/templates/components/discussion/layouts/user-entities-list.hbs
bkoval: Fortunately, sort service remembers the chosen sort and restores it :)

rafalkalinski:
please, fix the Linter bugs

rafalkalinski:
LGTM

kvas-damian:
LGTM

front/main/app/components/discussion-connection-error.js
```diff
@@ -1,5 +1,16 @@
...
+	classNames: ['discussion-error'],
+
+	actions: {
+		/**
+		 * Attempts to retry a previously failed transition
+		 *
+		 * @returns {void}
+		 */
+		retry() {
+			this.get('error.transition').retry();
```
rogatty: Shouldn't this send action to the route? Feels kinda wrong to call route methods from a component.

front/main/app/routes/discussion/base.js
```diff
@@ -1,26 +1,46 @@
...
+			 *
+			 * @param {Ember.Object} model
+			 * @param {Ember.Transition} transition
+			 *
+			 * @returns {boolean}
+			 */
+			error(model, transition) {
+				this.controllerFor('application').set('noMargins', true);
+
+				model.get('error').set('transition', transition);
```
rogatty: Looks hacky, related to https://github.com/Wikia/mercury/pull/2272/files#r58431695

front/main/app/templates/components/discussion-connection-error.hbs
```diff
@@ -1,6 +1,6 @@
 <div class="error-content">
 	{{i18n 'main.connection-error' ns='discussion'}}
-	<a class="action-item" {{action attrs.retry}} >{{svg 'icon-retry' role='img' class='icon-retry'}}
+	<a class="action-item" {{action 'retry'}}>{{svg 'icon-retry' role='img' class='icon-retry'}}
```
rogatty: How about `{{action (route-action 'retry')}}`?
See https://github.com/DockYard/ember-route-action-helper

front/main/app/components/discussion-connection-error.js
bkoval: There's no easy way to find an error subroute and set a property (failed transition) on it, (is there?)

front/main/app/templates/components/discussion-connection-error.hbs
bkoval: We'll add a ticket to port all our fitting actions to route-actions, thanks!

front/main/app/components/discussion-connection-error.js
rogatty: It seems that the suggested solution is to store transition on the controller:
https://guides.emberjs.com/v2.4.0/routing/preventing-and-retrying-transitions/#toc_storing-and-retrying-a-transition

Then you could send an action to the controller instead of calling route method here.

bkoval: In our case, we're not sure which controller will handle the transition retry (this is a base class). It could be a generic error controller, could be an error subcontroller for the current controller. The only refactor I can see that could be made here is passing the transition as an action param from component to controller and then triggering the action from the error controller. 

Does that sound better? :)

rogatty: I don't think this makes much sense, just adds more code. Let's leave it as it is and think about it again when we know more.

kvas-damian:
LGTM

PR #2271, opened 2016-04-01
XW-1308 escape fragile data on server side

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1308

## Description

Stop abusing `{{{`

## Reviewers

@Wikia/x-wing

PR #2276, opened 2016-04-01
Xw 1308 escape fragile data on server side

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1308
* https://github.com/Wikia/mercury/pull/2271/files

## Reviewers

@Wikia/x-wing


vforge:
LGTM

PR #2277, opened 2016-04-01
DAT-4041 setup state only if infobox data provided

idradm:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4041

## Description

Correctly support empty infobox state.

## Reviewers

@Wikia/west-wing


dianafa:
LGTM

PR #2280, opened 2016-04-04
CE-3552, Open Infobox builder from VE Context

RafLeszczynski:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3552
* [Related dev changes #app/pull/10150](https://github.com/Wikia/app/pull/10150)

## Description

Open Infobox builder from VE Context
## Reviewers

@Grunny @adamkarminski @lukaszkonieczny @Wikia/x-wing @dianafa


front/main/app/controllers/infobox-builder.js
```diff
@@ -13,7 +25,18 @@ export default Ember.Controller.extend({
...
 		cancel() {
-			this.get('target').send('redirectToPage');
+			if (this.get('isVEContext')) {
+				if (!this.get('isDirty') || window.confirm(this.confirmationMessage())) {
+					const model = this.get('model');
+
+					this.get('target').send('returnToVE');
+					this.set('isDirty', false);
+					model.initInfoboxState();
+					model.setupInitialState();
```
dianafa: what about moving above lines to a separate function?

front/main/app/controllers/infobox-builder.js
```diff
@@ -47,6 +70,15 @@ export default Ember.Controller.extend({
 			return this.get('model').getTemplateExists(title);
 		},
 
+		goBackToVE() {
+			const model = this.get('model'),
+				route = this.get('target');
+
+			route.send('returnToVE', true);
+			model.initInfoboxState();
+			model.setupInitialState();
```
dianafa: Why do we need to cleanup infobox state before (after :P ) leaving? `returnToVE` is a promise, doesn't it matter if we run initInfoboxState & setupInitialState before or after it resolves?

front/main/app/components/infobox-builder.js
```diff
@@ -384,15 +384,17 @@ export default Ember.Component.extend(
 		 */
 		handleSaveResults(data, shouldRedirectToPage) {
```
RafLeszczynski: The JsDoc description for this method is no longer valid after adding VE integration. Please update the JsDoc

front/main/app/components/infobox-builder.js
```diff
@@ -384,15 +384,17 @@ export default Ember.Component.extend(
...
 			if (data.success) {
-				this.set('showSuccess', true);
+				this.set('showSuccess', !this.get('isVEContext'));
```
RafLeszczynski: why you don't want to show success after save ?

front/main/app/components/infobox-builder.js
```diff
@@ -384,15 +384,17 @@ export default Ember.Component.extend(
...
 
 				track({
 					action: trackActions.success,
 					category: 'infobox-builder',
 					label: 'save-successful'
 				});
 
-				if (shouldRedirectToPage) {
+				if (this.get('isVEContext')) {
```
RafLeszczynski: please consider caching isVEContext in local const as it's used multiple times in this method

front/main/app/controllers/infobox-builder.js
```diff
@@ -47,6 +70,15 @@ export default Ember.Controller.extend({
 			return this.get('model').getTemplateExists(title);
 		},
 
+		goBackToVE() {
```
RafLeszczynski: missing JS doc

front/main/app/controllers/infobox-builder.js
```diff
@@ -13,7 +25,18 @@ export default Ember.Controller.extend({
 		 * @returns {void}
 		 */
 		cancel() {
-			this.get('target').send('redirectToPage');
+			if (this.get('isVEContext')) {
+				if (!this.get('isDirty') || window.confirm(this.confirmationMessage())) {
```
RafLeszczynski: As controllers are more on less deprecated in Ember and they will be soon removed in considered a good pattern to make them as light as possible. Usually we use them just for  handling actions and calling appropriate methods on route or doing some operations required for changing UI that cannot be done on component level. 

I'd propose moving this logic to separate method on route. 

In addition please try to avoid working on a single model instance for multiple infobox templates and resetting the model after successful save. AFAIK you need to have clear model state each time you are lunching IB from within VE. Instead doing operations on single model instance please create new instance each time you open IB in VE and pass this instance to controller.

front/main/app/models/infobox-builder.js
```diff
@@ -7,14 +7,20 @@ const InfoboxBuilderModel = Ember.Object.extend({
...
-			image: 0,
-			title: 0,
-			'section-header': 0
-		};
-		this.infoboxState = [];
-		this.itemInEditMode = null;
+		this.initInfoboxState();
+	},
+
+	initInfoboxState() {
```
RafLeszczynski: Please check the comment above. instead of resetting model instance to initial sate consider simply creating new model instance and pass it to controller

front/main/app/routes/infobox-builder.js
```diff
@@ -312,5 +324,19 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+	 * set VE context - true if IB opened inside visual editor
+	 * @param {Boolean} isVEContext
+	 * @returns {void}
+	 */
+	setVEContext(isVEContext = false) {
+		const controller = this.controllerFor('infobox-builder');
+
+		controller.set('isVEContext', isVEContext);
+		if (isVEContext) {
+			controller.set('isDirty', false);
```
RafLeszczynski: why do you set `isDirty` to `false` for  `isVEContext`?

RafLeszczynski:
Only part of new introduced logic is covered by tests, please add tests for remaining part.

front/main/app/models/infobox-builder.js
RafLeszczynski: Please consider triggering `refresh` on route: http://emberjs.com/api/classes/Ember.Route.html#method_refresh

You will need to refactor `beforeModel` hook in `infobox-builder.js` route so that ponto in not loaded second time and styles are not added to DOM also for the second time.

front/main/app/components/infobox-builder.js
bognix: is it because ponto iframe is closed and message about success is not shown?

front/main/app/components/infobox-builder.js
lukaszkonieczny: Not anymore ;)

front/main/app/routes/infobox-builder.js
lukaszkonieczny: Not to show confirmation dialog when there is no change made by user

front/main/app/controllers/infobox-builder.js
```diff
@@ -13,7 +14,11 @@ export default Ember.Controller.extend({
 		 * @returns {void}
 		 */
 		cancel() {
-			this.get('target').send('redirectToPage');
+			if (this.get('isVEContext')) {
+				this.get('target').send('returnToVE');
```
RafLeszczynski: to avoid code duplication you can just call `goBackToVE` action using `this.send('goBackToVE')`

front/main/app/routes/infobox-builder.js
RafLeszczynski: but if you create a new template it already has a default items added to the stack so there is a change and user will lost what's in a preview if she exits IB without save

front/main/app/routes/infobox-builder.js
```diff
@@ -207,9 +229,12 @@ export default Ember.Route.extend(ConfirmationMixin, {
 	 * @returns {JQueryXHR}
```
RafLeszczynski: return type is not valid after code changes

front/main/app/routes/infobox-builder.js
```diff
@@ -207,9 +229,12 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
 	 */
 	loadPonto() {
```
RafLeszczynski: please consider changing this method name as it currently loads ponto only if it's not already being loaded or what what would be even better move the conditional to the `beforeModel` hook to keep this method responsible only for single action

front/main/app/routes/infobox-builder.js
```diff
@@ -28,7 +28,7 @@ export default Ember.Route.extend(ConfirmationMixin, {
 						this.setupStyles(responses.dataAndAssets);
```
RafLeszczynski: `setupStyles` will be called each time user reopens dialog with IB in VE which will cause multiplication of CSS assets added to DOM. It should be called only ones for initial load of Mercury similar to Ponto

front/main/app/routes/infobox-builder.js
```diff
@@ -28,7 +28,7 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
 						this.setupInfoboxData(responses.dataAndAssets);
 					})
-					.then(this.isWikiaContext)
+					.then(this.isWikiaContext.bind(this))
```
RafLeszczynski: do we want to check context each time we reload model for this route or should it be done only ones as context don't change?

front/main/tests/unit/controllers/infobox-builder-test.js
```diff
@@ -137,3 +137,49 @@ test('test group items calculation', function (assert) {
 		assert.deepEqual(controller.get('groupItems'), testCase.expected);
 	});
 });
+
+test('check the appropriate action is sent on save and cancel', function (assert) {
+	const controller = this.subject(),
+		model = {
+			saveStateToTemplate: sinon.stub().returns(Ember.RSVP.Promise.resolve()),
+			initInfoboxState: sinon.stub(),
```
RafLeszczynski: `initInfoboxState` method is not declared in `infobox-builder` model

RafLeszczynski:
please fix failing tests

front/main/tests/unit/controllers/infobox-builder-test.js
```diff
@@ -137,3 +137,49 @@ test('test group items calculation', function (assert) {
...
+				isVEContext: true,
+				calledWith: 'returnToVE'
+			},
+			{
+				action: 'cancel',
+				isVEContext: false,
+				calledWith: 'redirectToPage'
+			},
+			{
+				action: 'save',
```
RafLeszczynski: `save` action doesn't check `isVEContext` any more so these tests should be refactored

front/main/app/components/infobox-builder.js
RafLeszczynski: please update `handleSaveResults` - https://github.com/Wikia/mercury/blob/CE-3552/front/main/tests/unit/components/infobox-builder-test.js#L503

front/main/app/routes/infobox-builder.js
```diff
@@ -1,10 +1,10 @@
 import Ember from 'ember';
-import {track, trackActions} from 'common/utils/track';
 import InfoboxBuilderModel from '../models/infobox-builder';
 import ConfirmationMixin from 'ember-onbeforeunload/mixins/confirmation';
+import {track, trackActions} from 'common/utils/track';
 
 export default Ember.Route.extend(ConfirmationMixin, {
-	pontoLoadingInitialized: false,
+	modelInitialized: false,
```
RafLeszczynski: you are using this flag to check if `ponto` should be loaded, assets added to DOM and  `wikiContext` set. It doesn't have to do anything with initialising the model so consider changing the name for this flag. Maybe `isEnvirionmentSet` or something else???

front/main/app/routes/infobox-builder.js
```diff
@@ -17,18 +17,24 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
 		return new Ember.RSVP.Promise((resolve, reject) => {
-			if (window.self !== window.top && (!window.Ponto || !this.get('pontoLoadingInitialized'))) {
+			if (window.self !== window.top) {
 				const promises = {
-					dataAndAssets: this.loadInfoboxDataAndAssets(templateName),
-					ponto: this.loadPonto()
+					dataAndAssets: this.loadInfoboxDataAndAssets(templateName)
 				};
 
+				if (!window.Ponto || !this.get('modelInitialized')) {
```
RafLeszczynski: please consider moving this condition into computed property `shouldLoadPonto`

front/main/app/routes/infobox-builder.js
```diff
@@ -17,18 +17,24 @@ export default Ember.Route.extend(ConfirmationMixin, {
 		const templateName = transition.params['infobox-builder'].templateName;
 
 		return new Ember.RSVP.Promise((resolve, reject) => {
-			if (window.self !== window.top && (!window.Ponto || !this.get('pontoLoadingInitialized'))) {
+			if (window.self !== window.top) {
```
RafLeszczynski: you could also create a computed property `isIframeContext` and move this check there

front/main/app/routes/infobox-builder.js
```diff
@@ -17,18 +17,24 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+				if (!window.Ponto || !this.get('modelInitialized')) {
+					promises.ponto = this.loadPonto();
+				}
+
 				Ember.RSVP.hash(promises)
 					.then((responses) => {
-						this.setupStyles(responses.dataAndAssets);
+						if (!this.get('modelInitialized')) {
+							this.setupStyles(responses.dataAndAssets);
+							this.isWikiaContext();
```
RafLeszczynski: `isWikiaContext` returns promise, are you sure that the whole promise chain will work properly if `isWikiaContext` will be rejected?

front/main/tests/unit/controllers/infobox-builder-test.js
```diff
@@ -137,3 +137,37 @@ test('test group items calculation', function (assert) {
 		assert.deepEqual(controller.get('groupItems'), testCase.expected);
 	});
 });
+
+test('check the appropriate action is sent on save and cancel', function (assert) {
```
RafLeszczynski: the name of the test doesn't correspond with the test itself - `save` action is not being tested here

front/main/app/routes/infobox-builder.js
```diff
@@ -17,18 +17,24 @@ export default Ember.Route.extend(ConfirmationMixin, {
 		const templateName = transition.params['infobox-builder'].templateName;
```
RafLeszczynski: `beforeModel` has become pretty complex with many nested `if` statements. Please consider covering this method with tests so we are sure that all cases work as expected. In addition it would be also great if you think about reducing `if` and / or breaking this into smaller methods that would easier to test. 

Additionally consider moving conditions into descriptive computed properties so its much easier to understand what's going on  - check comments below

front/main/app/controllers/infobox-builder.js
```diff
@@ -13,7 +14,11 @@ export default Ember.Controller.extend({
 		 * @returns {void}
 		 */
 		cancel() {
-			this.get('target').send('redirectToPage');
+			if (this.get('isVEContext')) {
+				this.send('goBackToVE');
```
SebastianMarzjan: Why is this needed?
could https://github.com/Wikia/mercury/pull/2280/files#diff-e18b400a25ffa404ccfd4fa8ca7fe5b6R61 be invoked directly here?

front/main/tests/unit/controllers/infobox-builder-test.js
RafLeszczynski: please address comment above

front/main/app/templates/components/infobox-builder.hbs
```diff
@@ -72,17 +72,19 @@
...
-			click=(action 'tryGoToSource')
-			icon='source'
-			iconSize=24
-			id='go-to-source'
-		}}
-		{{#pop-over for="go-to-source" on="hover" flow='leftCenter'}}
-			<p class="infobox-builder-go-to-source-tooltip-text">{{i18n 'main.source-editor-tooltip'
-			ns='infobox-builder'}}</p>
-		{{/pop-over}}
+		{{#unless isVEContext}}
```
RafLeszczynski: could we change this variable to `isGoToSoruceEnabled` or something similar. IB should now anything about its communication with VE on template basis

front/main/app/routes/infobox-builder.js
```diff
@@ -141,6 +164,43 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
 
 	/**
+	 * Setup infobox builder by loading infobox data and styles.
+	 * Also initialize ponto and checks in what context
+	 * infobox builder was opened
+	 *
+	 * @param {string} templateName
+	 * @returns {Promise}
+	 */
+	setupEnvironment(templateName) {
```
RafLeszczynski: what do you think about `setupEnvironmentAndInfoboxData` ?

front/main/app/routes/infobox-builder.js
```diff
@@ -141,6 +164,43 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+
+	/**
+	 * Setup infobox data.
+	 * It's invoke on model refresh
+	 * to avoid loading all existing resources again
+	 *
+	 * @param {string} templateName
+	 * @returns {Promise}
+	 */
+	updateInfoboxData(templateName) {
```
RafLeszczynski: what do you think about `loadAndSetupInfoboxData` ?

front/main/app/routes/infobox-builder.js
```diff
@@ -17,20 +20,19 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
-					.then((responses) => {
-						this.setupStyles(responses.dataAndAssets);
-						this.setupInfoboxData(responses.dataAndAssets);
-					})
-					.then(this.isWikiaContext)
-					.then(resolve)
-					.catch(reject);
+			if (this.get('isIframeContext')) {
+				if (!this.get('isEnvironmentSet')) {
+					this.setupEnvironment(templateName)
```
RafLeszczynski: I'd suggest doing the following:
- break getting assets and getting data into two separate requests so we do not need to request the same things twice as it being done right now.
- this will allow us for first setting the environment which is Ponto, assets and wikia context and later request for template data and create a model from it
- in addition I'd think about moving ponto communication and assets into separate `Ember.Service` or maybe more than one service

There is the question when and how this work should be done. As I am WFH today please talk to @SebastianMarzjan  about how we can handle that

front/main/app/controllers/infobox-builder.js
```diff
@@ -47,6 +52,15 @@ export default Ember.Controller.extend({
 			return this.get('model').getTemplateExists(title);
 		},
 
+		/**
+		 * Triggers action to communicate VE that creating infobox is completed
+		 *
+		 * @returns {void}
+         */
+		goBackToVE() {
```
RafLeszczynski: could you make action names consistent - either `goBackToVE` or `returnToVE`

front/main/app/routes/infobox-builder.js
```diff
@@ -17,20 +20,19 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
-
-				Ember.RSVP.hash(promises)
-					.then((responses) => {
-						this.setupStyles(responses.dataAndAssets);
-						this.setupInfoboxData(responses.dataAndAssets);
-					})
-					.then(this.isWikiaContext)
-					.then(resolve)
-					.catch(reject);
+			if (this.get('isIframeContext')) {
```
RafLeszczynski: this code is also a great candidate for being covered with tests :-)

front/main/app/routes/infobox-builder.js
kamilkoterba: Added TODO
https://github.com/Wikia/mercury/pull/2280/commits/9d83f6e464cc222bb3069818e76e569ba94e92fa
https://wikia-inc.atlassian.net/browse/CE-3600

front/main/app/routes/infobox-builder.js
kamilkoterba: Done https://github.com/Wikia/mercury/pull/2280/commits/35e9199896e88f05f1815b24611c1e0070a95126

front/main/app/components/infobox-builder.js
```diff
@@ -393,8 +393,8 @@ export default Ember.Component.extend(
 					label: 'save-successful'
 				});
 
-				if (this.get('isVEContext')) {
-					this.get('goBackToVE')();
+				if (this.get('isGoToSourceEnabled')) {
```
RafLeszczynski: ? is this intended?? why do want to check `isGoToSourceEnabled` here ?

front/main/app/templates/components/infobox-builder.hbs
```diff
@@ -72,7 +72,7 @@
 				}}
 			</div>
 		{{/liquid-bind}}
-		{{#unless isVEContext}}
+		{{#unless isGoToSourceEnabled}}
```
RafLeszczynski: something is wrong with this condition…???? unless `isGoToSourceEnabled === true` render go to source button ???? this means that go to source button should be rendered when `isGoToSourceEnabled === false`

front/main/app/routes/infobox-builder.js
kamilkoterba: It's valid now

front/main/app/components/infobox-builder.js
kamilkoterba: This part is related to isVEContext check. I think that this redirect/goBack to VE logic should be moved to controller...

front/main/app/components/infobox-builder.js
```diff
@@ -392,7 +393,10 @@ export default Ember.Component.extend(
 					label: 'save-successful'
 				});
 
-				if (shouldRedirectToPage) {
+				if (!this.get('isGoToSourceEnabled')) {
```
RafLeszczynski: why are we checking the value of `isGoToSourceEnabled` to decide if we want to get back to VE?

front/main/app/controllers/infobox-builder.js
```diff
@@ -5,6 +5,8 @@ export default Ember.Controller.extend({
 	isDirty: false,
 	groupItems: [],
 	lastGroupItem: null,
+	isVEContext: false,
+	isGoToSourceEnabled: Ember.computed.not('isVEContext'),
```
RafLeszczynski: consider moving this computed property to component as you still need `isVEContext` on component level to trigger the right action - check the comment above

front/main/app/components/infobox-builder.js
kamilkoterba: As I wrote above:

This part is related to isVEContext check and I think it doesn't belong here. I think that this redirect/goBack to VE logic should be moved to controller. Then we'll get rid of VE related functionality within this component, which would make it agnostic to VE context.

I'm working on it.

front/main/app/controllers/infobox-builder.js
kamilkoterba: per @SebastianMarzjan comment below decided to call directly

front/main/app/routes/infobox-builder.js
kamilkoterba: Addressed https://github.com/Wikia/mercury/pull/2280/commits/e8a6ac2aab75b378ab7ef2c6ae208ed25a4a31db

RafLeszczynski:
LGTM if build pass

PR #2283, opened 2016-04-04
XW-1282 track PV on /main/ pages

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1282

## Description

Track PVs on /main/* 

## Reviewers

@Wikia/x-wing


front/main/app/mixins/main-page-route.js
```diff
@@ -44,6 +45,28 @@ export default Ember.Mixin.create({
...
 		});
+
+		transition.then(() => {
+			this.updateTrackingData(model);
+		});
+	},
+
+	updateTrackingData(model) {
+		const uaDimensions = {},
+			mainPageData = $.extend({}, M.prop('mainPageData')),
```
rogatty: What happens when user opens an article and then navigates to the main page? From what I see we set this `M.prop` only in `server-data.hbs` so only for main page as a first page.

front/main/app/mixins/main-page-route.js
```diff
@@ -44,6 +45,28 @@ export default Ember.Mixin.create({
...
+			adsContext = Ember.get(mainPageData, 'adsContext');
+
+		if (adsContext) {
+			uaDimensions[3] = Ember.get(adsContext, 'targeting.wikiVertical');
+			uaDimensions[14] = Ember.get(adsContext, 'opts.showAds') ? 'yes' : 'no';
+		}
+
+		setTrackContext({
+			a: model.get('title'),
+			n: Ember.getWithDefault(mainPageData, 'details.ns', 0)
```
vforge: Please set up dimension 19 (articleType, if possible) and 25 (namespace) in UA too.

front/main/app/mixins/main-page-route.js
bognix: good catch.
Because of that Ads were not working as well.

front/main/app/mixins/main-page-route.js
```diff
@@ -1,7 +1,30 @@
 import Ember from 'ember';
 import {normalizeToWhitespace} from 'common/utils/string';
+import {setTrackContext, trackPageView} from 'common/utils/track';
 
 export default Ember.Mixin.create({
+	curatedMainPageData: Ember.inject.service(),
+	ns: Ember.computed('curatedMainPageData.ns', function() {
```
rogatty: This property seems complicated, let's add a test for it!

front/main/app/mixins/main-page-route.js
```diff
@@ -1,7 +1,30 @@
...
+		if (!Ember.isEmpty(this.get('curatedMainPageData.ns'))) {
+			return this.getWithDefault('curatedMainPageData.ns', 0);
```
rogatty: Does it make sense to use `getWithDefault` if it's already `!Ember.isEmpty`?

front/main/app/mixins/main-page-route.js
```diff
@@ -1,7 +1,30 @@
...
+		} else {
+			const ns = Ember.getWithDefault(M.prop('mainPageData'), 'details.ns', 0);
+
+			this.set('curatedMainPageData.ns', ns);
```
rogatty: Wouldn't it be cleaner to use computed properties in the service and just get them here?

front/main/app/utils/wiki-handlers/curated-main-page.js
```diff
@@ -31,6 +31,9 @@ function afterModel(router, model) {
 		title: Ember.getWithDefault(Mercury, 'wiki.siteName', 'Wikia')
 	});
 
+	router.set('curatedMainPageData.ns', model.get('ns'));
+	router.set('curatedMainPageData.adsContext', model.get('adsContext'));
```
rogatty: `setProperties`

vforge:
LGTM

front/main/app/utils/wiki-handlers/curated-main-page.js
```diff
@@ -31,6 +31,11 @@ function afterModel(router, model) {
 		title: Ember.getWithDefault(Mercury, 'wiki.siteName', 'Wikia')
 	});
 
+	router.setProperties({
+		'curatedMainPageData.ns': model.get('ns'),
+		'curatedMainPageData.adsContext': model.get('adsContext')
```
rogatty: This part doesn't look too clean. We use router (it actually should be `route`, right?) here as a proxy to the service. If (for example) its name changes in the route then it will have to be updated here or it will stop working silently. I see two different ways to approach this:
1. Use `alias` instead of `readOnly` and set their values here without knowledge about the service.
2. Pass the service here and use it directly.

What do you think?

front/main/app/services/curated-main-page-data.js
```diff
@@ -0,0 +1,19 @@
...
+			return M.prop('mainPageData.adsContext');
+		},
+		set(_, value) {
+			return value;
+		}
+	}),
+
+	ns: Ember.computed({
+		get() {
+			return M.prop('mainPageData.ns');
```
rogatty: Should we use some defaults? What will happen if it's not there? Can it happen?

front/main/app/routes/wiki-page.js
```diff
@@ -12,6 +12,7 @@ export default Ember.Route.extend(RouteWithAdsMixin, {
 	redirectEmptyTarget: false,
 	wikiHandler: null,
 	currentUser: Ember.inject.service(),
+	curatedMainPageData: Ember.inject.service(),
```
rogatty: What's this for?

bognix: it's used in wiki-handlers/curated-main-page.js

rogatty: It's very implicit then ;) Will using aliases make it more explicit?

rogatty:
Please fix unit tests.

front/main/app/services/curated-main-page-data.js
bognix: well it shouldn't because we send it always from app - there is no condition when we don't send it.
We can add a check and fallback if you feel like it's needed.

front/main/app/routes/wiki-page.js
bognix: I hope it's better this way :)

vforge:
LGTM

hakubo:
LGTM

PR #2284, opened 2016-04-04
(CE-3570) Add support for VE context to InfoboxBuilder

Grunny:
## Links

* https://wikia-inc.atlassian.net/browse/CE-3570

## Description

Adds support for returning to VE and hiding the go to source button
when InfoboxBuilder is launched from VE.

## Reviewers

@RafLeszczynski @Wikia/community-engineering @Wikia/x-wing

### DIFF
--- a/front/main/app/controllers/infobox-builder.js
+++ b/front/main/app/controllers/infobox-builder.js
@@ -14,11 +14,7 @@ export default Ember.Controller.extend({
 		 * @returns {void}
 		 */
 		cancel() {
-			if (this.get('isVEContext')) {
-				this.get('target').send('returnToVE');
-			} else {
-				this.get('target').send('redirectToPage');
-			}
+			this.get('target').send(this.get('isVEContext') ? 'returnToVE' : 'redirectToPage');
 		},
 
 		/**
@@ -36,10 +32,12 @@ export default Ember.Controller.extend({
 			this.set('isDirty', false);
 
 			return model.saveStateToTemplate().then((urls = {}) => {
+				const route = this.get('target');
+
 				if (this.get('isVEContext')) {
-					this.get('target').send('returnToVE');
+					route.send('returnToVE');
 				} else if (shouldRedirectToPage) {
-					this.get('target').send('redirectToPage', urls.templatePageUrl);
+					route.send('redirectToPage', urls.templatePageUrl);
 				}
 			});
 		},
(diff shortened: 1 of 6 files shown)


front/main/app/controllers/infobox-builder.js
```diff
@@ -13,7 +14,11 @@ export default Ember.Controller.extend({
 		 * @returns {void}
 		 */
 		cancel() {
-			this.get('target').send('redirectToPage');
+			if (this.get('isVEContext')) {
+				this.get('target').send('returnToVE');
```
RafLeszczynski: what do you thing about:
`this.get('target').send(this.get('isVEContext') ? 'returnToVE' : 'redirectToPage');`

front/main/app/controllers/infobox-builder.js
```diff
@@ -31,7 +36,9 @@ export default Ember.Controller.extend({
 			this.set('isDirty', false);
 
 			return model.saveStateToTemplate().then((urls = {}) => {
-				if (shouldRedirectToPage) {
+				if (this.get('isVEContext')) {
+					this.get('target').send('returnToVE');
```
RafLeszczynski: Small optimisation but we could store reference to route in local const and avoid duplicating `this.get('target')`. What do you think about it?

front/main/tests/integration/components/infobox-builder-test.js
```diff
@@ -19,3 +19,20 @@ test('reset item in edit mode on clicking preview background', function (assert)
 	assert.equal(setEditItemSpy.calledWith(null), true);
 });
 
+test('Go to source is rendered if not in VE context', function (assert) {
```
RafLeszczynski: please consider merging this two tests into one with two testCases to avoid code duplication

front/main/tests/unit/controllers/infobox-builder-test.js
```diff
@@ -137,3 +137,53 @@ test('test group items calculation', function (assert) {
 		assert.deepEqual(controller.get('groupItems'), testCase.expected);
 	});
 });
+
+/**
+ * @returns {Object}
+ */
+function getRouteMock() {
```
RafLeszczynski: as this function is used only for single test you could move the logic inside the test. For example create a spy as local const, make it part of route mock and reset spy state for each test case using `spy.reset()`. wh

PR #2287, opened 2016-04-04
PLATFORM-1017: use docker to build mercury

owend:
## Links

* http://wikia-inc.atlassian.net/browse/PLATFORM-1017

## Description

Adding dockerfile.  Updating package.json so that it doesn't entirely remove the www directory during the build process.  When building with docker, I am "mounting" the www directory from the build environment, so that the output files can then be extracted normally (as .tar.gz) without changing the way the build process currently works.  However, removing the directory removes the mount/link.

We probably want to change the way phantomjs dependency is handled.  The binary was being downloaded/installed every time.  The phantomjs binary from apt-get was too old so I couldn't pre-install it.  Since it's not needed for building, I just disabled it.  If we do want to run tests inside docker in the future, we will want to fix that by pre-installing a newer phantomjs build into the image.  

## Reviewers

@hakubo


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# phantomjs workaround
+RUN echo -e '#!/bin/sh\necho "2.1.1"' > /bin/phantomjs
+RUN chmod a+x /bin/phantomjs
+
+# bower workaround
+RUN echo '{ "allow_root": true }' > /root/.bowerrc
+
+# install dependencies
+RUN npm install -g bower && npm install -g ember-cli
```
rogatty: How about `npm run setup-for-local`? We keep global dependencies there and keep them up to date.
https://github.com/Wikia/mercury/blob/PLATFORM-1017/package.json#L9

hakubo:
apart npm run setup-for-local - LGTM, thanks!

PR #2288, opened 2016-04-05
Clean up svgs to remove duplicated ids

rogatty:
## Description

Fixes:
>
Phantomas BOT ​*Tested url: http://muppet.wikia.com/wiki/Kermit?noexternals=1&buckysampling=0 - FAILED*​
```
DOMidDuplicated: limit 0, got 1
errors:
* Page-1: 2 occurrences 
```

## Reviewers

@kvas-damian @bkoval I've traced `loading.svg` and found that it's used only in `auth/app/styles/module/_forms.scss` in a rule for `input[type="submit"], .signup-provider-facebook` where only the latter exists (there are no inputs with `type="submit"`). To test the change I clicked "Connect with Facebook" and confirmed that it looks the same way as before.

@adamkarminski @lukaszkonieczny `info-filled.svg` is used as a "blue dot" and I confirmed that it still looks the same.


kvas-damian:
LGTM

PR #2309, opened 2016-04-07
SOC-1302 | tracking for the force login popup enhancement

rafalkalinski:
## Links
https://wikia-inc.atlassian.net/browse/SOC-1302

## Reviewers
@kvas-damian @bkoval


front/auth/app/common/auth-tracker.js
```diff
@@ -19,6 +19,25 @@ export default class AuthTracker {
 			a: pageType,
 			n: -1
 		});
+
+		this.trackCloseWindow();
```
kvas-damian: We can have multiple instances of `AuthTracker`. It can cause many handlers to be registered to the same event.

front/auth/app/common/auth-tracker.js
```diff
@@ -19,6 +19,25 @@ export default class AuthTracker {
...
+	/**
+	 * @returns {void}
+	 */
+
+	trackCloseWindow() {
+		if (!window.opener) {
+			return;
+		}
+
+		window.onbeforeunload = function () {
```
kvas-damian: this looks little bit old school to me. I believe that modern web apps use `window.addEventListener()`. This way we won't override existing event handlers.

front/auth/app/common/auth-tracker.js
```diff
@@ -19,6 +19,25 @@ export default class AuthTracker {
...
 			n: -1
 		});
+
+		this.trackCloseWindow();
+	}
+
+	/**
+	 * @returns {void}
+	 */
+
```
bkoval: No newline needed

server/app/facets/auth/auth-view.js
```diff
@@ -185,11 +185,12 @@ export function getDefaultContext(request) {
 		standalonePage: (viewType === VIEW_TYPE_DESKTOP && !isModal),
 		pageParams: {
 			cookieDomain: localSettings.authCookieDomain,
-			isModal,
 			enableSocialLogger: localSettings.clickstream.social.enable,
+			forceLogin: (request.query.forceLogin || undefined),
+			isModal,
+			parentOrigin: (isModal ? getOrigin(request) : undefined),
```
bkoval: As developers we should avoid assigning undefined - is there a way to use null instead here and in line 189?

kvas-damian:
LGTM

PR #2315, opened 2016-04-08
XW-1350 | limit infobox height in scss

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1350

@rogatty

### DIFF
--- a/front/main/app/styles/component/_article-media-thumbnail.scss
+++ b/front/main/app/styles/component/_article-media-thumbnail.scss
@@ -1,3 +1,10 @@
+.article-media-placeholder img {
+	@extend %image-placeholder;
+
+	display: inline-block;
+	width: 100%;
+}
+
 .article-media-thumbnail {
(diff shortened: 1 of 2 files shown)


front/main/app/styles/component/_article-media-thumbnail.scss
```diff
@@ -1,3 +1,10 @@
+.article-media-placeholder img {
```
rogatty: Let's move it below `.article-media-thumbnail { ... }` which is more important rule.

front/main/app/styles/module/article/_portable-infobox.scss
```diff
@@ -363,3 +363,7 @@ $group-header-font-size: 16px;
 		}
 	}
 }
+
+.pi-hero img.article-media-placeholder {
```
rogatty: Please find it a place somewhere above. Surely there is a reason to put it all there.

PR #2327, opened 2016-04-11
SOC-1394 | As a user, I want to load the next page of newer posts when I jump to a specific reply

rafalkalinski:
## Links

https://wikia-inc.atlassian.net/browse/SOC-1394
## Description

button that loads the next page of newer posts when a user jumps to a specific reply

## Reviewers

@kvas-damian @bkoval


front/common/public/locales/en/discussion.json
```diff
@@ -70,7 +71,14 @@
...
     "unlock-post": "Unlock Post",
-    "locked-and-reported-to-moderators-text": "This post is locked, and has been reported."
+    "locked-and-reported-to-moderators-text": "This post is locked, and has been reported.",
+    "upvote-link-label": "__count__ Upvote",
+    "upvote-link-label_plural": "__count__ Upvotes",
+    "replies-link-label": "__count__ Reply",
+    "replies-link-label_plural": "__count__ Replies",
+    "share-header": "Share Post",
+    "share-label": "Share"
+
```
tomasznapieralski: remove that empty line

front/main/app/components/discussion-post.js
```diff
@@ -7,10 +7,12 @@ export default Ember.Component.extend(
...
-		// we can't use: Ember.computed.gt('model.replies.0.position', 1),
-		// because we don't change the 'position' property of first reply - we add new objects there
-		canShowOlder: Ember.computed('model.replies.@each.position', function () {
-			return this.get('model.replies.0.position') > 1;
+		canShowOlder: Ember.computed('model.replies.firstObject.position', function () {
+			return this.get('model.replies.firstObject.position') > 1;
+		}),
+
+		canShowNewer: Ember.computed('model.replies.lastObject.position', 'model.repliesCount', function () {
+			return this.get('model.replies.lastObject.position') < this.get('model.repliesCount');
```
tomasznapieralski: will this work for the last item?
if array has 5 elements, last position would be 4, while repliesCount would be 5, so it will return true, right?

rafalkalinski: position is a property set by the service and starts counting from 1 :)

front/main/app/models/discussion/post.js
```diff
@@ -8,36 +8,48 @@ import DiscussionReply from './domain/reply';
...
-	pivotId: null,
 	replyLimit: 10,
+	threadId: null,
+	loadDir: {
+		older: 'olderthan',
+		newer: 'newerthan',
+	},
 
 	/**
+	 * @param {Object} data - result from xhr request
```
tomasznapieralski: can you either provide more detailed description or remove it, please?

front/main/app/models/discussion/post.js
```diff
@@ -8,36 +8,48 @@ import DiscussionReply from './domain/reply';
...
+				});
+
+		if (this.get('data.replies.firstObject.position') > Ember.get(newReplies, 'lastObject.position')) {
+			this.get('data.replies').unshiftObjects(newReplies);
+		} else {
+			this.get('data.replies').pushObjects(newReplies);
+		}
+	},
+
+
```
tomasznapieralski: remove extra line

front/main/app/routes/discussion/reply.js
```diff
@@ -2,7 +2,24 @@ import DiscussionPostRoute from './post';
...
 export default DiscussionPostRoute.extend(
 	{
+		discussionEditor: Ember.inject.service(),
 		controllerName: 'discussion.post',
 		templateName: 'discussion.post',
+
+		actions: {
+			/**
+			 * Triggers new reply creation on a model
+			 * @param {object} replyData
```
tomasznapieralski: please use `Object` as all objects are Capitalised

front/main/app/styles/module/_discussion.scss
```diff
@@ -449,6 +440,18 @@ body.discussions {
 		}
 	}
 
+	.load-newer {
+		align-self: center;
+		box-align: ceneter;
+		display: flex;
```
tomasznapieralski: fix order

front/main/app/components/discussion-post.js
tomasznapieralski: 👍

front/main/app/models/discussion/post.js
bkoval: I believe the information is sufficient enough here - it's a payload from an XHR request, which is stated in the comment. More granular info is not needed, as deleting the comment, as it's not an obvious inline callback :)

tomasznapieralski:
minors, LGTM

PR #2342, opened 2016-04-13
Revert "Revert "SOC-1919 - FE | As a user, I want to jump to a specific reply""

rafalkalinski:
Reverts Wikia/mercury#2341

This should be the PR for https://wikia-inc.atlassian.net/browse/SOC-1919


rafalkalinski:
moved here: https://github.com/Wikia/mercury/pull/2369

PR #2352, opened 2016-04-14
SOC-1953 | A lot of new tracking for Discussions

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/JIRA-1953

## Reviewers

@kvas-damian @rafalkalinski


front/main/app/components/discussion-editor.js
```diff
@@ -22,13 +23,25 @@ export default Ember.Component.extend(ViewportMixin, {
 	bodyText: '',
 	layoutName: 'components/discussion-editor',
 
+	contentAction: trackActions.PostContent,
```
kvas-damian: what do you think about making it more clear that this variable is related to tracking?
`contentTrackingAction` ?

front/main/app/components/discussion-post-editor.js
```diff
@@ -81,5 +82,16 @@ export default DiscussionEditorComponent.extend({
 			Ember.$('html, body').animate({scrollTop: 0});
 			this.handleNewItemCreated(newPost);
 		}
+	},
+
+	actions: {
+		/**
+		 * @returns {void}
+		 */
+		close() {
```
kvas-damian: What do you think about moving this method to parent class and only change tracking labels?

front/main/app/components/share-feature.js
```diff
@@ -65,6 +66,15 @@ export default Ember.Component.extend(
 			]
 		},
 
+		/**
+		 * @returns {void}
+		 */
+		trackOpening: Ember.on('didInsertElement', function () {
```
kvas-damian: What do you think about defining `didInsertElement` method instead?

front/main/app/components/share-feature.js
```diff
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import TrackClickMixin from '../mixins/track-click';
 import LanguagesMixin from '../mixins/languages';
+import {track as discussionTrack, getTrackActionForShareNetwork, trackActions} from '../utils/discussion-tracker';
```
kvas-damian: For consistency I'd stick to one variable name for `track`. `track` vs `discussionTrack`?

front/main/app/mixins/discussion-moderation-model.js
```diff
@@ -157,6 +158,8 @@ export default Ember.Mixin.create({
 			url: M.getDiscussionServiceUrl(`/${this.wikiId}/threads/${post.threadId}/lock`),
 			success: () => {
 				post.set('isLocked', true);
+
+				track(trackActions.PostLock);
```
kvas-damian: What do you think about moving tracking to more frontend code? I'm thinking about some component maybe?

kvas-damian: The same comment applies to all model and routes tracking.

front/main/app/components/share-feature.js
bkoval: There's no way to call didInserElement already defined in a component by calling _super() it only works with base class, not with a statically extended object.

front/main/app/mixins/discussion-moderation-model.js
bkoval: I want to be sure that I'm tracking a succesful action. This is the first place that find out about a success of an operation.

kvas-damian: It's not consistent with tracking other actions like changing sortBy.

bkoval: Yup, that makes sense. I'll move tracking actions to the component level.

front/main/app/components/share-feature.js
kvas-damian: :+1:

kvas-damian:
LGTM :shipit:

front/main/app/components/discussion-editor.js
```diff
@@ -22,13 +23,26 @@ export default Ember.Component.extend(ViewportMixin, {
 	bodyText: '',
 	layoutName: 'components/discussion-editor',
 
+	closeTrackingAction: trackActions.PostClose,
+	contentTrackingAction: trackActions.PostContent,
+	startTrackingAction: trackActions.PostStart,
```
rafalkalinski: It took me a while to figure it out what this name refers to. Maybe some short comment here?

front/main/app/utils/discussion-tracker.js
```diff
@@ -5,6 +5,22 @@ const trackActions = {
 	ReplyCreate: 'ReplyCreate',
 	UndoUpvotePost: 'UndoUpvotePost',
 	UpvotePost: 'UpvotePost',
+	LatestPostTapped: 'LatestPostTapped',
+	TrendingPostTapped: 'TrendingPostTapped',
```
rafalkalinski: I would do: 'LatestPostsActivated' and 'LatestPostsActivated'

front/main/app/models/discussion/base.js
```diff
@@ -82,7 +82,13 @@ export default Ember.Object.extend({
 			hasUpvoted = entity.get('userData.hasUpvoted'),
 			method = hasUpvoted ? 'delete' : 'post';
 
-		if (this.upvotingInProgress[entityId] || typeof entity.get('userData') === 'undefined') {
+		if (this.upvotingInProgress[entityId]) {
+			return null;
+		}
+
+		if (typeof entity.get('userData') === 'undefined') {
```
rafalkalinski: Wouldn't be enough sth like this:
```if(!entity.get('userData'))
?

File: front/main/app/templates/components/discussion-share.hbs [hunk 9]
```diff
@@ -7,5 +7,6 @@
 
 	{{share-feature
 		sharedUrl=attrs.sharedUrl
+		enableDiscussionTracker=true
```
rafalkalinski: Why do we need this only here? If this is because share-feature can also be used outside the discussions, then this is not the right way, because right now inside the component there is a code closely connected to Discussions.
I would rather recommend the solution based on injecting needed methods and objects into the universal component.

bkoval: I agree that tracking method injection might be better here, but it's out of scope. It needs refactoring code owned by many teams and then regression tests. If you feel like this is a must-have, please file a separate ticket.

front/main/app/utils/discussion-tracker.js
bkoval: This is not ours to choose - https://docs.google.com/spreadsheets/d/1Gk1iqWkpjjVgZNiooIfYUP2X5MunFkcDt2RHQhK7b7A/edit

front/main/app/models/discussion/base.js
bkoval: That's a good point, it should be enough.

front/main/app/utils/discussion-tracker.js
rafalkalinski: :+1:

rafalkalinski:
Please fix the linter issues.
Beside that: LGTM :+1:

kvas-damian:
New pull request at: https://github.com/Wikia/mercury/issues/2401

PR #2353, opened 2016-04-14
XW-1302 | Add firstRenderTime measurement

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1302
* https://davidwalsh.name/css-animation-callback

## Description

We want to know when users actually see the content and it turns out that `domContentLoaded` isn't the correct event to measure that.

Instead, let's use an empty element with CSS transition applied and measure when the transition is finished. I've run some tests on http://webpagetest.org and it seems that this time is the correct one to measure.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/server/app/views/_partials/first-render-time.hbs
+++ b/server/app/views/_partials/first-render-time.hbs
@@ -4,7 +4,7 @@
 <style type="text/css">
 	.render-time-test {
 		transition-property: opacity;
-		transition-duration: .000001s;
+		transition-duration: 1ms;
 		opacity: 0;
 	}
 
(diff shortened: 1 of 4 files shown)


server/app/views/_partials/first-render-time.hbs
```diff
@@ -0,0 +1,50 @@
+{{! @todo (XW-1374) this should be an Ember addon }}
+{{! Idea from https://davidwalsh.name/css-animation-callback }}
+<span class="render-time-test"></span>
+<style type="text/css">
+	.render-time-test {
+		transition-property: opacity;
+		transition-duration: .000001s;
```
vforge: It's `0.001s` minimal, so you can use just `1ms`

vforge:
LGTM, one minor change is suggested though.

vforge:
LGTM

hakubo:
LGTM

PR #2354, opened 2016-04-14
XW-fix-pi-expand-button | increased z-index

Warkot:
Increased z-index to cover media component chevrons.

It was a bug and it was causing selenium tests to fail.

@Wikia/x-wing


bognix:
LGTM

PR #2356, opened 2016-04-14
XW-1373 | Fix front tests

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1367


vforge:
Nice about tests, it would be extremely helpful to know why this stopped working.

front/main/app/components/category-section.js
```diff
@@ -23,6 +27,14 @@ export default Component.extend(
 					window.document.getElementById(index).scrollIntoView();
 					window.scrollBy(0, -50);
 				});
+			},
+
+			trackClick(category, label) {
```
bognix: this doesn't look like it's needed. We don't use it inside `templates/components/category-section.hbs`

front/main/app/components/article-wrapper.js
```diff
@@ -192,6 +190,14 @@ export default Ember.Component.extend(
 			updateHeaders(headers) {
 				this.set('headers', headers);
 			},
+
+			trackClick(category, label) {
```
bognix: not sure if it should be done now but we could change the `collapsible-menu` so it takes as an argument track function or `label` and `category`. It's because of this component we have this action here. If it's used only inside `collapsible-component` it makes sense to move it there.

front/main/app/components/category-section.js
rogatty: https://github.com/Wikia/mercury/blob/XW-1373/front/main/app/templates/components/category-section.hbs#L17
You were saying...? ;)

front/main/app/components/article-comments.js
```diff
@@ -127,7 +126,12 @@ export default Ember.Component.extend(
 			toggleComments() {
 				this.set('page', this.get('page') ? null : 1);
 				this.toggleProperty('isCollapsed');
-				this.trackClick('comments', this.get('page') ? 'expanded' : 'collapsed');
+
+				track({
```
bognix: we should create a helper for trackClick - it's 80% of use cases of this util

rogatty: It's
```
trackClick(
   category,
   label
)
```
vs.
```
track({
    action: 'click',
    category,
    label
});
```

Do you think it's worth it?

front/main/app/components/category-section.js
bognix: you are right, I opened in IDE `category-sections.hbs` by accident.

front/main/app/components/article-comments.js
bognix: yes, given it's in 62 places

PR #2359, opened 2016-04-15
XW-1367 | Upgrade Ember to v2.3.2

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1367
* https://github.com/emberjs/ember.js/blob/master/CHANGELOG.md#v232-march-17-2016

## Description

Upgrade Ember from v2.1.0 to v2.3.2 and fix everything that stopped working, including:
- removing `mixins/track-click` which was causing issues with tests for long time
- removing `<script>` tag from `components/widget-flite` template and instead using JS to create it
- fixing `isReported` attribute used in discussions

## Reviewers

@Wikia/x-wing 
/cc @kvas-damian @bkoval


front/main/app/components/discussion-post-card-detail.js
```diff
@@ -1,7 +1,7 @@
 import Ember from 'ember';
-import DiscussionPostCardBaseComponent from './discussion-post-card-base';
+import DiscussionPostCardBaseComponent from '../mixins/discussion-post-card-base';
 
-export default DiscussionPostCardBaseComponent.extend({
+export default Ember.Component.extend(DiscussionPostCardBaseComponent, {
```
rogatty: Should we revert that change? If not then it should be `DiscussionPostCardBaseMixin`.

rogatty:
One thing and it'll LGTM.

front/main/app/components/article-comments.js
```diff
@@ -127,7 +126,12 @@ export default Ember.Component.extend(
 			toggleComments() {
 				this.set('page', this.get('page') ? null : 1);
 				this.toggleProperty('isCollapsed');
-				this.trackClick('comments', this.get('page') ? 'expanded' : 'collapsed');
+
+				track({
```
bognix: I would do the helper for trackClick. It's called 62 times, I believe it's worth it.

PR #2363, opened 2016-04-18
XW-1377 | remove Hapi API proxy calls

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/JIRA-XW1377

## Description

We're getting rid of Hapi and replacing it with Fastboot—that's a step in this direction.
First we need to remove all calls to Hapi API from client side.
In https://github.com/Wikia/mercury/pull/2365 we remove the server side 

## Reviewers
@Wikia/x-wing


front/main/app/models/curated-content.js
```diff
@@ -25,6 +25,46 @@ import Ember from 'ember';
  * @property {number} height
  */
 
+
```
tomasznapieralski: remove extra line

front/main/app/models/curated-content.js
```diff
@@ -25,6 +25,46 @@ import Ember from 'ember';
...
+		});
+	} else if (type === 'category') {
+		const query = {
+			controller: 'ArticlesApi',
+			method: 'getList',
+			expand: 'true',
+			abstract: 0,
+			width: 300,
+			height: 300,
+			category: `${decodeURIComponent(title)}`,
```
vforge: Why can't you write it as `category: decodeURIComponent(title)`?

front/main/app/components/side-nav-local-wikia-search.js
```diff
@@ -164,7 +164,14 @@ export default Ember.Component.extend(
 		 * @returns {string}
 		 */
 		getSearchURI(query) {
-			return `${M.prop('apiBase')}/search/${encodeURIComponent(query)}`;
+			return M.buildUrl({
+				path: '/wikia.php',
+				query: {
+					controller: 'MercuryApi',
+					method: 'getSearchSuggestions',
+					query
```
rogatty: Do we need to use `encodeURIComponent` here or is it irrelevant now?

bognix: `M.buildUrl` does URL encoding

front/main/app/models/curated-content.js
bognix: overlooked it - thanks

front/main/app/models/curated-content.js
```diff
@@ -36,28 +76,18 @@ CuratedContentModel.reopenClass({
 	/**
 	 * @param {string} title
 	 * @param {string} [type='section']
-	 * @param {string|null} [offset=null]
 	 * @returns {Ember.RSVP.Promise}
 	 */
-	find(title, type = 'section', offset = null) {
+	find(title, type = 'section') {
```
rogatty: We still want to use offset, right?

front/main/app/models/curated-content.js
```diff
@@ -36,28 +76,18 @@ CuratedContentModel.reopenClass({
...
-				params = {};
-
-			let url = `${M.prop('apiBase')}/main/`;
-
-			url += `${type}/${title}`;
-
-			if (offset) {
-				params.offset = offset;
-			}
+				url = getURL(...arguments);
```
rogatty: I wouldn't use `...arguments` without good reason. `getURL(title, type, offset)` will work just fine.

front/main/tests/unit/models/article-comments-test.js
```diff
@@ -1,15 +1,77 @@
...
+	beforeEach() {
+		stub = sinon.stub(M, 'buildUrl');
+	},
+
+	afterEach() {
+		M.buildUrl.restore();
+	}
+});
 
 test('url creates valid url to a resource', function (assert) {
```
rogatty: I would remove this test completely, IMO it's pointless.

rogatty:
LGTM

front/main/tests/unit/models/article-comments-test.js
bognix: Let's leave one test case. Better safe than sorry. But the rest of test cases doesn't make any sense

hakubo:
LGTM

PR #2370, opened 2016-04-19
XW-1369 | replace ember-cli-meta-tags with ember-cli-head

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1369

## Reviewers

@Wikia/x-wing


front/main/app/index.html
```diff
@@ -1,28 +1,28 @@
 <!DOCTYPE html>
 <html lang="{{wikiVariables.language.content}}" dir="{{wikiVariables.language.contentDir}}" {{#if preview}}class='article-preview'{{/if}}>
 <head>
-	<meta charset="utf-8">
-	<meta http-equiv="X-UA-Compatible" content="IE=edge">
-	<title>{{documentTitle}}</title>
-	<meta name="keywords"
+	<meta data-hapi-leftovers charset="utf-8">
```
rogatty: How about marking start and end instead? That's what ember-cli-head does:
https://github.com/ronco/ember-cli-head/blob/master/addon/templates/components/head-layout.hbs
https://github.com/ronco/ember-cli-head/blob/master/app/instance-initializers/browser/head.js

We wouldn't have to pollute all tags with this attribute.

front/main/app/mixins/head-tags.js
```diff
@@ -0,0 +1,101 @@
...
+		}
+	},
+
+	/**
+	 * This function sets static head tags defined in templates/head.hbs
+	 * This is for head tags which are set only once
+	 *
+	 * @returns {void}
+	 */
+	setStaticHeadTags() {
```
bognix: Don't set static tags on each transition. It has to be set only once.

front/main/app/mixins/head-tags.js
```diff
@@ -0,0 +1,101 @@
...
+
+	/**
+	 * This function removes all head tags with data-hapi-leftovers attribute to avoid having duplicates.
+	 * This function should be removed when fastboot will be introduced.
+	 *
+	 * @returns {void}
+	 */
+	removeServerTags() {
+		if (!this.controllerFor('application').get('serverTagsRemoved')) {
+			Ember.$('head *[data-hapi-leftovers]').remove();
```
bognix: * you don't need the `*`
* let's keep the logic for head tags inside head tags service - set the `serverTagsRemoved` on service not application controller

vforge: I would even use `'[data-hapi-leftovers]'`

front/main/app/mixins/head-tags.js
```diff
@@ -0,0 +1,101 @@
...
+	 * This function updates dynamic tags defined in templates/head.hbs
+	 * This is for head tags which could be changed on different routes
+	 *
+	 * @param {Object} model, this is model object from route::afterModel() hook,
+	 * it should be used ONLY in custom implementation of this function
+	 * @param {Object} data, object where you can pass data from custom implementation of this function
+	 * @returns {void}
+	 */
+	setDynamicHeadTags(model, data = {}) {
+		const wikiVariables = Ember.get(Mercury, 'wiki'),
```
bognix: how about using extend? (https://api.jquery.com/jquery.extend/)

front/main/app/templates/head.hbs
```diff
@@ -0,0 +1,56 @@
+<!--Tags set once (static head tags)-->
```
vforge: Please use `{{! Tags.... }}` for comment

front/main/app/templates/head.hbs
```diff
@@ -0,0 +1,56 @@
...
+
+{{#if model.qualarooScript}}
+	<link rel="dns-prefetch" href="{{model.qualarooScript}}">
+{{/if}}
+
+{{#if model.favicon}}
+	<link rel="shortcut icon" href="{{model.favicon}}">
+{{/if}}
+
+<meta charset="utf-8">
```
vforge: charset should be placed directly after `<head>` tag - see https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Introduction_to_HTML5

front/main/app/index.html
tomasznapieralski: I would rather leave data attributes on elements that i really want to remove, solution you propose is a little bit  'magic'

rogatty: It's just a simple range, no magic involved :) With the current solution, every time developer wants to add a new tag he must remember about this attribute.

Anyway, hopefully it's here only for a short time so I won't push. If you insist on doing it this way please rename it to something like `data-server-generated`, "leftovers" sounds bad.

bognix: on the other hand, every time developer wants to add a new tag he/she must remember about placing it in the right place. I agree with @tomasznapieralski - I believe in explicitly setting given tag as `server-generated`. 

However as you said - this is temporary and meta tags are not updated very often so I won't push on any of possible solutions.

front/main/package.json
```diff
@@ -31,15 +31,14 @@
...
+    "ember-cli-head": "0.0.5",
     "ember-cli-deprecation-workflow": "0.2.2",
     "ember-cli-htmlbars": "1.0.3",
     "ember-cli-htmlbars-inline-precompile": "0.3.1",
     "ember-cli-inject-live-reload": "1.4.0",
     "ember-cli-inline-content": "0.4.0",
-    "ember-cli-meta-tags": "1.0.1",
     "ember-cli-moment-shim": "1.0.0",
     "ember-cli-qunit": "1.4.0",
-    "ember-cli-qunit-eslint": "2.0.0",
```
bognix: I removed this because we run eslint on every PR check by ourselves. We use old version of ESlint so new addons we install may cause problems (like `ember-cli-head`) when are checked by `qunit-eslint`. `ember-cli-head` was expecting ESlint 2.0 syntax.

bognix:
LGTM

tomasznapieralski:
LGTM

front/main/app/mixins/head-tags-dynamic.js
```diff
@@ -0,0 +1,81 @@
...
+export default Ember.Mixin.create({
+	headData: Ember.inject.service(),
+
+	/**
+	 * @param {Object} model
+	 * @param {Ember.Transition} transition
+	 * @returns {void}
+	 */
+	afterModel(model, transition) {
+		this._super(...arguments);
```
vforge: Please use real arguments instead.

front/main/app/mixins/head-tags-dynamic.js
```diff
@@ -0,0 +1,81 @@
...
+		}
+	},
+
+	/**
+	 * This function updates dynamic tags defined in templates/head.hbs
+	 * This is for head tags which could be changed on different routes
+	 *
+	 * @param {Object} model, this is model object from route::afterModel() hook,
+	 * it should be used ONLY in custom implementation of this function
+	 * @param {Object} data, object where you can pass data from custom implementation of this function
```
vforge: `[data={}]

front/main/app/mixins/main-page-route.js
```diff
@@ -27,6 +28,8 @@ export default Ember.Mixin.create({
 	 * @returns {void}
 	 */
 	afterModel(model, transition) {
+		this._super(...arguments);
```
vforge: Please use real arguments instead.

front/main/app/routes/discussion/base.js
```diff
@@ -1,12 +1,27 @@
...
 export default Ember.Route.extend(
 	DiscussionLayoutMixin,
+	HeadTagsDynamicMixin,
 	{
 		postDeleteFullScreenOverlay: false,
 
+		/**
+		 * Custom implementation of HeadTagsMixin::setDynamicHeadTags
+		 * @param {Object} model, this is model object from route::afterModel() hook
+		 * @param {Object} data
```
vforge: `[data={}]`

front/main/app/routes/discussion/base.js
```diff
@@ -1,12 +1,27 @@
...
+import HeadTagsDynamicMixin from '../../mixins/head-tags-dynamic';
 
 export default Ember.Route.extend(
 	DiscussionLayoutMixin,
+	HeadTagsDynamicMixin,
 	{
 		postDeleteFullScreenOverlay: false,
 
+		/**
+		 * Custom implementation of HeadTagsMixin::setDynamicHeadTags
```
vforge: Is this description necessary - it doesn't bring anything.
Same for model's description here. Everything is already described in the mixin.

front/main/app/routes/discussion/post.js
```diff
@@ -57,6 +59,17 @@ export default DiscussionBaseRoute.extend(
 			this._super();
 		},
 
+		/**
+		 * Custom implementation of HeadTagsMixin::setDynamicHeadTags
```
vforge: Same as previous jsdoc comment

front/main/app/routes/main-page-category.js
```diff
@@ -14,14 +14,19 @@ export default Ember.Route.extend(MainPageRouteMixin, MetaTagsMixin, RouteWithAd
 	},
 
 	/**
-	 * @returns {*}
+	 * Custom implementation of HeadTagsMixin::setDynamicHeadTags
```
vforge: See previous comment.

vforge:
Minor issue with the comment, LGTM otherwise

front/main/app/routes/discussion/base.js
bognix: Well if it's obvious for you than it's great. On the other hand we believe that overriding method which is defined inside mixin and called inside mixin may be confusing. There are some routes where one route extends another route and each of this route includes bunch of mixins. In such case looking for a place where this method is called may be time consuming and confusing - hence we added the comment to avoid such situations.

vforge: 👍

PR #2372, opened 2016-04-20
JPN 396

d34th4ck3r:
## Links

* http://wikia-inc.atlassian.net/browse/JPN-396

Added new press release for Japan homepage.


homepage/front/styles/component/_companyinfo.scss
```diff
@@ -17,6 +17,10 @@
 		text-decoration: none;
 	}
 
+  	li {
```
perjg: Order here should be alphabetical, so this elements should go under aside.

d34th4ck3r:
@perjg Fixed the order according to your comment.

perjg:
LGTM

PR #2375, opened 2016-04-20
DAT-4045 infobox-builder input support for asian chars

idradm:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4045

## Description

Adds one-way-input ember module. 
Adds support for asian chars in infobox-builder inputs.

## Reviewers

@Wikia/west-wing


front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -53,6 +42,15 @@ export default Ember.Component.extend(
 					this.get('labelValue'),
 					this.get('labelFocusTrackingKey')
 				);
+			},
+
+			onLabelUpdate(value) {
+				console.info(value);
```
SebastianMarzjan: Each time you use console.* a kitty dies.
Please remove it :)

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -14,19 +14,8 @@ export default Ember.Component.extend(
...
-				const item = this.get('item');
-
-				// mark that user interacted with label input
-				this.set('wasLabelAltered', true);
-
-				this.get('editRowItem')(item, value);
-				return value;
-			}
+		labelValue: Ember.computed(function() {
+			return this.get('item.data.label');
```
rogatty: How about `labelValue: Ember.computed.readOnly('item.data.label')`?

idradm: This value is later overwritten by next component (text-field), therefore, it would cause exceptions.

rogatty: If it's supposed to be two-way binding then you can use `labelValue: Ember.computed.alias('item.data.label')`

front/main/app/components/infobox-builder-edit-item-row.js
```diff
@@ -14,19 +14,8 @@ export default Ember.Component.extend(
...
-			set(key, value) {
-				const item = this.get('item');
-
-				// mark that user interacted with label input
-				this.set('wasLabelAltered', true);
-
-				this.get('editRowItem')(item, value);
-				return value;
-			}
+		labelValue: Ember.computed(function () {
```
hakubo: Ember.computed.alias or
Ember.computed.reads

front/main/app/components/wikia-ui-components/text-field.js
```diff
@@ -76,15 +76,25 @@ export default Ember.Component.extend({
 		},
 
 		/**
+		 * @returns {void}
+		 */
+		onEnter() {
+			const onEnterHandler = this.get('onEnterHandler');
```
hakubo: You could set onEnterHandler to Ember.K();

this way you would not have to check if it exists here

front/main/app/components/wikia-ui-components/text-field.js
```diff
@@ -76,15 +76,25 @@ export default Ember.Component.extend({
...
+
+		/**
 		 * @param {String} value
-		 * @param {Event} event
 		 * @returns {void}
 		 */
-		onKeyUp(value, event) {
-			const onKeyUpHandler = this.get('onKeyUpHandler');
+		onUpdate(value) {
+			const onUpdateHandler = this.get('onUpdateHandler');
```
hakubo: same here - you could just provide default

front/main/tests/unit/components/wikia-ui-components/text-field-test.js
```diff
@@ -102,12 +101,35 @@ test('calls appropriate handler on focus, blur and keyUp', function (assert) {
...
+test('calls appropriate handler on enter key pressed and update actions', function (assert) {
+	const component = this.subject(),
+		onEnterHandler = sinon.spy(),
+		onUpdateHandler = sinon.spy(),
+		cases = [
+			{
+				action: 'onEnter',
+				handlerName: 'onEnterHandler',
+				handler: onEnterHandler,
+				calledWidthArguments: []
```
hakubo: is it calledWithArguments or calledWidthArguments?

PR #2376, opened 2016-04-20
XW | fix too much encoding

bognix:
## Description

We don't need to `encode` title which comes from MW. 
It was needed before because Hapi when sending data to the client was decoding it.
Right now we don't use Hapi as proxy for API calls.
The handlebars take care of encoding titles when passing them to the client
 
## Reviewers
@Wikia/x-wing


vforge:
LGTM

rogatty:
LGTM

PR #2379, opened 2016-04-20
XW-1380 | New article header

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1380
* http://wikia-inc.atlassian.net/browse/XW-1385

## Description

New article header and hero image

## Reviewers

@Wikia/x-wing


front/main/app/components/article-wrapper.js
```diff
@@ -109,6 +109,18 @@ export default Ember.Component.extend(
 			return this.get('currentUser.isAuthenticated') && !Ember.$.cookie('recent-edit-dismissed');
 		}),
 
+		heroImage: Ember.computed('model.media', function () {
+			let heroImage = false;
+
+			this.get('model.media.media').forEach((current) => {
```
rogatty: @vforge won't it break when there is no media in article?

front/main/app/components/wiki-page-header.js
```diff
@@ -0,0 +1,69 @@
...
+	ViewportMixin,
+	{
+		imageAspectRatio: 16 / 9,
+		classNames: ['wikia-page-header'],
+		isMainPage: false,
+		mainPageName: Ember.get(Mercury, 'wiki.siteName'),
+		mainPageTitle: Ember.get(Mercury, 'wiki.mainPageTitle'),
+
+
+		computedStyle: Ember.computed('heroImage', 'viewportDimensions.width', function () {
```
rogatty: Can we split it in smaller chunks? Something like https://github.com/Wikia/mercury/blob/dev/front/main/app/components/portable-infobox-hero-image.js but with `@todo` done ;)

front/main/app/templates/components/wiki-page-header.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="wiki-page-header {{if hasBlock 'has-action'}} {{if heroImage 'has-hero-image'}}" style={{computedStyle}}>
+    <div class="wiki-page-header__wrapper">
+		<div class="wiki-page-header__main">
+			{{#if isMainPage}}
+				<h1 class="wiki-page-header__title">{{mainPageName}}</h1>
+			{{else}}
+				<h2 class="wiki-page-header__site-name" {{action 'trackClick'}}>
```
rogatty: This action isn't necessary. You can use `trackingCategory` attribute in `link-to` helper below.

PR #2381, opened 2016-04-20
XW | clean up encoding

bognix:
## Description

Hapi was doing some special encoding/decoding when passing data to client. Because of that we had a lot of encoding/decoding on the client side. Without Hapi we can remove this logic.
All encoding is handled by:
* `M.buildUrl` when it comes to creating URLs
* Ember when it comes to transitions to different routes
* Handlebars when it comes to displaying data to user

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/utils/wiki-handlers/wiki-page.js
+++ b/front/main/app/utils/wiki-handlers/wiki-page.js
@@ -12,11 +12,13 @@ function getURL(params) {
 	const query = {
 		controller: 'MercuryApi',
 		method: 'getPage',
-		title: params.title,
+		// We need to decode title because MW sends encoded content
+		// It's only necessary in case of in-content links
+		title: decodeURIComponent(params.title),
 	};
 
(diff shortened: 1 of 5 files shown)


vforge:
lgtm

front/main/app/utils/wiki-handlers/wiki-page.js
```diff
@@ -12,11 +12,13 @@ function getURL(params) {
...
 		method: 'getPage',
-		title: params.title,
+		// We need to decode title because MW sends encoded content
+		// It's only necessary in case of in-content links
+		title: decodeURIComponent(params.title),
 	};
 
 	if (params.redirect) {
-		query.redirect = `?redirect=${encodeURIComponent(params.redirect)}`;
+		query.redirect = params.redirect;
```
tomasznapieralski: you sure, you don't need `
?redirect`?

PR #2383, opened 2016-04-21
Changed the order to decending

d34th4ck3r:
## Links

* http://wikia-inc.atlassian.net/browse/JPN-396

Changed the order to decreasing by year.


d34th4ck3r:
@perjg Can you check?

perjg:
LGTM

PR #2386, opened 2016-04-21
ramove style referencing to unexisting class

RafLeszczynski:
Fix broken ember build after node-sass dependency upgrade

## Reviewers

@Wikia/x-wing @SebastianMarzjan


rogatty:
LGTM

SebastianMarzjan:
Yep, removing a reference to non-existent class makes perfect sense

PR #2387, opened 2016-04-21
ramove style referencing to unexisting class - to release

RafLeszczynski:
## Description

Fix broken ember build after node-sass dependency upgrade. The removed line probably was always causing error but before last node-sass upgrade it was silent.

https://github.com/sass/libsass/issues/1670

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

PR #2389, opened 2016-04-21
XW-1422 | Bring back title encoding

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1422

## Description

We need to encode special characters like `?`, `&`, `=` and probably more.

MediaWiki uses `wfUrlencode` method which has a whitelist of characters to not encode but this is outside of the scope. For now, let's make it work again ASAP.

## Reviewers
@Wikia/x-wing


bognix:
LGTM

PR #2395, opened 2016-04-22
Nav redesign

SebastianMarzjan:
## Links
* https://wikia-inc.atlassian.net/browse/DAT-4060

## Description
This is a redesign of navigation in Mercury


front/main/app/components/article-wrapper.js
```diff
@@ -133,7 +133,7 @@ export default Ember.Component.extend(
 			 * @returns {void}
 			 */
 			expandSideNav() {
```
RafLeszczynski: please consider changing name of this action to `expandDrawer`

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,71 @@
...
+
+		currentUser: Ember.inject.service(),
+		isUserAuthenticated: Ember.computed.oneWay('currentUser.isAuthenticated'),
+
+		logoutLink: M.buildUrl({
+			namespace: 'Special',
+			title: 'UserLogout'
+		}),
+
+		userProfileLink: Ember.computed('currentUser.name', function () {
```
RafLeszczynski: you could cache `computed` in local const

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,71 @@
...
+			return M.buildUrl({
+				namespace: 'User',
+				title: this.get('currentUser.name')
+			});
+		}),
+
+		actions: {
+			onClick(item) {
```
RafLeszczynski: add JSDoc

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,71 @@
...
+				track({
+					action: trackActions.click,
+					category,
+					label
+				});
+			}
+		},
+
+		scrollTop() {
+			Ember.$('.wikia-drawer__content').scrollTop(0);
```
RafLeszczynski: if `scrollTop` is called more than ones please consider caching jQuery collection for `.wikia-drawer__content` in local variable

front/main/app/components/wikia-search.js
```diff
@@ -57,15 +57,6 @@ export default Ember.Component.extend(
 				window.location.assign(`${Mercury.wiki.articlePath}Special:Search?search=${value}&fulltext=Search`);
 			},
```
RafLeszczynski: does search get autofocus on drawer open???

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
+import Ember from 'ember';
+
+export default Ember.Object.extend({
+
+	init() {
+		this._super(...arguments);
+		this.state = [];
+	},
+
+	hubsLinks: Ember.get(Mercury, 'wiki.navigation2016.hubsLinks'),
```
RafLeszczynski: please consider caching `get` in local `const`

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+	localLinks: Ember.get(Mercury, 'wiki.navigation2016.localNav'),
```
RafLeszczynski: I wander if `M` is not an alias for global `Mercury` object???

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+	exploreWikiaLinks: Ember.get(Mercury, 'wiki.navigation2016.exploreWikiaMenu'),
+	exploreWikiaLabel: Ember.get(Mercury, 'wiki.navigation2016.exploreWikia.textEscaped'),
+	wikiName: Ember.get(Mercury, 'wiki.siteName'),
+	wikiLang: Ember.get(Mercury, 'wiki.language.content'),
+
+	items: Ember.computed('state.[]', function () {
+		const s = this.get('state');
+
+		if (s.length > 0 && !s[0]) {
```
RafLeszczynski: please consider moving the conditional logic to descriptive computed property

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+			// special case for exploration nav
+			return this.prepareExplorationItems();
+		}
+		// subNav, show only local
+		if (s.length > 0) {
```
RafLeszczynski: please check comment above

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+			return this.prepareLocalItems(s);
+		}
+		// if state == []
+		return this.prepareGlobalItems().concat(this.prepareLocalItems([]));
+	}),
+
+	header: Ember.computed('state.[]', function () {
```
RafLeszczynski: please consider caching `computed` in local `const`

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+		let nav = this.get('localLinks'),
+			header = this.get('exploreWikiaLabel');
+
+		for (const i of this.get('state')) {
+			if (nav[i - 1]) {
```
RafLeszczynski: please move `nav[i - 1]` to local `let` with descriptive name

front/main/app/components/discussion-header.js
```diff
@@ -4,6 +4,16 @@ import HeadroomMixin from '../mixins/headroom';
 export default Ember.Component.extend(
 	HeadroomMixin,
 	{
+		headroomOptions: {
+			classes: {
+				initial: 'discussion-headroom',
+				pinned: 'discussion-headroom-pinned',
+				unpinned: 'discussion-headroom-un-pinned',
+				top: 'discussion-headroom-top',
+				notTop: 'discussion-headroom-not-top'
```
dianafa: where is this `notTop` used?

front/main/app/components/wikia-nav.js
dianafa: it's used only once, so not necessarily

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+	},
+
+	hubsLinks: Ember.get(Mercury, 'wiki.navigation2016.hubsLinks'),
+	localLinks: Ember.get(Mercury, 'wiki.navigation2016.localNav'),
+	exploreWikiaLinks: Ember.get(Mercury, 'wiki.navigation2016.exploreWikiaMenu'),
+	exploreWikiaLabel: Ember.get(Mercury, 'wiki.navigation2016.exploreWikia.textEscaped'),
+	wikiName: Ember.get(Mercury, 'wiki.siteName'),
+	wikiLang: Ember.get(Mercury, 'wiki.language.content'),
+
+	items: Ember.computed('state.[]', function () {
```
dianafa: const {computed, Object, get} = Ember;

front/main/app/models/wikia-nav.js
dianafa: +1

front/main/app/models/wikia-nav.js
dianafa: can't we merge this `if (s.length > 0)` check from lines 20 and 25 into one?

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+				name: item.textEscaped,
+				trackLabel: `open-${item.trackingLabel}`
+			};
+		});
+	},
+
+	prepareGlobalItems() {
+		let global = [];
+
+		if (this.get('wikiLang') === 'en') {
```
dianafa: why we're doing this only for en wikis?

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,144 @@
...
+		return global.concat([{
+			// add exploration sub menu item
+			type: 'side-nav-menu-root',
+			index: 0,
+			name: this.get('exploreWikiaLabel'),
+			trackLabel: 'open-explore-wikia'
+		}]);
+	},
+
+	prepareLocalItems(state) {
```
dianafa: can we split this 50 line method into pieces?

front/main/app/templates/components/side-nav-menu-item.hbs
```diff
@@ -0,0 +1,11 @@
+<li class="side-nav-menu__item {{className}}" {{action attrs.itemClick preventDefault=false}}>
+	{{#if href}}
+		{{#link-to link href}}
+			{{name}}
+		{{/link-to}}
+	{{else}}
+		{{#link-to link}}
+			{{name}}
+		{{/link-to}}
+	{{/if}}
```
dianafa: what would happen without this `{{#if href}}` check?

front/main/app/styles/component/_site-head.scss
```diff
@@ -36,6 +50,22 @@ $site-head-height: 50; //unitless
 		}
 	}
 
+	.site-head-icon {
+		position: relative;
+		svg {
```
RafLeszczynski: could you use icon button classes instead of svg?

front/main/app/components/article-wrapper.js
dianafa: from what I can see it's not used anywhere, removing

front/main/app/components/wikia-search.js
dianafa: it doesn't, fixing.

front/main/app/templates/components/application-wrapper.hbs
```diff
@@ -1,30 +1,33 @@
...
-		toggleSideNav=attrs.toggleSideNav
-		toggleUserMenu=attrs.toggleUserMenu
-		toggleSiteHeadPinned=attrs.toggleSiteHeadPinned
+		toggleDrawer=this.attrs.toggleDrawer
 		themeBar=themeBar
+		wikiaHomepage=wikiaHomepage
+		drawerContent=activeDrawerContent
+		drawerVisible=drawerVisible
+		setDrawerContent=(action 'setDrawerContent')
+		drawerVisible=drawerVisible
```
dianafa: `drawerVisible=drawerVisible` 2 times :D

front/main/app/components/wikia-nav.js
```diff
@@ -43,6 +62,10 @@ export default Ember.Component.extend(
 					label
 				});
 			}
+		},
+
+		scrollTop() {
```
RafLeszczynski: consider moving this logic to one of component lifecycle hooks in `wikia-drawer` component

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,161 @@
...
+	},
+
+	hubsLinks: Ember.get(Mercury, 'wiki.navigation2016.hubsLinks'),
+	localLinks: Ember.get(Mercury, 'wiki.navigation2016.localNav'),
+	exploreWikiaLinks: Ember.get(Mercury, 'wiki.navigation2016.exploreWikiaMenu'),
+	exploreWikiaLabel: Ember.get(Mercury, 'wiki.navigation2016.exploreWikia.textEscaped'),
+	wikiName: Ember.get(Mercury, 'wiki.siteName'),
+	wikiLang: Ember.get(Mercury, 'wiki.language.content'),
+
+	parent: Ember.computed('state.[]', 'localLinks', function () {
```
RafLeszczynski: please consider changing names of local variables to make it clear that you are traversing the tree

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,161 @@
...
+		const s = this.get('state');
+		let nav = this.get('localLinks'),
+			parent, item;
+
+		if (s.length && s[0] !== 0) {
```
RafLeszczynski: use computed property for the conditional logic

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,161 @@
...
+					href: item.href,
+					name: item.textEscaped,
+					trackLabel: `open-${item.trackingLabel}`
+				};
+			}) || [];
+	}),
+
+	globalItems: Ember.computed('inRoot', 'wikiLang', 'hubsLinks', function () {
+		return this.get('inRoot') &&
+			this.get('wikiLang') === 'en' &&
```
RafLeszczynski: consider moving conditional to computed property

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,161 @@
...
+	}),
+
+	randomPageItem: Ember.computed('inRoot', function () {
+		return this.get('inRoot') &&
+			[{
+				type: 'side-nav-menu-item',
+				href: '#',
+				name: i18n.t('app.random-page-label'),
+				trackLabel: 'random-page',
+				clickHandler: 'loadRandomArticle'
```
RafLeszczynski: could we avoid hardcoding action name inside model?

front/main/app/components/discussion-header.js
idradm: I believe it's not, however, I would like to get it overridden. So we don't have default class applied.

front/main/app/models/wikia-nav.js
idradm: Acceptance criteria

front/main/app/templates/components/side-nav-menu-item.hbs
idradm: ``{{#link-to link href}}`` fails if href is null or undefined

dianafa: 👍

front/main/app/styles/_wikia-variables.scss
```diff
@@ -97,6 +97,7 @@ $smart-banner-android-height: rem-calc(66);
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // Site head
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+$site-head-height: rem-calc(50);
```
vforge: Please update this too: https://github.com/Wikia/mercury/pull/2395/files#diff-8fb07e7ac8ce3ce8809239087827cb5aR4
(and possibly other places)

front/main/app/components/wikia-drawer.js
```diff
@@ -9,6 +9,11 @@ export default Ember.Component.extend(
 
 		homeOfFandomLabel: Ember.get(Mercury, 'wiki.navigation2016.fandomLabel'),
 		isFandomVisible: Ember.computed(() => Mercury.wiki.language.content === 'en'),
+		scrollMenuToTop: Ember.observer('shouldBeVisible', function() {
```
RafLeszczynski: what do you thing about moving this to one of the component lifecycle hooks. They should be fired ones `shouldBeVisible` is changed. That way you won't need to use observer which is passe :-)

front/main/app/components/site-head.js
```diff
@@ -1,53 +1,58 @@
...
-			/**
-			 * @returns {void}
-			 */
-			expandSideNav() {
+const {computed, Component} = Ember;
+
+export default Component.extend({
+	classNames: ['site-head-wrapper'],
+	classNameBindings: ['themeBar'],
+	tagName: 'div',
```
RafLeszczynski: why did we change `nav` to `div` ?

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,72 @@
...
+
+			goBack() {
+				this.get('model').goBack();
+			},
+
+			goToSubNav(index) {
+				this.get('model').goToSubNav(index);
+			},
+
+			trackClick(category, label) {
```
RafLeszczynski: missing JSDocs

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+			for (const i of s) {
+				// local nav indexes are shifted by 1,
+				// 0 is reserved for exploration nav
+				node = localNav[i - 1];
+				// check if nav branch
+				if (node && node.children) {
+					parent = node;
+					localNav = node.children;
+				} else {
+					throw new Error('Incorrect navigation state');
```
RafLeszczynski: just wander if we for sure want to `throw` error as it will break whole JS execution and affect user interaction. Maybe we should just log error and provide user with some friendly output

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+
+	goRoot() {
+		this.set('state.[]', []);
+	},
+
+	goBack() {
+		this.get('state').popObject();
+	},
+
+	goToSubNav(index) {
```
RafLeszczynski: missing JS Doc

front/main/app/templates/components/wikia-nav.hbs
```diff
@@ -0,0 +1,25 @@
+{{#if model.inRoot}}
+	{{#if isUserAuthenticated}}
+		<div class="side-nav-menu__item main">
```
RafLeszczynski: `side-nav-menu__item` seams like a good candidate for a separate component. What do you think about it?

front/main/package.json
```diff
@@ -50,6 +50,7 @@
     "ember-in-viewport": "2.0.4",
     "ember-load-initializers": "0.5.1",
     "ember-onbeforeunload": "0.2.0",
+    "ember-polyfill-for-tests": "0.2.0",
```
RafLeszczynski: do we need this polyfill

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,72 @@
+import Ember from 'ember';
+import LoginLinkMixin from '../mixins/login-link';
+import WikiaNavModel from '../models/wikia-nav';
+import {track, trackActions} from 'common/utils/track';
+
+// noinspection JSUnresolvedFunction
```
rogatty: What's that for?

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,72 @@
...
+export default Ember.Component.extend(
+	LoginLinkMixin,
+	{
+		init() {
```
rogatty: Can we keep the unwritten convention of ordering properties in Ember objects? Hooks are usually placed right above actions.

front/main/app/components/wikia-nav.js
```diff
@@ -0,0 +1,72 @@
...
+				title: this.get('currentUser.name')
+			});
+		}),
+
+		actions: {
+			/**
+			 * Handles link items click, runs handler provided in item object
+			 * additionally to tracking and menu reset
+			 * @param {Object} item side menu item data
+			 * @returns {undefined}
```
rogatty: `@returns {void}`

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
+import Ember from 'ember';
+
+const {computed, get} = Ember;
```
rogatty: What about `Object` and `A`?

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+
+export default Ember.Object.extend({
+	init() {
```
rogatty: Same as https://github.com/Wikia/mercury/pull/2395/files#r62021406

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+	exploreWikiaLabel: get(Mercury, 'wiki.navigation2016.exploreWikia.textEscaped'),
+	wikiName: get(Mercury, 'wiki.siteName'),
+	wikiLang: get(Mercury, 'wiki.language.content'),
+
+	/**
+	 * Iteratively traverse local navigation tree to find out root node
+	 * of current nav state
+	 * @returns {Object} parent
+	 */
+	parent: computed('state.[]', 'localLinks', function () {
```
rogatty: When I read `parent` I think about parent of the current item / level / whatever. From what I see this property keeps the current state, right? I would call it `current` (or `currentSomething`) to be less confusing.

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+		const s = this.get('state');
```
rogatty: Please use a full name for readibility and let minifier do its job for optimization.

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+		let localNav = this.get('localLinks'),
+			parent, node;
+
+		if (!this.get('inExploreNav')) {
+			for (let i = 0; i < s.length; i++) {
+				// local nav indexes are shifted by 1,
+				// 0 is reserved for exploration nav
```
rogatty: This seems fragile. What if some day we will have to add another exception? Will we reserve 9999?

front/main/app/models/wikia-nav.js
rogatty: +1
I would just log it here and few lines below we will return an empty object.

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+					localNav = node.children;
+				} else {
+					throw new Error('Incorrect navigation state');
+				}
+			}
+		}
+		return parent || {};
+	}),
+
+	currentLocalLinks: computed.or('parent.children', 'localLinks'),
```
rogatty: Can you add a comment explaining when this fallback to `localLinks` might happen? Same for the `header` below.

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+	header: computed.or('parent.text', 'exploreWikiaLabel'),
+
+	inExploreNav: computed('state.[]', function () {
+		const s = this.get('state');
+
+		return s.length && s[0] === 0;
+	}),
+
+	inSubNav: computed('parent.children', function () {
+		return Boolean(this.get('parent.children') && this.get('parent.children').length);
```
rogatty: `inSubNav: computed.bool('parent.children.length')`

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+					href: item.href,
+					name: item.textEscaped,
+					trackLabel: `open-hub-${item.specialAttr}`
+				};
+			}) || [];
+	}),
+
+	exploreSubMenuItem: computed('inRoot', 'exploreWikiaLinks', function () {
+		return this.get('inRoot') &&
+			this.get('exploreWikiaLinks').length &&
```
rogatty: You can use `this.get('exploreWikiaLinks.length')` which is safer - doesn't break if `exploreWikiaLinks` is undefined.

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+				trackLabel: 'local-nav'
+			}] || [];
+	}),
+
+	localItems: computed('inExploreNav', 'currentLocalLinks', function () {
+		let index = 0;
+
+		return !this.get('inExploreNav') &&
+			this.get('currentLocalLinks').map((item) => {
+				index++;
```
rogatty: That's not necessary, you can use `.map((item, index) => {`
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

front/main/app/models/wikia-nav.js
```diff
@@ -0,0 +1,173 @@
...
+				href: '#',
+				name: i18n.t('app.random-page-label'),
+				trackLabel: 'random-page',
+				actionId: 'onRandomPageClick'
+			}] || [];
+	}),
+
+	init() {
+		this._super(...arguments);
+		this.state = A([]);
```
rogatty: I think that it would be useful if we had a comment here explaining that `this.state` is a stack of indexes that define the current item in `this.items` structure.

front/main/app/components/site-head.js
idradm: I've added div wrapper which is a nav placeholder on top of the page. Nav can be found in site-head.hbs.

front/main/app/components/wikia-nav.js
```diff
@@ -30,6 +30,11 @@ export default Ember.Component.extend(
 			};
 		},
 
+		didRender() {
+			this._super(...arguments);
+			this.element.scrollTop = 0;
```
RafLeszczynski: do we know why we need to do it manually?

front/main/app/models/wikia-nav.js
dianafa: it's because we count items from 1, not from 0

rogatty: Well, you can use `index + 0` then instead of introducing additional variable :)

PR #2396, opened 2016-04-22
remove special treatment for catchpoint

hakubo:
## Description

We stopped using catchpoint - so we do not need that part here

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

bognix:
LGTM

PR #2399, opened 2016-04-23
Ember-cli-blanket

dianafa:
## Links

* https://github.com/sglanzer/ember-cli-blanket

## Description

This PR introduces code coverage tool for ember cli apps. Blog post will go live after this PR will get merged.

## Reviewers

@rogatty @hakubo @tomasznapieralski


front/main/testem.json
```diff
@@ -1,6 +1,6 @@
 {
   "framework": "qunit",
-  "test_page": "tests/index.html?hidepassed",
+  "test_page": "tests/index.html?hidepassed&coverage=true",
```
hakubo: does that mean it is enabled by default?

I probably don't care that much about code coverage every time I run unit tests

front/main/tests/index.html
```diff
@@ -11,6 +11,7 @@
 	{{content-for 'test-head'}}
 
 	<link rel="stylesheet" href="assets/test-support.css">
+	<style>#blanket-main { position: relative; z-index: 99999; }</style>
```
hakubo: why is this needed?

front/main/testem.json
rogatty: And it's probably much slower than a regular test suit?

front/main/bower.json
```diff
@@ -30,5 +30,8 @@
     "ember": "2.3.2",
     "jquery": "2.1.4",
     "ember-qunit-notifications": "0.1.0"
+  },
+  "devDependencies": {
+    "blanket": "5e94fc30f2e694bb5c3718ddcbf60d467f4b4d26"
```
rogatty: Is this required? We already have `ember-cli-blanket` in `package.json` which has `blanket` in its `bower.json` (https://github.com/sglanzer/ember-cli-blanket/blob/master/bower.json).

If yes, why in that form?

front/main/tests/index.html
rogatty: https://github.com/sglanzer/ember-cli-blanket/blob/ae0a482e89ce8977ef85cf865962706488e11481/blueprints/ember-cli-blanket/index.js#L33

front/main/testem.json
dianafa: it's a bit slower (~20s vs ~19s measured by `time` tool) indeed. However, if it would be enabled by default, each developer would see how the tests he writes impact general test coverage and stability of our platform. Results look really nice and IMO it's interesting to see what's missing - if it would be turned off by default, nobody would bother to enable it.
What I mean here is that this tool can be used not only for showing numbers for people interested in it, but also for increasing awareness & test writing culture :)

front/main/bower.json
dianafa: This is required, it's automatically added by `ember-cli-blanket` here: https://github.com/sglanzer/ember-cli-blanket/blob/ae0a482e89ce8977ef85cf865962706488e11481/blueprints/ember-cli-blanket/index.js#L9
probably to be sure that we're using version of blanket which is always in sync with ember-cli-blanket.

front/main/testem.json
rogatty: Coverage is measured only when you run tests in the browser so `time` doesn't make much sense here. I've "measured" the time with stopwatch and it was ~2.5s and you can always disable "Enable coverage" checkbox. I would leave it as it is.

rogatty:
LGTM

hakubo:
LGTM

PR #2400, opened 2016-04-25
Make nav bar always visible and sticky

idradm:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4126

### DIFF
--- a/front/main/app/components/discussion-post-editor.js
+++ b/front/main/app/components/discussion-post-editor.js
@@ -36,7 +36,7 @@ export default DiscussionEditorComponent.extend({
 	 * @returns {boolean}
 	 */
 	isStickyBreakpointHeight() {
-		return window.pageYOffset >= this.get('offsetTop') - (this.get('siteHeadPinned') ? this.get('siteHeadHeight') : 0);
+		return window.pageYOffset >= this.get('offsetTop') - this.get('siteHeadHeight');
 	},
 
 	/**
--- a/front/main/app/components/site-head.js
+++ b/front/main/app/components/site-head.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import {track, trackActions} from 'common/utils/track';
 
 export default Ember.Component.extend({
-	classNames: ['site-head', 'border-theme-color'],
+	classNames: ['site-head'],
 	classNameBindings: ['themeBar'],
 	tagName: 'nav',
 	themeBar: false,
--- a/front/main/app/components/site-head.js
+++ b/front/main/app/components/site-head.js
@@ -1,66 +1,64 @@
 import Ember from 'ember';
-import HeadroomMixin from '../mixins/headroom';
 import {track, trackActions} from 'common/utils/track';
 
 export default Ember.Component.extend({
-		classNames: ['site-head', 'border-theme-color'],
-		classNameBindings: ['themeBar'],
-		tagName: 'nav',
-		themeBar: false,
-		wikiaHomepage: Ember.getWithDefault(Mercury, 'wiki.homepage', 'http://www.wikia.com'),
+	classNames: ['site-head', 'border-theme-color'],
+	classNameBindings: ['themeBar'],
+	tagName: 'nav',
+	themeBar: false,
+	wikiaHomepage: Ember.getWithDefault(Mercury, 'wiki.homepage', 'http://www.wikia.com'),
 
-		currentUser: Ember.inject.service(),
-		newFeaturesBadges: Ember.inject.service(),
-		isUserAuthenticated: Ember.computed.oneWay('currentUser.isAuthenticated'),
-		shouldDisplayNewBadge: Ember.computed('newFeaturesBadges.features.[]', function () {
-			return this.get('newFeaturesBadges').shouldDisplay('recent-wiki-activity');
-		}),
+	currentUser: Ember.inject.service(),
+	newFeaturesBadges: Ember.inject.service(),
+	isUserAuthenticated: Ember.computed.oneWay('currentUser.isAuthenticated'),
+	shouldDisplayNewBadge: Ember.computed('newFeaturesBadges.features.[]', function () {
+		return this.get('newFeaturesBadges').shouldDisplay('recent-wiki-activity');
+	}),
 
-		actions: {
-			/**
-			 * @returns {void}
-			 */
-			expandSideNav() {
-				if (this.get('shouldDisplayNewBadge')) {
-					track({
-						action: trackActions.click,
-						category: 'recent-wiki-activity-blue-dot',
-						label: 'open-navigation'
-					});
-				}
-				track({
-					action: trackActions.click,
-					category: 'side-nav',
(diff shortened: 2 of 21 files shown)


front/main/app/components/discussion-post-editor.js
```diff
@@ -36,7 +36,7 @@ export default DiscussionEditorComponent.extend({
 	 * @returns {boolean}
 	 */
 	isStickyBreakpointHeight() {
```
RafLeszczynski: seams like a good candidate for computed property. @bkoval, @rafalkalinski, @kvas-damian was there any objections  for making this computed property?

RafLeszczynski:
LGTM

RafLeszczynski:
@dianafa please take a second look at these changes

front/main/app/mixins/theme.js
```diff
@@ -69,7 +69,6 @@ export default Ember.Mixin.create(ColorUtilsMixin, {
 			return;
 		}
 
-		styles += `.discussions .border-theme-color {border-color: ${this.get('themeColors.color-buttons')};}`;
```
rafalkalinski: Why is this line removed?

front/main/app/components/discussion-post-editor.js
kvas-damian: `window.pageYOffset` isn't observed by Ember, that's why we decided to not use `Ember.computed`. Probably it could be handled that way but I'm not sure if it's more readable  than current code.

front/main/app/mixins/theme.js
kvas-damian: Why this rule was removed?

front/main/app/styles/component/_discussion-editor.scss
```diff
@@ -85,10 +85,6 @@ $editor-textarea-overflow-ios-hack: 30px;
 			transition: top 0.5s;
 			width: $sticky-editor-width;
 			z-index: 1;
-
-			&.pinned-top {
```
kvas-damian: Please make sure removing this block doesn't break discussion editor UI.

front/main/app/styles/component/_discussion-header.scss
```diff
@@ -10,11 +10,10 @@ $site-header-height: 50px;
 	position: fixed;
 	top: 0;
 	width: 100%;
-	z-index: $z-8; // drop-downs from top-bar component (which is $z-7) display over the discussion header
+	z-index: $z-6; // drop-downs from top-bar component (which is $z-7) display over the discussion header
```
kvas-damian: Is this comment still up-to-date?

front/main/app/templates/components/discussion-editor.hbs
```diff
@@ -1,5 +1,5 @@
 <label class="editor-label">{{i18n labelText ns='discussion'}}</label>
-<div class="editor-container {{if isSticky 'sticky'}} {{unless siteHeadPinned pinnedClassName}}">
+<div class="editor-container {{if isSticky 'sticky'}} {{pinnedClassName}}">
```
rafalkalinski: If siteHeadPinned is now always true, then beeing consistent with previous condition, pinnedClassName should never be here anymore.

idradm: `siteHeadPinned` is always true, however, this code was using headroom functionality to turn on/off class for two components `discussion-post-editor` and `discussion-reply-editor`.

This could be removed for the first component, although would break the second one, as it's used here: https://github.com/Wikia/mercury/blob/92e7a559eb7f199f7be5e130a59ee6ef69135336/front/main/app/components/discussion-reply-editor.js#L6

front/main/app/styles/component/_discussion-header.scss
idradm: Except the `display over...` part, will remove it.

front/main/app/styles/component/_discussion-editor.scss
idradm: This css was used to move editor to top on scroll down when the bar was hidden. Which is no longer a case.

front/main/app/mixins/theme.js
idradm: We will be adding `fandom bar` bellow the nav, therefore, it was removed. Additionally, it collided with `smart-banner`.

kvas-damian: This class wasn't only used on nav. See: https://github.com/Wikia/mercury/blob/04255056fa7ac165346b2be41d53dd61c02e7032/front/main/app/templates/components/discussion-sort.hbs#L2-L2

kvas-damian: :+1:

front/main/app/components/discussion-post-editor.js
RafLeszczynski: let's leave it as it is for now but this is something we could consider changing it the future

front/main/app/styles/component/_site-head.scss
```diff
@@ -44,19 +44,8 @@ $site-head-height: 50; //unitless
...
-
-.headroom-not-top {
-	&.pinned {
-		transform: none;
-	}
-
-	&.un-pinned {
-		transform: translateY(-100%);
-	}
+.application-wrapper {
```
RafLeszczynski: shouldn't this be put inside application-wrapper styles?

RafLeszczynski:
LGTM if no more remarks from Social Team. @kvas-damian @rafalkalinski are you ok with that?

front/main/app/components/site-head.js
```diff
@@ -1,73 +1,63 @@
...
+		 */
+		trackWordmarkClick() {
+			track({
+				action: trackActions.click,
+				category: 'wordmark'
+			});
+		}
+	},
+
+	didRender() {
```
bkoval: How about a small, ebautiful jsdoc for this method? :)

RafLeszczynski: this is ember component lifecycle hook

kvas-damian:
It LGTM although please check if it doesn't break discussion functionality ;)

front/main/app/styles/component/_site-head.scss
idradm: This should happen only if `site-head` is enabled, that's the reason it's here.

front/main/app/components/site-head.js
bkoval: What a method above has :)

RafLeszczynski:
@kvas-damian we will do regression testing

RafLeszczynski:
@idradm please address remaining comments and move to QA

bkoval:
LGTM!

front/main/app/mixins/theme.js
rafalkalinski: Sounds good,but it probably should be removed from here: https://github.com/Wikia/mercury/pull/2400/files#diff-3c1465af792c139083130ff66c2d96aeR5

PR #2409, opened 2016-04-27
DAT-4115 navigation menu refactoring

idradm:
## Links

* http://wikia-inc.atlassian.net/browse/JIRA-DAT-4115
* http://wikia-inc.atlassian.net/browse/JIRA-DAT-4118

## Description

Navigation menu refactoring.

## Reviewers

@Wikia/west-wing


front/main/app/components/wikia-nav.js
```diff
@@ -1,48 +1,182 @@
...
-	defaultNavContent: 'side-nav-global-navigation-root',
+export default Ember.Component.extend(
+	LoginLinkMixin,
+	{
+		hubsLinks: Ember.get(Mercury, 'wiki.navigation2016.hubsLinks'),
+		localLinks: Ember.get(Mercury, 'wiki.navigation2016.localNav'),
+		exploreWikiaLinks: Ember.get(Mercury, 'wiki.navigation2016.exploreWikiaMenu'),
+		exploreWikiaLabel: Ember.get(Mercury, 'wiki.navigation2016.exploreWikia.textEscaped'),
+		wikiName: Ember.get(Mercury, 'wiki.siteName'),
+		state: [],
```
dianafa: `state` and all non-primitive types should be declared via `init() {this._super(...arguments); HERE}`. Otherwise they'll share them among all instances of wikia-nav component, more info here: https://wikia-inc.atlassian.net/wiki/pages/viewpage.action?pageId=44007507

front/main/app/components/wikia-nav.js
```diff
@@ -1,48 +1,182 @@
...
+			if (s.length > 0 && !s[0]) {
+				// special case for exploration nav
+				return this.prepareExplorationItems();
+			}
+			// subNav, show only local
+			if (s.length > 0) {
+				return this.prepareLocalItems(s);
+			}
+			// if state == []
+			return this.prepareGlobalItems().concat(this.prepareLocalItems([]));
```
dianafa: can we have tests for this?

also, wouldn't be nice to check:

1. if `!s.length` then `return this.prepareGlobalItems().concat(this.prepareLocalItems([]));`
2. if `s[0]` then `return this.prepareLocalItems(s);`
3. `return this.prepareExplorationItems();` ?

front/main/app/components/wikia-nav.js
```diff
@@ -1,48 +1,182 @@
...
+		}),
```
dianafa: let's add \n between properties for more readable code

front/main/app/components/wikia-nav.js
```diff
@@ -1,48 +1,182 @@
...
+				header = this.get('exploreWikiaLabel');
+
+			for (const i of this.get('state')) {
+				if (nav[i - 1]) {
+					header = nav[i - 1].text;
+				}
+				nav = nav[i - 1] ? nav[i - 1].children : nav;
+			}
+			return header;
+		}),
```
dianafa: property above looks like a nice candidate for tests as well

front/main/app/components/wikia-nav.js
```diff
@@ -1,48 +1,182 @@
...
+			}]);
+		},
+
+		prepareLocalItems(state) {
+			let nav = this.get('localLinks'),
+				index = 0,
+				local;
+
+			if (state.length > 0) {
+				// look for correct subnav
```
dianafa: wouldn't this work without this check?

front/main/app/components/wikia-nav.js
idradm: 2. is not true, it should be if s[0] > 0. 
I'm still thinking about refactoring those preparation functions to some separate class or model, what do you think?

front/main/app/components/wikia-nav.js
idradm: Probably :P, I need tests

PR #2423, opened 2016-05-04
XW-1449 check if displayTitle exists before calling it	

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1449
* https://github.com/Wikia/mercury/pull/2424

## Description

We are calling `indexOf` on `displayTitle` and `title` when they are not set.
This case happens when category page has no article and on main pages

## Reviewers
@Wikia/x-wing


rogatty:
LGTM

vforge:
LGTM

PR #2427, opened 2016-05-06
XW-1124 | Bump eslint version and fix all errors

vforge:
/cc @Wikia/x-wing


front/common/baseline/tracker-universal-analytics.js
```diff
@@ -433,7 +433,7 @@ if (typeof window.M.tracker === 'undefined') {
 		initAccount(accountPrimary, domain);
 		initAccount(accountAds, domain);
 
-		if (Boolean(M.prop('isGASpecialWiki') || Mercury.wiki.isGASpecialWiki)) {
+		if (M.prop('isGASpecialWiki') || Mercury.wiki.isGASpecialWiki) {
```
hakubo: why do we need to change that?

server/app/facets/auth/birthdate-input.js
```diff
@@ -101,6 +101,7 @@ export default class BirthdateInput {
 				return [data.month, data.day, data.year];
 			case 'little':
 				// covered by default block
+				// eslint-disable no-fallthrough
```
hakubo: instead of working around 
do we have to keep case 'little' - is it more readable?

bognix: agree with @hakubo. I think we can achieve the same level of readability by adding comment to default

.eslintrc
```diff
@@ -194,7 +194,7 @@
     "no-var": 2,
     "object-shorthand": 2,
     "prefer-arrow-callback": 2,
-    "prefer-const": 2,
+    "prefer-const": 0,
```
bognix: why we are disabling this rule?

front/common/baseline/tracker-universal-analytics.js
vforge: 1) Conditional casts everything to `boolean` anyway.
2) "no-extra-boolean-cast" - this rule told us this.

.eslintrc
vforge: Talked offline

bognix:
LGTM

rogatty:
LGTM

PR #2428, opened 2016-05-06
XW-1306 | send articleId instead of title to internal tracker

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1306

## Reviewers

@Wikia/x-wing


server/app/facets/mediawiki-page.js
```diff
@@ -105,6 +105,7 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
 	}
 
 	result.mediaWikiNamespace = ns;
+	result.articleId = pageData.details ? pageData.details.id : 0;
```
bognix: I'm not sure if fallback to 0 is better than `undefined`.
We send a lot of `undefined`s to Internal when we track A/B testing.

server/app/views/_partials/server-data.hbs
```diff
@@ -12,6 +12,7 @@
 	M.prop('userId', {{userId}});
 	M.prop('displayTitle', '{{displayTitle}}');
 	M.prop('urlTitleParam', '{{urlTitleParam}}');
```
bognix: I wanted to remove urlTitleParam but unfortunately it's used in `tracking-ua-auth` and `tracking-ua`.

server/app/facets/mediawiki-page.js
drsnyder: @ameyamahajan what is your suggestion here regarding what the default should be for the article id?

ameyamahajan: Ideal we should have a default article id in wikicities / 'page' table, but we don't have anything. It would be safe to keep these article_id = 0 which would be treated as default article_id. 
Also, I verified this by spot checking on various Wikias' page table
```
[wikicities]>select * from page where page_id =0; 
Empty set (0.00 sec)
[muppet]>select * from page where page_id =0; 
Empty set (0.01 sec)
[runescape]>select * from page where page_id =0; 
Empty set (0.00 sec)
[starwars]>select * from page where page_id =0;
Empty set (0.00 sec)

```
Also, we don't have any other article id for title like 'default' 
```
[starwars]>select * from page where page_title like '% default %' limit 10 \G ;
Empty set (0.37 sec)
[runescape]>select * from page where page_title like '% default %' limit 10 \G ;
Empty set (0.42 sec)
[wikicities]>select * from page where page_title like '% default %' limit 10 \G ;
Empty set (0.17 sec)
```

hakubo: Thanks!

hakubo:
Summary
- articles/categories will send an ID as an 'a' parameter - actual change
- 'sub pages' of an curated main page will send ID of a main page as an 'a' parameter
- auth pages will send path as an 'a' parameter - stays as it was

@Wikia/x-wing review please

rogatty:
LGTM

PR #2431, opened 2016-05-10
SOC-1938 - As a user, I want to see URL OG cards added to posts and replies on the discussion post details page and on the post list page

rafalkalinski:
## Links

https://wikia-inc.atlassian.net/browse/SOC-1938

## Description

As a user, I want to see URL OG cards added to posts and replies on the discussion post details page and on the post list page

## Reviewers

@bkoval @kvas-damian


front/main/app/models/discussion/domain/open-graph.js
```diff
@@ -0,0 +1,47 @@
...
+
+OpenGraph.reopenClass({
+	/**
+	 * @param {Object} openGraphData
+	 *
+	 * @returns {array}
+	 */
+	create(openGraphData) {
+		if (!openGraphData) {
+			return this._super();
```
kvas-damian: Does it make sense to create empty object when data if empty? What do you think about just returning `null` here?

front/main/app/templates/components/discussion-open-graph.hbs
```diff
@@ -0,0 +1,10 @@
+<a class="og-container" href="{{openGraphData.url}}" title="{{openGraphData.title}}">
+	<img class="og-image" src="{{openGraphData.imageUrl}}" alt="{{openGraphData.title}}">
```
kvas-damian: Don't we need some condition if image isn't present?

front/main/app/styles/component/_discussion-open-graph.scss
```diff
@@ -0,0 +1,71 @@
+
+// -- theme --
+
+.og-container {
```
kvas-damian: Those selectors are very wide, it's possible that they can match classnames in other features. What do you think about wrapping them into a class name that relates to discussions only?

front/main/app/styles/component/_discussion-open-graph.scss
```diff
@@ -0,0 +1,71 @@
...
+	background-color: $discussion-open-graph-background-color;
```
kvas-damian: Do we use the same colors on dark theme?

front/main/app/components/discussion-open-graph.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	siteName: Ember.computed('openGraphData.domain', 'openGraphData.siteName', function () {
```
bkoval: Do we want to show the domain twice if there's no siteName?

front/main/app/models/discussion/domain/open-graph.js
bkoval: I'd rather return an object with default values :)

front/main/app/styles/component/_discussion-open-graph.scss
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.og-site-name {
+	font-size: $type-min;
+	line-height: $line-height-min;
+	margin: 0 18px 22px;
+}
+
+.og-image {
+	flex-grow: 2;
```
bkoval: Why 2? I don't see any other flexible elements.

front/main/app/templates/components/discussion-open-graph.hbs
```diff
@@ -0,0 +1,10 @@
+<a class="og-container" href="{{openGraphData.url}}" title="{{openGraphData.title}}">
```
bkoval: Do we need an additional div on top of that? Maybe we could move this to a component definition and remove it from here?

bkoval:
Thanks for unit tests?

front/main/app/models/discussion/domain/open-graph.js
rafalkalinski: It creates the object with default values.

front/main/app/templates/components/discussion-open-graph.hbs
rafalkalinski: SOC-1940 covers that (different OG card views depends on the data from server)

front/main/app/models/discussion/domain/open-graph.js
kvas-damian: I don't get it.
Why do we use default values when OpenGraph doesn't exist for the post?

front/main/app/styles/component/_discussion-open-graph.scss
rafalkalinski: +done

front/main/app/styles/component/_discussion-open-graph.scss
rafalkalinski: I planned to make the dark theme work as a part of SOC-1940, when there will be the final shape of those styles.

front/main/app/components/discussion-open-graph.js
rafalkalinski: So we want to show there the title and the siteName only. If there is no siteName, then we take the domain as a siteName.

front/main/app/styles/component/_discussion-open-graph.scss
rafalkalinski: removed

front/main/app/templates/components/discussion-open-graph.hbs
rafalkalinski: Actually we don't need that. It is changed as a part of SOC-1940 - there will be difference in this behavior between post-list (no link) and post-details (link).

front/main/app/models/discussion/domain/open-graph.js
bkoval: Yeah, that's true. Doesn't make sense.

PR #2434, opened 2016-05-12
XW-1480 | Add new header to mobile article preview

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1480

## Description

Redesign of the top of the article implementation, header to be more precise, changes how the header + infobox is handled. This fixes the issue when no hero image and title was shown.

## Reviewers

@Wikia/x-wing @Wikia/west-wing

### DIFF
--- a/front/main/app/controllers/article-preview.js
+++ b/front/main/app/controllers/article-preview.js
@@ -1,16 +1,6 @@
 import Ember from 'ember';
-export default Ember.Controller.extend({
+import HeroImageMixin from '../mixins/hero-image';
+
+export default Ember.Controller.extend(HeroImageMixin, {
 	application: Ember.inject.controller(),
-
-	heroImage: Ember.computed('model.media', function () {
-		let heroImage;
-
-		this.get('model.media.media').forEach((current) => {
-			if (current.hasOwnProperty('context') && current.context === 'infobox-hero-image') {
-				heroImage = current;
-			}
-		});
-
-		return heroImage;
-	})
 });
--- a/front/main/app/controllers/article-preview.js
+++ b/front/main/app/controllers/article-preview.js
@@ -1,6 +1,4 @@
 import Ember from 'ember';
-import HeroImageMixin from '../mixins/hero-image';
+import PortableInfoboxHeroImageMixin from '../mixins/portable-infobox-hero-image';
 
-export default Ember.Controller.extend(HeroImageMixin, {
-	application: Ember.inject.controller(),
-});
+export default Ember.Controller.extend(PortableInfoboxHeroImageMixin, {});
--- a/front/main/app/controllers/article-preview.js
+++ b/front/main/app/controllers/article-preview.js
@@ -1,0 +1,16 @@
+import Ember from 'ember';
+export default Ember.Controller.extend({
+	application: Ember.inject.controller(),
+
+	heroImage: Ember.computed('model.media', function () {
+		let heroImage;
+
+		this.get('model.media.media').forEach((current) => {
+			if (current.hasOwnProperty('context') && current.context === 'infobox-hero-image') {
+				heroImage = current;
+			}
+		});
+
+		return heroImage;
+	})
+});
--- a/front/main/app/mixins/hero-image.js
+++ b/front/main/app/mixins/hero-image.js
@@ -1,0 +1,17 @@
+import Ember from 'ember';
+
+export default Ember.Mixin.create({
+	heroImage: Ember.computed('model.media', function () {
+		const media = this.get('model.media.media');
+
+		if (Ember.isArray(media)) {
+			for (let i = 0; i < media.length; i++) {
+				if (media[i].hasOwnProperty('context') && media[i].context === 'infobox-hero-image') {
+					return media[i];
+				}
+			}
+		}
+
+		return null;
+	})
+});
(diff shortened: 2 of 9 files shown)


front/main/app/controllers/article-preview.js
```diff
@@ -0,0 +1,16 @@
+import Ember from 'ember';
+export default Ember.Controller.extend({
+	application: Ember.inject.controller(),
+
+	heroImage: Ember.computed('model.media', function () {
```
Warkot: What do you think is the best way to DRY this code? Because this is copied'n'pasted from here https://github.com/Wikia/mercury/blob/ea9276d9e55e755d0d8b102ed87a407e67ec539a/front/main/app/components/article-wrapper.js#L106.

front/main/app/controllers/article-preview.js
```diff
@@ -0,0 +1,16 @@
...
+		let heroImage;
+
+		this.get('model.media.media').forEach((current) => {
+			if (current.hasOwnProperty('context') && current.context === 'infobox-hero-image') {
+				heroImage = current;
```
dianafa: Maybe it's a micro-optimisation but we actually don't have to iterate over media array from the moment we found first hero image element.

Warkot: I was thinking about it but 
> There's no built-in ability to `break` in `forEach`. To interrupt execution you would have to throw an exception of some sort. eg.

http://stackoverflow.com/questions/2641347/how-to-short-circuit-array-foreach-like-calling-break

😿

Warkot: Something like this would work:
```javascript
heroImage: Ember.computed('model.media', function () {
	const media = this.get('model.media.media');

	if (Ember.isArray(media)) {
		for (let i = 0; i < media.length; i++) {
			if (media[i].hasOwnProperty('context') && media[i].context === 'infobox-hero-image') {
				return media[i];
			}
		}
	}

	return null;
})
```
What do you think?

Warkot: Okay, I did it here https://github.com/Wikia/mercury/pull/2434/commits/8136683e3ef19e1a61e30bfd9e4b3ab1f4f8dca4. 😼

front/main/app/mixins/hero-image.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
+
+export default Ember.Mixin.create({
+	heroImage: Ember.computed('model.media', function () {
+		const media = this.get('model.media.media');
+
+		if (Ember.isArray(media)) {
+			for (let i = 0; i < media.length; i++) {
+				if (media[i].hasOwnProperty('context') && media[i].context === 'infobox-hero-image') {
```
dianafa: hmm if we iterate over media array, `media[i]` will always exist, thus we don't need to check here if hasOwnProp - in the worst case `media[i].context` just will be undefined, what do you think?

front/main/app/mixins/hero-image.js
```diff
@@ -0,0 +1,17 @@
...
+					return media[i];
+				}
+			}
+		}
+
+		return null;
```
dianafa: do we have to explicitly return null here?

front/main/app/controllers/article-preview.js
```diff
@@ -0,0 +1,6 @@
+import Ember from 'ember';
+import HeroImageMixin from '../mixins/hero-image';
+
+export default Ember.Controller.extend(HeroImageMixin, {
+	application: Ember.inject.controller(),
```
rogatty: What's that for?

front/main/app/mixins/hero-image.js
Warkot: Agreed. I copy-pasted to much. :+1:

front/main/app/styles/component/wikia-ui-components/_wiki-page-header.scss
```diff
@@ -75,7 +75,7 @@
 }
 
 // component modifier-related changes to elements
-.wiki-page-header.has-action .wiki-page-header__wrapper {
+.wiki-page-header .wiki-page-header__wrapper {
```
rogatty: Why was this removed? We already have rules for `.wiki-page-header__wrapper` above.

front/main/app/mixins/hero-image.js
Warkot: No. It just felt right to me to always return something if I already return other values. I'll drop it.

rogatty: I like `null` better than `undefined` :)

rogatty:
JS LGTM but I'd prefer to let @vforge speak about SCSS.

front/main/app/styles/component/wikia-ui-components/_wiki-page-header.scss
```diff
@@ -9,6 +9,50 @@
...
 	background: $color-blue-light;
 	color: #fff;
+
+	.wiki-page-header__wrapper {
+		display: flex;
+	}
+
+	// component modifier-related changes to elements
+	&.has-hero-image {
+		background-color: #fff;
```
hakubo: background: {
color:
position:
...
}

vforge:
@rogatty Talked offline, LQGTM

PR #2436, opened 2016-05-12
style only the ad-slot not the wrapper

hakubo:
## Description

This will hide ad slot when there is no ad

## Reviewers

@Wikia/x-wing


rogatty:
Should we do the same for `.mobile-in-content, .mobile-prefooter`?

hakubo:
@rogatty I did not want to change that much in this small PR

rogatty:
Fair enough, LGTM

PR #2439, opened 2016-05-12
RECIRC-109: Add lateral tracking script

pauloslund:
## Links
* https://wikia-inc.atlassian.net/browse/RECIRC-109

## Description
Adding in a tracking script for Lateral so that we can get better content recommendations


front/main/app/index.html
```diff
@@ -45,6 +45,7 @@
 	{{> server-data}}
 	{{> tracking-quantserve}}
 	{{> tracking-comscore}}
+	{{> tracking-lateral}}
```
rogatty: @vforge @bognix how should we decide if this goes into `head` or the bottom of `body`? I see that we use both places.

server/app/views/_partials/tracking-lateral.hbs
```diff
@@ -0,0 +1,17 @@
+<script>
+(function() {
+	window.trackLateralPageView = function() {
+		if (!M.prop('queryParams.noexternals')) {
+			(function(i) {
+				var a = document.location.pathname.replace(/\/+$/,"").split("/").pop(),
+					b = RegExp("wikia_beacon_id=([A-Za-z0-9_-]{10})").exec(document.cookie);
```
rogatty: What happens when you run this code on iPhone with disabled cookies? We had some issues with that before.

front/main/app/index.html
vforge: @rogatty As a rule of thumb, If it's very important that we run this as soon as possible AND it doesn't need anything special from Mercury, it can be in the `head`.

server/app/views/_partials/tracking-lateral.hbs
pauloslund: From my testing it fails silently. What were the issues you've encountered in the past?

rogatty: If failing silently is fine then :+1: 
We had some errors thrown but it's most likely a matter of the code around.

rogatty:
LGTM

PR #2440, opened 2016-05-13
XW-1478 | server header matching fronted markup and style

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1478

## Description

One of refinements to be done to the new header.

## Reviewers

@Wikia/x-wing


vforge:
lgtm

PR #2447, opened 2016-05-16
DAT-4045 corectlly handle jap/zh input

idradm:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4045

## Description

Fix for inputs to correctly handle Japanese and Chinese keyboard tools. 
## Reviewers

@Wikia/west-wing


front/main/app/components/wikia-ui-components/text-field.js
```diff
@@ -46,6 +46,12 @@ export default Ember.Component.extend({
 	}),
 	isInvalid: Ember.computed.bool('errorMessage'),
 
+	init() {
+		// input shouldn't be reloaded on change, it's needed for jap/zh os tools
```
rogatty: Can we test that this is always true?

dianafa:
LGTM

PR #2450, opened 2016-05-18
SOC-2425 | Wikia Session - use local URL for Helios

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2425

## Description

So far we used external URL to validate `access_token` which required additional network routing to find the right host. We believe it caused many connection timeouts, that's why we decided to try using local network for session validating.

## Reviewers

@rafalkalinski @bkoval @hakubo


server/config/localSettings.base.js
```diff
@@ -165,7 +165,9 @@ const localSettings = {
 	// Targeted environment [prod|preview|verify|dev|testing]
 	environment: getEnvironment(process.env.WIKIA_ENVIRONMENT),
 	helios: {
-		path: '/auth'
+		localUrl: 'http://prod.helios.service.consul:9500/',
```
rogatty: I believe this should come from chef-repo just like `mediawikiDomain` comes from https://github.com/Wikia/chef-repo/blob/d44af54b33b6bb88f5796e441e7ee5e284652957/cookbooks/mercury/templates/default/sv-mercury-run.erb#L10

kvas-damian: What's the benefit of passing this url via environment variables?
I talked with @pchojnacki and we believe it isn't sensitive data - we already expose. On the other hand distributing config in many places makes it harder to maintain.

rogatty: Yup, I was thinking about the last one. These URLs have history of being changed and it's easier when ops have to worry about that ;)

kvas-damian: I hope we introduced consul service discovery to make it easy to change addresses at consul level. @pchojnacki please correct me if I'm wrong ;)

rafalkalinski:
except the one thing with helios.localUrl (I'm for the solution to have that defined in only one place)
LGTM

bkoval:
+1 for storing the url in chef - this way we don't need a whole Mercury release in case we want to change it in the future. 

+ the benefits of storing all consul urls in one place. Peace.

Other than that LGTM

kvas-damian:
During offline discussion we decided to keep non sensitive data like URLs to service and mediawiki in one place: mercury repo `server/config/localSettings.base.js`

Main advantages:
* Updating one repo makes release procedure easier. We don't need to wait for synchronization.
* it's easier to find and update configuration

//cc @bkoval @rafalkalinski @hakubo

rafalkalinski:
LGTM

bkoval:
LGTM

PR #2451, opened 2016-05-18
DAT-4206 | hide hero image/title even if first infobox fails to be moved to top

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4206

## Description

Using old app code (without moving portable infobox to top of the article) revealed a flaw, that stuff that should be hidden is not in such case. This updated selector fixes that. Still true fix for hiding it is on the way with the cleanup.

## Reviewers

@Wikia/x-wing @nikodamn


vforge:
GJ :+1:

nikodamn:
LGTM

PR #2453, opened 2016-05-18
ADEN-3358 BFAA unit

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3358


front/main/app/styles/component/_site-head.scss
```diff
@@ -6,6 +6,13 @@ $icon-size: 25px;
 	margin-top: $icon-site-head-width;
 }
 
+body.no-scroll {
+	.site-head {
+		top: 0 !important;
+		position: fixed !important;
+	}
+}
```
fraszczakszymon: 🙈

front/main/app/styles/module/_ads.scss
```diff
@@ -18,12 +18,26 @@ $ad-background: rgb(44, 52, 61);
...
 }
 
-.mobile-top-leaderboard .ad-slot {
-	background: $ad-background;
-	min-height: 50px;
-	padding: 10px;
-	position: relative;
-	z-index: $z-ads-top-leaderboard;
+.mobile-top-leaderboard {
+	.ad-slot {
```
rogatty: @hakubo would probably want to add a comment about this one :)
(https://github.com/Wikia/mercury/pull/2436/commits/c25200d75540f483d903ff052c8347ed39849ab9)

fraszczakszymon: Isn't it the same rule?

rogatty: Of course it is, I missed the space there :see_no_evil:

front/main/app/styles/component/_site-head.scss
```diff
@@ -6,6 +6,13 @@ $icon-size: 25px;
 	margin-top: $icon-site-head-width;
 }
 
+body.no-scroll {
+	.site-head {
+		top: 0 !important;
```
hakubo: what options were considered to not use important?

front/main/app/styles/module/_ads.scss
```diff
@@ -18,12 +18,26 @@ $ad-background: rgb(44, 52, 61);
...
+		background: $ad-background;
+		min-height: 50px;
+		padding: 10px;
+		position: relative;
+		z-index: $z-ads-top-leaderboard;
+	}
+
+	&.bfaa-template {
+		position: absolute;
```
hakubo: this is top of the page and position absolute is not the best for performance - any ideas how we could avoid that?

front/main/app/styles/component/_site-head.scss
rogatty: I think this is being done to move the site-head to the proper place when side-nav is opened. Am I right? If yes, then maybe we should hide the ad slot instead?

front/main/app/styles/module/_ads.scss
rogatty: Is a separate ad slot an option? Then we could just put it at the top of body.

fraszczakszymon:
OK, I'm closing it now. We have to do it in different way. Sorry and thanks for your time :)

PR #2454, opened 2016-05-18
SOC-1398 | Discussions - fetch posts from all categoris

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-1398

## Description

* Discussions - fetch posts from all categories
* Move `sort` parameter to query string

## Reviewers

@bkoval @rafalkalinski


front/main/app/mixins/discussion-contribution-model.js
```diff
@@ -11,7 +11,7 @@ export default Ember.Mixin.create({
 	 * @returns {Ember.RSVP.Promise}
 	 */
 	createPost(postData) {
-		return request(M.getDiscussionServiceUrl(`/${this.wikiId}/forums/${this.forumId}/threads`), {
+		return request(M.getDiscussionServiceUrl(`/${this.wikiId}/forums/${this.wikiId}/threads`), {
```
rafalkalinski: Hopefully this is a temporary change, because 'forumId' property name much better fits here, even if the 'wikiId' at the moment contains the same value.

rafalkalinski:
LGTM

front/main/app/mixins/discussion-contribution-model.js
bkoval: Shouldn't we be switching to siteId/threads resource for this?

server/app/routes.js
```diff
@@ -255,6 +255,13 @@ authenticatedRoutes.push({
 	handler: discussionsHandler
 });
 
+authenticatedRoutes.push({
```
bkoval: I think we might have too many separate routes defined for discussions right now (three, and they all point to the same handler)

Can't we just make sure /d* goes to discussionsHandler?

bkoval:
One comment. Other than that LGTM

front/main/app/mixins/discussion-contribution-model.js
kvas-damian: For short period of time we're completely dropping support for `forumId`. We'll introduce it back when we start adding categories into our code. For now we'll use `wikiId` which is in practice equal to what we had in `forumId`.

server/app/routes.js
kvas-damian: Personally I don't like wildcard routes that handles everything. It makes it impossible in future to find all possibilities that are handled by such route.

I slightly changed how we handle redirect and was able to remove changes in `show-discussions.js`

server/app/routes.js
```diff
@@ -242,23 +242,32 @@ mediaWikiPagePaths.forEach((path) => {
...
 
 authenticatedRoutes.push({
-	// Discussion index
+	// Make sure old discussion URLs are redirected to discussion main page
 	method: 'GET',
-	path: '/d',
+	path: '/d/f/{id}',
+	handler(request, reply) {
+		return reply.redirect('/d/f').permanent(true);
+	},
```
rafalkalinski: :+1:

rafalkalinski:
LGTM

bkoval:
LGTM 📦

PR #2458, opened 2016-05-19
DAT-4228 add !important to avoid overriding by other CSS rules

dianafa:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-4228

## Reviewers

@idradm


idradm:
okey

PR #2459, opened 2016-05-20
RECIRC 82: Recirculation placement test

pauloslund:
## Links

* https://wikia-inc.atlassian.net/browse/RECIRC-82

## Description

Similar to the placement test on desktop, we want to isolate each type of content and placement on mobile and understand the performance of each. The data providers and locations will change as we receive results


front/main/app/components/article-content.js
```diff
@@ -4,6 +4,11 @@ import {getRenderComponentFor, queryPlaceholders} from '../utils/render-componen
 import {track, trackActions} from 'common/utils/track';
 import {getGroup, inGroup} from 'common/modules/abtest';
 
+import RecirculationFooterComponent from './recirculation/footer';
+import RecirculationIncontentComponent from './recirculation/incontent';
```
rogatty: Please use `this.createComponentInstance(componentName)` instead as it's a more clean way to create components.

front/main/app/components/article-content.js
```diff
@@ -521,6 +527,73 @@ export default Ember.Component.extend(
...
+				group = getGroup(experimentName);
+
+			let view, component, model, location,
+				externalLink = false;
+
+			switch (group) {
+				case 'CONTROL':
+					component = RecirculationFooterComponent;
+					model = FandomPostsModel.create();
+					location = this.parentView.$('.article-footer');
```
rogatty: `parentView` is a [private method](https://github.com/emberjs/ember.js/blob/v2.5.0/packages/ember-views/lib/views/core_view.js#L69) so it's risky to use. Usually, when we need to access DOM element from outside of the current scope then we just use `Ember.$('.article-footer')`

front/main/app/components/article-content.js
```diff
@@ -521,6 +527,73 @@ export default Ember.Component.extend(
...
+				view = this.createChildView(component, {
+					model,
+					experimentName,
+					externalLink
+				});
+				view.createElement();
+				model.load();
+
+				location.after(view.$());
+				view.trackImpression();
```
rogatty: Shouldn't impression be tracked when component is actually in the viewport? You could use https://github.com/dockyard/ember-in-viewport for that.

front/main/app/templates/components/recirculation/footer.hbs
```diff
@@ -0,0 +1,23 @@
+{{#if items.length}}
+<section class="content-recommendations">
+	<h2>{{model.title}}</h2>
+	<ul class="thumbnails">
+		{{#each items as |item|}}
+			<li class="title-thumbnail" {{action 'trackExperimentClick' item.url}}>
+				<a href="{{if externalLink '' item.url}}">
+					<div class="image image-thumbnail">
+						{{#if item.thumbnail }}
+							<img src="{{item.thumbnail}}" alt="{{item.title}}">
```
rogatty: Can this be lazy loaded? Right now when user opens an article and Ember gets up we load following images all at once:
- images visible in the viewport (usually max one), always thumbnailed to have max width of the viewport
- avatars under the article (doesn't make sense, I've created XW-1512 to fix that)
- 5 Fandom images in the footer, each one 825px wide and currently of 417 kilobytes weight in total

It clogs up the bandwidth and slows down the rest of page unnecessarily.

front/main/app/components/article-content.js
pauloslund: Since we want to test what placement is the most effective, we want to track impressions if the element is rendered - even if the user never actually sees it.

pauloslund:
Closing in favor of https://github.com/Wikia/mercury/pull/2540

PR #2466, opened 2016-05-20
SOC-2495 | Tracking for OG Cards

bkoval:
Links

http://wikia-inc.atlassian.net/browse/SOC-2495
Reviewers

@rafalkalinski @kvas-damian


front/main/app/components/discussion-editor.js
```diff
@@ -508,6 +510,8 @@ export default Ember.Component.extend(ViewportMixin, {
 				showsOpenGraphCard: false,
 				openGraph: null,
 			});
+
+			track(trackActions.OGPreviewClosed);
```
kvas-damian: What do you think about changing this label to something that suggest OG was removed?
From my perspective `closed` doesn't properly describe the action that was performed.

bkoval: Yeah, I was torn apart with this one. From one point of view this might not indicate directly that the associated OG Card was removed. However, we all know how the app works, and the action which is done by the user is "closing" after the click on the close button. However, I have no preference and I guess we can find out what will be more convenient for our product owner.

WHat would you suggest? OGPreviewRemoved? OGRemoved?

kvas-damian: As discussed offline, we think `OGRemoved` is label that suites us most. Let's talk with Product if it's acceptable.

kvas-damian:
LGTM

PR #2476, opened 2016-05-25
XW-1510 | Introduce ember-cli-mirage and add first acceptance test

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1510
* https://guides.emberjs.com/v2.5.0/testing/acceptance/
* http://www.ember-cli-mirage.com/

## Description

It's time to start writing Ember acceptance tests. This PR:
* introduces ember-cli-mirage which allows to easily mock API responses
* adds first, simple acceptance test for Curated Content Editor
* fixes tests which started to fail in this new setup

## Reviewers

@Wikia/x-wing


tomasznapieralski:
good job, it looks pretty promising.
I have one question: Can we move object you created in fixtures to model (curated-content-editor-item.js)?
Because it looks like for each model we will have to create 2 files fixtures with object and nearly empty model

rogatty:
@tomasznapieralski don't forget about nearly empty serializer ;) That's how mirage is built, citing http://www.ember-cli-mirage.com/docs/v0.2.0-beta.9/:
>Mirage borrows concepts from typical backend systems like
>* routes
>* a database and models, for storing data and defining relationships
>* factories and fixtures, for stubbing data
>* serializers, for formatting responses

We could use a factory instead of fixture but IMO it doesn't make too much sense in this case with our not-at-all-standard data format.

hakubo:
lgtm

front/main/mirage/fixtures/curated-content-editor-item.js
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable max-len */
```
bognix: let's disable the rule only for the line that breaks the rule. Not for the all file

rogatty: I'm not sure if it makes sense in a fixture. I wonder if we could change it to json file and don't care about eslint at all.

PR #2477, opened 2016-05-25
XW | don't use class defined in curated content for non curated content

tomasznapieralski:
## Reviewers

@Wikia/x-wing


tomasznapieralski:
tested

Warkot:
LGTM

PR #2378, opened 2016-04-20
XW-904 | Replace $.ajax with ember-ajax

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-904
* https://github.com/ember-cli/ember-ajax
* https://wikia-inc.atlassian.net/wiki/display/MER/2016/04/22/ember-ajax+-+an+ajax+lib+that+also+works+with+FastBoot

## Description

This PR prepares us to use Ember Fastboot, but before we get there we need to have a way to call APIs with the same code both on client and server side. Ember ajax has a basic support for that and they are committed to improve that part of the lib.

## Reviewers

@kvas-damian @bkoval @Wikia/x-wing

PR #2478, opened 2016-05-26
XW | Fix image-review model

rogatty:
## Description

Image review stopped working after https://github.com/Wikia/mercury/pull/2378 was merged. It was caused by the fact that [ember-ajax/request](https://github.com/ember-cli/ember-ajax/blob/master/addon/request.js) resolves with payload only. I've changed it to [ember-ajax/raw](https://github.com/ember-cli/ember-ajax/blob/master/addon/raw.js) which resolves with `jqXHR`.

## Reviewers

@Wikia/x-wing @RafalWilinski


RafalWilinski:
Thanks @rogatty, LGTM!

vforge:
lgtm

PR #2482, opened 2016-05-29
DAT-4323 Search - no results page

dianafa:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-4323

## Reviewers

@Wikia/west-wing


front/main/app/controllers/search.js
```diff
@@ -1,15 +1,27 @@
...
+		},
+
+		onErrorPageClick() {
+			this.set('query', '');
+			this.focusSearchInput();
 		}
+	},
+
+	focusSearchInput() {
+		$('.side-search__input').focus();
```
dianafa: yes, having the same function in two places (here and in wikia-search component) is not DRY.
But, other option is passing property between controller and above mentioned component, and observing for change of this property on the controller side. Then, resetting it after focusing the input. Also, observer will not fire if this prop won't be used anywhere, so we would need to artificially fire this. That's why I decided for this solution. Happy to hear other ideas :)

front/main/app/controllers/search.js
```diff
@@ -1,15 +1,27 @@
 import Ember from 'ember';
 
-const {Controller, inject} = Ember;
+const {Controller, computed, inject, $} = Ember;
 
 export default Controller.extend({
 	application: inject.controller(),
 	queryParams: ['query'],
+	errorPageQueryMarkup: computed('model.{erroneousQuery,error}', function () {
+		return `<span class="${this.get('model.error')}__query">${this.get('model.erroneousQuery')}</span>`;
```
dianafa: query key has to be both part of the translated string and has to be styled differently than rest of this string

front/main/app/controllers/search.js
idradm: Add a comment in the code, pls.

front/main/app/templates/search.hbs
```diff
@@ -24,4 +23,14 @@
 			{{/wikia-ui-components/wikia-card}}
 		{{/each}}
 	</ul>
+	{{else if model.error}}
+		{{#wikia-ui-components/error-info
+			class=model.error
+		}}
+			{{{i18n (concat 'main.' model.error) ns='search' query=errorPageQueryMarkup}}}
```
idradm: Please do not concat translation strings. It will cause issue in maintaining them later on.

front/main/app/controllers/search.js
idradm: Let's make each error a separate template and load depending on their type instead.

front/common/public/locales/en/search.json
```diff
@@ -1,6 +1,10 @@
 {
   "main": {
-    "total-results": "About __totalItems__ results for ",
-    "searchInputLabel": "Search"
+    "search-error-general": "Can't connect, please try again",
+    "search-error-general-action": "Click to retry",
+    "search-error-not-found": "We couldn't find anything for __query__",
+    "search-error-not-found-action": "Try Another Search",
+    "search-input-label": "Search",
+    "total-results": "About __totalItems__ results for "
```
idradm: Let's add a __query__ here instead of concatenating later.

front/main/app/styles/component/wikia-ui-components/_error-info.scss
```diff
@@ -0,0 +1,18 @@
...
+$error-info-text-bubble-margin-vertical: 60px;
+
+.error-info {
+	font-size: $type-smallest;
+	line-height: $line-height-smallest;
+	padding: 0 $error-info-side-padding;
+}
+
+.error-info__text-bubble {
+	margin: $error-info-text-bubble-margin-vertical auto;
```
SebastianMarzjan: Nitpick: sort the properties please

front/main/app/components/wikia-search-error-not-found.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+
+const {Component, computed} = Ember;
+
+export default Component.extend({
+	classNames: ['search-error-not-found'],
+	errorPageStringMarkup: computed('erroneousQuery', function () {
+		return `<span class="search-error-not-found__query">${this.get('erroneousQuery')}</span>`;
```
SebastianMarzjan: Would it be possible to render this using a template? I know it requires a few more lines of code, but I strongly recommend that we keep the markup separate and use the logic only to generate the class/modifier

front/main/app/helpers/wrap-me.js
```diff
@@ -0,0 +1,38 @@
...
+
+	if (options.tagName) {
+		tagName = options.tagName;
+	}
+
+	if (options.className) {
+		className = ` class="${options.className}"`;
+	}
+
+	return new Handlebars.SafeString(`<${tagName}${className}>${name}</${tagName}>`).toHTML();
```
rogatty: Cool, can you add some tests too?

dianafa: done

front/main/app/helpers/wrap-me.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {Array} params
+ * @param {Object} options
+ * @returns {string}
+ */
+
+const {Handlebars, Helper} = Ember;
+
+export default Helper.helper((params, options) => {
+	const content = params[0] || '';
+	let tagName = 'div',
```
idradm: ``span`` seems more safe option as a default

PR #2487, opened 2016-05-30
DAT-4236 | Fix site head markup

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-4236

## Description

Unify pre-ember site head markup with the final one. Fixes the issue with article header's position.
Also removes the empty ad-slot placeholder because page jumps anyway and it looked bad.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/server/app/views/_layouts/error.hbs
+++ b/server/app/views/_layouts/error.hbs
@@ -19,19 +19,55 @@
 				flex-direction: column;
 			}
 
+			.site-head-container {
+				min-height: 3.4375rem;
+			}
+
+			.site-head-wrapper {
+				height: 3.4375rem;
+			}
+
 			.site-head {
-				border-bottom: 1px solid #E6EBF2;
-				display: block;
-				flex-shrink: 0;
-				padding: 14px 0 7px;
-			}
-
-			.site-head div {
+				background-color: #FFF;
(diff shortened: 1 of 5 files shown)


server/app/views/_layouts/error.hbs
```diff
@@ -19,19 +19,55 @@
...
+				min-height: 3.4375rem;
+			}
+
+			.site-head-wrapper {
+				height: 3.4375rem;
 			}
 
-			.site-head div {
+			.site-head {
+				background-color: #FFF;
```
Warkot: Are these styles copied from somewhere? If yes, they should just share a selector.

vforge: Also: we use prefixer, don't we?

And those look like copy/paste from inspector.

rogatty: This layout is used only on error pages and doesn't include `app.css` so we inline only required styles.

Yes, this is copy/pasted from inspector and no, we don't use prefixer for styles inlined in hbs files.

Warkot: Makes absolute sense! 👍

vforge:
LGTM

Warkot:
LGTM

PR #2494, opened 2016-06-01
DAT-4368 add a way to disable suggestions, disable them on srp

idradm:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4368

### DIFF
--- a/front/main/app/components/wikia-search.js
+++ b/front/main/app/components/wikia-search.js
@@ -27,7 +27,8 @@ export default Component.extend(NoScrollMixin,
 		// in ms
 		debounceDuration: 250,
 		inputFocused: false,
-		// Whether or not to display the loading search suggestion results message (en: 'Loading...')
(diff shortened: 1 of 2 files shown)


front/main/app/components/wikia-search.js
```diff
@@ -114,14 +116,19 @@ export default Component.extend(NoScrollMixin,
...
 			const query = this.get('query');
 
 			this.setProperties({
 				suggestions: [],
 				searchRequestInProgress: false
 			});
 
-			// If the query string is empty or shorter than the minimal length, return to leave the view blank
+			// If the query string is empty or shorter than the minimal length, return to leave the
+			// view blank
```
rogatty: Why would you break it like that? It's now harder to read ("the" at the end of line :confused:?) and our line length limit is 120.

dianafa:
LGTM

PR #2498, opened 2016-06-01
Introduce template linter

rogatty:
## Links
* https://wikia-inc.atlassian.net/browse/XW-1234
* https://github.com/rwjblue/ember-cli-template-lint
* https://github.com/rwjblue/ember-template-lint/#rules

## Description

Introduce linting for our templates using ember-cli-template-lint. Enable only a single rule - `'block-indentation': 'tab'` - to keep PR readable while still providing value and base for introducing further rules.

Linter will be run on each `ember build` and `ember test` and display errors in console. It also means that our PR check "Front tests" will fail when rules are not followed.

## Reviewers

@Wikia/x-wing @dianafa @bkoval @kamilkoterba


front/main/app/templates/components/portable-infobox.hbs
```diff
@@ -1,6 +1,6 @@
 {{{infoboxHTML}}}
 {{#if isLongInfobox}}
-<div class={{expandButtonClass}} {{action "toogleInfobox"}}>
-	<svg viewBox="0 0 12 7" class="icon pi-expand-icon"><use xlink:href="#chevron"></use></svg>
-</div>
+	<div class={{expandButtonClass}} {{action "toogleInfobox"}}>
+		<svg viewBox="0 0 12 7" class="icon pi-expand-icon"><use xlink:href="#chevron"></use></svg>
```
idradm: Shouldn't this be indented as well?

rogatty: Isn't it already? If you ask if we should put `<use>` in a seperate line then no, it seems like as long as everything is in a single line then this is fine. We can't do this, though:
```
<svg><use>
</use></svg>
```

front/main/ember-cli-build.js
```diff
@@ -74,7 +74,6 @@ module.exports = function (defaults) {
 			// By default vendor is not watched by Ember CLI and we want to rebuild when common scripts are modified
 			vendor: 'vendor'
 		},
-		hinting: false,
```
bkoval: Just out of curiosity - what did that do? :)

kvas-damian:
:+1:

front/main/ember-cli-build.js
rogatty: Yeah, it was required before to disable JSHint. At some point it got obsolete and now is not needed. Also you can take a look at https://github.com/rwjblue/ember-cli-template-lint/issues/118

hakubo:
LGTM

PR #2499, opened 2016-06-02
SOC-1400 | Discussion Categories

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-1400

## Reviewers

@kvas-damian @rafalkalinski


bkoval:
Fixing tests in progress :)

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+
+		visibleCategoriesCount: null,
+
+		init() {
+			this._super();
+			this.setLocalCategories();
+		},
+
+		disabledObserver: Ember.observer('disabled', function () {
+			if (this.get('disabled', true)) {
```
kvas-damian: What do you think about adding default value for this flag and remove default value from getter?

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+
+				this.sendAction('updateCategories', localCategories);
+			},
+
+			/**
+			 * @param {boolean} shouldTrack
+			 *
+			 * @returns {void}
+			 */
+			onAllCategoryClick(shouldTrack) {
```
kvas-damian: This action is used not only when `allCategory` is clicked. Probably we should have better name for this action.

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+			 * @param {Object} localCategory
+			 * @param {Event} event
+			 *
+			 * @returns {void}
+			 */
+			onCategoryClick(localCategory, event) {
+				const localCategories = this.get('localCategories');
+
+				this.trackCategory(false);
+				event.preventDefault();
```
kvas-damian: This looks like a hack, please dig for a moment and try to find why this action is triggered twice.

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+
+				Ember.set(localCategory, 'selected', !localCategory.selected);
```
kvas-damian: Why can't we simply use `localCategory.set()`?

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+
+				this.setAllCategorySelected(localCategories);
+
+				this.sendAction('updateCategories', this.get('localCategories'));
```
kvas-damian: we already have `localCategories` as variable, can we use it?

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+		 * @returns {void}
+		 */
+		setLocalCategories() {
+			const categories = this.get('categories'),
+				localCategories = new Ember.A();
+
+			categories.forEach((category) => {
+				localCategories.pushObject({
+					category,
+					selected: category.selected
```
kvas-damian: we already have `selected` attribute at `category` level. Can we think of a solution that doesn't have this information duplicated?

front/main/app/models/discussion.js
```diff
@@ -0,0 +1,108 @@
...
+const DiscussionModel = Ember.Object.extend({
+	categories: new Ember.A(),
+	data: null,
+	wikiId: null,
+
+	selectedCategoryIds: Ember.computed('categories.@each.selected', function () {
+		return this.getSelectedCategoryIds();
+	}),
+
+	isAllCategories: Ember.computed('categories.@each.selected', function () {
```
kvas-damian: What does `isAllCategories` mean?

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+				});
+			});
+
+			this.setProperties({
+				localCategories,
+				allCategorySelected: this.get('isAllCategories'),
```
kvas-damian: While we need those values separated what do you think about renaming it to let other know that this value is local?

front/main/app/components/discussion-categories-filter.js
```diff
@@ -0,0 +1,112 @@
...
+		/**
+		 * @param {Ember.Array} localCategories
+		 *
+		 * @returns {void}
+		 */
+		setAllCategorySelected(localCategories) {
+			const isNothingSelected = localCategories.isEvery('selected', false),
+				allCategorySelected = this.get('allCategorySelected');
+
+			if (!allCategorySelected && isNothingSelected) {
```
kvas-damian: Why do we need both checks in this `if` statement.

front/main/app/controllers/discussion/base.js
```diff
@@ -19,7 +19,21 @@ export default Ember.Controller.extend(
 			 *
 			 * @returns {void}
 			 */
-			applyFilters() {}
+			applyFilters() {},
+
+			/**
+			 * This method will be overwritten in the controller, but needs to be stubbed for the error subcontroller
```
kvas-damian: It suggests me that we're doing something when we need to do something like this to handle error state correctly.

front/main/app/components/discussion-filters.js
```diff
@@ -5,22 +5,72 @@ import {track, trackActions} from '../utils/discussion-tracker';
...
-		}),
-
 		trendingDisabled: Ember.computed('onlyReported', function () {
 			return this.get('onlyReported') === true ? 'disabled' : false;
 		}),
 
+		/**
+		 * @returns {void}
+		 */
+		init() {
```
kvas-damian: Please remove empty `init`

front/main/app/components/discussion-filters.js
```diff
@@ -5,22 +5,72 @@ import {track, trackActions} from '../utils/discussion-tracker';
...
+		/**
+		 * @param {string} sortBy
+		 *
+		 * @returns {void}
+		 */
+		trackSortByTapped(sortBy) {
+			const discussionSort = this.get('discussionSort');
+
+			if (sortBy !== discussionSort.get('sortBy')) {
+				if (sortBy === 'latest') {
```
kvas-damian: What do you think about using ternary operator?

front/main/app/models/discussion.js
```diff
@@ -0,0 +1,108 @@
...
+	 * @param {Ember.Array} changedCategories
+	 *
+	 * @returns {void}
+	 */
+	updateCategoriesFromFilters(changedCategories) {
+		if (!changedCategories || !changedCategories.length) {
+			return;
+		}
+
+		changedCategories.forEach((changedCategory) => {
```
kvas-damian: Please add a comment that we're updating local `categories` by reference.

front/main/app/models/discussion.js
```diff
@@ -0,0 +1,108 @@
...
+			const discussionInstance = DiscussionModel.create({
+				wikiId
+			});
+
+			request(M.getDiscussionServiceUrl(`/${wikiId}/forums`)).then((data) => {
+				discussionInstance.setNormalizedData(data);
+
+				resolve(discussionInstance);
+			}).catch(() => {
+				// TODO handle errors
```
kvas-damian: Can we handle this error somehow? ;)
It looks for me that for displaying categories selector we can just ignore error silently.

front/main/app/models/discussion/domain/category.js
```diff
@@ -0,0 +1,27 @@
...
+	/**
+	 * Creates a permissions dict from API's permissions array
+	 *
+	 * @param {object} categoryData
+	 *
+	 * @returns {Ember.Object}
+	 */
+	create(categoryData) {
+		const category = {
+			collapsed: false,
```
kvas-damian: While this attribute is only used in component, what do you think about moving it there?

front/main/app/routes/discussion/forum.js
```diff
@@ -49,14 +64,43 @@ export default DiscussionBaseRoute.extend(
...
-				this.modelFor(this.get('routeName')).loadPage(pageNum, this.get('discussionSort.sortBy'));
+				this.modelFor(this.get('routeName')).current.loadPage(pageNum, this.get('discussionSort.sortBy'));
+			},
+
+			updateCategories(updatedCategory) {
+				const model = this.modelFor(this.get('routeName'));
+
+				model.index.updateCategorySelected(updatedCategory);
+
+				const catId = model.index.getSelectedCategoryIds();
```
kvas-damian: there can be multiple catIds, what do you think about `catIds`?

rogatty: Even better - what do you think about `categoryIds`? ;)

front/main/app/templates/components/discussion-categories.hbs
```diff
@@ -0,0 +1,37 @@
...
+						}}
+					</li>
+				{{/unless}}
+			{{/each}}
+		</ul>
+		{{#if toggleButtonVisible}}
+			<button {{action 'toggleMore'}} class="secondary active-element-border-theme-color active-element-theme-color active-element-hover-theme-color active-element-hover-border-theme-color">{{toggleButtonLabel}}</button>
+		{{/if}}
+	</div>
+</fieldset>
```
kvas-damian: Do we need two separate files for almost the same template?

front/main/app/models/discussion.js
bkoval: It means whether or not application shows posts from all categories or not :)

kvas-damian: Shouldn't it be `isAllCategoriesSelected`?

kvas-damian:
As we agreed off-line. Let's refactor categories component in a way that doesn't involve model in changing which category is selected. Use actions only for making transitions, update model on transition.

front/main/app/components/discussion-filters.js
bkoval: We assume there might be more ways to sort posts in the future, that's why we don't want to default to one of these two when tracking whoch one was applied.

front/main/app/controllers/discussion/base.js
bkoval: That's correct, these methods need to be created for error states as Ember needs function declarations for all actions used like action=(action 'action') in hbs.

front/main/app/components/discussion-filters.js
kvas-damian: Ping?

front/main/app/controllers/discussion/base.js
kvas-damian: As we discussed offline - please create a ticket

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,172 @@
...
+					category,
+					collapsed: false,
+					selected: category.selected
+				}));
+			});
+
+			return localCategories;
+		}),
+
+		allCategorySelected: Ember.computed.oneWay('isAllCategoriesSelected'),
```
kvas-damian: Do we need this property to come from model?

front/main/app/routes/discussion/forum.js
```diff
@@ -49,14 +60,40 @@ export default DiscussionBaseRoute.extend(
...
+
+				const catId = updatedCategories.filterBy('selected', true).mapBy('category.id');
+
+				this.transitionTo({queryParams: {
+					catId,
+					sort: this.get('discussionSort.sortBy')
+				}});
+			},
+
+			resetCategories() {
```
kvas-damian: Do we need separate methods for updating and resetting categories? can't we just pass `[]` to `updateCategories()`?

front/main/tests/integration/components/discussion-categories-test.js
```diff
@@ -0,0 +1,69 @@
...
+
+
+test('should display up to 10 categories by default', function (assert) {
+	this.set('categories', getCategories(20));
+
+	this.render(hbs`{{discussion-categories categories=categories}}`);
+
+	// 11 = 10 categories + All
+	assert.equal(this.$('li').length, 11);
+});
```
kvas-damian: While we refactored the component a lot, is it possible to restore tests that we had before for this component?

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,172 @@
...
+				const localCategories = this.get('localCategories');
+
+				this.trackCategory(false);
+				event.preventDefault();
+
+				localCategory.set('selected', !localCategory.get('selected'));
+
+				this.setAllCategorySelected(localCategories);
+
+				this.sendAction('updateCategories', this.get('localCategories'));
```
kvas-damian: Is it possible to use local variable here?

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,172 @@
...
+		visibleCategoriesCount: null,
+
+		init() {
+			this._super();
+			this.collapseCategoriesAboveLimit(this.get('localCategories'));
+		},
+
+		disabledObserver: Ember.observer('disabled', function () {
+			if (this.get('disabled') === true) {
+				this.send('selectAllCategory', false);
```
kvas-damian: This actions send another action up. It looks that it's not required when we agreed that we update model only if change is submitted.

bkoval: This si correct. It will send action up to discussion-filters to inform this component that there might've been changes to categories, but model will find out only after applying filters.

front/main/tests/integration/components/discussion-categories-test.js
bkoval: Done.

front/main/app/components/discussion-categories.js
bkoval: Moved to a component.

kvas-damian:
Good Job @bkoval !
LGTM :octocat:

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,187 @@
...
+			}
+		}),
+
+		isAllCategoriesSelected: Ember.computed('categories.@each.selected', function () {
+			return this.get('categories').isEvery('selected', false);
```
rafalkalinski: Maybe I just don't get it, but... shouldn't it be `true` here?

rafalkalinski: Oh... I can see... If none of the categories is not selected, then it means the 'all-categories' is, right?

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,187 @@
...
+		}),
+
+		localCategories: Ember.computed('categories.@each.selected', function () {
```
rafalkalinski: localCategories? form reading this code it is not easy to tell what this method actually does. A word of documentation would be appreciated here.

front/main/app/components/discussion-categories.js
rafalkalinski: And why we can't compute it here instead of using isAllCategoriesSelected property?

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,187 @@
...
+				this.set('allCategorySelected', true);
+			} else if (allCategorySelected && !isNothingSelected) {
+				this.set('allCategorySelected', false);
+			}
+		},
+
+		collapseCategoriesAboveLimit(localCategories) {
+			const visibleCategoriesCount = this.get('visibleCategoriesCount');
+
+			if (typeof visibleCategoriesCount === 'number') {
```
rafalkalinski: More general comment:
If this is not an input from a user, then that kind of checks are good only for hiding our own mistakes (and we want to know about our mistakes)

rafalkalinski: And if this is because visibleCategoriesCount  may be `null`, then lets make it `0` by default

front/main/app/components/discussion-categories.js
```diff
@@ -0,0 +1,187 @@
...
+			 * @param {Object} localCategory
+			 * @param {Event} event
+			 *
+			 * @returns {void}
+			 */
+			onCategoryClick(localCategory, event) {
+				const localCategories = this.get('localCategories');
+
+				this.trackCategory(false);
+				event.preventDefault();
```
rafalkalinski: I'd like to see some comment here, which could help me understand why we're doing that.

front/main/app/models/discussion/domain/category.js
```diff
@@ -0,0 +1,26 @@
+import Ember from 'ember';
+
+const DiscussionCategory = Ember.Object.extend();
```
rafalkalinski: We have that pattern for defining all the properties here.

front/main/app/templates/components/discussion-categories.hbs
```diff
@@ -0,0 +1,15 @@
+{{discussion-categories-list
```
rafalkalinski: I can't see why we need this in separate component

front/main/app/components/discussion-categories.js
bkoval: Yes :)

front/main/app/templates/components/discussion-categories.hbs
bkoval: Talked offline, this is a deliberate separation between a view and logical container.

front/main/app/components/discussion-categories.js
bkoval: We could, but this way it's cleaner IMO.

front/main/app/components/discussion-categories.js
bkoval: As we spoke offline, we will not introduce any changes to this code, but we will strive to find patterns that will ensure any programmers errors will be caught / thrown but not fail silently.

rafalkalinski:
LGTM

PR #2504, opened 2016-06-03
DAT-4370

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4370
* more relevant links

## Description

hiding keyboard on transition between article and search page

## Reviewers

@Wikia/west-wing

### DIFF
--- a/front/main/app/controllers/search.js
+++ b/front/main/app/controllers/search.js
@@ -12,6 +12,7 @@ export default Controller.extend({
 
 	actions: {
 		onSearchEnter(query) {
+			this.blurSearchInput();
 			this.set('query', query);
 			this.get('model').search(query);
 		},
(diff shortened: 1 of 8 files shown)


dianafa:
as I mentioned offline, we can think about moving `this.$('.side-search__input')` as a property and only referencing to it when focusing or bluring.

Rest LGTM

front/main/app/controllers/search.js
```diff
@@ -5,33 +5,21 @@ const {Controller, computed, inject, $} = Ember;
...
 		},
 
 		onErrorPageClick() {
 			this.set('query', '');
-			this.focusSearchInput();
+			this.get('inputField').focus();
 		}
-	},
-
-	/**
```
SebastianMarzjan: I think this comment, while only regarding the action of caching the element, still holds true :)

SebastianMarzjan:
One remark re: removed comment, rest LGTM

front/main/app/controllers/search.js
rybmat: done

SebastianMarzjan:
LGTM

PR #2505, opened 2016-06-03
ADEN-3399 BFAA on mobile

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3399

## Description

Big Fancy Ad Above from Universal Ad Package. Changes:
* possibility to load BFAA unit and place it above headroom (using existing top-leaderboard slot)
* start all ad slots as hidden by default and show them on success (when ad is fetched)

## Reviewers

@Wikia/x-wing @Wikia/adeng


rogatty:
LGTM

nandy-andy:
LGTM :+1:

front/main/app/components/site-head.js
```diff
@@ -13,6 +13,8 @@ export default Component.extend(
 		themeBar: false,
 		closeIcon: 'close',
 
+		adsState: Ember.inject.service(),
```
nandy-andy: @fraszczakszymon: feel free to tell Ad Engineers more about it on Monday's parking lot :japanese_ogre:

front/common/modules/ads.js
```diff
@@ -279,8 +279,19 @@ class Ads {
 	}
 
 	/**
+	 * This method is called on each transition
+	 *
+	 * @returns {void}
+	 */
+	onTransition() {
+		if (this.adMercuryListenerModule && this.adMercuryListenerModule.runOnPageChangeCallbacks) {
```
nandy-andy: I wonder if `adMercuryListenerModule` can be `undefined` in some real scenario. Basically, it's being set once the ads are loaded from MW which means a user can open an app and quickly do a transition to another page, right? If yes, then we probably need this "make-up" here :crying_cat_face:

front/main/app/routes/application.js
```diff
@@ -27,6 +27,9 @@ export default Route.extend(
 			}
 		},
 
+		adsInstance: null,
```
rogatty: What do you think about moving this to `adsState` and use only the service from here? It seems like we pollute this route too much with ads stuff.

PR #2508, opened 2016-06-03
CE-3703 Refactor Infobox integration with VE

kamilkoterba:
MERGE TOGETHER WITH: https://github.com/Wikia/app/pull/10594
## Links

* http://wikia-inc.atlassian.net/browse/CE-3703

## Description

* Separated requests for data and assets
* Added a service for loading styles and scripts
* Skipped sending request for new infobox (no templateName provided) data as it's always empty object

## Reviewers

@Wikia/spitfires @Wikia/x-wing


kamilkoterba:
Hold on I'm on the failing test!!! :)

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,137 @@
...
+			bodyClass: 'infobox-builder-body-wrapper',
+			type: typeCss
+		}
+	},
+
+	/**
+	 * Loads resource for requested bundle, don't load if already loaded
+	 *
+	 * @param {string} assetsBundleName
+	 * @returns {Ember.RSVP.Promise}
```
Grunny: It also currently returns a bool if the asset has already been loaded.

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,137 @@
...
+ * Adds styles to the DOM
+ * and optionally a class to the body element
+ *
+ * @param {Object} assetsBundle
+ * @param {Object} serverResponse
+ * @returns {void}
+ */
+Css.prototype.setupStyles = function (assetsBundle, serverResponse) {
+	if (assetsBundle.bodyClass) {
+		$('body').addClass(assetsBundle.bodyClass);
```
Grunny: I'm not sure about this extra effect of loading an asset, doesn't seem like the job of the resource loader.

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,137 @@
...
+});
+
+/**
+ * Load CSS assets from MedaWiki
+ * @param {Object} assetsBundle
+ * @param {Object} ajax
+ * @throws {Error}
+ * @returns {Promise}
+ */
+Css.prototype.load = function (assetsBundle, ajax) {
```
Grunny: I think it would be better to group the prototype method definitions with the constructor above, rather than having them separated by the main code to improve readability. What do you think?

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,137 @@
...
+};
+
+
+/**
+ * Load JS assets from path
+ * @param {Object} assetsBundle
+ * @returns {Promise}
+ */
+Js.prototype.load = function (assetsBundle) {
+	return Ember.$.getScript(assetsBundle.path).then(() => {
```
Grunny: You aliases `Ember.$` to just `$`, so can just use that.

front/main/app/services/resource-loader.js
kamilkoterba: OK

front/main/app/services/resource-loader.js
kamilkoterba: Changed so it always returns promise.

front/main/app/routes/infobox-builder.js
```diff
@@ -173,16 +173,15 @@ export default Ember.Route.extend(ConfirmationMixin, {
 	 * @returns {Promise}
 	 */
 	setupEnvironmentAndInfoboxData(templateName) {
-		// TODO CE-3600 extract data and assets into services
 		const promises = {
-			dataAndAssets: this.loadInfoboxDataAndAssets(templateName),
-			ponto: this.loadPonto()
+			data: this.loadInfoboxData(templateName),
+			assets: this.get('resourceLoader').load('portableInfoboxBuilderCss'),
```
rogatty: Shouldn't this be called simply `css`? Ponto is an asset too IMO ;)

front/main/app/services/resource-loader.js
rogatty: How about 
```
const cssLoader = {
  load(assetsBundle, ajax) { ... },
  setupStyles(assetsBundle, serverResponse) { ... },
  appendToHead(html) { ... }
};

const jsLoader = {
  load(assetsBundle) { ... }
};
```
?

front/main/app/services/resource-loader.js
kamilkoterba: Now it returns only promises :)

front/main/app/routes/infobox-builder.js
```diff
@@ -165,6 +166,34 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+			data: {
+				controller: 'PortableInfoboxBuilderController',
+				method: 'getAssets',
+				format: 'json'
+			}
+		}).then((data) => {
+			if (data && data.css) {
+				return resourceLoader.load('portableInfoboxBuilderCss', {type: 'css', paths: data.css});
+			} else {
+				return Ember.RSVP.reject(Error('Invalid assets data was returned from MediaWiki API'));
```
hakubo: here you are inside a promise you can throw here.

front/main/app/routes/infobox-builder.js
```diff
@@ -165,6 +166,34 @@ export default Ember.Route.extend(ConfirmationMixin, {
...
+			}
+		}).then((result) => {
+			this.set('cssLoaded', true);
+			return Ember.RSVP.resolve(result);
```
hakubo: you can just return the result

front/main/app/routes/infobox-builder.js
```diff
@@ -240,16 +272,21 @@ export default Ember.Route.extend(ConfirmationMixin, {
 	 * @param {string} templateName
 	 * @returns {Ember.RSVP.Promise}
 	 */
-	loadInfoboxDataAndAssets(templateName) {
+	loadInfoboxData(templateName) {
+		if (!templateName) {
+			return Promise.resolve().then(() => {
+				return {data: '{}', isNew: true};
```
hakubo: I would break the lines here
also 
object as string?

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,84 @@
...
+
+			$(html).appendTo('head');
+
+			loaded[assetsBundleName] = true;
+			return RSVP.resolve(assetJustAddedStatusName);
+		},
+		js(assetsBundle, assetsBundleName, loaded) {
+			// return RSVP.hash(promises)
+			// TODO support all array items
+			return $.getScript(assetsBundle.paths[0]).then(() => {
```
hakubo: I think script.js does support that usecase, can we check?

front/main/app/routes/infobox-builder.js
```diff
@@ -173,16 +202,19 @@ export default Ember.Route.extend(ConfirmationMixin, {
 	 * @returns {Promise}
 	 */
 	setupEnvironmentAndInfoboxData(templateName) {
-		// TODO CE-3600 extract data and assets into services
+		const resourceLoader = this.get('resourceLoader');
```
hakubo: I think we follow one const per block rule now? but I am not sure now :)

front/main/app/routes/infobox-builder.js
hakubo: also I think you can pass result directly do promise.resolve

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,84 @@
...
+	 * @param {string} assetsBundle Optional bundle config - if not present, config for provided name
+	 *        is fetched from assets (config/resources)
+	 * @returns {Ember.RSVP.Promise}
+	 */
+	load(assetsBundleName, assetsBundle) {
+		if (!assetsBundle) {
+			assetsBundle = assets[assetsBundleName];
+		}
+
+		if (this.get('loaded')[assetsBundleName] === true) {
```
hakubo: ```this.get(`loaded.${assetsBundleName}`)```

front/main/tests/unit/routes/infobox-builder-test.js
```diff
@@ -107,3 +107,53 @@ test('test are evnironment resources not load again', function (assert) {
...
+	route.set('resourceLoader', resourceLoader);
+	route.set('cssLoaded', true);
+	route.set('ajax', {
+		request: () => {
+			return {then: requestSpy};
+		}
+	});
+
+	route.loadCss().then((result) => {
+		assert.ok(!requestSpy.called, 'Request on ajax shouldn\'t be called');
```
hakubo: https://api.qunitjs.com/notOk/

front/main/app/services/resource-loader.js
```diff
@@ -0,0 +1,84 @@
+import Ember from 'ember';
+import assets from './config/resources';
+
+const {$, inject, Service, RSVP} = Ember,
+	assetJustAddedStatusName = 'Styles freshly added to DOM',
+	assetAlreadyLoadedStatusName = 'Asset already loaded';
+
+export default Service.extend({
```
hakubo: can you add a bit of documentation here on how to use that service?

front/main/app/routes/infobox-builder.js
```diff
@@ -5,11 +5,12 @@ import {track, trackActions} from 'common/utils/track';
 
 export default Ember.Route.extend(ConfirmationMixin, {
 	ajax: Ember.inject.service(),
+	resourceLoader: Ember.inject.service(),
 	isIframeContext: Ember.computed(() => {
 		return window.self !== window.top;
```
hakubo: sidenote: this does not have to be computed prop I think - as this won't change in runtime :)

front/main/app/routes/infobox-builder.js
kamilkoterba: "object as string"
...apparently yes: https://github.com/Wikia/app/blob/f07d41c8569238c3daf4561f9ae7861814a48a18/extensions/wikia/PortableInfoboxBuilder/PortableInfoboxBuilderController.class.php#L40-L40

front/main/app/services/resource-loader.js
kamilkoterba: https://api.jquery.com/jquery.getscript/

which one exactly do you mean by script.js?

kamilkoterba: I'll reach you tomorrow to discuss it.

front/main/app/components/infobox-builder-edit-title-modal.js
```diff
@@ -15,18 +15,25 @@ export default Ember.Component.extend(
...
 		}),
 
-		errorMessage: Ember.computed('titleExists', function () {
-			return this.get('titleExists') ?
-				i18n.t('main.title-naming-conflict-error', {
-					ns: 'infobox-builder'
-				}) :
-				'';
+		errorMessage: Ember.computed('titleExists', 'titleInvalid', function () {
+			switch (true) {
```
rogatty: :fearful: 
Please don't obfuscate this simple logic by using code like that.

PR #2513, opened 2016-06-06
DAT-4180 smoother scrolling on ios

dianafa:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4180

## Reviewers

@rogatty @tomasznapieralski

### DIFF
--- a/front/main/app/styles/component/_wikia-nav.scss
+++ b/front/main/app/styles/component/_wikia-nav.scss
@@ -15,6 +15,8 @@ $nav-padding-vertical: 18px;
 	background-color: $color-gray-dark;
 	flex-grow: 1;
 	overflow: auto;
+	// smooth scrolling on ios
+	-webkit-overflow-scrolling: touch;
 }
 
 .wikia-nav__header {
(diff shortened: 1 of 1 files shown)


front/main/app/styles/component/_wikia-nav.scss
```diff
@@ -15,6 +15,8 @@ $nav-padding-vertical: 18px;
 	background-color: $color-gray-dark;
 	flex-grow: 1;
 	overflow: auto;
+	// smooth scrolling on ios
+	-webkit-overflow-scrolling: touch;
```
SebastianMarzjan: It may work, though...
```
This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.
```
( https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling )

dianafa: yes, I'm aware of it, that's why it needs to be thoroughly tested on all supported devices. Moreover, we're using it already in a couple of places in our app: `article-media-gallery.scss`, `media.scss` and `table.scss`. What do you think about testing it anyways?

SebastianMarzjan: As long as it's marked and linked from proper comment, I think it's a reasonable trade-off

vforge:
LGTM

PR #2518, opened 2016-06-06
DAT-4312 | add curated content acceptance test

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4312

## Reviewers

@Wikia/x-wing


front/main/mirage/config.js
```diff
@@ -7,7 +7,26 @@ export default function () {
...
+		if (controller === 'MercuryApi') {
+			if (method === 'getPage') {
+				// Curated Main Page Data
+				if (title === 'Mercury_CC_Wikia') {
+					return schema.curatedContents.first();
+				}
+			}
+
+			if (method === 'getCuratedContentSection') {
+				if (section === 'Categories') {
```
rogatty: Currently there is no need for this check, right? Let's not bloat the file with unnecessary code, it's gonna get big anyway.

front/main/mirage/config.js
```diff
@@ -7,7 +7,26 @@ export default function () {
...
+					return schema.curatedContentSections.first();
+				}
+			}
+		}
+
+		if (controller === 'ArticlesApi' && method === 'getList' && category === 'Articles') {
```
rogatty: `category` param isn't very useful here, we would return the same data mock if it had other value too.

front/main/mirage/fixtures/curated-content-category.js
```diff
@@ -0,0 +1,57 @@
+export default {
+	items: [
```
rogatty: Can you base it off https://github.com/Wikia/mercury/blob/5e4f87761b6c763847dae65a7c54d10fa124a225/front/main/mirage/fixtures/curated-content-editor-item.js?

Then it would be named `curated-content-category-item.js` and look like this:
```
export default [
  {
    id: ...
  }, ...
];
```

Then you can modify serializer to look like https://github.com/Wikia/mercury/blob/5e4f87761b6c763847dae65a7c54d10fa124a225/front/main/mirage/serializers/curated-content-editor-item.js but with `keyForCollection` returning `'items'`. Then you would be able to use `return schema.curatedContentCategoryItems.all();` which is more semantic than hacky `.first()`.

I'd ignore `basepath` as it's not used anyway.

rogatty: You can do the same for `curated-content-section.js` but not for `curated-content.js` which isn't a list.

front/main/mirage/serializers/curated-content.js
```diff
@@ -0,0 +1,7 @@
+import BaseSerializer from './application';
+
+export default BaseSerializer.extend({
+	serialize(response) {
+		return {data: response.attrs.data};
```
rogatty: What would happen if this file didn't exist?

front/main/tests/unit/components/ad-slot-test.js
```diff
@@ -99,7 +100,7 @@ test('behaves correctly depending on noAds value', function (assert) {
 		component.didInsertElement();
 		assert.equal(
 			require('common/modules/ads').default.getInstance().adSlots.length,
-			testCase.expectedLength,
+			testCase.expectedLength + offset,
```
rogatty: I think we shouldn't do that. Instead ads instance should be reset before this test.

front/main/mirage/fixtures/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "max-len": 0
```
hakubo: can we not override eslint rules for fixtures?

front/main/tests/unit/components/ad-slot-test.js
bognix: I agree with Igor. Right now this test isn't a unit test because it operates on some state of an application

front/main/mirage/fixtures/curated-content-category.js
tomasznapieralski: Talked offline, we've decided that we will stay with this solution. To be consistent i will update other tests

front/main/mirage/config.js
tomasznapieralski: i will merge those 2 if blocks, but i will leave check for `section === 'Categories'`, because data i've prepared will work only with this section name

front/main/mirage/config.js
tomasznapieralski: is is usefull (explanation in comment above)

front/main/mirage/serializers/curated-content.js
tomasznapieralski: test will fail, because attrs will be undefined

front/main/mirage/fixtures/.eslintrc
tomasznapieralski: this rule disables warning when you put more than 120 chars in line, but because fixtures also contains url such as: `thumbnail: 'http:\/\/vignette1.wikia.nocookie.net\/mercurycc\/images\/8\/8a\/400x400.png\/revision\/latest\/window-crop\/width\/300\/x-offset\/0\/y-offset\/0\/window-width\/401\/window-height\/400?cb=20150706154016',` this rule should be disabled

front/main/tests/helpers/reset-ads.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+
+export default Ember.Test.registerHelper('resetAds', function(app) {
+	const Ads = require('common/modules/ads')['default'].getInstance();
```
hakubo: .default

front/main/mirage/fixtures/.eslintrc
hakubo: since it's js - you can break lines. and that is something I would propose to do.

bognix:
@hakubo @rogatty - did we agree on something regarding `max-len` rule in fixtures? It blocks this PR so please agree of something and write how we should proceed.

front/main/mirage/serializers/curated-content-editor-item.js
```diff
@@ -1,9 +1,7 @@
 import BaseSerializer from './application';
 
 export default BaseSerializer.extend({
-	// without it keyForCollection isn't called (ember-cli-mirage bug?)
-	embed: true,
-	keyForCollection() {
-		return 'data';
+	serialize(response) {
```
hakubo:
```
serialize({attrs: {data}}) {
  return data;
}
```

front/main/mirage/serializers/curated-content-section.js
```diff
@@ -0,0 +1,7 @@
+import BaseSerializer from './application';
+
+export default BaseSerializer.extend({
+	serialize(response) {
```
hakubo:
```
serialize({attrs: {items}}) {
  return {items};
}
```

front/main/mirage/serializers/curated-content.js
hakubo: this could also be written as:
```
serialize({attrs: {data}}) {
  return {data};
}
```

front/main/mirage/serializers/search.js
```diff
@@ -1,6 +1,7 @@
 import BaseSerializer from './application';
 
 export default BaseSerializer.extend({
-	root: false,
-	embed: true
+	serialize(response) {
```
hakubo:
```
serialize({attrs}) {
  return attrs;
}
```

front/main/tests/helpers/reset-ads.js
```diff
@@ -0,0 +1,5 @@
+import Ember from 'ember';
+
+export default Ember.Test.registerHelper('resetAds', function() {
+	require('common/modules/ads').default.getInstance().removeAllSlots();
```
hakubo: actually can this be written as a 'normal' import?
`import Ads from 'common/modules/ads';`

hakubo:
couple of minor comments - LGTM otherwise

front/main/tests/helpers/reset-ads.js
bognix: I tried that but I couldn't access `Ads` inside helper. Maybe because it's called inside `afterEach` hook and context is different? I didn't dig deep into it, I can check it one more time and go back to you with more sophisticated response if you wish.

PR #2521, opened 2016-06-08
DAT-4227

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4227

## Description
making local nav header in hamburger menu clickable and redirecting to wikia main page after click

## Reviewers
@Wikia/west-wing

### DIFF
--- a/front/main/app/models/wikia-nav.js
+++ b/front/main/app/models/wikia-nav.js
@@ -115,6 +115,7 @@ export default Object.extend({
 			[{
 				type: 'nav-menu-header',
 				route: 'wiki-page',
+				//will generate href="/wiki/"
 				href: '',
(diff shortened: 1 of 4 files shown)


front/main/app/models/wikia-nav.js
```diff
@@ -114,6 +114,8 @@ export default Object.extend({
 			this.get('wikiName') &&
 			[{
 				type: 'nav-menu-header',
+				route: 'wiki-page',
+				href: '',
```
SebastianMarzjan: is it possible to have a real-life href here?

rybmat: what do you mean by "real-life href"?

SebastianMarzjan: One that is actually a relative url

PR #2527, opened 2016-06-08
SOC-2600 | Inline Editor styles for Discussions

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2600

## Description

Styles for inline desktop editor for discussions.

## Reviewers
@bkoval @rafalkalinski


front/main/app/components/discussion-inline-editor.js
```diff
@@ -0,0 +1,122 @@
...
+	/**
+	 * Set right height for editor placeholder when editor gets sticky
+	 *
+	 * @returns {void}
+	 */
+	style: Ember.computed('isSticky', function () {
+		if (this.get('isSticky')) {
+			const editorHeight = this.$('.discussion-inline-editor-floating-container').outerHeight(true);
+
+			return Ember.String.htmlSafe(`height: ${editorHeight}px`);
```
bkoval: Why do we need to escape it?

front/main/app/components/discussion-inline-editor.js
```diff
@@ -0,0 +1,122 @@
...
+	/**
+	 * @returns {void}
+	 */
+	onScroll() {
+		Ember.run.throttle(
+			this,
+			function () {
+				// we can't change it to
+				// this.set('isSticky', !this.get('isSticky') && this.isStickyBreakpointHeight())
+				// because it is important to fire the set method only when it's necessary, because there is observer
```
bkoval: I don't such an observer in the code, I know it's a copied comment but maybe we could update it?

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	&.is-sticky {
+		.discussion-inline-editor-floating-container {
+			border-top: 0;
+			margin-left: ($discussion-page-width - $sticky-editor-width) / 2;
+			position: fixed;
+			top: $site-head-height;
+			width: $sticky-editor-width;
+			z-index: 1;
```
bkoval: Why is that needed? Is editor unable to cover something on the page? If it's needed, then we could define a variable for it (and use one of style-guide values).

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+			border: 0;
+			min-height: 106px;
+			resize: vertical;
+		}
+	}
+
+	&.is-sticky {
+		.discussion-inline-editor-floating-container {
+			border-top: 0;
+			margin-left: ($discussion-page-width - $sticky-editor-width) / 2;
```
bkoval: Can this produce a margin < 0?

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+			position: fixed;
+			top: $site-head-height;
+			width: $sticky-editor-width;
+			z-index: 1;
+		}
+	}
+
+	button {
```
bkoval: What do you think about adding a classname to the submit button and using it in the SCSS? It's possible we'll add some more buttons to the editor in the future and this would prevent accidental inheritance of styles.

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+		background: none;
+		border: 0;
+		font-size: $type-smallest;
+		font-weight: $font-weight-bold;
+		height: 40px;
```
bkoval: Do we need to set a height on it? Maybe line-height and paddings would be enough?

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -0,0 +1,91 @@
...
+		letter-spacing: .25px;
+		margin: 0;
+		padding: 0 1.25rem;
```
bkoval: I think we should use pixel values for paddings :)

front/main/app/styles/component/_discussion-open-graph.scss
```diff
@@ -15,7 +15,8 @@
 
 .discussions {
 	// -- theme --
```
bkoval: Let's just remove this ridicoulus thing. If we want to separate theme and layout styles we should split them into two files and here we're mixing them anyway :)

front/main/app/templates/components/discussion-inline-editor.hbs
```diff
@@ -0,0 +1,41 @@
...
+			{{textarea
+				autoresize=true
+				class="discussion-inline-editor-textarea"
+				placeholder=(i18n placeholderMessageKey ns='discussion')
+				value=content
+			}}
+		</div>
+		{{#if showsOpenGraphCard}}
+			{{discussion-open-graph
+			close=(action 'removeOpenGraph')
```
bkoval: One indentation more ;>

front/main/app/templates/components/discussion-inline-editor.hbs
```diff
@@ -0,0 +1,41 @@
...
+			isLoading=isOpenGraphLoading
+			openGraphData=openGraph
+			}}
+		{{/if}}
+	</div>
+	<div class="discussion-inline-editor-toolbar">
+		<button class="active-element-theme-color" type="submit" {{action 'submit' bubbles=false}} disabled={{submitDisabled}}>{{i18n submitMessageKey ns='discussion'}}</button>
```
bkoval: Did we clear things up about bubbles=false yet? :)

front/main/app/templates/components/discussion-inline-editor.hbs
```diff
@@ -0,0 +1,41 @@
...
+	</div>
+
+	{{#if isLoading}}
+		{{loading-spinner
+		active=isLoading
```
bkoval: Indentation

front/main/app/components/discussion-inline-editor.js
kvas-damian: It's not escaping - it's marking string as safe that doesn't need to be escaped later.

front/main/app/components/discussion-inline-editor.js
kvas-damian: I'm not sure I understand you. Can you elaborate more?

front/main/app/styles/component/_discussion-inline-editor.scss
kvas-damian: It's required because we have containers with relative position on the same level. I'll add the variable.

front/main/app/styles/component/_discussion-inline-editor.scss
kvas-damian: We're expecting negative margin here to center sticky editor which is wider than regular one.

front/main/app/styles/component/_discussion-open-graph.scss
kvas-damian: I didn't touch this code, let's do it in separate PR to limit this PR only to required changes.

front/main/app/templates/components/discussion-inline-editor.hbs
kvas-damian: We have click action bound to entire editor component that opens editor. In order to avoid race condition we need to prevent bubbling on this button.

front/main/app/components/discussion-inline-editor.js
bkoval: 💛

front/main/app/styles/component/_discussion-inline-editor.scss
bkoval: 🆗

bkoval:
LGTM 🌴

PR #2528, opened 2016-06-09
JPN-565 | Fix for add photo button broken when logged out

perjg:
## Links

* https://wikia-inc.atlassian.net/browse/JPN-565

## Description

Add photo on mercury does not work when the user is logged out.

{{action 'addPhoto'}} was missing for when the user is logged out (addPhotoAllowed set to false).
When user is logged in, the same action needs to be on the input tag, so I changed the nesting to render differently based on the state of addPhotoAllowed. The event handler addPhoto handles both cases correctly.

## Reviewers

@hakubo

### DIFF
--- a/front/main/app/templates/components/article-contribution.hbs
+++ b/front/main/app/templates/components/article-contribution.hbs
@@ -1,10 +1,14 @@
 {{#if addPhotoIconVisible}}
-	<div class='upload-photo'>
-		{{svg 'camera' role='img' class='icon camera'}}
-		{{#if addPhotoAllowed}}
+	{{#if addPhotoAllowed}}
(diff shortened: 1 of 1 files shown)


Warkot:
Shouldn't be the same done for the `edit` action then?

perjg:
@Warkot The edit action works correctly both for logged out and logged in state, so this change is only necessary for addPhoto.

(addPhoto when logged in is handled through the input html tag, and when it is present (logged in state only) the action must be attached to the input tag. The problem was that in logged out state, there was no input tag and the action got lost.)

Warkot:
@perjg Thank for the explanation. 👍

rogatty:
LGTM

PR #2529, opened 2016-06-09
SER-198 Add history and extend context in image-review

RafalWilinski:
## Links

* http://wikia-inc.atlassian.net/browse/SER-198

## Description

This PR extends image context in image-review, adds image history table, fixes static-assets info URL and gets rid of computed property.

## Reviewers
@Wikia/x-wing


RafalWilinski:
https://github.com/Wikia/pandora/pull/1203 should be merged first

front/main/app/components/image-review.js
```diff
@@ -5,20 +5,21 @@ export default Ember.Component.extend({
...
+				popupModel.size = (data.size === 0 ? 'not provided' : data.size);
+				popupModel.dimensions = data.dimensions || 'not provided';
+				popupModel.originalFilename = data.originalFilename;
 				popupModel.isLink = new RegExp('(http|https)?:\/\/[^\s]+').test(data.context);
 			}).finally(() => {
 				this.setProperties({
 					thumbnailModel: popupModel,
 					isModalVisible: true
 				});
+				document.getElementById('image-history').innerHTML = popupModel.history;
```
rogatty: This should be rendered using the template (https://github.com/Wikia/mercury/blob/f26835928708ba310f98faa75a9805198da64ba7/front/main/app/templates/components/image-review.hbs#L39).

front/main/app/models/image-review.js
```diff
@@ -65,19 +63,39 @@ ImageReviewModel.reopenClass({
...
+
+		historyJson.forEach((entry) => {
+			tableContent += '<tr>';
+			tableContent += `<td>${entry.date}</td>`;
+			tableContent += `<td>${entry.action}</td>`;
+			tableContent += `<td>${entry.user}</td>`;
+			tableContent += `<td>${entry.status}</td>`;
+			tableContent += '</tr>';
+		});
+		return tableContent;
```
rogatty: This should be done using a template.

front/main/app/styles/component/_image-review.scss
```diff
@@ -30,3 +31,10 @@
 	clear: both;
 	display: block;
 }
+
+table {
```
rogatty: This is too generic, it would add styles to _all_ tables.

rogatty:
There is a lot of tests failing:
http://jenkins.wikia-prod:8080/job/Mercury-PR-Checks/7913/console

I would start with fixing template linter ones.

front/main/app/models/image-review.js
```diff
@@ -65,19 +63,39 @@ ImageReviewModel.reopenClass({
...
+					status: 'accepted',
+					history: ImageReviewModel.prepareHistoryDom(image.imageHistory)
 				}));
 			}
 		});
 		return ImageReviewModel.create({images, contractId, imagesToReviewCount});
 	},
 
+	prepareHistoryDom(historyJson) {
+		if (historyJson === null || typeof historyJson === 'undefined') {
```
bognix: would this work -> http://emberjs.com/api/classes/Ember.html#method_isEmpty? It looks cleaner than above solution

front/main/app/models/image-review.js
```diff
@@ -65,19 +63,39 @@ ImageReviewModel.reopenClass({
...
+			return 'No history';
```
bognix: We don't want to localize this message?

front/main/app/components/image-review.js
```diff
@@ -5,20 +5,21 @@ export default Ember.Component.extend({
...
-		return Boolean(this.get('thumbnailModel.context'));
-	}),
-
 	actions: {
 		showModal(popupModel) {
 			imageReviewItemModel.getImageContext(popupModel.imageId).then((data) => {
+				popupModel.isContextProvided = Boolean(data.context);
 				popupModel.context = data.context;
+				popupModel.size = (data.size === 0 ? 'not provided' : data.size);
+				popupModel.dimensions = data.dimensions || 'not provided';
```
bognix: don't we want to localize this message?

front/main/app/routes/image-review.js
```diff
@@ -59,7 +59,14 @@ export default Route.extend({
 		getFlaggedOnly() {
 			window.scrollTo(0, 0);
 
-			this.set('onlyFlagged', true);
+			this.set('status', 'FLAGGED');
+			this.refresh();
+		},
+
+		getFlaggedOnly() {
```
tomasznapieralski: You have function with this name already, this breaks your code.

front/main/app/components/image-review.js
```diff
@@ -5,14 +5,15 @@ export default Ember.Component.extend({
...
-		return Boolean(this.get('thumbnailModel.context'));
-	}),
-
 	actions: {
 		showModal(popupModel) {
 			imageReviewItemModel.getImageContext(popupModel.imageId).then((data) => {
+				popupModel.isContextProvided = Boolean(data.context);
 				popupModel.context = data.context;
+				popupModel.size = (data.size === 0 ? i18n.t('main.not-provided', {ns: 'image-review'}) : `${data.size}
+					${i18n.t('main.image-bytes', {ns: 'image-review'})}`);
```
rogatty: I think it would be easier to read if it was in a variable instead.

rogatty: Or even better - it's UI logic so I would move it to the template:
```
popupModel.size = data.size
```

```
{{#if thumbnailModel.size}}
  {{i18n 'main.image-size' ns='image-review'}} {{thumbnailModel.size}} {{i18n 'main.image-bytes' ns='image-review'}}
{{else}}
  {{i18n 'main.not-provided' ns='image-review'}}
{{/if}}
```

front/main/app/components/image-review.js
```diff
@@ -5,14 +5,15 @@ export default Ember.Component.extend({
...
+				popupModel.dimensions = data.dimensions || i18n.t('main.not-provided', {ns: 'image-review'});
```
rogatty: Same here, makes more sense to have it in the template.

front/main/app/models/image-review.js
```diff
@@ -65,19 +63,41 @@ ImageReviewModel.reopenClass({
...
-					status: 'accepted'
+					status: 'accepted',
+					history: ImageReviewModel.prepareHistoryDom(image.imageHistory)
 				}));
 			}
 		});
 		return ImageReviewModel.create({images, contractId, imagesToReviewCount});
 	},
 
+	prepareHistoryDom(historyJson) {
```
rogatty: It's not DOM anymore.

front/main/app/models/image-review.js
```diff
@@ -65,19 +63,41 @@ ImageReviewModel.reopenClass({
...
+		}
+
+		let tableContent = [];
+
+		historyJson.forEach((entry) => {
+			tableContent.push({
+				date: entry.date,
+				action: entry.action,
+				user: entry.user,
+				status: (typeof entry.status === 'undefined' ? i18n.t('main.history-undefined', {ns: 'image-review'}) :
```
rogatty: I don't think it belongs to the model, more like to the template.

rogatty: Actually this whole method doesn't do much other than this line. How about removing it entirely and just render `history: image.imageHistory`?

front/main/app/templates/components/image-review.hbs
```diff
@@ -23,20 +26,52 @@
...
 			<p>{{i18n 'main.image-context-not-provided' ns='image-review'}}</p>
 		{{/if}}
+		<span>{{i18n 'main.image-name' ns='image-review'}} {{thumbnailModel.originalFilename}}</span>
+		<br/>
+		<span>{{i18n 'main.image-size' ns='image-review'}} {{thumbnailModel.size}}</span>
+		<br/>
+		<span>{{i18n 'main.image-dimensions' ns='image-review'}} {{thumbnailModel.dimensions}}</span>
+		<br/>
+		{{#if thumbnailModel.history}}
+			<table id="image-history">
```
rogatty: We try to not use ids and I see it doesn't serve any purpose here. How about adding a class instead and using it in the CSS instead of generic:
```
.modal-dialog {
  table {
```
?

front/main/app/templates/components/image-review.hbs
```diff
@@ -23,20 +26,52 @@
 
 	{{#modal-dialog isVisible=isModalVisible}}
 		<img src="{{thumbnailModel.fullSizeImageUrl}}"/>
-		{{#if isContextProvided}}
+		{{#if thumbnailModel.isContextProvided}}
 			{{#if thumbnailModel.isLink}}
-				<a href="{{thumbnailModel.context}}" target="_blank">{{i18n 'main.image-context' ns='image-review'}} {{thumbnailModel.context}}</a>
+				<a href="{{thumbnailModel.context}}"
+				   target="_blank">
```
rogatty: This line break looks odd, I don't think there is any reason to have it.

front/main/app/components/image-review.js
```diff
@@ -5,14 +5,14 @@ export default Ember.Component.extend({
...
-	}),
-
 	actions: {
 		showModal(popupModel) {
 			imageReviewItemModel.getImageContext(popupModel.imageId).then((data) => {
+				popupModel.isContextProvided = Boolean(data.context);
 				popupModel.context = data.context;
+				popupModel.size = data.size;
+				popupModel.dimensions = data.dimensions;
+				popupModel.originalFilename = data.originalFilename;
```
rogatty: Wouldn't it be easier to do `popupModel = data` and then add the additional properties?

front/main/app/models/image-review.js
```diff
@@ -65,13 +63,15 @@ ImageReviewModel.reopenClass({
 		const images = [];
 
 		rawData.forEach((image) => {
-			if (image.reviewStatus === 'UNREVIEWED' || image.reviewStatus === 'FLAGGED') {
+			if (image.reviewStatus === 'UNREVIEWED' || image.reviewStatus === 'FLAGGED' ||
+				image.reviewStatus === 'REJECTED') {
```
rogatty: How about an array and `indexOf()`?

front/main/app/styles/component/_image-review.scss
```diff
@@ -30,3 +31,16 @@
...
+
+.modal-dialog {
+	.image-history {
+		td {
+			padding: 5px;
+			border: 1px solid;
+		}
+	}
+
+	p {
```
rogatty: I think this is too generic. This is `_image-review.scss` and it defines `.modal-dialog p` while `modal-dialog` is a component itself and shouldn't have its styles overriden by some other, random component. Make sure that styles you set here apply only to the image review elements.

front/main/app/templates/components/image-review.hbs
```diff
@@ -23,20 +26,69 @@
...
+				{{/if}}
 			{{else}}
-				<span>{{i18n 'main.image-context' ns='image-review'}} {{thumbnailModel.context}}</span>
+				{{i18n 'main.image-context-not-provided' ns='image-review'}}
 			{{/if}}
+		</p>
+		<p>{{i18n 'main.image-name' ns='image-review'}} {{thumbnailModel.originalFilename}}</p>
+		<p>
+			{{#if thumbnailModel.size}}
+				{{i18n 'main.image-size' ns='image-review'}} {{thumbnailModel.size}} {{i18n 'main.image-bytes' ns='image-review'}}
```
rogatty: `{{i18n 'main.image-size' ns='image-review'}}` can be moved before `{{#if}}` block.

front/main/app/templates/components/image-review.hbs
```diff
@@ -23,20 +26,69 @@
...
+			{{else}}
+				{{i18n 'main.image-size' ns='image-review'}} {{i18n 'main.not-provided' ns='image-review'}}
+			{{/if}}
+		</p>
+		<p>
+			{{#if thumbnailModel.dimensions}}
+				{{i18n 'main.image-dimensions' ns='image-review'}} {{thumbnailModel.dimensions}}
```
rogatty: `{{i18n 'main.image-dimensions' ns='image-review'}}` can be moved before `{{#if}}` block.

front/main/app/templates/components/image-review.hbs
```diff
@@ -23,20 +26,69 @@
 
 	{{#modal-dialog isVisible=isModalVisible}}
```
rogatty: This file got pretty big. What do you think about extracting this block to a separate component? E.g. `components/image-review-modal-dialog.hbs`

rogatty:
LGTM

RafalWilinski:
Thanks for cleaning my mess 😂

PR #2533, opened 2016-06-10
HACK | move discussion to in repo addon

hakubo:
## Links

* https://ember-cli.com/extending/#developing-addons-and-blueprints

## Description

Ember has a concept for addons. In repo addon is just a normal addon but stored inside repo. By laveraging it we can divde our app better so it's easier to navigate through the codebase. 
It's also a step toward fully separating all apps and possibly use Ember Engines to compose our app.

Currently code lives in app/ folder (which is merged with main app/). Ideally we could move all discussion code to addon/ folder that is namespaced and isolated from the main app. I cannot do that now - as that would mean that main app knows nothing about discussion and would break it.

Next steps would be:
- move styles also to in repo addon
- move code to addon folder
- all shared modules/components should be installed via bower (aka Discussion code should not use random utils and mixins from main app)

## Reviewers

@Wikia/x-wing @kvas-damian @bkoval @rafalkalinski


hakubo:
Now I've noticed that this relies on this PR also: https://github.com/Wikia/mercury/pull/2532/files

It has to be merged first or we can fix it here - if this PR is something that we want to merge

PR #2534, opened 2016-06-10
HACK | Dont bidi by default

hakubo:
## Links
- https://github.com/zeppelin/ember-cli-bidi
## Description

While very useful - it takes around 2-3 seconds on each scss build to generate rtl css. We should test these languages but I think we don't do that while developing. 

This disables this plugin by defualt on development environmen to save some precious time. 
When deployed - it will generate these files so we can test rtl languages on stable/sandbox etc
## Reviewers

@Wikia/x-wing @SebastianMarzjan @kvas-damian


kvas-damian:
This solution makes it impossible to test new code on dev on rtl wikis. Testing on stagging environment isn't always possible and in general is discouraged.
What do you think about having separate dev build that disables this plugin?

BTW during hackaton I checked Sass Parsing time by lib-sass and it's ~400ms (not including bidi) so maybe we should try to find why it takes so long bidi to re-parse production CSS.

front/main/ember-cli-build.js
```diff
@@ -103,6 +106,10 @@ module.exports = function (defaults) {
 		app.import(app.bowerDirectory + '/es5-shim/es5-shim.min.js');
 	}
 
+	if (app.env === 'development') {
+		app.options.bidi = false;
```
rogatty: Shouldn't this be `app.options.bidi.enabled = false`?

rogatty:
@kvas-damian did you test RTL wiki even once on devbox? ;) I think it makes the most sense to disable it by default but allow to easily override it. E.g. `npm run dev --enable-rtl`

kvas-damian:
I didn't, but disabling it completely closes a gate for doing that. Enabling via flag makes sense for me.

PR #2535, opened 2016-06-10
'global' ignore for npm and bower dependencies

hakubo:
## Description

Our current policy is to never check in dependencies - this should make it so for every folder and subfolder in the repo

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

vforge:
LGTM

PR #2536, opened 2016-06-10
call devbox instead of local url

hakubo:
## Links

* app: https://github.com/Wikia/app/pull/10657
* config: https://github.com/Wikia/config/pull/1818

## Description

When I enable reflective CORS all works apart from these two calls.
With these 3 in place you can access your local instance with xip.io

http://yourwiki.127.0.0.1.xip.io:8000/

## Reviewers

@Wikia/x-wing @kvas-damian @bkoval


bkoval:
What do you mean by "enabling reflective CORS"?

LGTM

rogatty:
LGTM

@bkoval he means merging https://github.com/Wikia/app/pull/10657 but yeah, "reflective cors" isn't a common term, Google finds nothing.

hakubo:
I might actually made up that term :)
CORS allows for two allowed domains
full wildcard (*)
or a specific domain (my.specifc.domain) and that would not accept (glee.my.specific.domain)

so reflective as in it'll add cors headers with a domain that is asking for the resource

PR #2543, opened 2016-06-13
New version of ember-wormhole

dianafa:
## Description

A new version of ember-wormhole is available so we can start using it:  https://github.com/yapplabs/ember-wormhole/releases/tag/0.4.0

## Reviewers

@Wikia/west-wing @rogatty 

## Test
Checked on `sandbox-s3` in every place where the ember-wormhole is used.


rogatty:
If it works then LGTM :)

PR #2545, opened 2016-06-13
Switch to using embedded images for e3 assets

pauloslund:
## Description
Images need to be inlined in order to work :(

## Reviewers
@tomasznapieralski


rogatty:
SVGs do have to be inlined but PNGs should work fine without that, see:
https://github.com/Wikia/mercury/blob/dev/front/main/app/components/discussion-hero-unit.js#L152

pauloslund:
From our testing, PNGs do not seem to be working though. This image exists in the repo but is not available on preview http://mercury.nocookie.net/front/common/images/recirc-e3.png

pauloslund:
Got a LGTM from @tomasznapieralski offline. Going to merge so that the images work. Then we can discuss why it wasn't working before and come up with a potential better solution

rogatty:
Correct URL is http://mercury.nocookie.net/mercury-static/common/images/recirc-e3.png (don't ask why ;)) and it should be converted during the ember build (fingerprinting).

pauloslund:
Any reason why this style declaration wouldn't be converted? https://github.com/Wikia/mercury/blob/dev/front/main/app/styles/component/_recirculation-experiment.scss#L94

rogatty:
I've found https://github.com/Wikia/mercury/blob/dev/front/main/app/styles/component/_potential-member-page-experiment.scss#L4 which does exactly the same thing. Seems like URLs are not converted in SCSS, only in JS and HTML but on production build they just resolve to http://starwars.wikia.com/front/common/images/potential-member-page-experiment-banner-background.png which works fine.

pauloslund:
So it would work on production but not preview?

rogatty:
By production build I mean the output of `npm run build-prod` which is what deploy tools use. This is where paths are converted.

After reading through the code I think that `potential-member-page-experiment` isn't a good example and it probably doesn't even work. Using `/front/common/images/whatever.png` in the CSS won't work because CSS is accessed from domain `http://mercury.nocookie.net` so `/front` in CSS will resolve to `http://mercury.nocookie.net/front` (which doesn't work) and not `http://*.wikia.com/front` (which would work fine). See http://stackoverflow.com/a/11498495/1050577.

What would be the simplest solution (and - I think - more efficient than inlining PNG in CSS) is reusing solution from [discussions hero unit](https://github.com/Wikia/mercury/blob/dev/front/main/app/components/discussion-hero-unit.js#L152) which sets style from JS. Why it works? Because paths in JS - contrary to CSS ones - are relative to HTML so `/front` resolves to `http://*.wikia.com/front`.

PR #2553, opened 2016-06-15
Discussion | Fixing contributor avatar sizes in hero image

bkoval:

### DIFF
--- a/front/main/app/styles/component/_discussion-hero-unit.scss
+++ b/front/main/app/styles/component/_discussion-hero-unit.scss
@@ -28,13 +28,6 @@ $discussion-hero-unit-minimized-height: 25px;
 	}
 
 	.discussion-contributors {
-		.avatar {
-			img, svg {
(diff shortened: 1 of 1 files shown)


front/main/app/styles/component/_discussion-hero-unit.scss
```diff
@@ -76,6 +69,19 @@ $discussion-hero-unit-minimized-height: 25px;
 	}
 }
 
+.discussions {
+	.discussion-hero-unit {
+		.discussion-contributors {
+			.avatar {
+				img, svg {
```
rogatty: Oh my... Does it have to be so nested?

bkoval: Guess!

bkoval: (yes, it's needed for now as we need to override styles from discussion-contributors unit in the left rail of a post-details page)

PR #2558, opened 2016-06-15
DAT-4151 Set tracking for search pages with query param

dianafa:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-4151

## Description

* Make explicit transition on search query update.
* In tracker, send query param when 'query' param present, do not send in other cases (like is right now)

## Reviewers
@Wikia/west-wing @bognix @rogatty


front/common/baseline/tracker-universal-analytics.js
```diff
@@ -260,7 +260,31 @@ if (typeof window.M.tracker === 'undefined') {
...
+	 *
+	 * Examples:
+	 * ?query=test&useskin=mercury -> ?query=test
+	 * ?one=two&three=four&query=test&five=six -> ?query=test
+	 * ?one=two&three=four -> ''
+	 *
+	 * @param {string} queryParams query params string
+	 * @returns {string}
+	 */
+	function getQueryParam(queryParams) {
```
SebastianMarzjan: Where is the test for this logic?

rogatty: "get query param from query params" is very unclear. If we don't have to use the word `query` as a param name I would change it to `search` because it would be obvious for everyone what it's about.

If we do have to use it then how about naming this method `getSearchParam` anyway? This function is written deliberately for search and IMO it should be visible at the first glance. Right now it's not.

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -260,7 +260,31 @@ if (typeof window.M.tracker === 'undefined') {
...
+			.filter((param) => {
+				return param.indexOf('query=') === 0;
+			})
+			.reduce((p, c) => c, '');
+
+		return query ? `?${query}` : '';
+	}
+
+	/**
+	 * Updates current page. For urls containing the query param 'query', updates them with this param.
```
rogatty: Can you also explain _why_?

front/main/app/routes/search.js
```diff
@@ -24,6 +29,12 @@ export default Route.extend({
 		Ember.$('body').removeClass('search-result-page');
 	},
 
+	afterModel(model, transition) {
+		transition.then(() => {
+			trackPageView();
```
rogatty: Won't it track two page views on the first render?

front/common/baseline/tracker-universal-analytics.js
dianafa: We do need to use the word `query`. I'm thinking about making it more general, that is define here an array of query params we want to accept.

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -260,7 +260,31 @@ if (typeof window.M.tracker === 'undefined') {
...
+	 *
+	 * @param {string} queryParams query params string
+	 * @returns {string}
+	 */
+	function getQueryParam(queryParams) {
+		const query = queryParams
+			.replace(/^\?/, '')
+			.split('&')
+			.filter((param) => {
+				return param.indexOf('query=') === 0;
```
bognix: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

front/main/app/routes/search.js
bognix: no it shouldn't because we have `initialPageView` set to `true`

front/common/baseline/tracker-universal-analytics.js
rogatty: Are you sure `startsWith` is transpiled by Babel? It's not supported in IE and probably that's why we don't use it anywhere else.

bognix:
LGTM, kudos for all the unit tests!

PR #2563, opened 2016-06-17
SOC-2650 | Fix GA Linker

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2650
* https://developers.google.com/analytics/devguides/collection/analyticsjs/linker

## Description

GA linker accepts array of domains - not one domain as a string. Our code was matching basically all domains - including external websites. Casting to array solves problem.

## Reviewers

@hakubo @bkoval


vforge:
lgtm

bkoval:
LGTM

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -97,7 +97,7 @@ if (typeof window.M.tracker === 'undefined') {
 		setupAccountOnce(accounts[trackerName].id, prefix, options);
 
 		if (domain) {
-			ga(`${prefix}linker:autoLink`, domain);
+			ga(`${prefix}linker:autoLink`, domain instanceof Array ? domain : [domain]);
```
rogatty: Wouldn't it be better to just change https://github.com/Wikia/mercury/blob/6af0bbf3ebde6ee2ddf610b07a7b60ef240e1fa2/front/common/baseline/tracker-universal-analytics.js#L453?

kvas-damian: Are you sure that no-one in future will call this method again with string parameter?
IMO it's safer to cast it here.

rogatty: This code seems like it was written to be "very elastic and reusable". Well, it's used only in one place and as we can see - broken. How about using
```
ga(`${prefix}linker:autoLink`, ['wikia.com'])
```
and removing the `domain` param? IMO the current state is unnecessary overengineering. When we need to parametrize it then we can complicate it. No point of doing it now because "someone in the future may need something else" and putting up safe measures because "some may use it in a different way".

kvas-damian: Done

rogatty:
Fix the linter and will LGTM!

PR #2565, opened 2016-06-17
SOC-2602 | Discussion editor refactoring - Add editor logic

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2602

## Description

Add logic for discussion editors, separate functionality into mixins, allow using discussion-editor for editing guidelines in future.

## Reviewers

@bkoval @rafalkalinski


front/main/app/components/discussion-editor.js
```diff
@@ -1,539 +1,78 @@
...
 	showSuccess: false,
+	isLoading: false,
+	editorType: 'contributeEditor',
 
-	offsetTop: 0,
-
-	showsOpenGraphCard: false,
-
-	siteHeadHeight: 0,
+	// Labels below needs to be override in subclasses
```
bkoval: *need to be overrode

front/main/app/components/discussion-editor.js
```diff
@@ -1,539 +1,78 @@
...
-		urls = text.match(/(https?:\/\/[^\s]+)/g);
-
-		if (!urls) {
-			return null;
-		}
-
-		return urls.pop();
-	},
+	submitDisabled: Ember.computed('content', 'currentUser.userId', function () {
+		return this.get('content').length === 0 || this.get('currentUser.userId') === null;
```
bkoval: I'd prefer to check 'currentUser.isAuthenticated' !== false as this is what we actuallyw ant to know here :)

front/main/app/components/discussion-more-options.js
```diff
@@ -37,14 +37,11 @@ export default Ember.Component.extend({
 		track(trackActions.MorePostActions);
 	},
 
-	discussionEditEditor: Ember.inject.service(),
-
 	actions: {
 		edit(post) {
-			const discussionEditEditor = this.get('discussionEditEditor');
+			this.sendAction('setEditDiscussionEntity', post);
+			this.sendAction('setEditorActive', 'editEditor', true);
```
bkoval: Could we handle opening editor in the previous action, so that we send only one action up?

front/main/app/components/discussion-editor.js
```diff
@@ -1,539 +1,85 @@
...
+	submitMessageKey: null,
 
-	bodyText: '',
-	contentLength: 0,
-	errorMessage: Ember.computed.alias('discussionEditor.errorMessage'),
-
-	layoutName: 'components/discussion-editor',
 	// Tracking action name of closing the editor
-	closeTrackingAction: trackActions.PostClose,
+	closeTrackingAction: null,
```
bkoval: Is there a need to define these here since they're already a part of the configuration mixin?

front/main/app/mixins/discussion-editor-opengraph.js
```diff
@@ -0,0 +1,148 @@
...
+import {track, trackActions} from '../utils/discussion-tracker';
+import {getLastUrlFromText} from 'common/utils/string';
+
+export default Ember.Mixin.create({
+	isOpenGraphLoading: false,
+	openGraph: null,
+	showsOpenGraphCard: false,
+	contentLength: 0,
+
+
```
bkoval: No need for this empty line

front/main/app/mixins/discussion-editor-opengraph.js
```diff
@@ -0,0 +1,148 @@
...
+			.on('paste', this.onPaste.bind(this));
+	},
+
+	/**
+	 * Track content changed
+	 *
+	 * @returns {void}
+	 */
+	onContentChange: Ember.observer('content', function () {
+		this.handleTyping();
```
bkoval: Why is there a need to handle typing once at the beginning? Maybe we could rename the method to e.g. detectOGUrl, what do you think?

front/main/app/components/discussion-editor.js
```diff
@@ -1,539 +1,85 @@
...
-				if (this.get('showsOpenGraphCard')) {
-					newDiscussionEntityData.openGraph = {
-						uri: this.get('openGraph.href')
-					};
-				}
-
-				this.get('create')(newDiscussionEntityData);
-			}
+		close() {
+			track(this.get('closeTrackingAction'));
```
rafalkalinski: it would be simpler to follow the flow it track method is called in the descendant methods (instead of calling _super)

front/main/app/components/discussion-inline-post-editor.js
```diff
@@ -0,0 +1,31 @@
+import DiscussionStickyComponentMixin from '../mixins/discussion-sticky-component';
+import DiscussionInlineEditor from './discussion-inline-editor';
+
+export default DiscussionInlineEditor.extend(
+	DiscussionStickyComponentMixin,
+	{
+		containerClassname: '.discussion-inline-editor-floating-container',
```
rafalkalinski: containerClassNameSelector

front/main/app/components/discussion-more-options.js
rafalkalinski: I think this should be one call and those both actions should be triggered somewhere in editor.

front/main/app/components/discussion-reply-editor-stub.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
```
rafalkalinski: I believe we are able to find a better module name here.

front/main/app/components/discussion-standalone-editor.js
```diff
@@ -1,51 +1,81 @@
...
-		handleKeyPress() {
+				if (!this.get('isEdit')) {
+					actionName = 'create';
+					discussionEntityData.creatorId = this.get('currentUser.userId');
+					discussionEntityData.siteId = Mercury.wiki.id;
+				} else {
+					const editEntity = this.get('editEntity');
 
-		}
+					if (editEntity.get('isReply')) {
```
rafalkalinski: this.get('editEntity.isReply')

front/main/app/components/discussion-standalone-editor.js
```diff
@@ -1,51 +1,81 @@
...
+						actionName = 'editReply';
+						discussionEntityData.id = editEntity.get('id');
```
rafalkalinski: this.get('editEntity.id')

front/main/app/components/discussion-standalone-editor.js
```diff
@@ -1,51 +1,81 @@
...
+					} else {
+						actionName = 'editPost';
+						discussionEntityData.id = editEntity.get('threadId');
```
rafalkalinski: this.get('editEntity.threadId')

front/main/app/components/discussion-standalone-editor.js
```diff
@@ -1,51 +1,81 @@
...
+				}
 
-		},
-		handleKeyPress() {
+				if (!this.get('isEdit')) {
+					actionName = 'create';
+					discussionEntityData.creatorId = this.get('currentUser.userId');
+					discussionEntityData.siteId = Mercury.wiki.id;
+				} else {
+					const editEntity = this.get('editEntity');
```
rafalkalinski: I can't see why we need this variable.

front/main/app/controllers/discussion/forum.js
```diff
@@ -1,13 +1,21 @@
 import DiscussionModerationControllerMixin from '../../mixins/discussion-moderation-controller';
 import DiscussionContributionControllerMixin from '../../mixins/discussion-contribution-controller';
 import DiscussionForumActionsControllerMixin from '../../mixins/discussion-forum-actions-controller';
-import DiscussionEditEditorMixin from '../../mixins/discussion-edit-editor';
+import ResponsiveMixin from '../../mixins/responsive';
```
rafalkalinski: Why import this when it is not used here?
And the same below.

front/main/app/mixins/discussion-contribution-controller.js
```diff
@@ -1,15 +1,182 @@
...
+	 *
+	 * @returns {object}
+	 */
+	getEditorState(editorType) {
+		if (editorType === 'contributeEditor') {
+			return this.get('editorState');
+		} else if (editorType === 'editEditor') {
+			return this.get('editEditorState');
+		} else {
+			throw new Error(`Editor type not supported: ${editorType}`);
```
rafalkalinski: :+1:

front/main/app/mixins/discussion-contribution-controller.js
```diff
@@ -1,15 +1,182 @@
...
+
+	/**
+	 * Get object that contains editor state
+	 *
+	 * @param {string} editorType type of editor - available 'contributeEditor' and 'editEditor'
+	 *
+	 * @returns {object}
+	 */
+	getEditorState(editorType) {
+		if (editorType === 'contributeEditor') {
```
rafalkalinski: We should have one place for the whole application where we'd keep the variables like this.

front/main/app/mixins/discussion-editor-configuration.js
```diff
@@ -0,0 +1,74 @@
...
+
+		this.configure();
+	},
+
+	configure() {
+		if (this.get('isEdit')) {
+			if (this.get('isReply')) {
+				// edit reply
+				this.setProperties({
+					labelMessageKey: 'editor.reply-edit-editor-label',
```
rafalkalinski: this method would look better if those objects with keys will be defined in the mixin scope.

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -1,11 +1,15 @@
 $sticky-editor-width: 529px;
+$success-icon-size: 80px;
+$editor-toolbar-height: 40px;
```
rafalkalinski: :abcd:

front/main/app/styles/component/_discussion-reply-editor-stub.scss
```diff
@@ -0,0 +1,37 @@
+.discussion-reply-editor-stub {
+	.stub-container {
+		background: #fff;
```
rafalkalinski: what about color variables in this module?

front/main/app/components/discussion-editor.js
kvas-damian: They're used in this file, so probably it's better to have them defined somewhere. If you feel different I can change it.

front/main/app/mixins/discussion-editor-opengraph.js
kvas-damian: I'm not sure I understand first sentence, can you clarify what do you mean?

front/main/app/components/discussion-standalone-editor.js
kvas-damian: We use this entity again few lines below. I believe it's better to cache it in local variable. What do you think?

front/main/app/mixins/discussion-editor-configuration.js
kvas-damian: Didn't we discuss similar case offline few day ago?
I'm not sure what was the outcome of that discussion ;)

front/main/app/templates/components/discussion-post.hbs
```diff
@@ -71,10 +75,19 @@
 		</button>
 	{{/if}}
 	{{#if canReply}}
-		{{discussion-reply-editor
-			create=(action this.attrs.createReply)
-			generateOpenGraph=(action this.attrs.generateOpenGraph)
-			replies=model.replies
-		}}
+		{{#if responsive.isDesktop}}
```
kvas-damian: @bkoval I think we missed label for editor on this view.

kvas-damian:
Please fix styles for dark theme.

front/main/app/components/discussion-standalone-editor.js
rafalkalinski: As we've talked offline, I just have a different opinion here.

front/main/app/components/discussion-editor.js
kvas-damian: Having this method defined in base class keeps all tracking method in the same place. If we decide to move it to subclass then it won't be clear which tracking methods are already implemented and which needs to be added in new subclass.
Do you think it still makes sense to move it somewhere else?

rafalkalinski: super

front/main/app/components/discussion-editor.js
bkoval: OK, let's leave it as is then.

kvas-damian:
@rafalkalinski @bkoval are you ready for 2nd round?

front/main/app/components/discussion-editor.js
```diff
@@ -1,539 +1,85 @@
...
-			lastChar = value.charCodeAt(textarea.selectionEnd - 1),
-			allowedChars = [10, 13, 32];
-
-		if (allowedChars.indexOf(lastChar) === -1 || value.length <= this.get('contentLength')) {
-			return;
-		}
-
-		const url = this.getLastUrlFromText(value.substring(0, textarea.selectionEnd));
+	onIsLoading: Ember.observer('isLoading', function () {
+		if (!this.get('isLoading')) {
```
bkoval: What about posting errors? This observer does not check for errors before showing success state.

rafalkalinski:
LGTM

front/main/app/components/discussion-editor.js
```diff
@@ -1,539 +1,85 @@
...
-		 *
-		 * @param {boolean} active
-		 *
-		 * @returns {void}
-		 */
-		toggleEditorActive(active) {
-			if (active && !this.get('wasStartTracked')) {
-				track(this.get('startTrackingAction'));
-				this.set('wasStartTracked', true);
+		handleKeyPress() {
```
kvas-damian: Because we change `content` when we initialize editing.

front/main/app/styles/component/_discussion-reply-editor-entry-point.scss
```diff
@@ -0,0 +1,37 @@
+.discussion-reply-editor-entry-point {
+	.discussion-editor-entry-point-container {
+		background: $discussion-editor-background-color;
+		border: 4px solid $discussion-editor-border-color;
+		align-items: center;
```
bkoval: Alphabetical order ;<<<

front/main/app/styles/component/_discussion-inline-editor.scss
```diff
@@ -88,4 +95,79 @@ $sticky-editor-width: 529px;
...
+			font-size: $type-smallest;
+			line-height: $line-height-larger;
+			margin: 0;
+			max-height: 2 * $line-height-larger;
+			overflow: hidden;
+			padding: 0;
+			position: absolute;
+			text-align: center;
+			width: 100%;
+			z-index: $z-discussion-editor + 1;
```
bkoval: Shouldn't it have it's own variable defined?

front/main/app/styles/component/_discussion-standalone-editor.scss
```diff
@@ -63,4 +65,84 @@
...
+	}
+}
+
+@media #{$mobile-range} {
+	.discussion-standalone-editor {
+		bottom: 0;
+		left: 0;
+		position: fixed;
+		top: 0;
+		right: 0;
```
bkoval: Alphabetical order :)

bkoval:
LGTM 👍 👍 👍 This is so freaking awesome, thanks!

front/common/utils/string.js
```diff
@@ -26,3 +26,20 @@ export function normalizeToWhitespace(str = '') {
 		.replace(/_/g, ' ')
 		.replace(/\s+/g, ' ');
 }
+
+/**
+ * Get last url from input text
+ * @param {string} text
+ * @returns {string}
+ */
+export function getLastUrlFromText(text) {
```
rogatty: It's used only in the main app, right? Can you put it in `main/app/utils` instead? See https://github.com/Wikia/mercury/pull/2532 for _why?_.

kvas-damian: It is. Although the PR you're referring to isn't merged yet. It means we don't have final decision which way we'd like to approach.

I'd suggest keeping it here and move together with other utils in #2532. What do you think?

PR #2571, opened 2016-06-22
New Auth Translations (blocker for new auth release)

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/LOC-167
* http://wikia-inc.atlassian.net/browse/SOC-2131

## Description

Social team is planning on releasing new auth screens to all russian and spanish communities and, hence, there's a need to update localization for these languages upfront.

## Reviewers

@kvas-damian

PR #2572, opened 2016-06-22
New Auth Translations (blocker for new auth release)

bkoval:
See #2571 for details

@kvas-damian


kvas-damian:
I believe we should include @dianafa code there ;)

bkoval:
Yup, seems like there's something in release-250 that was not merged to 251? ;>

PR #2575, opened 2016-06-22
SOC-2552 | Deactivate dark theme when exiting discussions

bkoval:
## Links
* http://wikia-inc.atlassian.net/browse/SOC-2552

@kvas-damian @rafalkalinski


front/main/app/routes/discussion/base.js
```diff
@@ -54,6 +54,12 @@ export default Ember.Route.extend(
 
 				if (!isDiscussionRoute) {
 					this.controllerFor('application').set('noMargins', false);
+
+					// deactivate dark theme after transition outside discussions to avoid
+					// a flash of unstyled content
+					transition.promise.then(() => {
+						this.deactivateTheming();
```
kvas-damian: I see noMargins modified before transition and here we're waiting for transition to finish. Is it possible to be consistent and fire those events in the same time?

bkoval: Yup, it works, just moved it :)

kvas-damian:
LGTM

PR #2580, opened 2016-06-29
Stop using mirror for PhantomJS

rogatty:
## Links

* https://github.com/Wikia/mercury/pull/2227

## Description

Mirror that we use is sometimes very slow and it takes few minutes to download PhantomJS. It doesn't make much sense to use it.

## Reviewers

@tomasznapieralski


hakubo:
lgtm

PR #2586, opened 2016-07-01
(DAT-4558) Use shrinkwrap for our dependencies in Mercury

Grunny:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4558
* https://docs.npmjs.com/cli/shrinkwrap

## Description

Adds commands to both shrinkwrap all the dependencies and update the
dependencies to the latest version en masse if desired.

We are using shonkwrap instead of shrinkwrap which removes the resolved
field so that it will fallback to the default and handle both our npm cache
proxy and installing from the npm registry when it's not available.

## Reviewers

@rogatty @hakubo


rogatty:
LGTM

README.md
```diff
@@ -38,14 +38,15 @@ Open http://muppet.127.0.0.1.xip.io:7000/wiki/Kermit
 * `npm run clean` - remove all dependencies and build directory
 * `npm run clean-deep` - remove all dependencies, remove build directory and clean the cache
 * `npm run svgo` - optimize all SVG files in the repository (check out `tasks/optimize-svg.sh` for details and configuration)
+* `npm run shrinkwrap` - updates shrinkwrap files which locks the version of dependencies we are using (should be run after updating a dependency with `npm install`)
```
hakubo: since you should always run npm install - does it make sense to make it part of the command?

package.json
```diff
@@ -24,6 +24,10 @@
     "watch-server": "gulp watch-server",
     "start": "nodemon --delay 2",
     "start-on-change": "nodemon --delay 2 --on-change-only",
+    "shrinkwrap": "npm run shrinkwrap-root && npm run shrinkwrap-main && npm run shrinkwrap-server",
+    "shrinkwrap-root": "shonkwrap --dev",
```
hakubo: what are the reasons for using shonkwrap?
I think shrinkwrap json is not really meant to be read by human so I don't see much value in cleaning it. 

Any other benefits?

README.md
Grunny: `npm install` by itself will make use of the version specified in npm-shrinkwrap.json and just changing package.json won't cause an update. With this line I meant updating an individual dependency which you need to do with `npm install <module>` not just plain `npm install`. After doing so and testing the new version of the dependency, running `npm run shrinkwrap` will update the shrinkwrap files. Do you have thoughts on how to improve that?

package.json
Grunny: `shrinkwrap` by itself sets a `resolved` field with a full URL to the package that will be used in future installs. On devboxes or elsewhere within our network, this will be set to our npm cache proxy when running shrinkwrap, and if you do it on your local machine it will be the npm registry. shonkwrap removes the resolved field so that it will fallback to the default (so on devboxes or within our network it will use our proxy, and locally it will go to the main registry.

hakubo: sounds good - 
last thing that worries me is https://github.com/skybet/shonkwrap#related-work
we are already using npm@2 and soon will be using npm@3 - will it support it?

README.md
hakubo: that make sense - the note was not obvious to me though maybe in parenthisis or somewhere we could write steps that have to be taken to update dependency?

package.json
Grunny: Hmm, shonkwrap primarily supports npm@3, so it should be fine once we start using that, with the fork [shrinkwarp](https://github.com/evocateur/shrinkwarp) aiming to better support npm@2 if you have issues. I haven't noticed any issues in the produced shrinkwrap file for npm@2 with shonkwrap though, or using the produced files to perform the setup of Mercury. We could consider shrinkwarp for now and switch back to shonkwrap once we start using npm@3 if there are any issues?

README.md
Grunny: Sounds good, I'll do that. Possibly a Confluence page with a link from the README?

package.json
hakubo: did not have any issues yet - since swapping implementations should be fairly easy
we can go with shonkwrap for now and see how it behaves

(these names are very confusing though :) shonkwrap, shrinkwrap, shrinkwarp ...)

hakubo:
LGTM

README.md
Grunny: Done in e089947388a7ae196b8b7e698b7dec7c07c4738d

rogatty:
LGTM :ship:

PR #2588, opened 2016-07-05
DAT-4560 Move TOP_MOBILE_LEADERBOARD below Portable Infobox

dianafa:
Links

http://wikia-inc.atlassian.net/browse/DAT-4560

Reviewers

@Wikia/adeng @Wikia/west-wing

### DIFF
--- a/front/main/app/mixins/ads.js
+++ b/front/main/app/mixins/ads.js
@@ -156,12 +156,10 @@ export default Ember.Mixin.create({
 		this.clearAdViews();
 
 		if ($pi.length) {
-			//inject top mobileTopLeaderBoard below infobox
-			console.log("inject top mobileTopLeaderBoard below infobox");
+			// inject top mobileTopLeaderBoard below infobox
 			this.appendAd(adsData.mobileTopLeaderBoard, 'after', $pi.first());
 		} else if ($pageHeader.length) {
-			//inject top mobileTopLeaderBoard below article header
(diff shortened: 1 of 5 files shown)


front/main/app/mixins/ads.js
```diff
@@ -141,27 +142,38 @@ export default Ember.Mixin.create({
...
+				articleBodyHeight > adsData.moreInContentAds.minPageLength;
 
 		this.clearAdViews();
 
+		if ($pi.length) {
+			// inject top mobileTopLeaderBoard below infobox
+			this.appendAd(adsData.mobileTopLeaderBoard, 'after', $pi.first());
+		} else if ($pageHeader.length) {
+			// inject top mobileTopLeaderBoard below article header
+			this.appendAd(adsData.mobileTopLeaderBoard, 'after', $pageHeader.first());
```
SebastianMarzjan: Do we want to inject ad if neither a PI nor header is found?

dianafa: If article will not contain any of this elements it means that something is broken. Generally- it shouldn't happen if page loaded correctly.

rogatty: Do we want to inject ads on broken pages?

SebastianMarzjan: IMO no, though it'd be useful to know how often we serve such pages

PR #2589, opened 2016-07-05
(DAT-4579) Treat local URLs with query parameters as external URLs

Grunny:
## Links

* https://wikia-inc.atlassian.net/browse/DAT-4579
* https://github.com/Wikia/app/pull/10787

## Description

If a local article URL has a query param, we should treat it as a full external
URL if it is not handled elsewhere. Manual local URLs with query params created by
users are already handled as the MediaWiki parser adds the external class which
is already checked in Mercury.

This is an extra safety check in case any extensions add URLs with query params
(as TabView did) so that they don't result in a broken page when clicked.

action=render is removed from the TabView links in https://github.com/Wikia/app/pull/10787.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/routes/application.js
+++ b/front/main/app/routes/application.js
@@ -106,7 +106,8 @@ export default Route.extend(
 					Mercury.wiki.basePath,
 					title,
 					target.hash,
-					target.href
+					target.href,
+					target.search
 				);
 
 				/**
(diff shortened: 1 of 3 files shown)


hakubo:
LGTM

rogatty:
LGTM

PR #2590, opened 2016-07-05
DAT-4573 | send error to Kibana only when translation for default language not found

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/DAT-4573

## Reviewers

@Wikia/x-wing


hakubo:
sounds like a default error
LGTM

rogatty:
I wonder if `error` is the right severity now. What will user see without the default translation? Will the app crash or display message keys?

We might consider using `Logger.fatal` instead.

PR #2591, opened 2016-07-05
ADEN-3483 Sync UAP

drets:
Push "mobile incontent", "mobile prefooter" and "mobile bottom leaderboard" on scroll if there is a UAP campaign in "top leaderboard".
(release after https://github.com/Wikia/app/pull/10790)
/cc @rogatty


front/common/modules/ads.js
```diff
@@ -103,6 +110,32 @@ class Ads {
...
+				noUapCallback();
+			});
+		} else if (this.uapResult) {
+			uapCallback();
+		} else {
+			noUapCallback();
+		}
+	}
+
+	getBtfSlots() {
```
fraszczakszymon: This getter is used only in places where we check whether slot is ATF/BTF. What do you think about moving the logic here and renaming method to `isAboveTheFold`/`isBelowTheFold`?

fraszczakszymon: Discussed offline 👍

front/common/modules/ads.js
```diff
@@ -103,6 +110,32 @@ class Ads {
 		});
 	}
 
+	addUapListeners(uapCallback, noUapCallback) {
+		if (!this.uapCalled) {
+			window.addEventListener('wikia.uap', () => {
```
rogatty: Do you think it's worth removing the listener after it's not needed anymore?

front/main/app/components/ad-slot.js
```diff
@@ -1,60 +1,104 @@
...
+		noAds: computed({
+			get() {
+				return false;
+			},
+			set(key, value) {
+				return value !== '' && value !== '0';
+			}
+		}),
+
+		didInsertElementOverride: Ember.on('didInsertElement', function () {
```
rogatty: It's not really an override as the original `didInsertElement` is still called. Maybe `onDidInsertElement`?

front/main/app/mixins/ads.js
```diff
@@ -1,14 +1,14 @@
 import Ember from 'ember';
-import AdSlotComponent from '../components/ad-slot';
 import Ads from 'common/modules/ads';
 
 export default Ember.Mixin.create({
 	adsData: {
 		minZerothSectionLength: 700,
 		minPageLength: 2000,
-		mobileTopLeaderBoard: 'MOBILE_TOP_LEADERBOARD',
+		mobileBottomleaderboard: 'MOBILE_BOTTOM_LEADERBOARD',
```
rogatty: Can it be consistent with the top one? `mobileBottomLeaderBoard` - one way or another.

front/main/app/mixins/ads.js
```diff
@@ -39,20 +39,27 @@ export default Ember.Mixin.create({
...
+			componentElement.$().insertBefore(element);
+		}
 
-		element[place](view.$());
-		this.adViews.push(view);
+		this.adViews.push(componentElement);
 
-		view.trigger('didInsertElement');
+		componentElement.didInsertElement();
+		componentElement.didInsertElementOverride();
```
rogatty: Oh my... If we really have to call it manually (do we?) then I would name it differently, e.g. `onElementManualInsert()`
Right now it's very misleading.

front/common/modules/ads.js
```diff
@@ -103,6 +110,32 @@ class Ads {
 		});
 	}
 
+	addUapListeners(uapCallback, noUapCallback) {
```
rogatty: How about `waitForUap` / `waitForUapResponse` / `waitForUapCallback`?

front/common/modules/ads.js
drets: :+1: It will be handled in separate ticket (ADEN-3542)

front/main/app/mixins/ads.js
fraszczakszymon: *Leaderboard please

drets: <3

fraszczakszymon:
LGTM

drets:
LGTM

PR #2604, opened 2016-07-11
RECIRC-168: Add placement test to renderedComponents for proper teardown

pauloslund:
## Links
* https://wikia-inc.atlassian.net/browse/RECIRC-168

## Description
Placement test components were not getting properly torn down in-between article transitions causing duplication to occur. We fix this by adding that component to article-contents `renderedComponents` array.

## Reviewers
@Wikia/x-wing


vforge:
lgtm

PR #2607, opened 2016-07-13
XW-1580 Add global footer messages for translations

kamilkoterba:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1580

## Description

Added links for global footer to be translated.
Message keys are following this RFC https://wikia-inc.atlassian.net/wiki/display/DESYS/RFC%3A+Global+Footer+Messages+Schema

## Reviewers
@Wikia/x-wing


crowdin/global-footer.conf
```diff
@@ -0,0 +1,9 @@
+[crowdin.project]
+projectid = "mercury"
+repo.branch = "i18n-update"
+source.base = "../front/common/public/locales"
+source.file = "${source.base}/en/global-footer.json"
```
rogatty: According to https://wikia-inc.atlassian.net/wiki/display/DESYS/RFC%3A+Storing+translations it should be `design-system.json`

crowdin/design-system.conf
```diff
@@ -0,0 +1,9 @@
+[crowdin.project]
+projectid = "mercury"
```
rogatty: I think we should use a separate project as this will be moved from here.

front/common/gulpfile.js
```diff
@@ -117,3 +117,28 @@ gulp.task('test-common', ['build-common-vendor-for-tests'], function (done) {
 		configFile: __dirname + '/tests/karma.conf.js'
 	}, done).start();
 });
+
+gulp.task('build-design-system-i18n', function () {
+	var fs = require('fs'),
```
hakubo: common pattern is to require 'all the things' at the top of a file

front/common/gulpfile.js
```diff
@@ -117,3 +117,28 @@ gulp.task('test-common', ['build-common-vendor-for-tests'], function (done) {
...
+		i18n = '',
+		i18nBuilt = {},
+		loadTree = require('../../gulp/utils/load-tree'),
+		tree = loadTree('node_modules/design-system-i18n/i18n');
+
+	for(var dir in tree) {
```
hakubo: one var per scope

gulp/utils/load-tree.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var fs = require('fs');
+
+var walk = function(path) {
```
hakubo: are we sure that there is nothing like that in npm

front/common/public/locales/en/design-system.json
```diff
@@ -0,0 +1,39 @@
+{
```
hakubo: hmm since this is installed - should that end up in mercury repo? might be messy to work with
especially if someone will edit this file

vforge: What do you think about adding this file to `.gitignore` then? It should always be build fresh from `design-system-i18n`.

gulp/utils/load-tree.js
kamilkoterba: Haven't found any

vforge:
LGTM

front/common/public/locales/README.md
```diff
@@ -0,0 +1,11 @@
+## Design System i18n messages
+Design system messages are stored centrally for all apps in https://github.com/Wikia/design-system-i18n repo.
+
+Those messages are pulled to Mercury as npm dependency on the setup phase (`npm run setup`) and copied with all other translation files to `/www/front/common/locales` on build phase.
```
rogatty: It would be easier to find if we put the name of the gulp task here:
`build-common-design-system-i18n`

front/common/public/locales/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+To lookup available messages take a look at `node_modules/design-system-i18n/i18n/en/design-system.json`
+
+To use in template just write `{{i18n 'global-footer-example-key' ns='design-system'}}`
+
+### Update Design System messages
+Once new messages are added to design-system-i18n repo version number (after # char) needs to be updated on the dependencies list in /package.json (e.g. `"design-system-i18n": "Wikia/design-system-i18n#0.0.1",`)
```
rogatty: It's actually not a _repo version number_ but rather _tag version_ as you can see on https://github.com/Wikia/mercury/releases/new.

kamilkoterba: Thanks, I'll adding that in next PR

vforge:
LGTM

PR #2610, opened 2016-07-14
XW-1576 | Add Design System to Mercury

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1576

## Description

* [x] Add Design System to Mercury.
* [x] Add (temporary) styling to test DS presence.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/bower.json
+++ b/front/main/bower.json
@@ -3,6 +3,7 @@
   "dependencies": {
     "Autolinker.js": "0.22.0",
     "cropper": "0.11.0",
+    "design-system": "Wikia/design-system#master",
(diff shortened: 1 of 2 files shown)


front/main/bower.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "Autolinker.js": "0.22.0",
     "cropper": "0.11.0",
+    "design-system": "Wikia/design-system#master",
```
rogatty: I think it's time for v0.0.1 :)
Let's not use master branch here.

PR #2612, opened 2016-07-14
ADEN-3493 Fix BFAA on pages with infobox

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3493


nandy-andy:
LGTM :+1:

PR #2613, opened 2016-07-15
ADEN-3467 implement PageFair

v3rth:


front/common/modules/ads.js
```diff
@@ -22,15 +22,18 @@ import load from '../utils/load';
...
  * @property {Ads} instance
- * @property {boolean|null} blocking
+ * @property {boolean|null} previousSourcePointDetectionResult
+ * @property {boolean|null} previousPageFairDetectionResult
  * @property {Array<string[]>} adSlots
  * @property {Object} adsContext
  * @property {*} adEngineRunnerModule
  * @property {*} adContextModule
  * @property {SourcePointDetectionModule} sourcePointDetectionModule
+ * @property {PageFairDetectionModule} pageFairDetectionModule
```
fraszczakszymon: Please add PageFairDetectionModule jsdoc at the top of this file

front/common/modules/ads.js
```diff
@@ -53,6 +56,16 @@ class Ads {
 			'MOBILE_PREFOOTER',
 			'MOBILE_BOTTOM_LEADERBOARD'
 		];
+		this.GASettings = {
```
fraszczakszymon: What do you think about renaming this variable and move here *DetectionResult flag for each module? Then `trackBlocking` would require only one argument.

v3rth: You mean `previousSourcePointDetectionResult`/`previousPageFairDetectionResult`? I'm not sure because this data is only for GA settings and contains consts. These vars are related with logic and represent state

front/common/modules/ads.js
```diff
@@ -22,15 +27,18 @@ import load from '../utils/load';
  * @class Ads
  *
  * @property {Ads} instance
- * @property {boolean|null} blocking
+ * @property {boolean|null} previousSourcePointDetectionResult
```
bognix: it's not clean. This variable is boolean but in fact has three values -> false, true and null. It would be cleaner to have a flag which would indicate if the detection was already run or not. What I mean is:
* one variable to store the result
* one variable to store the status if the detection was already run

front/common/modules/ads.js
```diff
@@ -179,30 +202,44 @@ class Ads {
 	}
 
 	/**
-	 * @param {string} value
+	 * @param {Object} GAOption
+	 * @param {Boolean} isABDetected
```
bognix: ab is AdBlock or AB test? Consider naming the parameter in a less confusing way

front/common/modules/ads.js
v3rth: I'm aware of this fact, but for me it's pretty clear when there is no previous result it equals `null` in other way it's `true`/`false`. If it isn't for you I will change it ;)

bognix: However, if I understood correctly if you won't store state in those variables you can do as Szymon suggested here -> https://github.com/Wikia/mercury/pull/2613#discussion_r71118429

bognix: I prefer the "two variables" solution but I won't be pushing on that one. IMHO it's cleaner.

bognix:
Do we have to roll out App before Mercury?

bognix:
LGTM

bognix:
LGTM

PR #2614, opened 2016-07-15
XW-1581 Global Footer in Mercury

kamilkoterba:
## Links

* With the development this PR became feature branch for: https://wikia-inc.atlassian.net/browse/XW-1581

## Description

Integrating Global Footer with Mercury:
- [x] wrapper partial added to page
- [x] supplied with mocked data
- [x] links helper generator based on link type provided by API
- [x] design system messages loaded for server side
- [x] html from reference page specification implemented
- [x] [@Warkot] all messages supplied with data from API (global-footer-licensing-and-vertical-description)
- [x] [@Warkot] support global-footer-licensing-description message on Mercury side when there's no vertical
- [x]  [@tomasznapieralski]unit tests for markup
- [x] SVGs integration
- [x] [@kvas-damian] i18n update to v0.0.3 
- [x] [@rogatty] Footer renders only where it should
- [x] [@kamilkoterba] Support types for elements returned by API to decide how to render
- [x] [@rogatty] Remove empty place which SVG element takes at the top of page
- [x] [@kvas-damian] Add loading spinner between global nav and footer

## Reviewers
@Wikia/x-wing


server/app/lib/utils.js
```diff
@@ -191,6 +191,14 @@ export function getCachedWikiDomainName(localSettings, request) {
 }
 
 /**
+ * @param {string} wikiDomain
+ * @returns {string}
+ */
+export function getCorporatePageUrlFromWikiDomain(wikiDomain) {
+	return `www${wikiDomain.substring(wikiDomain.indexOf('.'))}`;
```
kvas-damian: Are we sure it will work for all wiki domains? Including language prefixed domains like `http://pl.gta.wikia.com/`

front/main/app/index.html
```diff
@@ -93,6 +93,7 @@
 {{> tracking-ivw3}}
 
 {{content-for 'body-footer'}}
+{{> global-footer/global-footer model=globalFooter}}
```
rogatty: I think this should be above `{{content-for 'body-footer'}}`. This is where Ember addons put things that should be at the bottom of the body element.

server/app/lib/mediawiki-page.js
```diff
@@ -42,7 +42,8 @@ export class PageRequestHelper {
 				.page(this.params.title, this.params.redirect, this.params.sections),
 			new MediaWiki.WikiRequest({
 				wikiDomain: this.params.wikiDomain
-			}).wikiVariables()
+			}).wikiVariables(),
+			new MediaWiki.DesignSystemRequest(this.params).getFooter()
```
kvas-damian: reusing `this.params` results in [adding `wikicities_session` cookie into the API request](https://github.com/Wikia/mercury/blob/ece88fd196894df8df9c8af11633621cb9329c39/server/app/facets/mediawiki-page.js#L244-L244). Finally it tells our caches to skip cache for logged in users. Is it intended?

server/app/views/_helpers/global-footer-link.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Currently Hapi doesn't recognize ES6 syntax on exports (ie: "default" keyword)
+ *
+ * @param {object} context
+ * @param {object} link
+ * @returns {string}
+ */
+module.exports = function i18nHelper(context, link) {
```
Grunny: `i18nHelper` copy and paste leftover.

server/app/lib/mediawiki.js
```diff
@@ -239,17 +239,46 @@ class BaseRequest {
...
+	constructor(params) {
+		super(params);
+
+		this.corporatePageUrl = params.corporatePageUrl;
+	}
+
+	getFooter() {
+		const url = createUrl(this.corporatePageUrl, 'wikia.php', {
+			controller: 'DesignSystemApi',
+			method: 'getFooter'
```
Grunny: This request should supply the `wikiId` for the current wiki as well.

server/app/lib/utils.js
rogatty: @kvas-damian how do you like it now?
https://github.com/Wikia/mercury/pull/2614/commits/cf07040507c11487a5816810a26a7fe87e93a995

server/app/views/_partials/global-footer/global-footer-link-image.hbs
```diff
@@ -0,0 +1,4 @@
+<a href="{{model.href}}" class="global-footer__link">
+	<img alt="{{i18n model.title.key ns='design-system'}}"
+		 src="http://public.diana.wikia-dev.com/design-system-reference-page/bower_components/design-system/assets/{{model.image}}.svg">
```
kvas-damian: Probably URLs shouldn't be hardcoded...

server/app/views/_partials/global-footer/global-footer-link-branded.hbs
```diff
@@ -0,0 +1,7 @@
+<a href="{{model.href}}" class="global-footer__link is-{{model.brand}}">
+	<div>{{i18n model.title.key ns='design-system'}}</div>
+	<svg class="global-footer__image icon">
```
kvas-damian: Can we use svg helper?

server/app/views/_partials/global-footer/global-footer-section.hbs
```diff
@@ -0,0 +1,21 @@
...
+	{{/if}}
+
+	<ul class="global-footer__links-list">
+		{{#each model.links as |link|}}
+			<li class="global-footer__links-list-item">
+				{{!--
+					First parameter of called helper has to be .. to preserve parent context within called helper.
+					Without it context would be loops context i.e. current loop item
+				--}}
+				{{{global-footer-link .. link}}}
```
kvas-damian: We have only 3 types of links. Maybe we could simply use `if ... else` syntax to render them instead of introducing our own rendering helper?

server/app/views/_partials/global-footer/global-footer.hbs
```diff
@@ -0,0 +1,41 @@
...
+		</h2>
+	{{/if}}
+	<div class="global-footer__main">
+		{{#unless isInternational}}
+			<h2 class="global-footer__fandom-header">
+				{{> global-footer/global-footer-image-header model=model.fandom.header section='fandom'}}
+			</h2>
+		{{/unless}}
+		<div class="global-footer__fandom-sections">
+				{{> global-footer/global-footer-section model=model.fandom_overview name='fandom-overview' parentName='fandom'}}
```
kvas-damian: Please fix indentation.

server/app/lib/utils.js
```diff
@@ -191,6 +191,25 @@ export function getCachedWikiDomainName(localSettings, request) {
 }
 
 /**
+ * @param {LocalSettings} localSettings
+ * @param {string} wikiDomain
+ * @returns {string}
+ */
+export function getCorporatePageUrlFromWikiDomain(localSettings, wikiDomain) {
```
kvas-damian: Can we add support for xip.io domains?

rogatty: `wikiDomain` that is passed here goes through `getWikiDomainName()` first, so it should work fine. Didn't test it though as currently no one is using xip.io anyway.

server/app/lib/mediawiki-page.js
rogatty: Good catch, IMO we shouldn't send this cookie.

server/app/lib/mediawiki.js
```diff
@@ -239,17 +239,46 @@ class BaseRequest {
...
+		return this
+			.fetch(url, this.corporatePageUrl)
+			.then((footerData) => {
+				if (footerData) {
+					return Promise.resolve(footerData);
+				} else {
+					return Promise.reject();
+				}
+			}, () => {
+				return Promise.reject();
```
rogatty: I think this would work too and look cleaner:
```
.then((footerData) => {
  if (footerData) {
    return footerData;
  } else {
    throw new Error('No footer data returned from API');
  }
});
```
I don't see the point of catching failed promise to create a new one and then reject it.

server/app/views/_partials/global-footer/global-footer-link-branded.hbs
rogatty: It's not available on the server side so we can either leave it like that, write a new helper or use a partial instead.

server/app/views/_partials/global-footer/global-footer-section.hbs
rogatty: We can also use a [concat helper](http://stackoverflow.com/a/35862620/1050577) which is simpler and more generic than the current one.

server/tests/acceptance/design-system.js
```diff
@@ -0,0 +1,73 @@
...
+		done();
+	});
+
+	after(function (done) {
+		mediawiki.__ResetDependency__('Wreck');
+		done();
+	});
+
+	it('compare global footer with it\'s baseline', function (done) {
+		var footerMarkupBaseline = fs.readFileSync(__dirname + '/../fixtures/design-system/footer.html', 'utf-8')
```
rogatty: Let's keep it consistent and name it `global-footer.html`

server/tests/acceptance/design-system.js
```diff
@@ -0,0 +1,73 @@
...
+		wreckGetStub.onCall(1).yields(null, {statusCode: 200}, clone(wikiVariables));
+		wreckGetStub.onCall(2).yields(null, {statusCode: 200}, clone(footer));
+
+		server.inject(requestParams, function (response) {
+			jsdom.env(response.payload, function (errors, window) {
+				var footerHTML = window.document.getElementsByClassName('global-footer')[0],
+					footerWrapper = window.document.createElement('div'),
+					footerMarkup;
+
+				footerWrapper.appendChild(footerHTML.cloneNode(true));
```
rogatty: `Cannot read property 'cloneNode' of undefined`

server/app/facets/show-application.js
```diff
@@ -79,15 +80,26 @@ export default function showApplication(request, reply, wikiVariables, context =
...
+			return getGlobalFooterData(data, request);
+		})
+		/**
+		 * @param {*} contextData
 		 * @returns {void}
 		 */
-		.then((openGraphData) => {
-			// Add OpenGraph attributes to context
-			context.openGraph = openGraphData;
+		.then((context) => {
```
rogatty: Right now `context` is used for two different variables. Even if they should be equal, I still think it's a bad idea to do that.

server/app/facets/show-curated-content.js
```diff
@@ -72,6 +73,15 @@ export default function showCuratedContent(request, reply) {
 	mainPage.setTitle(request.params.title);
 	mainPage.getWikiVariablesAndDetails()
 		/**
+		 * Get data for Global Footer
```
rogatty: That's obvious when looking at the name of `getGlobalFooterData` function ;)

server/app/lib/global-footer.js
```diff
@@ -0,0 +1,18 @@
+import {
+	getCachedWikiDomainName,
+	getCorporatePageUrlFromWikiDomain
+} from './utils';
+import * as MediaWiki from './mediawiki';
+import localSettings from '../../config/localSettings';
+
+export default function getGlobalFooterData(data, request) {
```
rogatty: It modifies `data` so maybe `injectGlobalFooterData`?

server/app/lib/mediawiki-page.js
```diff
@@ -65,6 +65,7 @@ export class PageRequestHelper {
 			.then((results) => {
 				const mediaWikiPagePromise = results[0],
 					wikiVariablesPromise = results[1],
+					globalFooterPromise = results[2],
```
rogatty: This shouldn't be here, right?

server/app/lib/global-footer.js
rogatty: Actually it can be anonymous here but I'd change it in the files where it's imported.

server/app/facets/show-curated-content.js
kvas-damian: Great to heap, I like when things are obvious ;)

server/app/lib/global-footer.js
kvas-damian: I'm not sure I want it to be anonymous. File name doesn't match function name and what it does. Maybe it'd be better to convert this method into `globatlFooterUtils` class. WHat do you think?

rogatty: How about renaming the file to `inject-global-footer-data.js`?

kvas-damian: Done!

server/app/lib/utils.js
kvas-damian: :+1:

server/app/views/_partials/global-footer/global-footer.hbs
```diff
@@ -28,6 +28,6 @@
 		</div>
 	</div>
 	<div class="global-footer__licensing-and-vertical">
-		<span>{{i18n model.licensing_and_vertical.description.key ns='design-system'}}</span>
+		<span>{{{i18n model.licensing.description.key license=(wrap-me model.licensing.description.params.license.title.value tagName='a' href=model.licensing.description.params.license.href) ns='design-system'}}}</span>
```
Warkot: `licensing` is reverted to `licensing_and_vertical` in the next commit (f6c94beccfecc2a3526603ee3a95ac21720dac2b).

front/main/app/mixins/route-with-body-class-name.js
```diff
@@ -0,0 +1,23 @@
+import Ember from 'ember';
+
+const {$, on, Mixin} = Ember;
+
+export default Mixin.create({
+	addBodyClassOnActivate: on('activate', function () {
+		const $body = $('body'),
+			bodyClassNames = this.get('bodyClassNames') || [];
```
kvas-damian: Instead of using `|| []` we could consider to define default value for this attribute in the mixin.

server/app/views/_helpers/wrap-me.js
```diff
@@ -0,0 +1,44 @@
...
+	let tagName = 'span',
+		href = '',
+		className = '';
+
+	if (hash.tagName) {
+		tagName = escapeHtml(hash.tagName);
+	}
+
+	if (hash.tagName === 'a' && hash.href) {
+		href = ` href="${escapeHtml(hash.href)}"`;
```
kvas-damian: Did we make sure that `escapeHtml` is sufficient here?
I believe `" onclick="aler(1)` is HTML that won't be escaped here and allow XSS.

server/app/views/_partials/global-footer/global-footer.hbs
```diff
@@ -0,0 +1,41 @@
+<footer class="global-footer is-{{wikiVariables.language.content}}">
+	{{#if isInternational}}
```
kvas-damian: We have at least 3 checks for `isInternational` in this file. What do you think about separating internation and non-internation template into separate partials?

server/app/views/_helpers/wrap-me.js
rogatty: Here's the source of `escapeHtml`:
https://github.com/hapijs/hoek/blob/v2.12.0/lib/escape.js

I ran:
```
var Hoek = require('hoek');
var test = "\" onclick=\"alert(1)";
console.log(Hoek.escapeHtml(test));
```
and it produced:
```
&quot; onclick&#x3d;&quot;alert&#x28;1&#x29;
```
so it seems fine.

server/app/views/_partials/global-footer/global-footer.hbs
rogatty: I don't link this idea. `isInternational` is used to control very small parts of this template. I prefer doing that than duplicating ~40 lines.

kvas-damian: Ok, let's keep it as it is for now. We can revise this decision in future.

server/tests/acceptance/design-system.js
```diff
@@ -0,0 +1,73 @@
...
+				footerMarkup = footerWrapper.innerHTML
+					.replace(/>\s+</g, '><')
+					.replace(/(\r\n|\n|\r)/gm, '');
```
kvas-damian: I see the same operation performed on the HTML string. Can we extract it into a separate method?

front/common/baseline/load-dom-resource.js
```diff
@@ -14,7 +14,8 @@ if (typeof window.M === 'undefined') {
 			const div = document.createElement('div');
 
 			div.innerHTML = ajax.responseText;
-			document.body.insertBefore(div.childNodes[0], document.body.firstChild);
+			div.style = 'height: 0; width: 0; position: absolute;';
```
rogatty: This is read-only on Safari.

server/app/views/_partials/loading-spinner.hbs
```diff
@@ -0,0 +1,5 @@
+<svg class='spinner block' width='79' height='80vh' viewBox='0 0 79 79' xmlns='http://www.w3.org/2000/svg'>
```
rogatty: `80vh` isn't a correct value for this attr. We can use `style` instead.

server/app/views/_partials/global-footer/global-footer.hbs
Grunny: What do you think of making this conditional on the presence of the items that may or may not appear, so that the response from the API determines what to display, rather than needing to set an `isInternational` (and so if we decided to have some languages use the fandom header instead, the change in the API would handle it without separate changes here)? i.e. since one of the AC for the API is that it doesn't return sections that shouldn't be rendered, you could do `{{#if model.international_header}} ... {{/if}}` in theory.

server/app/lib/utils.js
```diff
@@ -266,14 +285,15 @@ export function getStaticAssetPath(localSettings, request) {
 	const env = localSettings.environment || Environment.Dev;
 
 	return env !== Environment.Dev ?
-		// The CDN path should match what's used in https://github.com/Wikia/mercury/blob/dev/gulp/options/prod.js
+		// The CDN path should match what's used in
+        // https://github.com/Wikia/mercury/blob/dev/gulp/options/prod.js
```
Grunny: Minor: whitespace.

server/tests/acceptance/design-system.js
tomasznapieralski: yes we can, i can do this later today

server/app/views/_partials/global-footer/global-footer.hbs
```diff
@@ -0,0 +1,41 @@
...
+	{{#if model.international_header}}
+		<h2 class="global-footer__header">
+			{{#with model.international_header.header as |header|}}
+				{{> svg class='global-footer__large-wikia-logo' image=header.image}}
+				<span class="global-footer__home-of-fandom">{{i18n header.subtitle.key ns='design-system'}}</span>
+			{{/with}}
+		</h2>
+	{{/if}}
+	<div class="global-footer__main">
+		{{#unless model.international_header}}
```
rogatty:
```
{{#unless model.fandom.header}}
```

server/app/views/_partials/global-footer/global-footer.hbs
```diff
@@ -0,0 +1,41 @@
...
+			{{> global-footer/global-footer-image-header model=model.fandom.header section='fandom'}}
+		{{/unless}}
+		<div class="global-footer__fandom-sections">
+			{{> global-footer/global-footer-section model=model.fandom_overview name='fandom-overview' parentName='fandom'}}
+			{{> global-footer/global-footer-section model=model.follow_us name='follow-us' parentName='fandom'}}
+		</div>
+		{{#unless model.international_header}}
```
rogatty:
```
{{#unless model.wikia.header}}
```

server/app/views/_partials/global-footer/global-footer.hbs
kvas-damian: I'd say `{{#if model.fandom.header}}`. What do you think?

front/common/baseline/load-dom-resource.js
```diff
@@ -7,14 +7,20 @@ if (typeof window.M === 'undefined') {
...
 
 		ajax.onload = () => {
 			const div = document.createElement('div');
 
 			div.innerHTML = ajax.responseText;
-			document.body.insertBefore(div.childNodes[0], document.body.firstChild);
+			div.style.cssText = 'height: 0; width: 0; position: absolute;';
+
+			if (!visible) {
+				div.style.cssText += ' visibility: hidden;';
```
kvas-damian: I see we use different approach for other svgs: `display: none` on svg tag. Can we agree on one way of solving the same problem and start using it everywhere?

front/main/app/mixins/route-with-body-class-name.js
```diff
@@ -0,0 +1,23 @@
...
+
+		bodyClassNames.forEach((className) => {
```
hakubo: we could limit here operations on DOM
```
$body.addClass(bodyClassNames.join(' '));
```

front/main/app/mixins/route-with-body-class-name.js
```diff
@@ -0,0 +1,23 @@
...
+			$body.addClass(className);
+		});
+	}),
+
+	removeBodyClassOnDeactivate: on('deactivate', function () {
+		const $body = $('body'),
+			bodyClassNames = this.get('bodyClassNames') || [];
+
+		bodyClassNames.forEach((className) => {
```
hakubo: here as well

front/main/app/mixins/route-with-body-class-name.js
hakubo: or actually, won't that mean that the state will be shared?
so that should be set in init I guess

alternative idea:
this mixin just exposes addBodyClass, removeBodyClass]
and a route is responsible for setting it and clearing the state
this should create more explicit code

what do you think?

front/main/app/routes/discussion.js
```diff
@@ -2,34 +2,40 @@ import Ember from 'ember';
...
+			 * @param {String} value - the new value for the attribute
+			 * @returns {void}
+			 */
+			editAttribute(name, value) {
+				const attributesModel = this.modelFor(this.get('routeName')).index.attributes;
+
+				this.setEditorError(null, true);
+
+				attributesModel.editAttribute(value).then(() => {
+					// action after successful edit
```
hakubo: then is not required - does it have to be here?

rogatty: We just changed the indentation here, let's not touch it in this PR.

front/main/app/styles/component/_global-footer.scss
```diff
@@ -0,0 +1,11 @@
+// These are overrides for Mercury only
+// If you look for Global Footer styles go here:
+// https://github.com/Wikia/design-system/blob/master/components/_global-footer.scss
+
+.global-footer {
+	display: none;
```
hakubo:
```scss
.global-footer { 
  .show-global-footer & {
  	display: block;
  }
}
```

package.json
```diff
@@ -50,7 +50,7 @@
     "babel-template": "6.7.0",
     "babel-types": "6.7.2",
     "code": "1.5.0",
-    "design-system-i18n": "Wikia/design-system-i18n#0.0.1",
+    "design-system-i18n": "Wikia/design-system-i18n#0.0.3",
```
hakubo: it's a small step for npm
but a giant leap... :)

front/common/baseline/load-dom-resource.js
Warkot: `display: none;` is *not* added to the svg symbol palette built by the Design System, whereas we have that added by scripts which build Mercury. To be consistent with what we have, we would have to add that in the Design System.

However, while browsing for solution I stumbled upon https://github.com/FWeinb/grunt-svgstore/issues/22, and if I understand correctly, svg gradients won't work if we use `display: none;`.

I'm also not sure in which place it's better to add the mentioned styles—to the palette itself or to the wrapper.

What are your opinions on that?

server/app/facets/show-curated-content.js
```diff
@@ -72,6 +73,15 @@ export default function showCuratedContent(request, reply) {
...
+		 * @param {CuratedContentPageData} data
+		 * @returns {CuratedContentPageData}
+		 *
+		 */
+		.then((data) => {
```
hakubo: you don't need to wrap this function in {}
```js
.then((data) => injectGlobalFooterData(data, request, true))
```

server/app/views/_helpers/concat.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Currently Hapi doesn't recognize ES6 syntax on exports (ie: "default" keyword)
+ * Accepts any number of string params to join.
+ *
+ * @returns {string}
+ */
+module.exports = function concat() {
```
hakubo: how about:
```js
function concat(...strings) {
  return strings.join('')
}
```

would that work?

server/app/views/_helpers/wrap-me.js
```diff
@@ -0,0 +1,44 @@
+import {escapeHtml} from 'hoek';
+
+/**
+ * Helper to generate HTML from passed string and additional options.
+ * By default, if no tagName specified, wraps passed string in <span> tags.
```
hakubo: is it a sensible default?
what do you think about tagName being required?

server/tests/acceptance/design-system.js
```diff
@@ -0,0 +1,73 @@
+var Lab = require('lab'),
```
hakubo: what do we think of renaming to global-footer.js test?

front/main/app/routes/discussion.js
hakubo: in which PR we will touch it then? :)

front/main/app/mixins/route-with-body-class-name.js
rogatty: Good catch with the shared state, that's true. In this case I would prefer to stay with `|| []` instead of overriding `init()` in mixin.

As for the explicit code - I think that name of this mixin is explicit enough and as long as we keep it as simple as it's now it's fine. If we only keep the methods we can drop the mixin at all, as most of the code would be duplicated anyway.

server/app/views/_helpers/wrap-me.js
rogatty: It's mostly a copy-paste from Ember helper. I wouldn't spend too much time on it. If it's ever needed we can adjust. Right now it's used in one place.

server/tests/acceptance/design-system.js
tomasznapieralski: let's leave it as it is for now, we will think how to split this when we will implement global-nav

front/main/app/routes/discussion.js
rogatty: Ask Social team, maybe they already have one.

kvas-damian: I think Social Team is working on it. @bkoval can you confirm that?

front/common/baseline/load-dom-resource.js
Warkot: Fixed here: 8d436f6515b77bdd974e6ce7bc74dbd91e4c5222

front/main/app/routes/discussion.js
bkoval: It's already merged to dev. https://github.com/Wikia/mercury/pull/2640 - can anyone of you CR this change? :)

server/app/lib/inject-global-footer-data.js
```diff
@@ -0,0 +1,38 @@
...
+import * as MediaWiki from './mediawiki';
+import localSettings from '../../config/localSettings';
+import Logger from './logger';
+
+/**
+ * @param {object} wikiVariables
+ * @returns {string}
+ */
+function getContentLanguage(wikiVariables) {
+	return wikiVariables.language && wikiVariables.language.content ? wikiVariables.language.content : 'en';
```
kvas-damian: What do you think about moving `en` into a variable? Maybe even extracting it into configuration file?

server/app/lib/utils.js
```diff
@@ -315,3 +335,13 @@ export function redirectToOasis(request, reply) {
...
 	reply.redirect(`${request.url.pathname}?${queryParams}`);
 }
+
+/**
+ * @param {Hapi.Request} request
+ * @param {object} wikiVariables
+ */
+export function setI18nLang(request, wikiVariables) {
+	if (wikiVariables.language && wikiVariables.language.content) {
+		request.server.methods.i18n.getInstance().setLng(wikiVariables.language.content);
```
kvas-damian: This change breaks support for `uselang` query param. Can we fix it somehow? It makes debugging easier.

server/app/lib/inject-global-footer-data.js
rogatty: Meh, I don't see a scenario where we want to change the default to Chinese.

server/app/lib/utils.js
tomasznapieralski: Currently we are not supporting this query param on Mercury, i'm not sure if that was working at some point.

tomasznapieralski: for debugging you can pass language here

kvas-damian:
@tomasznapieralski We do support `uselang` currently: http://fallout.wikia.com/wiki/hgfhdfhfhfd?uselang=pl vs http://fallout.wikia.com/wiki/hgfhdfhfhfd (you can see different translations during loading state). We also use `uselang` on auth pages.

server/app/lib/utils.js
kvas-damian: @tomasznapieralski currently we do support `uselang` query param on server, see: http://fallout.wikia.com/wiki/hgfhdfhfhfd vs http://fallout.wikia.com/wiki/hgfhdfhfhfd?uselang=pl (different translations are presented during loading state). The same feature is used for translating Auth Pages.

front/main/app/styles/component/_global-footer.scss
```diff
@@ -0,0 +1,16 @@
...
+
+	.show-global-footer & {
+		display: block;
+	}
+
+	&__home-of-fandom {
+		line-height: 1;
```
rogatty: This is to override [a rule from style-guide](https://github.com/Wikia/style-guide/blob/a93cd1c200ec7669c4c3b3f0064d46c76133b270/dist/css/lib/components/type.css#L317). @hakubo what do you think? Is it fine for now and we should make it better when working on CSS resets?

hakubo: ok, 
can you add a comment and create a ticket for the CSS reset?
Thanks!

server/app/views/_helpers/wrap-me.js
hakubo: mkay

front/main/app/styles/component/_global-footer.scss
rogatty: Added a comment in https://wikia-inc.atlassian.net/browse/XW-1759

tomasznapieralski:
Tested

PR #2643, opened 2016-07-22
XW-1753 | set content lang

tomasznapieralski:
content is set on:
- article page
- curated content
- discussions
- RWA

### DIFF
--- a/server/app/facets/mediawiki-page.js
+++ b/server/app/facets/mediawiki-page.js
@@ -14,7 +14,8 @@ import {PageRequestHelper} from '../lib/mediawiki-page';
 import {
 	getCachedWikiDomainName,
 	redirectToCanonicalHostIfNeeded,
-	redirectToOasis
+	redirectToOasis,
+	setI18nLang
 } from '../lib/utils';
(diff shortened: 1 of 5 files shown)


server/app/facets/mediawiki-page.js
```diff
@@ -144,6 +144,10 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
 	// @todo XW-596 we shouldn't rely on side effects of this function
 	Tracking.handleResponse(result, request);
 
+	if (result.wikiVariables && result.wikiVariables.language && result.wikiVariables.language.content) {
```
rogatty: How about extracting it to an util?

Warkot:
LGTM

server/app/facets/mediawiki-page.js
Warkot: I was also thinking about it. We're talking about it now. Wouldn't it make code less readable actually?

Warkot: Talked offline—let's DRY it! :smi

rogatty:
LGTM

PR #2645, opened 2016-07-25
ADEN-3584 Remove BLB from non-UAP scenarios

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3584


front/main/app/mixins/ads.js
```diff
@@ -3,6 +3,14 @@ import Ads from 'common/modules/ads';
 
 export default Ember.Mixin.create({
 	adsData: {
+		additionalConfig: {
+			MOBILE_TOP_LEADERBOARD: {
+				isAboveTheFold: true
```
rogatty: This isn't true for all top leaderboards. On http://starwars.wikia.com/wiki/Yoda it's under the infobox, way below the fold.

fraszczakszymon: In AdEngine's eyes it's still ATF slot :) we flag slots in this way because they are at the top and they should be called immediately/without any delay (or even in single request with other slots). Thanks!

rogatty: It can be misleading for everyone outside of the AdEng team. How about a comment clarifying that?

bognix:
LGTM, @fraszczakszymon consider addressing Igor's comment.

nandy-andy:
LGTM :+1:

PR #2647, opened 2016-07-25
WW-91 Provide entryPointWikiId var only when exists

dianafa:
## Links

* https://wikia-inc.atlassian.net/browse/WW-91

## Description

Use this var only when exists

## Reviewers

@Wikia/west-wing @bkoval @rafalkalinski 

ask for var only when avaiable

### DIFF
--- a/front/auth/views/_layouts/auth.hbs
+++ b/front/auth/views/_layouts/auth.hbs
@@ -85,8 +85,11 @@
 					user: 'en',
 				},
 			});
-			// place where we can check wikiId of the entrypoint
-			M.provide('entryPointWikiId', {{entryPointWikiId}});
+
+			// wikiId of the entrypoint if available
+			{{#if entryPointWikiId}}
+				M.provide('entryPointWikiId', {{entryPointWikiId}});
+			{{/if}}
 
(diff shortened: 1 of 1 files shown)


SebastianMarzjan:
```
{{#if LGTSocial}}
  LGTM
{[/if}}
```

front/auth/views/_layouts/auth.hbs
```diff
@@ -85,8 +85,11 @@
 					user: 'en',
 				},
 			});
-			// place where we can check wikiId of the entrypoint
-			M.provide('entryPointWikiId', {{entryPointWikiId}});
+
+			// wikiId of the entrypoint if available
+			{{#if entryPointWikiId}}
```
bkoval: Shouldn't we have a default value for it?

dianafa: we don't need such for now. If value is not set in the place we need it:
- means something is terribly wrong and we need to fix it anyways
- default value will not bring us any benefit/more information here - we're fine with undefined

rafalkalinski:
LGTM

PR #2648, opened 2016-07-25
WW-91 Provide entryPointWikiId var only when exists

nikodamn:
https://wikia-inc.atlassian.net/browse/WW-91

## Reviewers

@Wikia/west-wing


SebastianMarzjan:
:shipit:

PR #2651, opened 2016-07-27
XW-1733 | fix sending events to GA

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1733

## Changes

* google analytics script is downloaded over https (google recommends)
* google script is loaded in head instead of after body (google recommends)
* events send before GA initializer are stored and resend after initializer

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/common/baseline/tracker-universal-analytics.js
+++ b/front/common/baseline/tracker-universal-analytics.js
@@ -207,6 +207,8 @@ if (typeof window.M.tracker === 'undefined') {
 	 * @returns {void}
 	 */
 	function track(category, action, label, value, nonInteractive) {
+		console.log(">>>>>>>>>>>>>>>>> AFTER TRACK <<<<<<<<<<<<<<<<<<<<<<<");
+
 		syncDimensions();
 
 		tracked.forEach((account) => {
@@ -319,6 +321,7 @@ if (typeof window.M.tracker === 'undefined') {
 	 * @returns {void}
 	 */
 	function trackPageView(uaDimensions, overrideUrl) {
+		console.log(">>>>>>>>>>>>>>>>> TRACK PV <<<<<<<<<<<<<<<<<<<<<<<");
 		/**
 		 * We have some dimensions that are changing per-page - those include
 		 * articleType, namespace and so on. We can't unset once sent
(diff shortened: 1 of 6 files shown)


front/common/baseline/tracker-universal-analytics.js
```diff
@@ -492,6 +506,6 @@ if (typeof window.M.tracker === 'undefined') {
 		_getDimensionsSynced: getDimensionsSynced,
 		_updateTrackedUrl: updateTrackedUrl,
 		_filterQueryParams: filterQueryParams,
-		_dimensions: dimensions,
+		_dimensions: dimensions
```
vforge: There's need to remove that comma?

server/app/views/_partials/tracking-ua-auth.hbs
```diff
@@ -1,8 +1,5 @@
 {{#unless queryParams.noexternals}}
 	<script>
-		// Load UA async and flag $script when ready
```
vforge: Remember about this https://github.com/Wikia/mercury/blob/dev/server/app/views/_partials/tracking-ua-auth.hbs#L4 - and check auth tracking too

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -30,10 +30,12 @@ if (typeof window.M.tracker === 'undefined') {
 
 (function (M) {
 	let tracked = [],
+		trackEvents = [],
```
rogatty: How about `eventsQueue`? Would be more descriptive.

front/common/baseline/tracker-universal-analytics.js
```diff
@@ -460,6 +466,14 @@ if (typeof window.M.tracker === 'undefined') {
 			initAccount(accountSpecial);
 		}
 
+		isInitialized = true;
+
+		trackEvents.forEach((event) => {
+			track(event.category, event.action, event.label, event.value, event.nonInteractive);
+		});
+
+		trackEvents = [];
```
rogatty: I suggest moving this line and the loop above to a separate function named like `flushEventsQueue`, just for the readibility.

rogatty:
Apart from two minor comments LGTM.

PR #2656, opened 2016-07-28
New Auth | Remove tracking dimensions call not to break js on auth

bkoval:
## Description

We cannot send a request to http:// domain from https:// served page, it triggers mixed content error and breaks javascript on the page.

## Reviewers

@vforge @rogatty @Wikia/social-frontend 

https://wikia-inc.atlassian.net/browse/SOC-2819


server/app/views/_partials/tracking-ua-auth.hbs
```diff
@@ -5,44 +5,25 @@
...
-							14: data.dimensions[14],
-							15: 'No',
-							19: data.dimensions[19],
-							25: data.dimensions[25]
-						});
-					}
-				}
-			});
+			if (ua.initialize({
+						4: 'mercury',
```
vforge: Can you hardcode missing dimensions? If you'll do we'll have same dimensions sent on all our pages - that was the point of having the call in the first place.

rogatty: FYI:
- 3: wiki vertical
- 14: show ads
- 19: article type
- 25: namespace

bkoval: I don't feel like we ever needed these on auth pages. Do you feel like adding them all set to 'false' or null makes much sense?

vforge: I would still add them - even with `null/false` - just for the sake of next person who will read the code and wonder why we don't have some dimensions.

bkoval: 👍

server/app/views/_partials/tracking-ua-auth.hbs
```diff
@@ -2,44 +2,29 @@
...
-							14: data.dimensions[14],
-							15: 'No',
-							19: data.dimensions[19],
-							25: data.dimensions[25]
-						});
-					}
-				}
-			});
+			if (ua.initialize({
+						3: null,
```
rogatty: Too much tabs, I think.

rogatty:
Minor formatting issue but other than that LGTM

server/app/views/_partials/tracking-ua-auth.hbs
bkoval: My IDE formats it this way. You sure? ;>

vforge:
LGTM

rogatty:
LGTM

PR #2659, opened 2016-07-28
RECIRC 104: Setup Liftigniter for placement test

pauloslund:
## Links

* https://wikia-inc.atlassian.net/browse/RECIRC-104
* https://liftigniter.readme.io/docs/introduction

## Description

Liftigniter is a content recommendation service that we have been seeing decent results with on desktop so we are extending the test to mobile. The experiment will be run for two weeks.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -659,15 +659,17 @@ export default Ember.Component.extend(
 					externalLink
 				});
 
+				this.renderedComponents.push(view);
+
 				return model.load()
 					.then(() => {
-						view.createElement();
+						if (!view.isDestroyed && !view.isDestroying) {
+							view.createElement();
 
-						location.after(view.$());
-						view.trigger('didInsertElement');
-						view.trackImpression();
-
-						this.renderedComponents.push(view);
+							location.after(view.$());
+							view.trigger('didInsertElement');
+							view.trackImpression();
+						}
 
 						return view;
 					});
--- a/front/main/app/components/article-content.js
+++ b/front/main/app/components/article-content.js
@@ -555,28 +555,33 @@ export default Ember.Component.extend(
 		},
 
 		/**
+		 * Create component instance using container lookup.
+		 * @param {String} componentName
+		 * @returns {Ember.Component}
+		 */
+		createComponentInstance(componentName) {
+			return this.get('container').lookup(`component:${componentName}`, {
+				singleton: false
+			});
+		},
+
+		/**
 		 * TO BE THROWN AWAY AFTER RECIRCULATION_MERCURY_PLACEMENT AB TEST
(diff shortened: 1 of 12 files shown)


front/main/app/components/article-content.js
```diff
@@ -588,6 +594,27 @@ export default Ember.Component.extend(
...
+					model = LiftigniterModel.create({
+						widget: 'fandom-rec',
+						thumbHeight: 160,
+						thumbWidth: 280
+					});
+					location = Ember.$('.article-footer');
+					break;
+				case 'LI_BOTH':
+					this.injectPlacementTest('LI_INCONTENT').then(() => {
+						this.injectPlacementTest('LI_FOOTER');
```
rogatty: What happens when user opens article and navigates away before this is called? Can you check it? I think we might need to use `this.isDestroyed()` here.

front/main/app/components/article-content.js
```diff
@@ -628,13 +656,14 @@ export default Ember.Component.extend(
...
 				view.createElement();
-				model.load();
 
 				location.after(view.$());
 				view.trigger('didInsertElement');
 				view.trackImpression();
 
 				this.renderedComponents.push(view);
+
+				return model.load();
```
rogatty: Because it's done like that we need to use observer in `mixins/recirculation-experiment.js`. How about loading the model first and then rendering the component? We would have one re-render less and there would be no need for the observer.

front/main/app/models/liftigniter.js
```diff
@@ -0,0 +1,102 @@
...
+	thumbHeight: 120,
+
+	init() {
+		this._super(...arguments);
+		this.items = [];
+	},
+
+	title: Ember.computed('widget', function () {
+		if (this.get('widget') === 'fandom-rec') {
+			return 'Trending Fandom Articles';
```
kvas-damian: Is this test limited to EN wikis only?

pauloslund: Yes, it's actually only being run on 5 communities (starwars, marvel, gameofthrones, fallout, elderscrolls)

front/main/app/components/article-content.js
```diff
@@ -627,14 +658,19 @@ export default Ember.Component.extend(
...
-				model.load();
 
-				location.after(view.$());
-				view.trigger('didInsertElement');
-				view.trackImpression();
+				return model.load()
+					.then(() => {
+						view.createElement();
+
+						location.after(view.$());
```
rogatty: What happens when user navigates away from article before `model` is loaded? `location` might not exist anymore, right? I think we should check `isDestroyed` here too.

rogatty:
One thing and will LGTM

PR #2665, opened 2016-08-01
XW-1786 | get google script from http instead of https (temporary)

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1786

## Description

Our selenium tests are using library which intercepts all browser requests, but because we are using also some kind of proxy, this library is not catching https requests but just http.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/server/config/localSettings.base.js
+++ b/server/config/localSettings.base.js
@@ -238,7 +238,7 @@ const localSettings = {
 				id: 'UA-32132943-1',
 				sampleRate: 100
 			},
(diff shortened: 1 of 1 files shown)


rogatty:
LGTM

Grunny:
LGTM.

kvas-damian:
Doesn't it break auth pages?
They are served over HTTPS, so probably we'll end up with browser blocking request due to mixed content. Maybe we could use `//www.google-analytics.com/analytics.js` ?

PR #2669, opened 2016-08-01
SOC-2618 | Customizing discussions for Trek Initiative

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-2618
* more relevant links

## Description

We'll remove this code after sufficient time

## Reviewers

@Wikia/social-frontend


mleczey:
LGTM

rafalkalinski:
LGTM 
One minor, more general comment:
I would like better in the future to see changes in less places. Maybe deeper in the core of Mercury we could put the exceptional logic hidden behind one condition. 
One way or another, if this is a hack, then why not try to limit places where the code is changed?

PR #2670, opened 2016-08-02
SOC-2665 | New link for PT terms of use

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-2665

## Description

Updating a link to terms of use to be localized in PT instead of linking to the EN one.

## Reviewers

@Wikia/social-frontend


mleczey:
LGTM

PR #2672, opened 2016-08-02
SER-364 wrong number of images to review

KolorowyAleksander:
## Links

* http://wikia-inc.atlassian.net/browse/SER-364

## Description

The refreshing of page, and getting number of images must happen after all request reviewing images have been sent (either succesfully or not) to update number of images.


front/main/app/routes/image-review.js
```diff
@@ -43,16 +44,13 @@ export default Route.extend({
...
 			window.scrollTo(0, 0);
 
-			if (!Ember.isNone(model.images)) {
-				ImageReviewModel.reviewImages(model.images, model.contractId).then(() => {}, (data) => {
-					this.controllerFor('application').addAlert({
-						message: data,
-						type: 'warning',
-						persistent: false
-					});
+			ImageReviewModel.reviewImages(model.images, model.contractId).then(() => {}, (data) => {
```
rogatty: What if `Ember.isNone(model.images)`? We don't care anymore?

rogatty:
One question but LGTM

front/main/app/routes/image-review.js
KolorowyAleksander: Yes, it was not really necessary in the first place. Also, situation in which this function is called, and model.images is undefined or null should never happen.

PR #2683, opened 2016-08-04
Added meta tag for Japanese language

d34th4ck3r:
@perjg


homepage/server/views/_layouts/default.hbs
```diff
@@ -1,8 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="jp" dir="ltr">
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1, minimal-ui">
+	<meta name="language" content="japanese">
```
perjg: Looks like we don't need this: http://stackoverflow.com/questions/4417923/html-meta-tag-for-content-language

It can be removed.

perjg:
LGTM

PR #2684, opened 2016-08-04
Fix translations for pt-br and zw-tw lang

kvas-damian:
## Description

We export `pt-br` as default `pt` lang and we use it for all dialects of `pt` lang that are not defined. When we use old translation file for `pt-br` we end up having missing translations. Removing this old directory solves the problem. The same situation applies to `zh-tw`.

## Reviewers
@bkoval @rogatty


rogatty:
LGTM

PR #2687, opened 2016-08-04
SER-256 fix image-review layout

KolorowyAleksander:
## Links

* http://wikia-inc.atlassian.net/browse/SER-256

## Description
Image review layout fixed, the images are now centered. Added a button to get unreviewed queue aswell.


front/main/app/routes/image-review.js
```diff
@@ -53,6 +53,16 @@ export default Route.extend({
...
 		},
 
+		getUnreviewedOnly() {
+			const model = this.modelFor('imageReview');
+			window.scrollTo(0, 0);
+
+			ImageReviewModel.endSession(model.contractId);
+
+			this.set('status', 'UNREVIEWED');
+			this.refresh();
```
rogatty: How about a function that takes status as an argument and does all those things to keep it DRY?

front/main/app/styles/component/_image-review.scss
```diff
@@ -1,38 +1,72 @@
...
+
+			.img-container {
+				padding: 5px;
+				img {
+					min-width: 250px;
+					min-height: 250px;
+				}
+			}
+
+			.image-review-button-group {
```
rogatty: Does it need to be so nested? How about keeping it as flat as possible?
E.g. `.flagged` should be inside `.image-review-item` because it's not very specific but `.image-review-item` doesn't need to be inside `.image-review-list`, right?

front/main/app/routes/image-review.js
```diff
@@ -40,7 +40,7 @@ export default Route.extend({
 
 		reviewAndGetMoreImages() {
 			const model = this.modelFor('imageReview');
-
+			Logger.error('contractId', model.contractId);
```
rogatty: Doesn't look like error.

front/main/app/styles/component/_image-review.scss
```diff
@@ -1,13 +1,34 @@
...
+.image-review-text {
+	padding: 1em;
+	display: block;
+
+	p {
+		margin: 0.5em 0;
+	}
+}
+
+.img-container {
```
rogatty: Please add `image-review-` prefix or put it inside some other more precisely scoped class.

rogatty:
Two things and will LGTM

rogatty:
LGTM

PR #2689, opened 2016-08-04
XW-1812 | track footer clicks

Warkot:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1812

## Description

We need to know the amount of clicks on the Full site link, so why not to check all links in current footer?

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/components/wikia-footer.js
+++ b/front/main/app/components/wikia-footer.js
@@ -74,7 +74,7 @@ export default Ember.Component.extend({
 
 			Ember.run.later(() => {
 				window.location = href;
-			}, 100)
+			}, 100);
 		}
 	},
 
--- a/front/main/app/components/wikia-footer.js
+++ b/front/main/app/components/wikia-footer.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import {getDomain} from '../utils/domain';
+import {track, trackActions} from '../../common/utils/track';
 
 export default Ember.Component.extend({
 	classNames: ['wikia-footer'],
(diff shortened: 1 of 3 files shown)


front/main/app/components/wikia-footer.js
```diff
@@ -61,6 +62,11 @@ export default Ember.Component.extend({
 		 * @returns {void}
 		 */
 		handleFooterLinkClick(text, href) {
+			track({
```
rogatty: Don't we need to add some delay to let tracking request being sent before opening new page?

front/main/app/components/wikia-footer.js
```diff
@@ -61,6 +62,11 @@ export default Ember.Component.extend({
...
+				action: trackActions.click,
+				category: text
```
kvas-damian: What do you think about putting `text` into label and create category that has name related to footer?

kvas-damian: BTW using text will be really hard to analyze when users start using it in different languages

Warkot: Would `category: 'footer'` be ok?

Warkot: `text` is actually the message key—you can see it in the file few lines above.

kvas-damian: `footer` sounds good to me.
:+1: for `text`

front/main/app/components/wikia-footer.js
Warkot: What do you think now? Or maybe I should move the `later` logic to the `track` function, as where you pointed me in https://github.com/Wikia/app/blob/dev/resources/wikia/modules/tracker.js#L318-L330?

rogatty:
LGTM

tomasznapieralski:
LGTM

PR #2690, opened 2016-08-04
RECIRC-104: Setup Liftigniter placement test

pauloslund:
## Links

* https://wikia-inc.atlassian.net/browse/RECIRC-104
* #2659

## Description

Liftigniter is a content recommendation service that we have been seeing decent results with on desktop so we are extending the test to mobile. The experiment will be run for ~two weeks.

Reviewed and back-ported from #2659

## Reviewers

@Wikia/x-wing


rogatty:
LGTM

PR #2675, opened 2016-08-03
Aden 3696 intrestitial mercury - Proof of concept

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3696
* https://github.com/Wikia/app/pull/10998

## Description

Make INVISIBLE_HIGH_IMPACT_2 work on Mercury

PR #2695, opened 2016-08-05
Aden 3712 interstitial on transition

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3712
* Dependent on:
    * https://github.com/Wikia/app/pull/10998
    * https://github.com/Wikia/mercury/pull/2675

## Description

Show interstitial ad when user clicks a link

## Reviewers


front/common/modules/ads.js
```diff
@@ -290,11 +296,11 @@ class Ads {
 	/**
 	 * Adds ad slot
 	 *
-	 * @param {string} name - name of the slot
+	 * @param {AdSlot} adSlot - name of the slot
 	 * @returns {number} index of the inserted slot
 	 */
-	addSlot(name) {
-		return this.adSlots.push([name]);
+	addSlot(adSlot) {
```
hakubo: is it ad slot itself? 
name of it?
dom element?
a bird? :)

front/main/app/components/ads/invisible-high-impact-2.js
```diff
@@ -0,0 +1,13 @@
+const {Component, inject, computed} = Ember;
```
hakubo: import Ember!

front/main/app/styles/module/_ads.scss
```diff
@@ -82,3 +82,62 @@
 .wikia-ad-iframe {
 	border: 0;
 }
+
+#InvisibleHighImpactWrapper.interstitial {
```
hakubo: are this our
we should not use IDs in scss
if you have a good reason to do so
comment it please

front/main/app/templates/components/ads/invisible-high-impact-2.hbs
```diff
@@ -0,0 +1,11 @@
+{{#unless noAds}}
+	<div id="InvisibleHighImpactWrapper" class="hidden">
+		<div class="background"></div>
+		<a class="close">
+			{{svg 'close' role='img' class='icon close-button'}}
+		</a>
+		<div class="label">Advertisement</div>
+
+		<div id="{{name}}" class="ad-slot hidden {{if noAds "no-ads"}} {{nameLowerCase}}"></div>
```
hakubo: given +{{#unless noAds}}
will this be ever true? be {{if noAds "no-ads"}}

front/common/modules/ads.js
bognix: see the doc - there is definition of ad slot, if it's not enough then what do you expect?

front/main/app/components/ads/invisible-high-impact-2.js
bognix: decide on something! I am lost which convention is "the current one"

front/main/app/templates/components/ads/invisible-high-impact-2.hbs
bognix: good point 👍

front/main/app/styles/module/_ads.scss
bognix: right now this styles are copied and pasted from Oasis. I would like to use modal component here so ID styling shouldn't be needed. On the other hand we hare rule above

```
.ad-slot[id^="INVISIBLE_"] {
	display: none !important;
}
```
which mess up my rules :(

front/common/modules/ads.js
```diff
@@ -305,7 +311,7 @@ class Ads {
 	 */
 	removeSlot(name) {
 		this.adSlots = $.grep(this.adSlots, (slot) => {
-			return slot[0] && slot[0] === name;
+			return slot[0].name ? (slot[0].name === name) : (slot[0] && slot[0] === name);
```
bognix: do we need to support both versions? maybe we should only support adding slot as an object not as a string?
(question do AdEng)

front/common/modules/ads.js
```diff
@@ -348,11 +354,11 @@ class Ads {
...
 	 * Adds ad slot
 	 *
-	 * @param {string} name - name of the slot
+	 * @param {AdSlot} adSlot - object representing ad slot
 	 * @returns {number} index of the inserted slot
 	 */
-	addSlot(name) {
-		return this.adSlots.push([name]);
+	addSlot(adSlot) {
+		return this.adSlots.push([adSlot]);
```
fraszczakszymon: Is array needed here?

front/common/modules/ads.js
fraszczakszymon: This code is used in mercury repo only. Feel free to remove that

front/main/app/components/ad-slot.js
```diff
@@ -54,13 +36,19 @@ export default Component.extend(
 
 			if (this.get('isAboveTheFold')) {
 				Logger.info('Injected ad', name);
-				ads.addSlot(name);
+				ads.addSlot({
+					slotName: name,
+					onSuccess: this.get('onSuccess')
```
fraszczakszymon: Should we define this property in component?

front/main/app/services/ads-high-impact.js
```diff
@@ -0,0 +1,34 @@
...
+		});
+	},
+
+	/**
+	 * This method is called by callAllCallbacks method defined in mercury_ads_js package from MW.
+	 * This method's context is object created in adSlot.js in MW (from mercury_ads_js),
+	 * hence `this` is different than expected
+	 */
+	onSuccess() {
+		const iframe = document.getElementById(this.name).querySelector('div:not(.hidden) > div[id*="_container_"] iframe'),
```
fraszczakszymon: What do you think about using require/calling ads module here to get `ext.wikia.adEngine.slotTweaker` module and call `adjustIframeByContentSize` on it? (instead of reimplementing this)

fraszczakszymon:
LGTM 👍

PR #2696, opened 2016-08-05
SOC-2843 | fix: categories are preserved on mobile when sorting is ch…

mleczey:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2843

## Description

Selected categories are reset to 'All' on mobile when sort order is changed

## Reviewers

@bkoval


bkoval:
LGTM 👏

PR #2699, opened 2016-08-08
SOC-2802 | fix: prevent scrolling when edit categories modal is shown…

mleczey:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2802

Inspired by:
* http://stackoverflow.com/questions/9280258/prevent-body-scrolling-but-allow-overlay-scrolling
## Description

Fix for editing categories display incorrectly on mobile.

## Reviewers

@bkoval

### DIFF
--- a/front/main/app/styles/state/_full-screen.scss
+++ b/front/main/app/styles/state/_full-screen.scss
@@ -2,6 +2,7 @@
 	.mobile-full-screen {
 		height: 100%;
 		overflow: hidden;
+		position: fixed;
 
 		.site-head {
 			display: none;


rogatty:
Would it make sense to use  https://github.com/Wikia/mercury/blob/d17ecbda83dd041a0b5afb2c6fd23ec24ef1a36d/front/main/app/mixins/no-scroll.js? I think it solves the same problem.

bkoval:
@mleczey Let's give a shot @rogatty 's solution and if it fails to help in our case we'll continue with this one 🎃

bkoval:
LGTM

PR #2702, opened 2016-08-09
XW-1713 | add LyricFind tracking

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1713

## Reviewers

@Wikia/x-wing


front/main/app/routes/wiki-page.js
```diff
@@ -160,6 +162,36 @@ export default Ember.Route.extend(RouteWithAdsMixin, HeadTagsDynamicMixin, {
...
+					method: 'track',
+					title: model.title,
+					amgid: amgId,
+					gracenoteid: gracenoteId,
+					rand: (`${Math.random()}`).substr(2, 8)
+				},
+				dataType: 'text'
+			}).catch((error) => {
+				/**
+				 * MediaWiki is taking care of handling errors.
```
kvas-damian: What if for some reason we're not able to hit MW server? I'd at least log it.

tomasznapieralski: New Relic will do this for us (i spoke with @rogatty)

rogatty: Not sure if this one won't be ignored. See https://github.com/sir-dunxalot/ember-new-relic/blob/master/addon/instance-initializers/new-relic.js for the exact implementation.

front/main/app/routes/wiki-page.js
```diff
@@ -160,6 +162,36 @@ export default Ember.Route.extend(RouteWithAdsMixin, HeadTagsDynamicMixin, {
...
+				 * All we have to do is to send page view to MediaWiki.
```
rogatty: Please add all the info we have. E.g. that the 404s are normal when there is no lyrics.

front/main/app/routes/wiki-page.js
```diff
@@ -89,6 +90,7 @@ export default Ember.Route.extend(RouteWithAdsMixin, HeadTagsDynamicMixin, {
 			if (handler) {
 				transition.then(() => {
 					this.updateTrackingData(model);
+					this.sendLyricsPageView(handler, model);
```
rogatty: How about `handler.trackPageView(model)`? If it's only for articles then IMO it should live in `ArticleHandler`.
We wouldn't need `if (handler.controllerName === 'article'`

front/main/app/routes/wiki-page.js
```diff
@@ -160,6 +162,36 @@ export default Ember.Route.extend(RouteWithAdsMixin, HeadTagsDynamicMixin, {
 	},
 
 	/**
+	 * This function tracks page view only on articles on Lyrics Wiki (id: 43339)
+	 *
+	 * @param {Object} handler
+	 * @param {Ember.Model} model
+	 */
+	sendLyricsPageView(handler, model) {
```
rogatty: I'd extract this method to a separate util and just call it from `ArticleHandler`.

front/main/app/utils/wiki-handlers/article.js
```diff
@@ -16,11 +19,50 @@ function afterModel(route, model) {
...
+			gracenoteId = parseInt($('#gracenoteid').text(), 10) || 0;
+
+		request(M.buildUrl({path: '/wikia.php'}), {
+			data: {
+				controller: 'LyricFind',
+				method: 'track',
+				title: model.title,
+				amgid: amgId,
+				gracenoteid: gracenoteId,
+				rand: (`${Math.random()}`).substr(2, 8)
```
rogatty: What's this for?

front/main/app/utils/wiki-handlers/base-handler.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * This object acts like the interface, but implementation of fields ond methods is optional
```
rogatty: Typo

front/main/app/utils/wiki-handlers/article.js
```diff
@@ -16,11 +19,50 @@ function afterModel(route, model) {
...
+ */
+function afterTransition(model) {
+	sendLyricsPageView(model);
+}
+
+/**
  * Export Article handler
  */
-export default {
+export default Ember.$.extend({}, BaseHandler, {
```
rogatty: I'd like to hear @hakubo and @kvas-damian thoughts about this solution versus class inheritance.

front/main/app/utils/wiki-handlers/article.js
```diff
@@ -16,11 +19,50 @@ function afterModel(route, model) {
...
 
 /**
+ * This function tracks page view only on articles on Lyrics Wiki (id: 43339)
+ *
+ * @param {Ember.Model} model
+ */
+function sendLyricsPageView(model) {
+	if (Ember.get(Mercury, 'wiki.id') === 43339) {
+		const amgId = parseInt($('#lyric').data('amg-id'), 10) || 0,
+			gracenoteId = parseInt($('#gracenoteid').text(), 10) || 0;
```
rogatty: If we want to use jQuery it's more complicated than that. Article content is injected into DOM in `components/article-content.js`:
https://github.com/Wikia/mercury/blob/dev/front/main/app/components/article-content.js#L122-L137
and it happens after this code is run.

Did you check if it works on consecutive page views? If it did work for you then I think it just won a race and is not reliable.

front/main/app/utils/wiki-handlers/article.js
tomasznapieralski: Concept taken from https://github.com/Wikia/app/blob/dev/extensions/3rdparty/LyricWiki/LyricFind/js/modules/LyricFind.Tracker.js

front/main/app/utils/wiki-handlers/article.js
hakubo: why exactly do we need basehandler? what has changed that we require it now?

front/main/app/utils/wiki-handlers/article.js
```diff
@@ -16,11 +19,50 @@ function afterModel(route, model) {
...
+
+		request(M.buildUrl({path: '/wikia.php'}), {
+			data: {
+				controller: 'LyricFind',
+				method: 'track',
+				title: model.title,
+				amgid: amgId,
```
hakubo: http://es6katas.org/
```
{
amdid,
gracenoteid
}
```

hakubo: omg or actually omG

maybe instead of a var, grab the data in place? the code will be more succinct

front/main/app/utils/wiki-handlers/article.js
SeanColombo: The rand param was someone trying to avoid caching.

tomasznapieralski:
@hakubo we have `base-handler.js` because wiki-page route is using `afterTansition` hook which is implemented in `article-handler` but not in `category-handler`. All handlers extends `base-handler` which makes code more secure

rogatty:
LGTM

front/main/app/utils/wiki-handlers/article.js
```diff
@@ -16,11 +19,51 @@ function afterModel(route, model) {
...
+			data: {
+				controller: 'LyricFind',
+				method: 'track',
+				title,
+				amgid: 0,
+				gracenoteid: 0,
+				rand: (`${Math.random()}`).substr(2, 8)
+			},
+			dataType: 'text'
+		}).catch((error) => {
```
hakubo: what is this catch for?
What is a point for adding empty callback?

tomasznapieralski: if we do not catch error it will be thrown and visible in console

kvas-damian:
LGTM

front/main/app/utils/wiki-handlers/article.js
Warkot: I don't see what's wrong with that.

PR #2703, opened 2016-08-09
SOC-categories | fix: add object check

mleczey:
## Description

Fix for previously stored query parameters as "null"

## Reviewers

@bkoval


bkoval:
LGTM

PR #2730, opened 2016-08-24
JPN-584 Embeddable discussions tag for Mercury

perjg:
## Links

* https://wikia-inc.atlassian.net/browse/JPN-584
* Related issue for Oasis: https://wikia-inc.atlassian.net/browse/JPN-443

## Description

Mercury support for EmbeddableDiscussions extension developed in JPN-443.
Renders discussions posts inline in an article.

## Reviewers

@rogatty @hakubo @bkoval @rafalkalinski @d34th4ck3r

### DIFF
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -1,0 +1,49 @@
+import Ember from 'ember';
+
+import DiscussionForumModel from '../models/discussion/forum';
+
+import DiscussionContributionModelMixin from '../mixins/discussion-contribution-model';
+
+export default Ember.Component.extend(
+  DiscussionContributionModelMixin,
+	{
+		classNames: ['widget-discussions'],
+		layoutName: 'components/widget-discussions',
+		data: null,
+		upvotingInProgress: {},
+
+		/**
+		 * @returns {void}
+		 */
+		didInsertElement() {
+			const posts = DiscussionForumModel.find(Mercury.wiki.id, [], this.get('show'));
+
+			$('.widget-discussions').addClass('forum-wrapper discussion forum');
+
+			posts.then((result) => {
+				const entities = result.data.entities.slice(0, this.get('itemCount'));
+				this.set('posts', entities);
+				this.set('title', i18n.t('main.discussions-header-title', {ns: 'discussion'}));
+				this.set('view-all', i18n.t('main.all-discussions-link-mobile', {ns: 'discussion'}));
+			});
+		},
+
+		actions: {
+			upvote(entity) {
+				this.upvote(entity);
+			},
+
+			// Placeholders. These action must be present for discussion-post-card-detail to render,
+			// but are not used directly by this component
+			delete() {},
+			lock() {},
+			openEditEditor() {},
+			report() {},
+			undelete() {},
+			unlock() {},
+			setEditorActive() {},
+			reply() {},
+			share() {},
+		}
+	}
+);
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -2,8 +2,10 @@ import Ember from 'ember';
 
 import DiscussionForumModel from '../models/discussion/forum';
 import WidgetDiscussionsModel from '../models/widget-discussions';
+import InViewportMixin from 'ember-in-viewport';
 
 export default Ember.Component.extend(
+	InViewportMixin,
 	{
 		classNames: ['widget-discussions'],
 		classNameBindings: ['forumWrapper', 'discussion', 'forum'],
@@ -14,22 +16,14 @@ export default Ember.Component.extend(
 		data: null,
 		isLoading: true,
 
-		model: Ember.computed('model', () => {
+		model: Ember.computed(() => {
 			return WidgetDiscussionsModel.create();
-		}),
-
-		title: Ember.computed('title', () => {
-			return i18n.t('main.discussions-header-title', {ns: 'discussion'});
-		}),
-
-		viewAllPosts: Ember.computed('viewAllPosts', () => {
-			return i18n.t('main.all-discussions-link-mobile', {ns: 'discussion'});
 		}),
 
 		/**
 		 * @returns {void}
 		 */
-		didInsertElement() {
+		didEnterViewport() {
 			const category = this.get('category'),
 				posts = DiscussionForumModel.find(
 					Mercury.wiki.id,
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -1,20 +1,22 @@
 import Ember from 'ember';
 
 import DiscussionForumModel from '../models/discussion/forum';
-
-import DiscussionContributionModelMixin from '../mixins/discussion-contribution-model';
+import WidgetDiscussionsModel from '../models/widget-discussions';
 
 export default Ember.Component.extend(
-  DiscussionContributionModelMixin,
 	{
-		classNameBindings: ['forumWrapper', 'discussion', 'forum'],
 		classNames: ['widget-discussions'],
-		data: null,
+		classNameBindings: ['forumWrapper', 'discussion', 'forum'],
 		discussion: true,
 		forumWrapper: true,
 		forum: true,
+
+		data: null,
 		isLoading: true,
-		upvotingInProgress: {},
+
+		model: Ember.computed('model', () => {
+			return WidgetDiscussionsModel.create();
+		}),
 
 		title: Ember.computed('title', () => {
 			return i18n.t('main.discussions-header-title', {ns: 'discussion'});
@@ -45,7 +47,7 @@ export default Ember.Component.extend(
 
 		actions: {
 			upvote(entity) {
-				this.upvote(entity);
+				this.get('model').upvote(entity);
 			},
 
 			// Placeholders. These action must be present for discussion-post-card-detail to render,
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -13,13 +13,14 @@ export default Ember.Component.extend(
 		discussion: true,
 		forumWrapper: true,
 		forum: true,
+		isLoading: true,
 		upvotingInProgress: {},
 
 		title: Ember.computed('title', () => {
 			return i18n.t('main.discussions-header-title', {ns: 'discussion'});
 		}),
 
-		viewAllPosts: Ember.computed('title', () => {
+		viewAllPosts: Ember.computed('viewAllPosts', () => {
 			return i18n.t('main.all-discussions-link-mobile', {ns: 'discussion'});
 		}),
 
@@ -27,11 +28,18 @@ export default Ember.Component.extend(
 		 * @returns {void}
 		 */
 		didInsertElement() {
-			const posts = DiscussionForumModel.find(Mercury.wiki.id, [], this.get('show'));
+			const category = this.get('category'),
+				posts = DiscussionForumModel.find(
+					Mercury.wiki.id,
+					category ? [category] : [],
+					this.get('show')
+				);
 
 			posts.then((result) => {
-				const entities = result.data.entities.slice(0, this.get('itemCount'));
-				this.set('posts', entities);
+				this.setProperties({
+					posts: result.data.entities.slice(0, this.get('itemCount')),
+					isLoading: false,
+				});
 			});
 		},
 
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -7,10 +7,21 @@ import DiscussionContributionModelMixin from '../mixins/discussion-contribution-
 export default Ember.Component.extend(
   DiscussionContributionModelMixin,
 	{
+		classNameBindings: ['forumWrapper', 'discussion', 'forum'],
 		classNames: ['widget-discussions'],
-		layoutName: 'components/widget-discussions',
 		data: null,
+		discussion: true,
+		forumWrapper: true,
+		forum: true,
 		upvotingInProgress: {},
+
+		title: Ember.computed('title', () => {
+			return i18n.t('main.discussions-header-title', {ns: 'discussion'});
+		}),
+
+		viewAllPosts: Ember.computed('title', () => {
+			return i18n.t('main.all-discussions-link-mobile', {ns: 'discussion'});
+		}),
 
 		/**
 		 * @returns {void}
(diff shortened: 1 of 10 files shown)


front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,49 @@
...
+
+import DiscussionForumModel from '../models/discussion/forum';
+
+import DiscussionContributionModelMixin from '../mixins/discussion-contribution-model';
+
+export default Ember.Component.extend(
+  DiscussionContributionModelMixin,
+	{
+		classNames: ['widget-discussions'],
+		layoutName: 'components/widget-discussions',
```
bkoval: Seems like this .hbs should be automagically found by Ember without adding path explicitly here.

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,49 @@
+import Ember from 'ember';
+
+import DiscussionForumModel from '../models/discussion/forum';
+
+import DiscussionContributionModelMixin from '../mixins/discussion-contribution-model';
+
+export default Ember.Component.extend(
+  DiscussionContributionModelMixin,
```
bkoval: I don't think it's a good idea for a component to extend a Model. Probably what you want to do is to extend model for a currently used route with this Mixin and then call it's methods through action passing. What do yout hink @hakubo @rogatty ?

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,49 @@
...
+		data: null,
+		upvotingInProgress: {},
+
+		/**
+		 * @returns {void}
+		 */
+		didInsertElement() {
+			const posts = DiscussionForumModel.find(Mercury.wiki.id, [], this.get('show'));
+
+			$('.widget-discussions').addClass('forum-wrapper discussion forum');
```
bkoval: You should not use jQuery here. Instead, please take a look at https://guides.emberjs.com/v1.10.0/components/customizing-a-components-element/ - especially the part about classNameBindings (dynamically changing classNames).

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,49 @@
...
+
+			posts.then((result) => {
+				const entities = result.data.entities.slice(0, this.get('itemCount'));
+				this.set('posts', entities);
```
bkoval: Use setProperties for multiple setters at once.

front/main/app/styles/module/article/_discussions-widget.scss
```diff
@@ -0,0 +1,36 @@
+
+$discussions-widget-heading-background: #f6f6f6;
```
bkoval: Move this color declaration to _settings.scss

front/main/app/styles/module/article/_discussions-widget.scss
```diff
@@ -0,0 +1,36 @@
...
+		height: 70px;
+	}
+
+	.heading {
+		font-size: 27px;
+		padding-left: 18px;
+	}
+
+	.heading-viewall {
+		color: blue;
```
bkoval: Move this color declaration to _settings.scss

front/main/app/styles/module/article/_discussions-widget.scss
```diff
@@ -0,0 +1,36 @@
...
+		margin-left: auto;
+		padding-right: 14px;
+	}
+
+	.loading-overlay {
+		padding-top: 20px;
+		position: relative;
+	}
+
+	margin: 0 -$article-horizontal-padding;
```
bkoval: Direct styles of a class should go to the top of the class declaration in scss. Also, why would you want to expand the widget outside of the article?

front/main/app/templates/components/widget-discussions.hbs
```diff
@@ -0,0 +1,35 @@
+{{#if posts}}
+	<div class="heading-wrapper">
+		<div class="heading">
```
bkoval: Please, use h<1/2/3/4/5/6> for semantics.

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,49 @@
...
+				this.set('title', i18n.t('main.discussions-header-title', {ns: 'discussion'}));
```
bkoval: Now this doesn't require any data to be rendered so it can be a class property, calculated once for all instances of this component. Same with the line below.

front/main/app/styles/module/article/_discussions-widget.scss
```diff
@@ -0,0 +1,36 @@
+
+$discussions-widget-heading-background: #f6f6f6;
+
+.widget-discussions {
+	.heading-wrapper {
+		svg {
+			fill: blue;
```
bkoval: Move this color declaration to _settings.scss

front/main/app/templates/components/widget-discussions.hbs
```diff
@@ -0,0 +1,35 @@
+{{#if posts}}
+	<div class="heading-wrapper">
```
bkoval: Please, use <header> for semantics.

front/main/app/templates/components/widget-discussions.hbs
```diff
@@ -0,0 +1,35 @@
...
+		<div class="heading">
+			{{title}}
+		</div>
+		{{#link-to 'discussion' class='heading-viewall'}}
+			{{view-all}}{{svg 'arrow-right-long'}}
```
bkoval: What is "view-all"? Is it a label? A short text? Maybe this name could be more descriptive? :) Also, I think it would be a good idea to set a proper viewBox on the svg.

bkoval:
I'm hesitant on a "component being it's own model" aproach. I think we should care about separating the data layer from controller / view (in this case, a component). Least we could do here is to create DiscussionWidgetModel extending ContributionModelMixin and then use it as a "model" property on this component. Frankly, I'd try extending models for routes where this component will be rendered with this mixin and pass all the actions through .hbs, since this is a default way to go with models we have.

front/main/app/styles/module/article/_discussions-widget.scss
perjg: That is how it is specified in the Zeplin design:
https://app.zeplin.io/project.html#pid=56c75e4b1e0e6b5360844252&sid=57a270f23ed01b2d507f8223

I believe the design came from Richard Combs. If you think doing it like this is a bad idea, could you comment in the Jira issue and tag him?

front/main/app/templates/components/widget-discussions.hbs
perjg: If I use h1,h2 here I get all the global styles for headers included, since this is inside an article. I just want a plain style defined from scratch, which is why I went with plain divs.

perjg:
@bkoval I've taken care of most of your minor comments.

Next, I'll try to create a DiscussionWidgetModel as you suggest and set it as the model for the widget-discussion model.

perjg:
@bkoval I've addressed your main concern by creating a model that inherits from DiscussionContributionModelMixin, which I then added as a model property to the component. This avoids the issue of having the component inheriting from an unrelated model class.

Is this solution ok?

I've left another couple of comments too regarding your request to use h tags instead of div and why margins were changed.

front/main/app/templates/components/widget-discussions.hbs
bkoval: I believe semantic HTML is important - for SEO (to help crawlers understand our content properly), for screen readers for disabled people etc. I don't think overwriting default styles should be a problem.

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,66 @@
...
+		classNames: ['widget-discussions'],
+		classNameBindings: ['forumWrapper', 'discussion', 'forum'],
+		discussion: true,
+		forumWrapper: true,
+		forum: true,
+
+		data: null,
+		isLoading: true,
+
+		model: Ember.computed('model', () => {
```
rogatty: `model: Ember.computed('model'` doesn't really make sense.

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,66 @@
...
+			return WidgetDiscussionsModel.create();
+		}),
+
+		title: Ember.computed('title', () => {
+			return i18n.t('main.discussions-header-title', {ns: 'discussion'});
+		}),
+
+		viewAllPosts: Ember.computed('viewAllPosts', () => {
+			return i18n.t('main.all-discussions-link-mobile', {ns: 'discussion'});
```
rogatty: There is no logic here so it fits better in the template.

front/main/app/styles/_settings.scss
```diff
@@ -64,6 +64,8 @@ $discussion-filters-bullet-active-border: #fff;
 $discussion-post-actions-border-color: #e6e6e6;
 $discussion-error-color: #ed1c24;
 $discussion-submit-icon-color: #5ba200;
+$discussions-widget-heading-background: #f6f6f6;
```
rogatty: `discussion` vs `discussions`, I think it's better to be consistent.

front/main/app/components/widget-discussions.js
```diff
@@ -0,0 +1,66 @@
...
+			});
+		},
+
+		actions: {
+			upvote(entity) {
+				this.get('model').upvote(entity);
+			},
+
+			// Placeholders. These action must be present for discussion-post-card-detail to render,
+			// but are not used directly by this component
```
rogatty: @bkoval is it possible to modify `discussion-post-card-detail` so we don't need these empty functions? It doesn't feel right.

rogatty:
>I'd try extending models for routes where this component will be rendered (...), since this is a default way to go with models we have.

Actually, I think we should go with the solution proposed by @perjg. In this case it doesn't make a lot of sense to load discussions model in the article route. They aren't really related, right? Article is just a place that we want to render something based on a params different than the URL.

If we put it in the route what would happen when there is more than one widget instance? Also, I think it's perfectly fine to load an article without discussions data and lazy load the model after user scrolls to the widget.

I suggest we add lazy loading to improve the performance. @perjg you can use https://github.com/DockYard/ember-in-viewport which we already have implemented in few places, e.g. in https://github.com/Wikia/mercury/blob/dev/front/main/app/mixins/recirculation-experiment.js which is similar to what you try to achieve.

Nice article about this problem:
https://emberigniter.com/should-components-load-data/

front/main/app/components/widget-discussions.js
bkoval: Not sure. TBH I'm not aware of a way to conditionally trigger actions only if they exist.

front/main/app/templates/components/widget-discussions.hbs
perjg: Got it, I'll change to use h1, h2 and then override the styles properly.

perjg:
@rogatty I take that to mean that you are ok with the solution in my last commit: i.e have models/widget-discussions.js which implements DiscussionContributionModelMixin and using this from the component for upvoting. Everything else is contained in the component.

I'll have a look at lazy loading after I take care of the other pending issues (if I run out of time I may push to have that handled as a separate issue though, but I'll comment here if that turns out to be the case).

front/main/app/components/widget-discussions.js
perjg: I may have misunderstood how to use Ember.computed. What would be the proper way of doing it? Can I just do
model: WidgetDiscussionsModel.create(),
?

rogatty: Not sure what are the consequences of creating object in a property like this. For sure you can use:
```
model: Ember.computed(() => WidgetDiscussionsModel.create())
```
Just don't make it dependant on itself.

Quoting http://emberjs.com/api/classes/Ember.ComputedProperty.html:
>By default the function backing the computed property will only be called once and the result will be cached. You can specify various properties that your computed property depends on. This will force the cached result to be recomputed if the dependencies are modified.

perjg: Thanks for clearing that up. I had made an incorrect assumption about the purpose of passing a string as the first parameter to Ember.computed. I'll just use what you suggested, that's exactly what I want :)

perjg:
@bkoval @rogatty I've taken care of all issues mentioned in the PR now.

perjg:
@rogatty, as for lazy loading:

I tried to extend InViewportMixin, but didEnterViewport / didExitViewport never triggers
(I tried with a long article with the discussions widget far down in the page).

Could this be because the Ember component is injected _after_ Mercury has initialized?

Also I have a concern: What if you put a discussions widget in the top of the page?
When Mercury loads, it will just be the empty placeholder text, after which it is replaced
by the Ember component. Since that part of the page is already visible at that point,
will didEnterViewport actually trigger?

rogatty:
>I tried to extend InViewportMixin, but didEnterViewport / didExitViewport never triggers
(I tried with a long article with the discussions widget far down in the page).

They never triggered because you overrode `didInsertElement` hook which is essential for ember-in-viewport, see https://github.com/pauloslund/ember-in-viewport/blob/e47d88c2e96eecadf1511a0c10c377819f19b35d/addon/mixins/in-viewport.js#L38

In your usecase it's enough to just rename `didInsertElement` to `didEnterViewport`. I've tried on my devbox and it works just fine.

For the future reference if you need to use `didInsertElement` without overriding it you can use `Ember.on('didInsertElement'`. For example see https://github.com/Wikia/mercury/blob/cce45d0418fba90368123edef4f93ab86edacc0e/front/main/app/components/ad-slot.js#L24

rogatty:
Will LGTM when renamed `didInsertElement` -> `didEnterViewport`

perjg:
@rogatty Changing didInsertElement to didEnterViewport works great! Thanks for helping with debugging.

front/main/app/styles/module/article/_discussion-widget.scss
```diff
@@ -0,0 +1,38 @@
+
+$article-negative-padding: -$article-horizontal-padding;
+
+.widget-discussions {
```
rogatty: I think this file would fit better in `styles/component/` directory as those are styles specifically for a component.
`styles/module/` is kinda thing of the past and harder to reason about.

perjg:
@rogatty I moved the styles file to the components directory.
I had to tweak the style overrides a bit to make sure it renders the same (article-content overrides)

front/main/app/styles/component/_discussion-widget.scss
```diff
@@ -0,0 +1,42 @@
...
+			vertical-align: middle;
+			width: 11px;
+		}
+
+		align-items: center;
+		background-color: $discussion-widget-heading-background;
+		display: flex;
+		height: 70px;
+		font-size: 12px;
+		padding-top: 0 !important;
```
rogatty: Is it possible to force these rules using stronger selectors? `!important` should be the last resort when everything else fails. If it's not possible then I think it's a lesser evil to have it in `/module/article`

perjg:
@rogatty Sorry, the !important was a mistake, I've removed them.

(I changed h1 to .heading-wrapper, which eliminates the need for the !important)

rogatty:
LGTM

PR #2747, opened 2016-08-29
Aden 3776 localize advertisement

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3776


front/common/public/locales/de/main.json
```diff
@@ -1,166 +1,168 @@
 {
-  "app": {
```
rogatty: Did you do some reformatting or was it downloaded like this from Crowdin?

bognix: It was downloaded like this from crowdin

fraszczakszymon:
LGTM 👍

PR #2756, opened 2016-09-06
ADEN-3857 | make slots on n-th pageviews viewable for DFP

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3857
* https://github.com/Wikia/app/pull/11277


front/common/modules/ads.js
```diff
@@ -65,6 +65,9 @@ class Ads {
 				dimension: 7
 			}
 		};
+		this.googleTag = {
```
bognix: this is only for unit tests purposes, I don't like this approach but couldn't think of anything better :(

The problem is that we have a `reset-ads` helper which calls `removeSlots` method. This method calls `this.googleTag.destroySlots` which is undefined. Best way would be to mock the googleTag inside this helper, any ideas are welcome

front/common/modules/ads.js
```diff
@@ -294,6 +297,11 @@ class Ads {
 				onContextLoadCallback();
 			}
 
+			// TODO get rid of this hack
+			if (typeof this.googleTag.newPageView === 'function') {
```
bognix: this thing is undefined when `reload` method is called for the first time, later "it becomes" defined.

front/common/tests/unit/modules/ads.js
```diff
@@ -38,16 +38,14 @@ QUnit.module('mercury/modules/ads', function (hooks) {
 		instance.trackKruxPageView = sinon.stub();
 
 		origRequire = window.require;
-		window.require = function (modules, callback) {
-			callback();
+		window.require = function (modules) {
```
bognix: I needed to remove callback because of `this.googleTag = helper.getGoogleTag()` line. Maybe you can think of something how I can mock/make this part work?

front/common/tests/unit/modules/ads.js
```diff
@@ -38,16 +38,14 @@ QUnit.module('mercury/modules/ads', function (hooks) {
...
 		};
 
 		instance.init(testAdsUrl);
 
 		window.require = origRequire;
 
 		assert.ok(loadStub.calledWith(testAdsUrl));
-		assert.ok(reloadWhenReadyStub.calledOnce);
```
bognix: this one had to be removed when I removed `callback` from `require` mock

front/main/app/mixins/ads.js
```diff
@@ -173,8 +162,7 @@ export default Ember.Mixin.create({
 				articleBodyHeight > adsData.moreInContentAds.minPageLength,
 
 			$globalFooter = $('.wds-global-footer');
-
-		this.clearAdViews();
+		this.set('adViews', []);
```
bognix: I was trying to get rid of adViews variable but it was used by something else as well. This ticket is in progress for so long that I decided not to change it right now.

front/main/app/routes/wiki-page.js
```diff
@@ -208,6 +209,7 @@ export default Ember.Route.extend(
 				// notify a property change on soon to be stale model for observers (like
 				// the Table of Contents menu) can reset appropriately
 				this.notifyPropertyChange('displayTitle');
+				this.get('ads').destroyInContentAds();
```
bognix: instead of removing ads before injecting new ones I decided to remove them when leaving the wiki-page route. This way transition process is smother because there are no ads running in background.

front/common/modules/ads.js
```diff
@@ -65,6 +65,11 @@ class Ads {
 				dimension: 7
 			}
 		};
+		this.adLogicPageViewCounterModule = null;
+		this.googleTagModule = {
```
bognix: this one here is done because we call this method from ember acceptance tests. I couldn't find a way to mock this property from those tests. Maybe `destroySlots` method should be defined as static one? it would be easier to remove it then.

front/main/app/routes/wiki-page.js
rogatty: What happens when new article can't be loaded? It's an edge case but I'm curious ;)

front/common/modules/ads.js
```diff
@@ -101,6 +106,7 @@ class Ads {
 					'ext.wikia.adEngine.mobile.mercuryListener',
 					'ext.wikia.adEngine.pageFairDetection',
 					'ext.wikia.adEngine.sourcePointDetection',
+					'ext.wikia.adEngine.provider.gpt.googleTag',
```
nandy-andy: Minor comment: alphabetical order, please :)

front/main/app/mixins/ads.js
```diff
@@ -18,28 +18,8 @@ export default Ember.Mixin.create({
...
-
-			// Inject at max this many extra slots:
-			maxSlots: 3,
-
-			// ... and name them:
-			slotNamePrefix: 'MOBILE_IN_CONTENT_EXTRA_'
-		}
 	},
-	adViews: [],
+	ads: Ember.inject.service(),
```
nandy-andy: Ember services magic. Looks clean :bath:

front/main/app/routes/wiki-page.js
nandy-andy: What is the `wiki-page` route? Is it only for articles and main pages?

rogatty: It's for all mediawiki pages that Mercury support. Currently it means articles and categories. "Classic main pages" fit into articles group.

front/main/app/mixins/ads.js
```diff
@@ -68,91 +48,8 @@ export default Ember.Mixin.create({
...
-			adView = this.adViews.pop();
-		}
-	},
-
-	/**
-	 * Inject MOBILE_IN_CONTENT_EXTRA_* ads on selected wikis
-	 *
-	 * @returns {void}
-	 */
-	injectMoreInContentAds() {
```
nandy-andy: Thank you for clean-up :gem:

front/common/tests/unit/modules/ads.js
```diff
@@ -75,41 +75,47 @@ QUnit.module('mercury/modules/ads', function (hooks) {
...
 		instance.adSlots = [];
+		instance.googleTagModule = {};
 	});
 
 	QUnit.test('Add/remove slots works', function (assert) {
-		var instance = Ads.getInstance();
+		var instance = Ads.getInstance(),
+			destroySlotsSpy = sinon.spy();
+
+		instance.googleTagModule = {
```
nandy-andy: :+1:

PR #2763, opened 2016-09-08
SEO-509 Prerender for main pages

gabrys:
## Links

* http://wikia-inc.atlassian.net/browse/SEO-509
* http://github.com/Wikia/hapi-prerender our fork of hapi-prerender that moves some of the logic from the plugin to the app. This way it's more extensible without requiring plugin updates
* https://wikia-inc.atlassian.net/wiki/display/SEO/Prerender+for+Curated+Main+Pages

## Description

Curated Main Pages have image-based navigation and the links do not load when JS is disabled, making it difficult or impossible for bots to move through this page. Our goal is to make these pages crawlable.

Need to add the actual logic of deciding if the page if covered by the test. Currently we only selected 3 wikis main pages (including /main/section/* and /main/category/* subpages). Eventually we'll want to run this for all CMPs.

## Reviewers

@vforge, @bognix, @saipetch, @rogatty


bognix:
just for the reference - this is PR with last implementation of prereder https://github.com/Wikia/mercury/pull/1931/files

server/app/lib/prerender.js
```diff
@@ -0,0 +1,115 @@
...
+
+function rewriteUrl(originalUrl) {
+    return originalUrl.replace('.127.0.0.1.xip.io:7000/', '.wikia.com/') + '?useskin=mercury';
+}
+
+export const prerender = require('hapi-prerender');
+export const prerenderOptions = {
+    shouldShowPrerenderedPage: shouldShowPrerenderedPage,
+    updateRequestedUrl: rewriteUrl,
+    token: 'MvKpNOouVLhkFSk9sJ7F'
```
bognix: consider using the similar approach to one from https://github.com/Wikia/mercury/pull/1931/files. Move this to `localSettings.base.js` and add something like:
```
prerenderToken: process.env.PRERENDER_TOKEN || 'MvKpNOouVLhkFSk9sJ7F'
```

server/app/lib/prerender.js
```diff
@@ -0,0 +1,115 @@
+// TODO: Store the actual token outside the code (the current doesn't have the billing info put to prerender.io)
+// TODO: Filter by URLs: only enable on bleach.wikia.com, dc.wikia.com and onepiece.wikia.com main pages
+// TODO: Also make the tiles in the sections actual proper <a href> links and make those sections index,follow
+
+function shouldShowPrerenderedPage(req) {
+    // googlebot, yahoo, and bingbot are not in this list because
+    // we support _escaped_fragment_ and want to ensure people aren't
+    // penalized for cloaking.
+    var crawlerUserAgents = [
```
bognix: one `var` per block and don't use `var` use `const`

server/app/lib/prerender.js
gabrys: I wouldn't even supply the default. I'm seeing `prerenderApiKey: process.env.PRERENDER_API`. Where is the environment set, is it defined in chef?

gabrys:
@bognix we have a different goal. Your goal was to get some gains in organic search pageviews, and our goal is just crawlability of those pages (including for non-Google bots).

Also I believe the previous experiment was replace _some_ () HTML with the proper prerendered HTML, the current experiment is to replace _no_ real HTML on the main pages with _some_ meaningful HTML

server/app/lib/prerender.js
vforge: @gabrys Yup - here was it added int he last test: https://github.com/Wikia/chef-repo/pull/8163/files

gabrys: Perfect, thanks for the info.

front/main/app/components/curated-content-item.js
```diff
@@ -67,6 +67,7 @@ export default Ember.Component.extend(
 
 			if (itemType && itemType === 'section' || itemType === 'category') {
 				this.sendAction('openCuratedContentItem', this.get('model'));
+				return false;
```
bognix: what do you want to achieve by returning false? Stop bubbling? It's not bubbling by default.
http://emberjs.com/api/classes/Ember.ActionHandler.html#toc_bubbling

server/app/lib/prerender.js
```diff
@@ -0,0 +1,68 @@
...
+		'linkedinbot',
+		'embedly',
+		'quora link preview',
+		'showyoubot',
+		'outbrain',
+		'pinterest',
+		'developers.google.com/\+/web/snippet'
+	].join('|'), 'i'),
+	urlsToPrerenderRegExp = new RegExp([
+		'^bleach\.[^/]*/wiki/Bleach_Wiki',
```
bognix: can't you simply change it to `^bleach.wikia.com/wiki/Bleach_Wiki` ?

what about stagings?
does `req.url.host` return `sandbox-*.bleach.wikia.com`?. I know we do some stuff with modifying(removing) staging prefixes for request's host header on Varnish level.

My question is if `req.url.host` returns the value with staging prefix or without it?

bognix:
@gabrys, regarding https://github.com/Wikia/mercury/pull/2763#issuecomment-245679892. For articles we were replacing the content but for main pages we were replacing nothing with something so doing exactly the same what you do here. I understand that your test's scope is smaller, hence it would be easier to understand the results.

front/main/app/components/curated-content-item.js
gabrys: Without it, the browser opens a new tab with the new section/category as well as opening the same thing in the same tab. So it's more `preventDefault` in this context than bubbling (if I understand correctly).

server/app/lib/prerender.js
gabrys: This is mostly for wikia-dev and xip.io scenarios.

This conditions will go away when we release the feature globally to all curated content, so I don't think it's worth much discussion :-)

gabrys:
@bognix yes, that's the understanding. We're also aiming to make the main pages crawlable by tools like Botify.

front/main/app/components/curated-content-item.js
rogatty: This is not an action but an event handler and I think it bubbles without `return false;` all the way to https://github.com/Wikia/mercury/blob/dev/front/main/app/components/application-wrapper.js#L91-L110. This change probably fixes `undefined` error which is visible in console on production for a long time now (just a message, doesn't break anything).

rogatty:
LGTM

front/main/app/components/curated-content-item.js
bognix: my bad, thanks for fixing! 👍

bognix:
LGTM

PR #2771, opened 2016-09-09
XW-1693 | Introduce Global Navigation from Design System for Desktop

Grunny:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1693

## Description

Introduce Global Navigation from Design System for Desktop.


front/main/app/components/global-navigation-link-authentication.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: '',
+	linkClasses: Ember.computed('model.title.key', function () {
+		const messageKey = this.get('model.title.key');
+
+		if (messageKey === 'global-navigation-anon-sign-in') {
```
kvas-damian: Maybe we could use `switch` here?

front/main/app/components/global-navigation-link-branded.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default Ember.Component.extend({
+	tagName: 'a',
+	attributeBindings: ['href'],
+	classNames: ['wds-global-navigation__link'],
+	classNameBindings: ['linkClassName'],
+
+	href: Ember.computed.oneWay('model.href'),
+	linkClassName: Ember.computed('model.brand', function () {
+		return `wds-is-${this.get('model.brand')}`;
```
kvas-damian: It'll be hard to be grepped. Maybe we could list all available possibilities in the comment?

front/main/app/components/global-navigation-account-navigation.js
```diff
@@ -0,0 +1,6 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: 'div',
```
hakubo: isn't that default?

front/main/app/templates/components/global-navigation-links.hbs
```diff
@@ -0,0 +1,21 @@
+<div class="wds-dropdown__toggle wds-global-navigation__dropdown-toggle">
+	{{#if showAvatar}}
+		<img class="wds-avatar wds-is-circle" src="{{model.header.url}}" alt="model.header.username.value" />
```
kvas-damian: Probably `alt` value should should be changed ;)

front/main/app/components/global-navigation-link-authentication.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
```
hakubo: const {Component, computed} = Ember;

front/main/app/components/global-navigation-link-authentication.js
hakubo: or an object?
```var myObj = {
  'global-navigation-anon-sign-in': 'wds-button wds-is-full-width'
  ...
}

return myObj[this.get('model.title.key')] || '';
```

front/main/app/components/global-navigation-link-branded.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember';
```
hakubo: const {Component, computed} = Ember;

front/main/app/components/global-navigation-link-dropdown.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
```
hakubo: const ... :)

front/main/app/components/global-navigation-links.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
```
hakubo: const ...

front/main/app/components/global-navigation.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: 'div',
+	classNames: ['wds-global-navigation'],
+	model: Ember.Object.create(M.prop('globalNavigationData'))
```
kvas-damian: I'm not sure about taking data from `M.prop()` at this level. Is it possible to move it to real model and pass this data through route object?

front/main/app/components/global-navigation.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: 'div',
```
hakubo: div is default

front/main/app/templates/components/global-navigation.hbs
```diff
@@ -0,0 +1,37 @@
...
+				{{svg 'wds-icons-cross' class='wds-icon wds-icon-small wds-global-navigation__search-close-icon'}}
+			</button>
+		</div>
+		<button class="wds-button wds-global-navigation__search-submit">
+			{{svg 'wds-icons-arrow' class='wds-icon wds-icon-small wds-global-navigation__search-submit-icon'}}
+		</button>
+	</form>
+	{{#if model.user}}
+		{{global-navigation-links model=model.user type='user-menu' dropdownRightAligned=true}}
+	{{else if model.anon}}
```
hakubo: if model has no user and no anon - what then?

tomasznapieralski: Then we will display nothing, but if that situation occurs then the API is broken, because it has to return one of 2 objects: anon|user

hakubo: should we have fallback then?

hakubo:
all files for global navigation could live in
components/global-navigation folder

front/main/app/templates/components/global-navigation.hbs
kvas-damian: I believe not displaying anything is good fallback here. We won't confuse users with wrong state of application.
Other items on nav will be still fully operational.

rogatty:
@hakubo there is **some** issue with keeping components in a subdirectory but I don't remember what it's exactly and possibly it doesn't matter in our case. Worth a try.

front/main/app/templates/components/global-navigation-links.hbs
```diff
@@ -0,0 +1,21 @@
+<div class="wds-dropdown__toggle wds-global-navigation__dropdown-toggle">
```
rogatty: We add a `title` attribute with the user name in the app:
https://github.com/Wikia/app/blob/ds-global-navigation/extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalNavigation_links.php#L4-L6

front/main/app/templates/components/global-navigation-links.hbs
```diff
@@ -0,0 +1,21 @@
...
+	{{#if showAvatar}}
+		<img class="wds-avatar wds-is-circle" src="{{model.header.url}}" alt="{{model.header.username.value}}" />
+	{{else}}
+		<span>{{i18n model.header.title.key ns='design-system'}}</span>
+	{{/if}}
+	{{svg 'wds-icons-dropdown-tiny' class='wds-icon wds-icon-tiny wds-dropdown__toggle-chevron'}}
+</div>
+<div class="wds-dropdown__content wds-global-navigation__dropdown-content {{extraDropdownClasses}}">
+	<ul class="wds-list">
```
rogatty: `wds-list wds-is-linked`

front/main/app/templates/components/global-navigation-links.hbs
```diff
@@ -0,0 +1,21 @@
...
+			<li>
+				{{#if (equal link.type 'link-authentication')}}
+					{{global-navigation-link-authentication model=link}}
+				{{else}}
+					{{global-navigation-link-dropdown model=link}}
+				{{/if}}
+			</li>
+		{{/each}}
+	</ul>
+</div>
```
rogatty: We're missing the logic which is in https://github.com/Wikia/app/blob/ds-global-navigation/extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalNavigation_links.php

Copied from [a comment](https://github.com/Wikia/app/pull/11246#discussion_r77150413):
- If section has a header - display a dropdown with a list of links
- If section doesn't have a header - display the list of links, doesn't matter if there is one or many (right now there will be only one for international wikis)

This way we have a flexible solution and are able to control this section easily from the API.

front/main/app/templates/components/global-navigation.hbs
```diff
@@ -0,0 +1,37 @@
+<div class="wds-global-navigation__content-bar">
+	<a href="http://fandom.wikia.com" class="wds-global-navigation__logo wds-global-navigation__logo-wikia">
```
rogatty: :speak_no_evil: 
`{{model.logo.header.href}}`
Also, we need a way to change the logo using the API, see https://github.com/Wikia/app/blob/ds-global-navigation/extensions/wikia/DesignSystem/services/templates/DesignSystemGlobalNavigation_logo.php

front/main/app/templates/components/global-navigation-links.hbs
Grunny: Ah, I hadn't added that because it wasn't used anywhere, but sure we can add that.

front/main/app/templates/components/global-navigation.hbs
```diff
@@ -0,0 +1,37 @@
...
+		{{/each}}
+	{{/if}}
+	{{global-navigation-links model=model.wikis type='wikis-menu'}}
+	<form class="wds-global-navigation__search" action="{{model.search.module.results.url}}">
+		<div class="wds-global-navigation__search-input-wrapper">
+			<label class="wds-global-navigation__search-label">
+				{{svg 'wds-icons-magnifying-glass' class='wds-icon wds-icon-small wds-global-navigation__search-label-icon'}}
+				<input class="wds-global-navigation__search-input" name="{{model.search.module.results.param-name}}" placeholder="{{i18n model.search.module.placeholder-inactive.key ns='design-system'}}"/>
+			</label>
+			<button class="wds-button wds-is-text wds-global-navigation__search-close" type="reset">
```
Warkot: I like the `type="reset"`. 👍  We might want to update that in Design System (app's feature branch already has this type).

front/main/app/templates/components/global-navigation.hbs
```diff
@@ -0,0 +1,37 @@
...
+				{{svg 'wds-icons-cross' class='wds-icon wds-icon-small wds-global-navigation__search-close-icon'}}
+			</button>
+		</div>
+		<button class="wds-button wds-global-navigation__search-submit">
```
Warkot: We have it neither in Design System or on the app's feature branch, but maybe we should add `type="submit"`?

front/main/app/templates/components/global-navigation-links.hbs
rogatty: Isn't this already required to properly render `wikis` section on non-english communities?

front/main/app/templates/components/global-navigation.hbs
```diff
@@ -0,0 +1,37 @@
...
+			{{svg 'wds-icons-arrow' class='wds-icon wds-icon-small wds-global-navigation__search-submit-icon'}}
+		</button>
+	</form>
+	{{#if model.user}}
+		{{global-navigation-links model=model.user type='user-menu' dropdownRightAligned=true}}
+	{{else if model.anon}}
+		{{global-navigation-account-navigation model=model.anon}}
+	{{/if}}
+	<div class="wds-global-navigation__start-a-wiki ">
```
Warkot: Just one space to much after the class name.

front/main/app/templates/components/global-navigation.hbs
rogatty: Quoting http://www.w3schools.com/tags/att_button_type.asp:
>Tip: Always specify the type attribute for the <button> element. Different browsers may use different default types for the <button> element.

server/app/facets/show-application.js
```diff
@@ -12,7 +12,7 @@ import {
 } from '../lib/custom-errors';
 import {isRtl, getUserId, getLocalSettings} from './operations/page-data-helper';
 import showServerErrorPage from './operations/show-server-error-page';
-import injectGlobalFooterData from '../lib/inject-global-footer-data';
+import injectGlobalFooterData from '../lib/inject-design-system-data';
```
rogatty: Import name should be updated too. The same in `show-curated-content.js`

front/main/app/styles/app.scss
```diff
@@ -22,8 +22,3 @@
 
 // Load Design System
 @import '../../bower_components/design-system/dist/styles';
-
-// Temporary solution that will prove that DS is included.
```
hakubo: :D

PR #2776, opened 2016-09-13
XW-2005 | implement sticky state to global navigation on discussions

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2005
* https://github.com/Wikia/design-system/pull/60

## Reviewers

@Wikia/x-wing


front/main/app/components/discussion-categories.js
```diff
@@ -178,6 +178,7 @@ export default Ember.Component.extend(
 				Ember.$('body').toggleClass('mobile-full-screen', shouldEnable);
 
 				this.set('isEditMode', shouldEnable);
+				this.triggerDiscussionCategoriesEditModeStateChange(shouldEnable);
```
rogatty:
```
const shortName = this.get('triggerDiscussionCategoriesEditModeStateChange');

if (typeof shortName === 'function') {
    shortName(shouldEnable);
}
```

front/main/app/styles/app.scss
```diff
@@ -1,6 +1,7 @@
 // DO NOT CHANGE:
 // load order of files below is intentional
 @import '../../bower_components/wikia-style-guide/src/scss/main';
+@import '../../bower_components/design-system/dist/styles';
```
rogatty: This causes DS styles more sensitive to be overriden. Are you sure nothing breaks (in the footer too)?

front/main/app/styles/component/design-system/index.scss
```diff
@@ -0,0 +1 @@
+@import 'global-navigation';
```
rogatty: Pull the latest feature branch, it was already done in https://github.com/Wikia/mercury/pull/2777

front/main/app/templates/components/discussion-inline-editor.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="discussion-inline-editor-floating-container">
+<div class="discussion-inline-editor-floating-container" style="top: {{floatingContainerTop}}px">
```
rogatty: :see_no_evil:

front/main/app/styles/app.scss
kvas-damian: I believe we should be able to adjust styles provided by design-system in some cases. It's not always possible to use components by the book and sometimes we need to override some of them. Additionally having those style higher on a list allows us to use variables exported by DS.

When it come to testing we'll perform QA when we're done.

front/main/app/templates/components/discussion-inline-editor.hbs
kvas-damian: Any ideas for better solution?
We wanted to avoid coupling nav and editor styles, so we don't need to update both when height of global nav changes. By moving it to JS we're able to dynamically get height of the element when we stick to. We can't use `attributeBindings` as this isn't the root element of this component.

rogatty: It's fine how it is.

front/main/app/templates/discussion/forum.hbs
```diff
@@ -82,6 +82,7 @@
 		postsDisplayed=model.current.data.entities.length
 		report=(action 'report')
 		setEditorActive=(action 'setEditorActive')
+		stickEditorToGlobalNav=application.isGlobalNavigationPositionFixed
```
rogatty: Let's use the same solution as with smart banner in base.js:
```
isGlobalNavigationPositionFixed: Ember.computed.oneWay('application.isGlobalNavigationPositionFixed'),
```

front/main/app/styles/component/design-system/_global-navigation.scss
```diff
@@ -1,13 +1,22 @@
...
 	box-sizing: content-box;
 
 	.dark-theme & label {
 		color: $color-dark-blue-gray;
 	}
+
+	&.global-navigation-fixed {
+		position: fixed;
+		width: 100%;
+		z-index: $z-side-nav;
```
rogatty: This seems to be a strange coupling. Why side nav?

front/main/app/styles/component/design-system/_global-navigation.scss
```diff
@@ -1,13 +1,22 @@
...
+}
+
 .wds-global-navigation {
 	box-sizing: content-box;
 
 	.dark-theme & label {
 		color: $color-dark-blue-gray;
 	}
+
+	&.global-navigation-fixed {
```
rogatty: How about a simple `&.is-fixed`?

rogatty:
Two minor things but generally LGTM

Grunny:
LGTM.

PR #2779, opened 2016-09-13
JPN-584 Style for discussions widget error messages

perjg:
## Links

* https://wikia-inc.atlassian.net/browse/JPN-584

## Description

A style to use for showing error messages when the <discussions /> tag can not be rendered due to incorrect parameters (e.g. setting columns=foo, where it should be either 1 or 2)

## Reviewers

@rogatty @bkoval

### DIFF
--- a/front/main/app/styles/_settings.scss
+++ b/front/main/app/styles/_settings.scss
@@ -65,8 +65,8 @@ $discussion-post-actions-border-color: #e6e6e6;
 $discussion-error-color: #ed1c24;
 $discussion-submit-icon-color: #5ba200;
 $discussion-widget-heading-background: #f6f6f6;
-$discussion-widget-errorcolor: #f00;
-$discussion-widget-linkcolor: #00f;
+$discussion-widget-error-color: #f00;
+$discussion-widget-link-color: #00f;
 
 $discussion-rail-title-color: #333;
 
--- a/front/main/app/styles/_settings.scss
+++ b/front/main/app/styles/_settings.scss
@@ -65,6 +65,7 @@ $discussion-post-actions-border-color: #e6e6e6;
 $discussion-error-color: #ed1c24;
 $discussion-submit-icon-color: #5ba200;
 $discussion-widget-heading-background: #f6f6f6;
+$discussion-widget-errorcolor: #f00;
 $discussion-widget-linkcolor: #00f;
 
 $discussion-rail-title-color: #333;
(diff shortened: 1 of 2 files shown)


front/main/app/styles/_settings.scss
```diff
@@ -65,6 +65,7 @@ $discussion-post-actions-border-color: #e6e6e6;
 $discussion-error-color: #ed1c24;
 $discussion-submit-icon-color: #5ba200;
 $discussion-widget-heading-background: #f6f6f6;
+$discussion-widget-errorcolor: #f00;
```
rogatty: `$discussion-widget-error-color`?

rogatty:
One minor thing, LGTM

perjg:
@rogatty In context:

$discussion-widget-errorcolor: #f00;
$discussion-widget-linkcolor: #00f;

If I change one I should probably change both: discussion-widget-error-color and discussion-widget-link-color ok?

rogatty:
Yup, all the other variables in this file use dasherized names.

perjg:
@rogatty Changed both variables to use dasherized names.

rogatty:
LGTM

front/main/app/styles/component/_discussion-widget.scss
```diff
@@ -40,3 +40,7 @@ $article-negative-padding: -$article-horizontal-padding;
 		padding-top: 0;
 	}
 }
+
+.widget-discussions-error {
```
bkoval: Hey, it seems like you're using a mix of "discussion-widget" and "widget-discussions" - is that the same thing? :)

bkoval:
LGTM

front/main/app/styles/component/_discussion-widget.scss
perjg: Yeah, I used widget-discussions for everything initially, but used discussion-widget for the colors I added in _settings.scss to make it line up better with existing names there.

bkoval: Not a blocker, but I'd think about a refinement at some point - would be nice to have the names in SCSS correspond to the component name :)

perjg: Understood. We're discussing some design changes for how error messages are shown, which will probably be done as a separate task after initial release, so I'll look into cleaning up the naming then.

PR #2780, opened 2016-09-13
XW-2028 Fix label color on dark theme

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2028

## Description

Override color set in https://github.com/Wikia/mercury/blob/dev/front/main/app/styles/theme/_dark.scss#L75

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/styles/component/design-system/_global-navigation.scss
+++ b/front/main/app/styles/component/design-system/_global-navigation.scss
@@ -2,6 +2,12 @@
 // If you look for Global Navigation styles go here:
 // https://github.com/Wikia/design-system/blob/master/components/_global-navigation.scss
 
+@import '../../../../bower_components/design-system/base/colors';
+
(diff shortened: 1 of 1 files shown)


Warkot:
LGTM

front/main/app/styles/component/design-system/_global-navigation.scss
```diff
@@ -2,6 +2,12 @@
 // If you look for Global Navigation styles go here:
 // https://github.com/Wikia/design-system/blob/master/components/_global-navigation.scss
 
+@import '../../../../bower_components/design-system/base/colors';
+
 .wds-global-navigation {
 	box-sizing: content-box;
+
+	.dark-theme & label {
```
kvas-damian: What do you think about limiting scope of the selector that we need to override?

PR #2786, opened 2016-09-14
Remove unused addon files from app tree

hakubo:
## Links

* https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L497

## Description

Override `EmberApp.prototype.addonTreesFor` to remove from the final build files which are added by addons but not used by us. For now, it's just few files from ember-rl-dropdown but this can be easily configured for all the addons we use.

We found no better way to do this, there are no public hooks that we could use. After asking on https://embercommunity.slack.com/messages/-ember-cli the only response was to do what we did.

## Reviewers

@Wikia/x-wing @bognix @bkoval


kvas-damian:
LGTM

PR #2787, opened 2016-09-14
XW-1702 | Support search input behaviour in desktop nav

Grunny:
## Links

* http://wikia-inc.atlassian.net/browse/XW-1702

## Description

Add support for search input behaviour in global nav on desktop.

## Reviewers

/cc @Wikia/x-wing


front/main/app/templates/components/design-system/global-navigation-search.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="wds-global-navigation__search-input-wrapper">
+	<label class="wds-global-navigation__search-label">
+		{{svg 'wds-icons-magnifying-glass' class='wds-icon wds-icon-small wds-global-navigation__search-label-icon'}}
+		{{input
+			class='wds-global-navigation__search-input'
+			name=model.results.param-name
+			placeholder=searchPlaceholder
+			focus-in=(action 'focusSearch')
+			type='search'
+			value=query
```
kvas-damian: It's two way binding. I know @rogatty doesn't like it ;)

front/main/app/templates/components/design-system/global-navigation-search.hbs
```diff
@@ -0,0 +1,19 @@
...
+		}}
+	</label>
+	<button class="wds-button wds-is-text wds-global-navigation__search-close" type="reset" {{action 'closeSearch'}}>
```
kvas-damian: Don't know why, but this button doesn't clean input for me.

Grunny: Looks `type="reset"` doesn't work for these inputs, can just clear the field on the close event.

front/main/app/templates/components/design-system/global-navigation-search.hbs
Grunny: What would be the preferred alternative for this case?

front/main/app/templates/components/design-system/global-navigation-search.hbs
kvas-damian: It's strange it doesn't work here. It works on Oasis: https://github.com/Wikia/app/blob/180846d780c344f65b3e0fde58e9d05538e75892/extensions/wikia/DesignSystem/scripts/DesignSystemGlobalNavigationSearch.js#L29-L33

front/main/app/templates/components/design-system/global-navigation-search.hbs
kvas-damian: I'm not sure but heard `one-way-input` is a solution here.

rogatty: What @kvas-damian said. We currently use an old, deprecated version of ember-one-way-input in https://github.com/Wikia/mercury/blob/dev/front/main/app/templates/components/wikia-ui-components/fancy-check-box.hbs. It's now renamed to ember-one-way-controls and we probably should update it.

rogatty: Or actually... We don't need this `query` for anything else than checking if it's empty, right? We can use [keyUp action](http://emberjs.com/api/classes/Ember.TextField.html#method_keyUp) to handle that instead.

front/main/app/templates/components/design-system/global-navigation-search.hbs
rogatty: I think the button clears the input but then Ember immediately restores it back to the value it is bound to. We may need to use ember-one-way-controls after all.

rogatty: Not sure why this happens but I added manual reset in https://github.com/Wikia/mercury/pull/2787/commits/e4b5006321633d4fad94c2b1a1952edc7a246f5d.
This attribute has to stay here anyway because without it this button becomes the default one and input is cleared on enter keypress.

front/main/app/templates/components/design-system/global-navigation-search.hbs
rogatty: I used one-way-input to handle the reset button.

rogatty:
LGTM

front/main/app/templates/components/design-system/global-navigation-search.hbs
kvas-damian: I found a fix: https://github.com/Wikia/mercury/pull/2787/commits/ca90bfd8640fb7a7cf1b68458387494c7f398b67 ;)

front/main/app/components/design-system/global-navigation.js
```diff
@@ -1,6 +1,19 @@
 import Ember from 'ember';
 
-export default Ember.Component.extend({
+const {Component, computed} = Ember;
+
+export default Component.extend({
 	classNames: ['wds-global-navigation-wrapper'],
-	model: Ember.Object.create(M.prop('globalNavigation'))
+	model: Ember.Object.create(M.prop('globalNavigation')),
```
hakubo: if we wanted we could
```
const {Object: EmberObj} = Ember;
```

front/main/tests/integration/components/design-system/global-navigation-search-test.js
```diff
@@ -0,0 +1,109 @@
...
+	const activateSearch = sinon.spy(),
+		deactivateSearch = sinon.spy();
+
+	this.set('model', model);
+
+	this.on('activateSearch', activateSearch);
+	this.on('deactivateSearch', deactivateSearch);
+
+	this.render(hbs`
+		{{design-system/global-navigation-search
```
hakubo: design-system**/**global-navigation-search or design-system**.**global-navigation-search?

PR #2788, opened 2016-09-14
ADEN-3890 | make floor adhesion work with different sizes

bognix:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3890

## Description

Floor Adhesion wasn't worknig with different sizes. With this change it should look fine with different creatives

## Reviewers
@v3rth


v3rth:
LGTM

PR #2789, opened 2016-09-15
XW-1910 | Add Search Suggestions to Global Navigation

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1910

## Description

Reuse logic from `wikia-search` component. We should remove the old component when working on the mobile version of new Global Navigation.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/components/design-system/global-navigation-search.js
+++ b/front/main/app/components/design-system/global-navigation-search.js
@@ -1,16 +1,43 @@
 import Ember from 'ember';
+import wrapMeHelper from '../../helpers/wrap-me';
 
-const {Component, computed} = Ember;
+const {Component, computed, observer, inject, run, $} = Ember;
 
 export default Component.extend({
 	tagName: 'form',
-	classNames: ['wds-global-navigation__search'],
+	classNames: ['wds-dropdown', 'wds-global-navigation__search'],
+	classNameBindings: ['hasSuggestions:wds-is-active'],
 	attributeBindings: ['action'],
 	searchIsActive: false,
+	// key: query string, value: Array<SearchSuggestionItem>
+	cachedResults: {},
+	// How many items to store in the cachedResultsQueue
+	cachedResultsLimit: 100,
+	// string[] which holds in order of insertion, the keys for the cached items
+	cachedResultsQueue: [],
+	// in ms
+	debounceDuration: 250,
+	// Whether or not to display the loading search suggestion results message (en: 'Loading...')
+	isLoadingResultsSuggestions: false,
+	/**
+	 * A set (only keys used) of query strings that are currently being ajax'd so
+	 * we know not to perform another request.
+	 */
+	requestsInProgress: {},
+	searchRequestInProgress: false,
+	/**
+	 * This is what's currently displayed in the search results
+	 * @member {SearchSuggestionItem[]}
+	 */
+	suggestions: [],
+	suggestionsEnabled: true,
+	queryMinimalLength: 3,
 	query: '',
 
 	action: computed.oneWay('model.results.url'),
-	submitDisabled: computed.empty('query'),
+	ajax: inject.service(),
+	hasSuggestions: computed.notEmpty('suggestions'),
+	emptyQuery: computed.empty('query'),
 	searchPlaceholder: computed('searchIsActive', function () {
 		if (this.get('searchIsActive')) {
 			return i18n.t(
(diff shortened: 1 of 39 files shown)


front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +53,308 @@ export default Component.extend({
...
+		keyPress(event) {
+			const numSuggestions = this.get('suggestions.length'),
+				currentSelection = this.get('selectedSuggestion');
+
+			if (!numSuggestions) {
+				return;
+			}
+
+			if (event.keyCode === 40 && currentSelection < numSuggestions - 1) {
+				this.set('selectedSuggestion', currentSelection + 1);
```
rogatty: We can use `this.incrementProperty('selectedSuggestion')`

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
 		return i18n.t(this.get('model.placeholder-inactive.key'), {ns: 'design-system'});
 	}),
 
+	didInsertElement() {
+		Ember.$(document).bind('click.global-navigation-search-suggestions', {
```
hakubo: you can remove `Ember.`

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+			component: this
+		}, this.onClickOutsideSearch);
+	},
+
+	willDestroyElement() {
+		Ember.$(document).unbind('click.global-navigation-search-suggestions', this.onClickOutsideSearch);
```
hakubo: here as well you can remove `Ember.`

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+			}
+		},
+
+		onKeyUp() {
+			if (this.get('suggestions.length') && this.get('selectedSuggestionIndex') > -1) {
+				this.decrementProperty('selectedSuggestionIndex');
+			}
+		},
+
+		suggestionClick(suggestion) {
```
hakubo: suggestionClick({title})

front/main/app/templates/components/design-system/global-navigation-search.hbs
```diff
@@ -1,20 +1,36 @@
...
 	</label>
 	<button class="wds-button wds-is-text wds-global-navigation__search-close" type="reset" {{action 'closeSearch' preventDefault=false}}>
 		{{svg 'wds-icons-cross' class='wds-icon wds-icon-small wds-global-navigation__search-close-icon'}}
 	</button>
+	<div class="wds-dropdown__content wds-global-navigation__search-suggestions">
+		<ul class="wds-list wds-has-ellipsis wds-is-linked">
+			{{#each suggestions as |suggestion index|}}
+				<li class="{{if (equal index selectedSuggestionIndex) 'wds-is-selected'}}">
+					<a href="{{suggestion.uri}}" class="wds-global-navigation__dropdown-link" {{action 'suggestionClick' suggestion preventDefault=false}}>
+						{{{suggestion.text}}}
```
hakubo: this is coming from API right? are we safe using not escaped text here?

front/main/app/templates/components/wikia-ui-components/fancy-check-box.hbs
```diff
@@ -1,11 +1,10 @@
...
-	{{one-way-input
+	{{one-way-checkbox
 		checked=isChecked
 		class='fancy-checkbox'
 		click=onClick
 		id=inputId
 		name=inputId
-		type='checkbox'
 	}}
 	{{#if isChecked}}
```
hakubo: as we are at it, what do you think of:
`{{svg 'checkbox-checkmark' class=(concat 'icon ' (unless isChecked 'hidden') viewBox="0 0 20 14"}}`

This way we don't add additional sub template

front/main/app/utils/url.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+
```
hakubo: const {$} = Ember;

front/main/tests/unit/components/design-system/global-navigation-test.js
```diff
@@ -0,0 +1,312 @@
...
+	component.send('focusSearch');
+
+	assert.equal(component.get('searchIsActive'), true, 'focus on search sets searchIsActive to true');
+});
+
+test('close search', function (assert) {
+	const component = this.subject(),
+		setSearchSuggestionItems = sinon.stub(component, 'setSearchSuggestionItems');
+
+	component.set('searchIsActive', true);
```
hakubo: here also we can use setProperties

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+			});
+
+		this.setProperties({
+			suggestions,
+			isLoadingResultsSuggestions: false
+		});
+	},
+
+	/**
+	 * returns uri to send an ajax request to
```
hakubo: comments! yay!

front/main/app/templates/components/design-system/global-navigation-search.hbs
Grunny: Article titles won't be vulnerable here because they can't contain `<>`, but we shouldn't rely on that always being true and besides escaping also provides "correctness" of the output. Since we need to highlight the search phrase in the title with `<strong></strong>` the escaping needs to be done in global-navigation-search.js rather than in the template unfortunately.

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -1,16 +1,53 @@
...
+ */
 
 export default Component.extend({
 	tagName: 'form',
 	classNames: ['wds-global-navigation__search'],
 	attributeBindings: ['action'],
+
 	searchIsActive: false,
+	// key: query string, value: Array<SearchSuggestionItem>
+	cachedResults: {},
```
kvas-damian: Can we export search suggestions to separate Component/Mixin?

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+
+		queryChanged(query) {
+			this.set('query', query);
+			this.updateSuggestions(query);
+		},
+
+		onKeyDown() {
+			const numSuggestions = this.get('suggestions.length');
+
+			if (numSuggestions && this.get('selectedSuggestionIndex') < numSuggestions - 1) {
```
kvas-damian: can we simplify this condition to `this.get('selectedSuggestionIndex') < this.get('suggestions.length') - 1`?
`suggestions.length` can't be negative so condition won't change.

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+		Ember.$(document).bind('click.global-navigation-search-suggestions', {
+			component: this
+		}, this.onClickOutsideSearch);
+	},
+
+	willDestroyElement() {
+		Ember.$(document).unbind('click.global-navigation-search-suggestions', this.onClickOutsideSearch);
+	},
+
+	submit(event) {
```
kvas-damian: Instead of submit event I'd use `enter` key press.

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+	/**
+	 * returns uri to send an ajax request to
+	 *
+	 * @param {string} query - search string
+	 * @returns {string}
+	 */
+	getSearchSuggestionsUrl(query) {
+		const params = {};
+
+		params[this.get('model.suggestions.param-name')] = query;
```
kvas-damian: While params have only one key, we could instantiate it when calling a method ;)

front/main/app/components/design-system/global-navigation-search.js
Grunny: But `suggestions.length` can be 0 (i.e. no suggestions), in which case we don't want to do anything here.

front/main/tests/unit/components/design-system/global-navigation-test.js
```diff
@@ -253,8 +273,9 @@ test('eviction tests', function (assert) {
 test('setSearchSuggestionItems - set suggestions to be empty', function (assert) {
 	const component = this.subject();
 
-	component.set('isLoadingResultsSuggestions', true);
-	component.set('suggestions', ['test']);
+	component.setProperties({
+		isLoadingResultsSuggestions: true,
+		suggestions: ['test']
```
rogatty: `setProperties` isn't closed.

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,311 @@ export default Component.extend({
...
+		}
+	},
+
+	/**
+	 * @param {string[]} [rawSuggestions = []]
+	 * @returns {void}
+	 */
+	setSearchSuggestionItems(rawSuggestions = []) {
+		const query = this.get('query'),
+			highlightRegexp = new RegExp(`(${escapeRegex(query)})`, 'ig'),
```
Grunny: We need to escape the entire string before adding the `<strong>` tags otherwise the strong tags would also be escaped, so can you clarify "If we can go with highlight first solution then we should be fine with String.replace"?

And we're not actually escaping the Regexp, we're escaping the phrase before it is turned into a Regexp so that both it and the string are escaped and so match if they have a quote or ampersand in them.

front/main/app/components/design-system/global-navigation-search.js
rogatty: I've tried to handle enter on suggestion but the form is always submitted. Browser does it and there is no way to prevent it other than `event.preventDefault()`

front/main/app/components/design-system/global-navigation-search.js
rogatty: We have a unit tests so it's easy to check if it'll keep working!

front/main/app/components/design-system/global-navigation-search.js
Grunny: Oh and I realised this was an old version you were reviewing. To clarify `escapeRegex` turns the string to valid regex to fix https://wikia-inc.atlassian.net/browse/XW-2050 i.e if the string is "Some title (subtitle)" it turns it into "Some title \(subtitle\)" otherwise the regex fails. It's a failing of the JS standard lib that it doesn't have functionality to do this built in. :) Also, regex is needed because String.replace is case sensitive, while we need a case-insensitive replacement.

front/main/app/components/design-system/global-navigation-search.js
Grunny: Done.

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,309 @@ export default Component.extend({
 		return i18n.t(this.get('model.placeholder-inactive.key'), {ns: 'design-system'});
 	}),
 
+	didInsertElement() {
+		$(document).bind('click.global-navigation-search-suggestions', {
+			component: this
```
kvas-damian: Can we try using arrow function to keep the right scope?

front/main/tests/unit/components/design-system/global-navigation-test.js
```diff
@@ -0,0 +1,334 @@
...
+});
+
+test('setSearchSuggestionItems - correctly set suggestions array', function (assert) {
+	const component = this.subject(),
+		suggestionsFromApi = [
+			'suggestion1',
+			'suggestion 2!',
+			'sug GES %^&*^%&>?<tion>',
+			'no query inside',
+			'sug sug suggestions sug sug'
```
kvas-damian: important part of processing suggestions is being case insensitive. Can we add a test-case for it?

front/main/app/components/design-system/global-navigation-search.js
kvas-damian: We agreed offline that search and suggestions are closely coupled so it creates many problems to separate them.
For now we decided keep it as it is and refactor when it's required.

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,329 @@ export default Component.extend({
 		return i18n.t(this.get('model.placeholder-inactive.key'), {ns: 'design-system'});
 	}),
 
+	didInsertElement() {
+		$(document).on('click.global-navigation-search-suggestions', (event) => {
+			this.onClickOutsideSearch(event);
+		});
+	},
+
+	willDestroyElement() {
```
hakubo: +1 point for cleaning events!

front/main/app/components/design-system/global-navigation-search.js
```diff
@@ -25,15 +62,329 @@ export default Component.extend({
...
+		$(document).off('click.global-navigation-search-suggestions');
+	},
+
+	/**
```
hakubo: comment! +1 point :)

PR #2791, opened 2016-09-15
XW-2018 | animations for global navigation

Warkot:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2018
* https://github.com/Wikia/design-system/pull/63
* https://github.com/Wikia/app/pull/11344

## Reviewers

@Wikia/x-wing


front/main/bower.json
```diff
@@ -3,7 +3,7 @@
   "dependencies": {
     "Autolinker.js": "0.22.0",
     "cropper": "0.11.0",
-    "design-system": "Wikia/design-system#0.9.0",
+    "design-system": "Wikia/design-system#XW-2018",
```
tomasznapieralski: remember to use DS version

front/main/app/templates/components/design-system/global-navigation.hbs
```diff
@@ -10,26 +10,28 @@
...
-				<button class="wds-button wds-is-text wds-global-navigation__search-close" type="reset">
-					{{svg 'wds-icons-cross' class='wds-icon wds-icon-small wds-global-navigation__search-close-icon'}}
-				</button>
-			</div>
-			<button class="wds-button wds-global-navigation__search-submit" type="submit">
-				{{svg 'wds-icons-arrow' class='wds-icon wds-icon-small wds-global-navigation__search-submit-icon'}}
-			</button>
-		</form>
+		<div class="wds-global-navigation__links-and-search">
+			{{#if model.fandom_overview}}
```
hakubo: we don't need this if here each will do that for us

Warkot: We just added a wrapper, but I agree we can fix it. 😉  👍

Warkot: Just to be sure, are we safe to access one step deeper (`model.fandom_overview.links`) without checking the middle level (`model.fandom_overview`)? Vanilla JS would rise an exception `Cannot read property 'links' of undefined`.

hakubo: yup: https://ember-twiddle.com/f7eaf52aacf9ef0b7d8267f741c732bd?openFiles=templates.application.hbs%2C

PR #2792, opened 2016-09-15
XW-2034 | Fix the search acceptance test

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2034

## Description

We render the old search only on the mobile breakpoint. Let's make sure that it doesn't break the tests when run in a desktop browser.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/tests/helpers/mock-service.js
+++ b/front/main/tests/helpers/mock-service.js
@@ -1,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Test.registerHelper('mockService', function (app, newService, serviceName) {
+	const instance = app.__deprecatedInstance__,
+		registry = instance.register ? instance : instance.registry;
+
+	return registry.register(`service:${serviceName}`, newService);
+});
--- a/front/main/tests/helpers/start-app.js
+++ b/front/main/tests/helpers/start-app.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Application from '../../app';
 import config from '../../config/environment';
+import './mock-service';
 import './reset-ads';
 
 export default function startApp(attrs) {
--- a/front/main/mirage/config.js
+++ b/front/main/mirage/config.js
@@ -2,6 +2,8 @@
  * @returns {void}
  */
 export default function () {
+	this.passthrough('/write-blanket-coverage');
+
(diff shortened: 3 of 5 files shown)


front/main/tests/helpers/mock-service.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Test.registerHelper('mockService', function (app, newService, serviceName) {
+	const instance = app.__deprecatedInstance__,
+		registry = instance.register ? instance : instance.registry;
+
+	return registry.register(`service:${serviceName}`, newService);
```
bognix: why not simply pass `this` as the instance reference instead of doing this "magic"? 

small note: in order to pass `this` and make it 'correct' `this` you need to use `function (assert) {}` instead of `()=>{}` in test code.

front/main/tests/helpers/start-app.js
```diff
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Application from '../../app';
 import config from '../../config/environment';
+import './mock-service';
```
bognix: we mock services in different places in code as well, could you make them use this helper as well?

rogatty:
@bognix I tried to address your comments and found out that this method:
```
this.register('service:ads-state', adsStateStub);
this.inject.service('ads-state', {as: 'adsState'});
```
doesn't work in the acceptance tests. `this` is something totally different than in the unit tests context and I couldn't make it work.

I also tried the other way - to use this new helper in `unit/components/ad-slot-test`. It doesn't work neither because helpers are registered only for acceptance tests and not available in unit tests ¯\_(ツ)_/¯ 

Thanks for your suggestions anyway, I agree that it would be great to have this unified. Unfortunately Ember testing environment is pretty complex and not ready for what you proposed. Quoting https://github.com/rwjblue/rfcs/blob/42/text/0000-grand-testing-unification.md

>Unfortunately, there is a massive disparity between different types of tests. In acceptance tests, you use well designed global helpers to deal with async related interactions; whereas in integration and unit tests you are forced to manually deal with this asynchrony. 

>It is currently possible to register custom factories in integration and unit tests, but not in acceptance tests (without using private API's that is).

tomasznapieralski:
LGTM

PR #2796, opened 2016-09-19
XW-2048 | Support redirect parameter in authentication links

Grunny:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2048

## Description

Support the redirect parameter in authentication links. The returnto parameter doesn't currently support
Discussions, so we're leaving it out for now.

## Reviewers

/cc @Wikia/x-wing

### DIFF
--- a/front/main/app/components/design-system/global-navigation-link-authentication.js
+++ b/front/main/app/components/design-system/global-navigation-link-authentication.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import {addQueryParams} from '../../utils/url';
 
 const {Component, computed} = Ember;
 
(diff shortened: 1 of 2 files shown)


Grunny:
Also, this is based on XW-1910 to make use of `addQueryParams`, but we can change the target branch once that's merged.

front/main/app/components/design-system/global-navigation-link-authentication.js
```diff
@@ -12,5 +13,14 @@ export default Component.extend({
 		};
 
 		return classMap[this.get('model.title.key')] || '';
+	}),
+	href: computed('model.param-name', 'model.href', function () {
+		let url = this.get('model.href');
+
+		if (this.get('model.param-name') === 'redirect') {
```
rogatty: Shouldn't we make the decision based on `model.title.key` and just use `this.get('model.param-name')` as a param name instead of hardcoding it?

kvas-damian: For me it makes sense to have a condition based on `model.param-name`.
It's limited to `link-authentication` type returned by API and I believe there shouldn't be links in this type where `redirect` param name means something different.

PR #2798, opened 2016-09-19
XW-2003 | update icons in Global Navigation

Warkot:
## Links
- https://wikia-inc.atlassian.net/browse/XW-2003
- Wikia/app#11381
- https://github.com/Wikia/design-system/pull/65
## Reviewers

@Wikia/x-wing


hakubo:
hmm or actually should we merge it? or rather wait for release? :)

PR #2801, opened 2016-09-20
XW-2013 Use DS API for mobile global navigation

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2013

## Description

Instead of using the old Global Navigation API (data from it comes to Mercury together with the wiki variables) for the mobile navigation we want to use a mix:
- hubs and global links controlled by the Design System API
- local nav controlled by the old API

This is only a temporary solution until we rewrite the mobile nav to be compliant with the Design System.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/main/app/models/wikia-nav.js
+++ b/front/main/app/models/wikia-nav.js
@@ -3,13 +3,21 @@ import Ember from 'ember';
 const {Object, A, Logger, computed, get} = Ember;
 
 export default Object.extend({
-	hubsLinks: get(Mercury, 'wiki.navigation2016.hubsLinks'),
+	dsGlobalNavigation: M.prop('globalNavigation'),
+	hubsLinks: computed(function () {
+		return this.getWithDefault('dsGlobalNavigation.fandom_overview.links', []);
+	}),
+	exploreWikiaLinks: computed(function () {
+		return this.getWithDefault('dsGlobalNavigation.wikis.links', []);
+	}),
+	exploreWikiaLabel: computed(function () {
+		return i18n.t(this.get('dsGlobalNavigation.wikis.header.title.key'), {
+			ns: 'design-system'
+		});
+	}),
 	localLinks: get(Mercury, 'wiki.navigation2016.localNav'),
-	exploreWikiaLinks: get(Mercury, 'wiki.navigation2016.exploreWikiaMenu'),
-	exploreWikiaLabel: get(Mercury, 'wiki.navigation2016.exploreWikia.textEscaped'),
 	discussionsEnabled: get(Mercury, 'wiki.enableDiscussions'),
 	wikiName: get(Mercury, 'wiki.siteName'),
-	wikiLang: get(Mercury, 'wiki.language.content'),
 
 	/**
 	 * Iteratively traverse local navigation tree to find out root node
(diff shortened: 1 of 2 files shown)


front/main/app/models/wikia-nav.js
```diff
@@ -3,13 +3,21 @@ import Ember from 'ember';
...
 export default Object.extend({
-	hubsLinks: get(Mercury, 'wiki.navigation2016.hubsLinks'),
+	dsGlobalNavigation: M.prop('globalNavigation'),
+	hubsLinks: computed(function () {
+		return this.getWithDefault('dsGlobalNavigation.fandom_overview.links', []);
+	}),
+	exploreWikiaLinks: computed(function () {
+		return this.getWithDefault('dsGlobalNavigation.wikis.links', []);
+	}),
+	exploreWikiaLabel: computed(function () {
```
kvas-damian: This variable is used only once. Maybe we can convert it to local variable?

rogatty: It's used twice for `header` and `exploreSubMenuItem`.

kvas-damian: ok, makes sense then

front/main/app/models/wikia-nav.js
```diff
@@ -3,13 +3,21 @@ import Ember from 'ember';
 const {Object, A, Logger, computed, get} = Ember;
 
 export default Object.extend({
-	hubsLinks: get(Mercury, 'wiki.navigation2016.hubsLinks'),
+	dsGlobalNavigation: M.prop('globalNavigation'),
```
tomasznapieralski: how about `Ember.Object.create(M.prop('globalNavigation'))`?

front/main/app/models/wikia-nav.js
```diff
@@ -3,13 +3,21 @@ import Ember from 'ember';
...
+	hubsLinks: computed(function () {
```
tomasznapieralski: shouldn't this be `computed(''dsGlobalNavigation, function () {...}`?

Grunny: Or even just `computed.oneWay('dsGlobalNavigation.fandom_overview.links')`?

front/main/app/models/wikia-nav.js
```diff
@@ -3,13 +3,21 @@ import Ember from 'ember';
...
+		return this.get('dsGlobalNavigation.fandom_overview.links');
+	}),
+	exploreWikia: computed(function () {
```
Grunny: Possibly worth renaming these two from `Wikia` to `Wikis` now?

front/main/app/models/wikia-nav.js
rogatty: We don't use `dsGlobalNavigation.get()` anywhere so it doesn't give us any value, just an overhead.

front/main/app/models/wikia-nav.js
rogatty: Done, would be nice if GitHub hide this comment :disappointed:

front/main/app/models/wikia-nav.js
weronika: we tried to to it this way, but unfortunately we got this error:
`ember.debug.js:18851 Uncaught TypeError: Cannot define property:__ember_meta__, object is not extensible.`
We're not really sure why it occurs and how to fix it.

rogatty: Tried to debug, didn't find any reasonable explanation.

PR #2802, opened 2016-09-20
XW-2076 | support new logo in footer

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2076
## Description

This way we can support all variations of logo in footer in the meantime

I branched of from dev, we should merge dev to feature branch if we want to add it there, can do!

## Reviewers
@Wikia/x-wing


Warkot:
I wanted to update translations by changing to line below in `package.json`
```
"design-system-i18n": "git://github.com/Wikia/design-system-i18n.git#XW-2076",
```
but I get a wall of errors while running
```
npm run shrinkwrap
```
😞

PR #2803, opened 2016-09-21
XW-1727 | Click tracking on Global Navigation

kvas-damian:
## Links
* https://wikia-inc.atlassian.net/browse/XW-1727

## Reviewers
@Wikia/x-wing


front/main/app/components/design-system/global-navigation-link-branded.js
```diff
@@ -11,5 +11,9 @@ export default Component.extend({
 	href: computed.oneWay('model.href'),
 	linkClassName: computed('model.brand', function () {
 		return `wds-is-${this.get('model.brand')}`;
-	})
+	}),
+
+	click: function() {
```
Grunny: `click() {`

PR #2805, opened 2016-09-21
ADEN-4008 Fix ooyala no ads parameter

fraszczakszymon:
Fix ooyala no ads parameter

## Links

* http://wikia-inc.atlassian.net/browse/ADEN-4008


bulaszenko:
Checked works good

PR #2814, opened 2016-09-23
rename name to appname in logs

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1970

@Wikia/x-wing

PR #2815, opened 2016-09-23
add appname to logs

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1970

## Description

We don't care about name, but this is needed for bunyan

## Reviewers

@Wikia/x-wing

### DIFF
--- a/server/app/lib/logger.js
+++ b/server/app/lib/logger.js
@@ -88,6 +88,7 @@ export function createLogger(loggerConfig) {
 	}
 
 	return bunyan.createLogger({
+		appname: 'mercury',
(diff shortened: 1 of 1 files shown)


server/app/lib/logger.js
```diff
@@ -88,6 +88,7 @@ export function createLogger(loggerConfig) {
 	}
 
 	return bunyan.createLogger({
+		appname: 'mercury',
 		name: 'mercury',
```
rogatty: Yes, look at https://github.com/Wikia/mercury/pull/2814

PR #2823, opened 2016-09-26
XW-2113 | add `stable` environment type to server

Warkot:
## Links

* On Stable Env. Global Navigation and Footer are not working on Mercury

## Description

Make Mercury server understand the stable environment, so Discussions can function properly on `stable.wikiname.wikia.com`.

## Reviewers

@Wikia/x-wing


Grunny:
Worth adding test cases for stable as well to `server/tests/unit/lib/utils.js` for completion?

PR #2828, opened 2016-09-28
SEO-509 Second round of updates for testing prerender.io for Curated Main Pages

gabrys:
What's done here (see individual comments for details)

* Move the bot detection logic to VCL to avoid cache poisoning (VCL pull request https://github.com/Wikia/wikia-vcl/pull/1063)
* Set canonical on /main/category/XYZ to /wiki/Category:XYZ to signal duplicate content to bots
* Add `<meta fragment="!">` to /main/* pages
* Fix the URL requested from prerender.io

Reviewers:

@rogatty, @bognix, @vforge


server/app/lib/prerender.js
```diff
@@ -38,23 +38,15 @@ function canPrerender(req) {
...
+	return req.url.query._escaped_fragment_ !== undefined && canPrerender(req);
 }
 
 function updateRequestedUrl(url) {
-	// Direct prerender.io to production if initiated from dev environments
+	// Direct prerender.io to production if initiated from dev or sandbox environments
 	url = url.replace('.127.0.0.1.xip.io:7000/', '.wikia.com/');
 	url = url.replace(new RegExp('\.[a-z]*\.wikia-dev\.com\/'), 'wikia.com/');
-	return `$(url)?useskin=mercury`;
+	url = url.replace(new RegExp('^http://sandbox-[^.]*.'), 'http://');
```
rogatty: Would be nice if there were some unit tests.

server/app/lib/prerender.js
```diff
@@ -38,23 +38,15 @@ function canPrerender(req) {
...
-
-	return canPrerender(req);
+	return req.url.query._escaped_fragment_ !== undefined && canPrerender(req);
 }
 
 function updateRequestedUrl(url) {
-	// Direct prerender.io to production if initiated from dev environments
+	// Direct prerender.io to production if initiated from dev or sandbox environments
 	url = url.replace('.127.0.0.1.xip.io:7000/', '.wikia.com/');
 	url = url.replace(new RegExp('\.[a-z]*\.wikia-dev\.com\/'), 'wikia.com/');
```
vforge: I would replace `new RegExp(...)` with `/.../`, but that's just me.

gabrys: Here's the funny thing: it seems you cannot have a `/` as the last character of a regular expression, because it becomes `//` which starts a comment. I haven't checked if this works, but it confuses PhpStorm for sure :-)

PR #2830, opened 2016-09-28
SOC-1340 | Pagination for post list pages

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-1340

## Description

New querystring param: page=<number> and it's handling.
Also, adding it dynamically to the QS when the user is browsing through list pages.

## Reviewers

@Wikia/social-frontend


front/main/app/mixins/discussion-forum-actions-controller.js
```diff
@@ -2,6 +2,14 @@ import Ember from 'ember';
 
 export default Ember.Mixin.create(
 	{
+		queryParams: ['page'],
+
+		page: 1,
+
+		pageUpdater: Ember.observer('model.current.data.pageNum', function () {
```
bkoval: We cannot use page as a computed property because of the following bug in Ember 2+ https://github.com/emberjs/ember.js/issues/11592

front/main/app/mixins/discussion-wrapper-component.js
```diff
@@ -6,9 +6,8 @@ export default Ember.Mixin.create(
 		classNames: ['forum-wrapper', 'discussion', 'forum'],
 		isLoading: false,
 		loadOnScrollEnabled: false,
-		pageNum: 0,
+		page: 0,
```
mleczey: as discussed here should be `1` as default

front/main/app/mixins/discussion-wrapper-component.js
```diff
@@ -27,8 +31,8 @@ export default Ember.Mixin.create(
 			}
 		},
 
-		hasMore: Ember.computed('totalPosts', 'postsDisplayed', function () {
-			return this.get('totalPosts') > this.get('postsDisplayed');
+		hasMore: Ember.computed('totalPosts', 'page', function () {
+			return this.get('totalPosts') > this.getWithDefault('page', 1) * 20;
 		}),
 
 		loadingPageResolveObserver: Ember.observer('postsDisplayed', 'minorError', function () {
```
mleczey: Maybe we could get rid of `postsDisplayed` here? And then remove it from `*.hbs`?

front/main/app/models/discussion/forum.js
```diff
@@ -99,6 +100,13 @@ DiscussionForumModel.reopenClass({
 			}).then((data) => {
 				forumInstance.setNormalizedData(data);
 
+				if (page === 1) {
```
mleczey: Maybe this could be extracted to method and reused in models?

front/main/app/mixins/discussion-forum-actions-route.js
```diff
@@ -3,6 +3,11 @@ import localStorageConnector from '../utils/local-storage-connector';
 
 export default Ember.Mixin.create(
 	{
+		queryParams: {
+			catId: {
```
mleczey: This is probably not needed here.

front/main/app/mixins/discussion-forum-actions-model.js
```diff
@@ -24,4 +25,14 @@ export default Ember.Mixin.create({
 				return '';
 		}
 	},
+
+	setStartPageNumber(page) {
+		debugger;
```
mleczey: this should be removed

PR #2833, opened 2016-09-29
ADEN-3972 Load interstitial and floor adhesion on load

fraszczakszymon:
Simplify the logic and load interstitial/floor adhesion on page load/after each transition.

## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3972

## Reviewers

@bognix


front/main/app/mixins/ads.js
```diff
@@ -62,6 +63,10 @@ export default Ember.Mixin.create({
 			adsData = this.get('adsData'),
 			firstSectionTop = ($firstSection.length && $firstSection.offset().top) || 0,
 			articleBodyHeight = $articleBody.height(),
+			highImpactComponent = this.get('container').lookup('component:ads/invisible-high-impact-2', {
```
bognix: you need to do the same inside injectMainPageAds method in order to have ads on curated main pages

front/main/app/services/ads-high-impact.js
```diff
@@ -3,41 +3,25 @@ import Ember from 'ember';
...
-	},
-
-	loadFloorAdhesion() {
-		if (this.isFloorAdhesionEnabled()) {
-			this.get('adsModule').addSlot(this.get('name'));
-			this.get('ads').pushInContentAd(this.get('name'), this.get('component'));
-		}
-	},
+	load(componentElement) {
+		this.get('ads.module').onReady(() => {
```
bognix: this one is a little bit tricky because of the order of "hooks" triggered by `$script`

This is what happens
* we load `mercury_ads_js` using `$script.load(..., callback)`
* when script is loaded the `onReady` is called (`$script.ready(callback)`)
* the `callback` from `$script.load` is called

Because of this order there is a slim chance that the isProperGeo would be undefined, however we call `load` method inside `injectAds` which is called on `afterRender` which most likely would happen after `mercury_ads_js` is loaded. There is a race condition but because of how long particular tasks take it shouldn't occur but it's good to be aware of that fact

front/main/app/mixins/ads.js
fraszczakszymon: Right, thanks! 👍

front/main/app/components/ads/invisible-high-impact-2.js
```diff
@@ -7,20 +7,20 @@ export default Component.extend({
 	ads: inject.service(),
 
 	name: computed.readOnly('adsHighImpact.name'),
+	layoutName: 'components/ads/invisible-high-impact-2',
```
nandy-andy: `layoutName` is only used by Ember when we do lookups like in https://github.com/Wikia/mercury/pull/2833/files#diff-cae0ff4cb7f360842d84be513a119f88R56?

nandy-andy:
LGTM :+1:

PR #2840, opened 2016-09-30
Social | Fandomizing auth

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/MAIN-8108

## Reviewers

@Wikia/social-frontend


front/common/public/locales/en/auth-front.json
```diff
@@ -1,6 +1,6 @@
 {
   "errors" : {
-    "email_already_exists": "Email is already registered on Wikia",
+    "email_already_exists": "Email is already registered on Fandom",
```
rogatty: What about the non-en languages?

bkoval: My plan is to submit this branch to crowdin and wait for translations with merging.

PR #2843, opened 2016-09-30
XW-2147 | use DS-i18n v0.6.2

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2147

## Reviewers

@Wikia/x-wing

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "babel-template": "6.7.0",
     "babel-types": "6.7.2",
     "code": "1.5.0",
-    "design-system-i18n": "git://github.com/Wikia/design-system-i18n.git#0.6.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,7 +50,7 @@
     "babel-template": "6.7.0",
     "babel-types": "6.7.2",
     "code": "1.5.0",
-    "design-system-i18n": "git://github.com/Wikia/design-system-i18n.git#0.6.1",
+    "design-system-i18n": "git://github.com/wikia/design-system-i18n.git#8da617846639703c34e15b6089ab247f6ab7112e",
```
Warkot: Maybe you could change that back to `0.6.2` format, so it's obvious from the first look what version it is. A minor thing is to use the capital `Wikia`—how it actually is in our github link.

rogatty: Why not 0.6.2?

PR #2847, opened 2016-10-03
XW-1737 | Handle wikiVariables non-JSON API response as a redirect

kvas-damian:
## Links
* https://wikia-inc.atlassian.net/browse/XW-1737

## Description
While handling `wikiVariables` response we can get redirect to a HTML page instead of JSON API response. Instead of trying to display that page in mercury let's redirect user to that page so browser can render it.

## Reviewers
@Wikia/x-wing

### DIFF
--- a/server/app/lib/custom-errors.js
+++ b/server/app/lib/custom-errors.js
@@ -61,15 +61,16 @@ export class WikiVariablesRequestError {
 WikiVariablesRequestError.prototype = Object.create(Error.prototype);
 
 /**
- * @class WikiVariablesNotValidWikiError
+ * @class NonJsonApiResponseError
  */
-export class WikiVariablesNotValidWikiError {
+export class NonJsonApiResponseError {
 	/**
 	 * @returns {void}
 	 */
-	constructor() {
+	constructor(url) {
 		Error.apply(this, arguments);
+		this.url = url;
 	}
 }
 
-WikiVariablesNotValidWikiError.prototype = Object.create(Error.prototype);
+NonJsonApiResponseError.prototype = Object.create(Error.prototype);
--- a/server/app/lib/custom-errors.js
+++ b/server/app/lib/custom-errors.js
@@ -67,9 +67,9 @@ export class NonJsonApiResponseError {
 	/**
 	 * @returns {void}
 	 */
-	constructor(url) {
+	constructor(redirectLocation) {
 		Error.apply(this, arguments);
-		this.url = url;
+		this.redirectLocation = redirectLocation;
 	}
 }
 
--- a/server/app/lib/mediawiki.js
+++ b/server/app/lib/mediawiki.js
@@ -6,17 +6,7 @@ import Logger from './logger';
 import Wreck from 'wreck';
 import Promise from 'bluebird';
 import Url from 'url';
-import {WikiVariablesNotValidWikiError, WikiVariablesRequestError} from './custom-errors';
-
-/**
- * @typedef {Object} CallbackParams
- * @property {Function} resolve
- * @property {Function} reject
- * @property {Object} err
- * @property {*} payload
- * @property {Object} response
- * @property {string} url
- */
+import {NonJsonApiResponseError, WikiVariablesRequestError} from './custom-errors';
 
 /**
  * Create request URL
@@ -70,13 +60,18 @@ export function sanitizeRejectData(payload, response) {
 
 /**
  * Handle request response
+ * @param {Function} resolve
+ * @param {Function} reject
+ * @param {Object} err
+ * @param {*} payload
+ * @param {Object} response
+ * @param {string} url
+ * @param {string} host
+ * @param {string} redirectedUrl
  *
- * @param {CallbackParams} params
  * @returns {void}
  */
-function requestCallback(params) {
-	const {resolve, reject, err, payload, response, url, host} = params;
-
+function requestCallback({resolve, reject, err, payload, response, url, host, redirectedUrl}) {
 	if (err) {
 		Logger.error({
 			url,
@@ -91,7 +86,10 @@ function requestCallback(params) {
 			}
 		});
 	} else if (response.statusCode === 200) {
-		resolve(payload);
+		resolve({
+			payload,
+			redirectedUrl
+		});
 	} else {
 		// Don't flood logs with 404s
 		if (response.statusCode !== 404) {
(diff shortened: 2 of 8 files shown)


server/tests/acceptance/wiki-page.js
```diff
@@ -195,7 +200,15 @@ describe('wiki-page', function () {
 	});
 
 	it('redirects to community wikia on /wiki/ when requested wiki does not exist', function (done) {
-		wreckGetStub.onCall(0).yields(null, {statusCode: 200}, 'not a valid wikia');
+		mediawiki.__Rewire__('Wreck', {
+			get: function (url, options, callback) {
+				if (url.indexOf('getWikiVariables') > -1) {
+					options.redirected(null, 'http://community.wikia.com/wiki/Community_Central:Not_a_valid_Wikia');
```
kvas-damian: If anyone knows how to achieve similar result using sinon spys I'm for your comments.

server/app/lib/custom-errors.js
```diff
@@ -61,15 +61,16 @@ export class WikiVariablesRequestError {
 WikiVariablesRequestError.prototype = Object.create(Error.prototype);
 
 /**
- * @class WikiVariablesNotValidWikiError
+ * @class NonJsonApiResponseError
  */
-export class WikiVariablesNotValidWikiError {
+export class NonJsonApiResponseError {
```
hakubo: NonJsonApiResponseError

server/app/lib/mediawiki.js
```diff
@@ -308,15 +322,15 @@ export class WikiRequest extends BaseRequest {
...
 			.fetch(url)
 			/**
-			 * @param {*} wikiVariables
+			 * @param {payload, redirectedUrl}
 			 * @returns {Promise}
 			 */
-			.then((wikiVariables) => {
-				if (wikiVariables.data) {
-					return Promise.resolve(wikiVariables.data);
+			.then(({payload, redirectedUrl}) => {
```
hakubo: You could use destructuring on payload
`.then(({{data}, redirectedUrl})`

server/app/lib/mediawiki.js
```diff
@@ -353,7 +367,14 @@ export class PageRequest extends BaseRequest {
 			urlParams.sections = sections;
 		}
 
-		return this.fetch(createUrl(this.wikiDomain, 'wikia.php', urlParams));
+		return this.fetch(createUrl(this.wikiDomain, 'wikia.php', urlParams))
+			/**
+			 * @param {payload, redirectedUrl}
+			 * @returns {Promise}
+			 */
+			.then(({payload}) => {
```
hakubo:
```js
.then(({payload}) => Promise.resolve(payload));
```

server/app/lib/mediawiki.js
```diff
@@ -365,7 +386,14 @@ export class PageRequest extends BaseRequest {
 			method: 'getMainPageDetailsAndAdsContext'
 		});
 
-		return this.fetch(url);
+		return this.fetch(url)
+			/**
+			 * @param {payload, redirectedUrl}
+			 * @returns {Promise}
+			 */
+			.then(({payload}) => {
```
hakubo:
```js
.then(({payload}) => Promise.resolve(payload));
```

server/app/lib/custom-errors.js
rogatty: Naming isn't very obvious here. This exception is thrown when the reponse has code 200 but it's not JSON. It can happen after redirects or without them. Let's think about something that better describes this logic.

server/app/lib/mediawiki.js
hakubo: or actually why do we have to do this?
it already return promise right?

server/app/lib/custom-errors.js
rogatty: We would need additional Babel plugin, see http://stackoverflow.com/a/35858868/1050577

server/app/lib/mediawiki.js
```diff
@@ -70,13 +60,18 @@ export function sanitizeRejectData(payload, response) {
 
 /**
  * Handle request response
+ * @param {Function} resolve
```
rogatty: See https://github.com/eslint/eslint/blob/74b2e29a7c43ca22f97bcaa89392e3ac36f6d0fc/tests/lib/rules/valid-jsdoc.js#L518-L527

server/app/lib/mediawiki.js
```diff
@@ -196,21 +200,31 @@ export function post(url, formData, host = '', headers = {}) {
...
 		 * @param {*} err
 		 * @param {*} response
 		 * @param {*} payload
 		 * @returns {void}
 		 */
-		Wreck.request('POST', url, {payload: formData, headers}, (err, response) => {
+		Wreck.request('POST', url, {
+			payload: formData,
+			headers,
+			redirected: (statusCode, location, req) => {
```
rogatty: We don't allow redirects for POST so there is no need to handle them here.

server/app/lib/mediawiki.js
rogatty: I'd rename `redirectedUrl` to `redirectLocation`.

server/tests/acceptance/curated-content.js
```diff
@@ -80,16 +74,26 @@ describe('curated-content', function () {
...
+					callback(null, {statusCode: 200}, clone(mainPageDetailsAndContext));
+				}
+			}
+		});
 
 		server.inject(requestParams, function (response) {
 			expect(response.statusCode).to.equal(302);
 			expect(response.headers.location).to.equal(
-				'http://community.wikia.com/wiki/Community_Central:Not_a_valid_Wikia'
+				'http://google.com/'
```
rogatty: If you want to make this change then update the test's name.

server/app/lib/mediawiki.js
kvas-damian: Why do you think we don't allow redirects?
By default wreckjs follows `3` redirects. Additionally I  wanted to have the same API for both `post` and `get`.

server/tests/acceptance/curated-content.js
kvas-damian: Done

server/app/lib/mediawiki.js
kvas-damian: Additionally I checked `typedef` and unfortunately it's also not supported by webstorm when using destructing.
The good side of current solution is fact that webstorm autocomplete suggestions are right - to use object as a parameter.

PR #2850, opened 2016-10-05
JPN-660 Support for filtering by multiple categories in discussions tag

perjg:
## Links

* https://wikia-inc.atlassian.net/browse/JPN-660

## Description

Support for filtering by multiple categories. Goes together with this PR for Oasis:
https://github.com/Wikia/app/pull/11559

## Reviewers

@Warkot @bkoval @rogatty

### DIFF
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -25,10 +25,10 @@ export default Ember.Component.extend(
 		 * @returns {void}
 		 */
 		didEnterViewport() {
-			const category = this.get('category'),
+			const categoryIds = this.get('categoryIds'),
 				posts = DiscussionForumModel.find(
 					Mercury.wiki.id,
-					category ? [category] : [],
+					categoryIds,
 					this.get('show')
 				);
 
--- a/front/main/app/components/widget-discussions.js
+++ b/front/main/app/components/widget-discussions.js
@@ -25,7 +25,7 @@ export default Ember.Component.extend(
 		 * @returns {void}
 		 */
 		didEnterViewport() {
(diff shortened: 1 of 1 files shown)


front/main/app/components/widget-discussions.js
```diff
@@ -25,10 +25,10 @@ export default Ember.Component.extend(
 		 * @returns {void}
 		 */
 		didEnterViewport() {
-			const category = this.get('category'),
+			const categoryIds = this.get('categoryIds'),
 				posts = DiscussionForumModel.find(
 					Mercury.wiki.id,
-					category ? [category] : [],
+					categoryIds,
```
Warkot: Would Discussions still work if `categoryIds` are `undefined` or `null`? Maybe you could use `this.getWithDefault('categoryIds', [])` to be sure?

perjg: @Warkot Actually, the parameter starts our as [] on the PHP side, and is filled with any category ids if present, so when there are no categories it will already be [].

I can still change it to getWithDefault if you prefer, but as used now it will have no effect.

perjg: [] is actually the default behavior, it means no catid parameter is present in the discussions wikitext tag.

Warkot: Yeah, I saw the other PR, so I know it's supposed to be a `[]` by default on the `app` side. But should you trust APIs? What if it returns malformed data? Should the application throw an error (expected or not) or handle it gracefully?

Warkot: If you're sure it will be fine, it's OK with me to go either way.

PR #2826, opened 2016-09-27
RECIRC-216: Retire collapsed sections test

pauloslund:
## Links

* http://wikia-inc.atlassian.net/browse/RECIRC-216
* https://wikia-inc.atlassian.net/browse/RECIRC-29 (Original card for collapsed sections)
* https://github.com/Wikia/mercury/pull/2180 (Original PR)

## Description

This removes code that was brought in with RECIRC-29 and should allow us to update to the latest version of `ember-in-viewport`

## Reviewers

@rogatty

PR #2857, opened 2016-10-06
JPN-662 | Added Collapse testing for Japanese community.

d34th4ck3r:
## Links
- https://wikia-inc.atlassian.net/browse/JPN-662
## Description

Reimplementing the test that was removed as https://github.com/Wikia/mercury/commit/680b8d52be30b38830b3bd92a33979c333eaf23c for Japanese community.
## Reviewers

@Warkot @bkoval @rogatty @perjg


rogatty:
What about the styles?
https://github.com/Wikia/mercury/pull/2826

d34th4ck3r:
I have added them now. :)

perjg:
@rogatty Original code (modified to use new A/B test) and styles are restored now. Can you have another look at this PR?

front/main/app/components/article-content.js
```diff
@@ -27,12 +29,13 @@ export default Ember.Component.extend(
 		displayTitle: null,
 
 		articleContentObserver: Ember.on('init', Ember.observer('content', function () {
-			const content = this.get('content');
+			let content = this.get('content');
```
bkoval: Is there any need to assign it here since you don't use it until line 38?

front/main/app/components/article-content.js
```diff
@@ -246,6 +250,22 @@ export default Ember.Component.extend(
 		/**
 		 * @returns {void}
 		 */
+		bindHeaderClicks() {
+			if (!this.$('.collapsible-section-header').length) {
+				return;
+			}
+
+			this.$('.collapsible-section-header').click(function () {
```
bkoval: Don't call the DOM twice. Make sure to cache the result of this.$('.collapsible-section-header') first and reuse the reference, this should upgrade performance :)

front/main/app/components/article-content.js
```diff
@@ -246,6 +250,22 @@ export default Ember.Component.extend(
...
+				const $header = $(this);
+
+				$header.toggleClass('open');
```
bkoval: You can make a one liner out of it: 
$header.toggleClass('open').next('.collapsible-section-body').toggleClass('hidden')

front/main/app/components/article-content.js
```diff
@@ -632,5 +652,55 @@ export default Ember.Component.extend(
 					});
 			}
 		},
+
+		/**
+		 * TO BE THROWN AWAY AFTER JPN_MERCURY_COLLAPSE AB TEST
```
bkoval: Is there any cleanup ticket? It could be referenced here for better understanding of the timeline.

front/main/app/components/article-content.js
```diff
@@ -632,5 +652,55 @@ export default Ember.Component.extend(
...
+		 *
+		 * @param {string} content
+		 * @returns {documentFragment}
```
bkoval: Make sure to capitalize the first letter: DocumentFragment. Complex type constructors in JS always start with the capital letter.

front/main/app/components/article-content.js
bkoval: Also, is there no way to avoid jQuery here? Maybe using the Ember action binding would be a better choice?

front/main/app/components/article-content.js
bkoval: Could you also add more description to it? It's a pretty complicated method so would be nice to know what it does :)

perjg:
@bkoval Thanks for the comments. This code is mostly just the original collapsible section A/B test that was run some time back reinstated. I'd like to consider this throwaway code, and just use as is for the A/B test.

Right now, the task has been put on hold. Will keep the PR open for the time being.

PR #2858, opened 2016-10-06
SOC-3101 | fix: willTransition from base route fired by willtransitio…

latata:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-3101

## Description

Fixing an error with no margins after exiting discussions. WillTransition method from DiscussionBase route was not fired because it was overridden by DiscussionContributionRoute mixin.

## Reviewers

@Wikia/social-frontend


bkoval:
LGTM

PR #2864, opened 2016-10-07
SER-473 'flagged' queue name should be 'questionable'

KolorowyAleksander:
## Links

* http://wikia-inc.atlassian.net/browse/SER-473

## Description

Refactor of queue name both in service and front-end part of image review. We want to have 'questionable' or 'q' everywhere to keep consistency.

@pwojcik86 @rogatty


sqreek:
LGTM

@rogatty can you also take a look

PR #2869, opened 2016-10-11
SOC-2911 | fix: inline editor textarea is readonly if editor isn't active

latata:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-2911

## Description

Textareas in inline editor is readonly by default. It's editable only when editor is active.

## Reviewers

@Wikia/social-frontend

### DIFF
--- a/front/main/app/components/discussion-inline-editor.js
+++ b/front/main/app/components/discussion-inline-editor.js
@@ -28,6 +28,10 @@ export default DiscussionMultipleInputsEditor.extend(
 			return !this.get('isReply');
 		}),
 
+		isReadonly: Ember.computed('isActive', function () {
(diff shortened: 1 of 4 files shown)


front/main/app/components/discussion-inline-editor.js
```diff
@@ -28,6 +28,10 @@ export default DiscussionMultipleInputsEditor.extend(
 			return !this.get('isReply');
 		}),
 
+		isReadonly: Ember.computed('isActive', function () {
```
mleczey: Would this work using `Ember.computed.not`?

latata: Unfortunately not, readonly property has to be undefined (not false) if we want to have editable input

PR #2871, opened 2016-10-12
SOC-3186 Fix for empty guidelines

garthwebb:
Recently the discussion service was deployed and depended on a version of site-attributes that was not deployed.  This caused a bunch of newly created discussion sites to not have a guidelines attribute.

This change makes the FE more resilient to this type of situation, so that users can still update guidelines.  Before this change users could never hit the save button because it was greyed out due to a JS error.

Also, fixed spelling of "toggle".

https://wikia-inc.atlassian.net/browse/SOC-3186


front/main/app/components/discussion-editor.js
```diff
@@ -38,10 +38,10 @@ export default Ember.Component.extend(
 		wasStartTracked: false,
 
 		onIsActive: Ember.observer('isActive', function () {
-			this.toogleActiveState(this.get('isActive'));
+			this.toggleActiveState(this.get('isActive'));
```
mleczey: Can you revert this if no change was done? (found few more in this PR)

front/main/app/mixins/discussion-editor-opengraph.js
```diff
@@ -23,7 +23,12 @@ export default Ember.Mixin.create({
 	onContentChange: Ember.observer('content', function () {
 		this.handleTyping();
 
-		this.set('contentLength', this.get('content').length);
+		let length = 0;
```
mleczey: This could probably be replaced with `const length = this.getWithDefault('content.length', 0);`

front/main/app/components/discussion-editor.js
```diff
@@ -58,7 +58,8 @@ export default Ember.Component.extend(
 
 		submitDisabled: Ember.computed('content', 'currentUser.isAuthenticated', 'showOverlayMessage',
 			'isReply', 'category', 'isGuidelinesEditor', function () {
-				return this.get('content').length === 0 ||
+				return this.get('content') === null ||
```
mleczey: Maybe this could also be replaced with `this.getWithDefault('content.length', 0) === 0`?

front/main/app/components/discussion-editor.js
garthwebb: I changed all occurrences in our area of the codebase.  There are two others that appear that I didn't change because I didn't want QA to expand into non-Social features.

PR #2875, opened 2016-10-17
XW-2208 | Stop using `performance` global directly

rogatty:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2208

## Description

Use `window.performance` instead of `performance`. This should fix `Can't find variable` issue in Safari 9.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/auth/app/common/auth-utils.js
+++ b/front/auth/app/common/auth-utils.js
@@ -31,10 +31,10 @@ export default class AuthUtils {
 
 				window.location.reload();
 			},
-			preferencesRequestStartTime = performance.now();
+			preferencesRequestStartTime = window.performance.now();
 
 		xhr.onload = () => {
-			const preferencesRequestEndTime = performance.now();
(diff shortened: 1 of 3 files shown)


front/auth/app/signup/signup-form.js
```diff
@@ -198,10 +198,10 @@ export default class SignupForm {
 						challengeStartTime,
 						challengeEndTime;
 
-					challengeStartTime = performance.now();
+					challengeStartTime = window.performance.now();
```
hakubo: This might still break if performance is not available on window object

front/auth/app/common/auth-utils.js
```diff
@@ -31,10 +31,10 @@ export default class AuthUtils {
 
 				window.location.reload();
 			},
-			preferencesRequestStartTime = performance.now();
+			preferencesRequestStartTime = window.performance.now();
```
hakubo: This might still break if performance is not available on window object

PR #2883, opened 2016-10-20
SOC-3208 | Do not override computed property - local categories.

mleczey:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-3208

## Description

Do not override computed property - local categories.

## Reviewers
@Wikia/social-frontend


front/main/app/components/discussion-categories-edit.js
```diff
@@ -71,7 +71,7 @@ export default Ember.Component.extend(ResponsiveMixin,
 				}
 
 				this.setProperties({
-					categoryToDelete: category,
+					categoryToDelete: this.get('categories').findBy('id', category.get('id')),
```
latata: Is that necessary?

mleczey: No, it's object form `localCategories` and I want here object from `categories`.

PR #2893, opened 2016-10-21
CONN-738 Fixing the uglyness of Curated Main Pages

gabrys:
No-one wants to fix it, so here I come


front/main/app/templates/components/trending-articles-item.hbs
```diff
@@ -1,6 +1,10 @@
 <figure>
 	<div class="item-image" style={{imageStyle}}>
-		<img src="{{currentlyRenderedImageUrl}}" alt="{{title}}" />
+		{{#if currentlyRenderedImageUrl}}
+			<img src="{{currentlyRenderedImageUrl}}" alt="{{title}}" />
+		{{else}}
+			<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAQAIBRAA7" alt="{{title}}" style="width: 56%" />
```
gabrys: The 56% causes the tiles with no image to be of the proper aspect ratio. For videos we don't need it for two reasons: I images all videos have their thumbnails; and they occupy 100% width, so there's no misalignment when mixed square tiles with non-square ones.

gabrys:
@rogatty thank you for the review. By uglyness I mean how it looks on bigger screens, the fact the viewport is used as the base for dimensions and not the actual width and the fact it's for no apparent reason coded in JavaScript; and duplication of the code to handle the dimensions in different components :-). The fact this fixes those issues by removing 96 lines of code and adding just 15 is a nice illustration of the situation I think.

PR #2894, opened 2016-10-24
ADEN | fix mercuryPV key-val not appearing for regular slots

bognix:
## Description

MercuryPV key-val was not sent becasue it was added too late. Moving it three lines above fixes the issue (above `startOnLoadQueue` call)
 
...and it has to be a string not an int.


jacekjursza:
LGTM

fraszczakszymon:
LGTM

bognix:
synthetic tests passed, merging.

PR #2902, opened 2016-10-25
WW-453: Track clicks only on given element, not it's whole area

lukaszkonieczny:
## Links

* https://wikia-inc.atlassian.net/browse/WW-453
* https://wikia-inc.atlassian.net/browse/WW-456
* https://wikia-inc.atlassian.net/browse/WW-435

## Description

We want to track clicks only on real element, not whole block element are

## Reviewers

@Wikia/west-wing @Wikia/social-frontend


front/main/app/styles/component/discussion/_discussion-contributors.scss
```diff
@@ -4,13 +4,9 @@ $discussion-contributors-border-width: 2px;
...
 	display: inline-block;
 
-	.event-wrapper {
-		display: inline-block;
-		cursor: unset;
-	}
-
 	.user-avatar {
 		display: inline-block;
+		margin: 0 2px;
```
sebastianmarzjan: What is being accomplished by this change?

lukaszkonieczny: Keeping space between avatars. I'm not sure why, adding `div` wrapper removes it. I suppose it's connected with spaces between inline elements, because I didn't find any `padding` or `maring` values in styles.

idradm:
It seems (at least to me) that action cannot be passed through helper, so my last comment is no longer valid.

PR #2906, opened 2016-10-26
XW-2253 | handle wikia.org logo in top bar

tomasznapieralski:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2253
* https://github.com/Wikia/design-system/pull/96
* https://github.com/Wikia/app/pull/11723

## Reviewers

@Wikia/x-wing


front/main/app/components/site-head-fandom-bar.js
```diff
@@ -1,15 +1,13 @@
...
 import {track, trackActions} from 'common/utils/track';
 
-const {computed, Component, get} = Ember;
+const {Component} = Ember;
 
 export default Component.extend({
 	classNames: ['site-head-fandom-bar-wrapper'],
 
-	homeOfFandomLabel: get(Mercury, 'wiki.navigation2016.fandomLabel'),
-	isVisible: computed.not('isSearchPage'),
```
hakubo: does that mean we want to show it on a search page?

front/main/app/components/site-head.js
```diff
@@ -27,6 +27,12 @@ export default Component.extend(
 
 		wikiaHomepage: M.prop('globalNavigation.logo.header.href') || 'http://fandom.wikia.com',
 
+		displayFandomBar: computed(() => {
```
hakubo: What is the benefit of using computed property here as opposed to:
Boolean(M.prop('globalNavigation.logo-small-tagline'))

front/main/app/templates/components/site-head.hbs
```diff
@@ -27,8 +27,9 @@
 		</div>
 	</nav>
 </div>
-{{site-head-fandom-bar
-	wikiaHomepage=wikiaHomepage
-	isSearchPage=isSearchPage
-}}
+{{#if displayFandomBar}}
```
hakubo: wouldn't it better if we passed isVisible to component?
We recently learned that if's are heavy and isVisible was already imlemented

server/app/views/_partials/site-head.hbs
```diff
@@ -7,8 +7,13 @@
...
+				{{#if (equal globalNavigation.logo-small.header.image-data.name 'wds-company-logo-wikia-org')}}
```
hakubo: before the logo was hard coded both on Front-End and Back-end
now we have control on the front end
but if we'll change API to a possible third logo variant - this will also have to be updated, right?
Can we do it in such a way, that we won't have to do it?

front/main/app/components/site-head-fandom-bar.js
rogatty: We don't.

tomasznapieralski: This property was not used.

server/app/views/_partials/site-head.hbs
tomasznapieralski: added comment in file

front/main/bower.json
```diff
@@ -3,7 +3,7 @@
   "dependencies": {
     "Autolinker.js": "0.22.0",
     "cropper": "0.11.0",
-    "design-system": "Wikia/design-system#1.1.0",
+    "design-system": "Wikia/design-system#XW-2253",
```
rogatty: Remember, remember.

PR #2907, opened 2016-10-26
XW-2188 | added SCSS variable 

weronika:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2188

## Description

this variable is moved from design system, as it is Mercury specific


front/main/app/styles/component/design-system/_global-navigation.scss
```diff
@@ -2,6 +2,8 @@
 // If you look for Global Navigation styles go here:
 // https://github.com/Wikia/design-system/blob/master/components/_global-navigation.scss
 
+$wds-global-navigation-outer-height: $nav-height + $border-top-width;
```
kvas-damian: As we agreed off-line, let's define this variable in DS and hide parts as privates.

PR #2909, opened 2016-10-26
SOC-3196 | introducing following modal

latata:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-3196

## Description

Introducing following modal

## Reviewers

@Wikia/social-frontend


front/main/app/styles/component/discussion/_discussion-introducing-following.scss
```diff
@@ -0,0 +1,50 @@
+.introducing-following {
+
+	.header {
+		align-items: center;
+		display: flex;
+		height: 84px;
+		justify-content: center;
+
+		.follow-white {
+			fill: #fff;
```
bkoval: If you could, define a var for it in _settings.scss and reference it here.

front/main/app/styles/component/discussion/_discussion-introducing-following.scss
```diff
@@ -0,0 +1,50 @@
+.introducing-following {
+
+	.header {
+		align-items: center;
+		display: flex;
+		height: 84px;
+		justify-content: center;
+
+		.follow-white {
```
bkoval: I'd rather not use color name directly in the className, I'd rather try to name it after the class purpose, not it's look, which can actually change in the future.

front/main/app/templates/components/modal-dialog.hbs
```diff
@@ -1,4 +1,4 @@
 <div class="modal-dialog-layover" {{action 'close'}}></div>
 <div class="modal-dialog">
-	{{yield}}
+	{{yield (action 'close')}}
```
latata: We talked about it, works good.

front/main/app/templates/discussion/forum.hbs
```diff
@@ -37,3 +37,11 @@
 			setEditorActive=(action 'setEditorActive')
 	}}
 {{/discussion-feed}}
+{{#modal-dialog
+		class='introducing-following-modal'
```
latata: We talked about it. It's ok

PR #2913, opened 2016-10-27
XW-1888 | Fix scrolling after discussion reply is posted

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1888

## Description

After a reply is posted, scroll to the end of the page while keeping the Global Footer below the fold.
Bonus: replaced the old badges SVGs with the DS ones. They look the same so there is no point of loading them twice.

## Reviewers

@Wikia/x-wing @Wikia/social-frontend


front/main/app/components/discussion-inline-reply-editor.js
```diff
@@ -17,7 +20,11 @@ export default DiscussionInlineEditor.extend(
 		 * @returns {void}
 		 */
 		scrollAfterEntityAdded() {
-			Ember.$('html, body').animate({scrollTop: window.document.body.scrollHeight});
+			// Scroll to the end of the page while keeping the Global Footer below the fold
+			const scrollTop = window.document.body.scrollHeight -
```
hakubo: did we consider: https://developer.mozilla.org/en/docs/Web/API/Element/scrollIntoView ?

rogatty: Interesting but I see that `scrollIntoViewOptions` is supported only in Firefox and it seems to be required for animation.

PR #2914, opened 2016-10-28
XW-2146 | Be smart about Design System strings

rogatty:
## Links
- https://wikia-inc.atlassian.net/browse/XW-2146
- https://github.com/Wikia/app/blob/1eaf966ed10b6be6305635b285381429de86f19f/extensions/wikia/DesignSystem/DesignSystemHelper.class.php#L98-L158
## Description

Render text based on value of the `type` field, support following types:
- `text`
- `translatable-text`
- `link-text`
## Reviewers

@Wikia/x-wing


front/main/app/utils/wds-text.js
```diff
@@ -0,0 +1,56 @@
+// Keep in sync with /server/app/views/_helpers/wds-text.js
+
+const maxRecursionDepth = 10;
+
+/**
+ * @see http://shebang.brandonmintern.com/foolproof-html-escaping-in-javascript/
+ * @param {String} unsafe
+ * @returns {String}
+ */
+function escapeAttribute(unsafe) {
```
kvas-damian: Is it possible to use one of existing functions?
For Handlebars we have `Handlebars.Utils.escapeExpression()`.
Maybe we can find something similar for HTMLBars?

front/main/app/utils/wds-text.js
```diff
@@ -0,0 +1,56 @@
...
+
+			return i18n.t(fields.key, params);
+		} else {
+			return i18n.t(fields.key, {
+				ns: 'design-system'
+			});
+		}
+	} else if (fields.type === 'link-text') {
+		return `<a href="${escapeAttribute(fields.href)}">${wdsText(fields.title, recursionDepth + 1)}</a>`;
+	} else {
```
kvas-damian: Is it normal state that we got here?
I believe it's an edge-case that isn't handled yet so probably we need to add support for. Can we add logging for it?

front/main/app/utils/wds-text.js
```diff
@@ -0,0 +1,56 @@
...
+			}
+
+			return i18n.t(fields.key, params);
+		} else {
+			return i18n.t(fields.key, {
+				ns: 'design-system'
+			});
+		}
+	} else if (fields.type === 'link-text') {
+		return `<a href="${escapeAttribute(fields.href)}">${wdsText(fields.title, recursionDepth + 1)}</a>`;
```
kvas-damian: Please an unit test for it.

kvas-damian: BTW safety of this method depend on input. We let developer decide whether to escape it's output or not. Currently  `text` and `translatable-text` are not escaped (escaping is delegated to method that calls it) and `link-text` can't be escaped.
Can we unify the output and make it clear if it should be escaped or not?

server/app/views/_helpers/wds-text.js
```diff
@@ -0,0 +1,52 @@
...
+
+module.exports = function wdsText(fields, recursionDepth) {
+	if (!fields || recursionDepth > maxRecursionDepth) {
+		return '';
+	}
+
+	if (fields.type === 'text') {
+		return fields.value;
+	} else if (fields.type === 'translatable-text') {
+		const translateWithCache = this.i18n.translateWithCache,
```
kvas-damian: Can we use https://github.com/Wikia/mercury/blob/3d234bc25de97dc2a86d71f5557e5377bdaaa7a0/server/app/views/_helpers/i18n.js#L8-L8 ?

rogatty:
Quoting @kvas-damian on Slack:
>I think our `shouldEscape` approach won't work.
>The problem is that in `translatable-text` we can have a link as a parameter. We can only escape parts of the link not entire HTML, so when we pass a link as a i18next parameter and then escape it then link won't work.
>Previously we discussed escaping and unescaping when we use output of `wdsText()` method. It won't work for the same reason.  We can't unescape and escape a link HTML because it won't be the same string.

We've decided it's too complicated and not worth the trouble. Closing and won't fix.

PR #2915, opened 2016-10-28
SOC-3205 | feat: add error message when following not succeeded

mleczey:
## Links
- https://wikia-inc.atlassian.net/browse/SOC-3205
## Description

Add error message when following not succeeded
## Reviewers

@Wikia/social-frontend

### DIFF
--- a/front/main/app/styles/component/discussion/_discussion-alert-notification.scss
+++ b/front/main/app/styles/component/discussion/_discussion-alert-notification.scss
@@ -1,6 +1,6 @@
 .discussions {
 	.alert-notifications {
-		z-index: $z-1;
+		z-index: $z-8;
 
 		.discussions-action-failed {
 			background-color: $discussion-alert-notification-background-color;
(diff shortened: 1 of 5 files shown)


front/main/app/styles/component/discussion/_discussion-alert-notification.scss
```diff
@@ -0,0 +1,25 @@
+.discussions {
+	.alert-notifications {
+		z-index: $z-1;
```
bkoval: I fe don't need it, let's consider removing it.

front/main/app/styles/component/discussion/_discussion-alert-notification.scss
```diff
@@ -0,0 +1,25 @@
...
+			font-weight: $font-weight-bold;
+			margin: 0 auto;
+			padding-bottom: 9px;
+			padding-top: 9px;
+			text-align: center;
+		}
+	}
+}
+
+@media #{$desktop-range} {
```
bkoval: If the only width here is 1176px then we can remove the media query whatsoever, since mobile widths end up at 1063px. 🎲

front/main/app/styles/component/discussion/_discussion-alert-notification.scss
mleczey: Actually had to give $z-8 to show at front of reply editor.

PR #2916, opened 2016-11-02
SER-340 added csv download button

KolorowyAleksander:
## Links

* http://wikia-inc.atlassian.net/browse/SER-340

## Description

Added a download as csv button, which connects to the csv endpoint.

## Reviewers
@sqreek

### DIFF
--- a/front/main/app/models/image-review/index.js
+++ b/front/main/app/models/image-review/index.js
@@ -33,12 +33,6 @@ ImageReviewModel.reopenClass({
 		});
 	},
 
-	endSession(batchId) {
-		return request(M.getImageReviewServiceUrl(`/batch/${batchId}`, {}), {
-			method: 'DELETE',
-		});
-	},
-
 	getImages(batchId) {
 		return request(M.getImageReviewServiceUrl(`/batch/${batchId}`, {}));
 	},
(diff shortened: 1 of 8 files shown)


front/main/app/routes/image-review/index.js
```diff
@@ -85,9 +85,10 @@ export default Route.extend({
 		},
 
 		willTransition(transition) {
-			const isStayingOnEditor = transition.targetName.indexOf('image-review.summary') > -1;
+			const isStayingInReview = transition.targetName.indexOf('image-review') > -1;
+			Logger.error(isStayingInReview);
```
sqreek: ?

front/main/app/models/image-review/index.js
```diff
@@ -27,20 +27,12 @@ ImageReviewModel.reopenClass({
...
 				return ImageReviewModel.createEmptyModelWithPermission(status);
-			} else if (jqXHR.status === 404) {
-				return ImageReviewModel.createEmptyModelWithPermission(status);
 			} else {
 				return ImageReviewModel.getImagesAndPermission(payload.id, status);
 			}
 		});
 	},
 
-	endSession(batchId) {
```
sqreek: maybe we want to have endSession?

KolorowyAleksander: For now, there is no 'end session/ delete batch' endpoint in the service. If we want to consider adding that, we'll need to add such an endpoint aswell.

PR #2919, opened 2016-11-02
XW-2209 | Update to new DS version

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2209

## Description
Newest version of DS removes square avatars so we're removing this class.

## Reviewers

@Wikia/x-wing


kvas-damian:
LGTM ;)

PR #2921, opened 2016-11-02
SOC-3115 | Add thread ID to tracking data for post details pageviews

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-3115

## Description

Also, fixes a pollution of tracking context for all mercury tracking requests.

## Reviewers

@rogatty @Wikia/social-frontend


front/common/utils/track.js
```diff
@@ -150,7 +150,7 @@ export function track(params) {
 	}
 
 	if (trackingMethod === 'both' || trackingMethod === 'internal') {
-		params = $.extend(context, params);
+		params = $.extend({}, context, params);
```
bkoval: @rogatty @hakubo Was this intended before? Whenever you use $.extend in this (old) way, it will copy all the params from each tracking request into the context object. This way all the subsequent tracking requests were polluted (prefilled?) with data coming from a previous request.

This is not visible if all the requests have the same params set, but once you want to add a custom param it will also add it to all subsequent requests, which is unwanted.

I figured this might be a bug so I'm adding this change.

front/main/app/routes/discussion/post.js
```diff
@@ -3,6 +3,8 @@ import DiscussionPostModel from '../../models/discussion/post';
 import DiscussionContributionRouteMixin from '../../mixins/discussion-contribution-route';
 import DiscussionModerationRouteMixin from '../../mixins/discussion-moderation-route';
 import DiscussionModalDialogMixin from '../../mixins/discussion-modal-dialog';
+import {trackPageView} from 'common/utils/track';
+import {track, trackActions} from '../../utils/discussion-tracker';
```
mleczey: Do you need those two imports here?

front/common/utils/track.js
rogatty: I'd say it's 50/50 bug/feature :wink: 
I agree that this looks like a bug but I worry that some code might rely on it to prefill the context. Please test if the most critical events like pageviews are still sent with the same params as on production. If yes, then it LGTM.

PR #2927, opened 2016-11-04
SOC-3218 | As a user, I want to see an indicator than my post or reply was edited by a admin+

mleczey:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-3218

## Description

As a user, I want to see an indicator than my post or reply was edited by a admin+

## Reviewers

@Wikia/social-frontend

### DIFF
--- a/front/main/app/components/discussion-post-card-detail.js
+++ b/front/main/app/components/discussion-post-card-detail.js
@@ -16,7 +16,12 @@ export default DiscussionPostCardBaseComponent.extend(
 		// Whether the component is displayed on the post details discussion page
 		isDetailsView: false,
 
-		lastEditedBy: Ember.computed('post.userData.permissions.canModerate', function () {
+		showLastEditedByMessage: Ember.computed('post.lastEditedBy', function() {
+			return this.get('post.lastEditedBy')
+				&& this.get('post.createdBy.id') !== this.get('post.lastEditedBy.id');
+		}),
+
+		lastEditedByMessage: Ember.computed('post.userData.permissions.canModerate', function () {
 			const userName = this.getWithDefault('post.lastEditedBy.name', '');
 
 			return this.get('post.userData.permissions.canModerate')
--- a/front/main/app/components/discussion-post-card-detail.js
+++ b/front/main/app/components/discussion-post-card-detail.js
@@ -16,12 +16,13 @@ export default DiscussionPostCardBaseComponent.extend(
 		// Whether the component is displayed on the post details discussion page
 		isDetailsView: false,
 
-		showLastEditedByMessage: Ember.computed('post.lastEditedBy', function () {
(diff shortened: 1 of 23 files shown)


front/main/app/models/discussion/domain/contributor.js
```diff
@@ -25,13 +25,18 @@ DiscussionContributor.reopenClass({
...
-		return this._super({
-			avatarUrl: data.avatarUrl,
-			badgePermission: data.badgePermission,
-			id: data.id,
-			name: data.name,
-			profileUrl: DiscussionContributor.getProfileUrl(data.name)
-		});
+		let result = null;
+
+		if (data) {
```
bkoval: Why do we need to change it?

front/main/app/templates/components/discussion-post-card-detail.hbs
```diff
@@ -70,6 +70,11 @@
 			isListView=true
 		}}
 	{{/if}}
+	{{#if post.lastEditedBy}}
+		<div class="row post-edited-by-row">
+			{{lastEditedBy}}
```
bkoval: Shouldn't it be post.lasEditedBy?

front/main/app/models/discussion/domain/contributor.js
mleczey: In case if data is null.

front/main/app/templates/components/discussion-post-card-detail.hbs
mleczey: As discussed, variable name changed to `lastEditedByMessage`.

front/main/app/components/discussion-post-card-detail.js
```diff
@@ -16,6 +16,19 @@ export default DiscussionPostCardBaseComponent.extend(
 		// Whether the component is displayed on the post details discussion page
 		isDetailsView: false,
 
+		showLastEditedByMessage: Ember.computed('post.lastEditedBy', function() {
+			return this.get('post.lastEditedBy')
```
latata: It will return boolean or value returned by `post.lastEditedBy`. Don't we want to return boolean always?

front/main/app/components/discussion-post-card-detail.js
```diff
@@ -16,6 +16,19 @@ export default DiscussionPostCardBaseComponent.extend(
...
+				&& this.get('post.createdBy.id') !== this.get('post.lastEditedBy.id');
+		}),
+
+		lastEditedByMessage: Ember.computed('post.userData.permissions.canModerate', function () {
+			const userName = this.getWithDefault('post.lastEditedBy.name', '');
```
latata: Shoudn't computed property be dependent on `post.lastEditedBy.name` value?

PR #2929, opened 2016-11-07
XW-2203 | using DS variable for body font family

weronika:
* using $wds-font-family for defining body font family

https://wikia-inc.atlassian.net/browse/XW-2203

### DIFF
--- a/front/main/app/styles/component/design-system/_wikia-styleguide-overrides.scss
+++ b/front/main/app/styles/component/design-system/_wikia-styleguide-overrides.scss
@@ -1,7 +1,3 @@
-body {
-	font-family: $wds-font-family;
-}
-
 .wds-button {
 	height: auto;
 	margin-bottom: 0;
--- a/front/main/app/styles/component/design-system/_wikia-styleguide-overrides.scss
+++ b/front/main/app/styles/component/design-system/_wikia-styleguide-overrides.scss
@@ -1,3 +1,7 @@
+body {
+	font-family: $wds-font-family;
+}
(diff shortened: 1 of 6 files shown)


kvas-damian:
I'm not 100% sure if we should change it, but I found `font-family definition in https://github.com/Wikia/mercury/blob/dev/server/app/views/_layouts/error.hbs#L11

What do you think?

kvas-damian:
The same question about auth pages: https://www.wikia.com/signin

weronika:
We decided not to change font-family in https://github.com/Wikia/mercury/blob/dev/server/app/views/_layouts/error.hbs#L11, as styling there is defined in head, so it's not possible to use DS variable.

rogatty:
@weronika please just copy the variable's value to `error.hbs`. Chinese errors will look better :smiley:

front/main/app/styles/component/design-system/_wikia-styleguide-overrides.scss
```diff
@@ -1,3 +1,7 @@
+body {
+	font-family: $wds-font-family;
```
rogatty: I don't think it belongs here. We don't override DS styles but use a variable from it.
How about https://github.com/Wikia/mercury/blob/dev/front/main/app/styles/base/_body.scss?

weronika: definitions in this file don't override ds styles, but wikia-styleguide styles

PR #2933, opened 2016-11-08
SER-340 fixed setting start and end dates in model

KolorowyAleksander:
## Links
* http://wikia-inc.atlassian.net/browse/SER-340

## Description
Model properties need to be set by a setter, not directly.

## Reviewers
@nikodamn

### DIFF
--- a/front/main/app/models/image-review/summary.js
+++ b/front/main/app/models/image-review/summary.js
@@ -20,19 +20,19 @@ const ImageReviewSummaryModel = Ember.Object.extend({
 	},
 
 	setStartDate(startDate) {
-		this.startDate = moment(startDate).format('YYYY-MM-DD');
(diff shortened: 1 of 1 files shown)


front/main/app/models/image-review/summary.js
```diff
@@ -20,19 +20,19 @@ const ImageReviewSummaryModel = Ember.Object.extend({
 	},
 
 	setStartDate(startDate) {
-		this.startDate = moment(startDate).format('YYYY-MM-DD');
-		this.csvLink = M.getImageReviewServiceUrl('/statistics/csv', {
+		this.set('startDate', moment(startDate).format('YYYY-MM-DD'));
+		this.set('csvLink', M.getImageReviewServiceUrl('/statistics/csv', {
 			startDate: this.startDate,
```
rogatty: Use getters, too. Will save you fixing next bug ;)

PR #2943, opened 2016-11-16
ADEN-4183 Animate VUAP on Mercury.

drets:
* http://wikia-inc.atlassian.net/browse/ADEN-4183

### DIFF
--- a/front/main/app/styles/module/_ads.scss
+++ b/front/main/app/styles/module/_ads.scss
@@ -1,6 +1,4 @@
 @import 'ads-floor';
-// Use Wikia style guide settings
-@import '../../../bower_components/wikia-style-guide/src/scss/lib/settings';
 
 $closeButtonHeight: 40px;
 
@@ -34,9 +32,6 @@ $closeButtonHeight: 40px;
 
 .vuap-loaded {
 	transition: padding 0.4s ease;
-	-webkit-transition: padding 0.4s ease;
-	-moz-transition: padding 0.4s ease;
-	-o-transition: padding 0.4s ease;
 }
 
 .bfaa-template .mobile-top-leaderboard {
@@ -52,9 +47,6 @@ $closeButtonHeight: 40px;
 		padding: 0;
 		position: relative;
 		transition: height 0.4s ease;
-		-moz-transition: height 0.4s ease;
-		-o-transition: height 0.4s ease;
-		-webkit-transition: height 0.4s ease;
 
 		iframe {
 			height: 100%;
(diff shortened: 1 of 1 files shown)


front/main/app/styles/module/_ads.scss
```diff
@@ -39,9 +46,14 @@ $closeButtonHeight: 40px;
 	width: 100%;
 
 	.ad-slot {
-		background: none;
+		background: #000;
```
nandy-andy: What happens if there is no background color set? I wonder if we could use here some pre-defined SCSS variable.

nandy-andy: How about [$color-black](https://github.com/Wikia/style-guide/blob/dev/src/scss/lib/_settings.scss#L114) used for example for [menu list text color](https://github.com/Wikia/mercury/blob/dev/front/main/app/styles/_settings.scss#L130)?

front/main/app/styles/module/_ads.scss
```diff
@@ -1,6 +1,7 @@
 @import 'ads-floor';
+// Use Wikia style guide settings
+@import '../../../bower_components/wikia-style-guide/src/scss/lib/settings';
```
bognix: are you sure you have to import that?
You should use colors from front/main/bower_components/design-system/dist/scss/wds-variables/_colors.scss and this file is already imported.

front/main/app/styles/module/_ads.scss
```diff
@@ -31,6 +32,13 @@ $closeButtonHeight: 40px;
 	margin: 0 -20px;
 }
 
+.vuap-loaded {
+	transition: padding 0.4s ease;
+	-webkit-transition: padding 0.4s ease;
+	-moz-transition: padding 0.4s ease;
+	-o-transition: padding 0.4s ease;
```
bognix: no need for all the prefixes. http://caniuse.com/#feat=css-transitions

front/main/app/styles/module/_ads.scss
```diff
@@ -39,9 +47,14 @@ $closeButtonHeight: 40px;
 	width: 100%;
 
 	.ad-slot {
-		background: none;
+		background: $color-black;
```
bognix: why the black color as background?

drets: https://github.com/Wikia/mercury/pull/2943#discussion_r88451142

bognix: name of black color in DS is `$wds-color-black: #1a1a1a;`. Are you sure color-black is going to work?

drets: y

front/main/app/styles/module/_ads.scss
nandy-andy: As discussed in bigger group on Slack: doesn't make sense for background.

PR #2937, opened 2016-11-14
SOC-1975 | Link to d/f?sort=latest on all post list variations

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-1975

## Description

Affects canonical links on post list pages.

## Reviewers

@Wikia/social-frontend

PR #2944, opened 2016-11-16
SOC-1975 | Canonical link for Forum + adding acceptance tests for Discussion

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-1975

## Description

New pull request, needed to close https://github.com/Wikia/mercury/pull/2937

## Reviewers

@Wikia/social-frontend

### DIFF
--- a/front/main/app/routes/discussion/base.js
+++ b/front/main/app/routes/discussion/base.js
@@ -17,7 +17,10 @@ export default Ember.Route.extend(
 		 * @returns {void}
 		 */
 		setDynamicHeadTags(model, data = {}) {
+			const shouldSetDefaultCanonical = data.canonical === undefined;
+
 			data.documentTitle = 'Discussions';
+
 			if (data.canonical === undefined) {
 				data.canonical = `${Ember.get(Mercury, 'wiki.basePath')}${window.location.pathname}`;
 			}
--- a/front/main/app/routes/discussion/base.js
+++ b/front/main/app/routes/discussion/base.js
@@ -18,7 +18,13 @@ export default Ember.Route.extend(
 		 */
 		setDynamicHeadTags(model, data = {}) {
 			data.documentTitle = 'Discussions';
-			data.canonical = data.canonical || `${Ember.get(Mercury, 'wiki.basePath')}${window.location.pathname}`;
+
+			// We do not want to set a canonical for pages other than first
+			if (this.get('controller.page') === 1) {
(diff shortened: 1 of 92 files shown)


front/main/app/routes/discussion/base.js
```diff
@@ -18,7 +18,13 @@ export default Ember.Route.extend(
 		 */
 		setDynamicHeadTags(model, data = {}) {
 			data.documentTitle = 'Discussions';
-			data.canonical = `${Ember.get(Mercury, 'wiki.basePath')}${window.location.pathname}`;
+
+			// We do not want to set a canonical for pages other than first
+			if (this.get('controller.page') === 1) {
```
latata: Why do you check it in base route instead of forum route?

bkoval:
@rogatty I would like to leave double quotes for our fixtures, since otherwise it would be problematic to convert these from JSON (double quotes) to JS objects (singles where needed, rest without quotes) each time we add something new. What do you think?

front/main/mirage/config.js
```diff
@@ -3,6 +3,7 @@
  */
 export default function () {
 	this.passthrough('/write-blanket-coverage');
+	this.passthrough('https://localhost/**');
```
rogatty: How about a const for `https://localhost/`?

rogatty:
@bkoval would JSON files work here? I think they should.

front/main/app/routes/discussion/base.js
```diff
@@ -18,7 +18,9 @@ export default Ember.Route.extend(
 		 */
 		setDynamicHeadTags(model, data = {}) {
 			data.documentTitle = 'Discussions';
-			data.canonical = `${Ember.get(Mercury, 'wiki.basePath')}${window.location.pathname}`;
+			if (data.canonical === undefined) {
```
mleczey: I assume that you want to execute this code when `canonical` is only `undefined` and not when is `null`?

bkoval:
@rogatty We'll try it next time. I just copied JSON here and exported it from fixtures since valid JSON is a valid JS object :)

front/main/app/routes/discussion/base.js
```diff
@@ -17,8 +17,13 @@ export default Ember.Route.extend(
 		 * @returns {void}
 		 */
 		setDynamicHeadTags(model, data = {}) {
+			const shouldSetDefaultCanonical = data.canonical === undefined;
```
rogatty: @bkoval you forgot about something? :wink: It's a shame that linter didn't shout at this unused variable.

PR #2947, opened 2016-11-17
SER-604 allow to display images in rejected queue after vignette change

KolorowyAleksander:
## Links

* http://wikia-inc.atlassian.net/browse/SER-604

## Description

Changing frontend to add ?status=REJECTED to rejected images, so they can be shown.

## Reviewers

@sqreek 
@africazulu 
@tmszdmsk

### DIFF
--- a/front/main/app/models/image-review/index.js
+++ b/front/main/app/models/image-review/index.js
@@ -76,11 +76,11 @@ ImageReviewModel.reopenClass({
 				images.push(Ember.Object.create({
 					batchId,
 					imageId: image.imageId,
-					fullSizeImageUrl: image.imageUrl,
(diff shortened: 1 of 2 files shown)


front/main/app/models/image-review/index.js
```diff
@@ -76,11 +76,11 @@ ImageReviewModel.reopenClass({
 				images.push(Ember.Object.create({
 					batchId,
 					imageId: image.imageId,
-					fullSizeImageUrl: image.imageUrl,
+					fullSizeImageUrl: status === 'REJECTED' ? `${image.imageUrl}?status=REJECTED` : image.imageUrl,
```
KolorowyAleksander: In my opinion we should send a link with ?status=REJECTED from service when asked for rejected images. This way only staff/reviewer (trusted) users can see the query param.

PR #2958, opened 2016-11-21
SOC-3195 | followed by post list backend integration

latata:
## Links

* https://wikia-inc.atlassian.net/browse/SOC-3195

## Description

Integration with backend

## Reviewers

@Wikia/social-frontend

### DIFF
--- a/front/main/app/models/discussion/domain/post.js
+++ b/front/main/app/models/discussion/domain/post.js
@@ -76,7 +76,9 @@ DiscussionPost.reopenClass({
 				// A hack to compensate for API sometimes returning numbers and sometimes strings
 				categoryId: String(threadData.forumId),
 				createdBy: DiscussionContributor.create(threadData.createdBy),
-				creationTimestamp: threadData.creationDate.epochSecond,
+				// We need to support both epoch object and iso string dates
+				creationTimestamp: typeof threadData.creationDate === 'string' ?
+					(new Date(threadData.creationDate)).getTime() / 1000 : threadData.creationDate.epochSecond,
 				id: threadData.firstPostId,
 				isDeleted: threadData.isDeleted,
 				isFollowed: threadData.isFollowed,
@@ -94,8 +96,12 @@ DiscussionPost.reopenClass({
 				upvoteCount: parseInt(threadData.upvoteCount, 10),
 				userBlockDetails: DiscussionUserBlockDetails.create(threadData.userBlockDetails)
 			}),
-			userData = Ember.get(threadData, '_embedded.userData.0'),
-			openGraphData = Ember.get(threadData, '_embedded.openGraph.0');
+			/*
+			 * if it's gotten from followed-by end-point we get userData and openGraph in plain object,
+			 * not in an array, we still need to support /threads end-point with an arrays result
+			 */
+			userData = Ember.get(threadData, '_embedded.userData.0') || Ember.get(threadData, '_embedded.userData'),
+			openGraphData = Ember.get(threadData, '_embedded.openGraph.0') || Ember.get(threadData, '_embedded.openGraph');
 
 		if (openGraphData) {
 			post.set('openGraph', OpenGraph.create(openGraphData));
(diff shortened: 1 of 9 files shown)


front/main/app/models/discussion/domain/post.js
```diff
@@ -76,7 +76,9 @@ DiscussionPost.reopenClass({
 				// A hack to compensate for API sometimes returning numbers and sometimes strings
 				categoryId: String(threadData.forumId),
 				createdBy: DiscussionContributor.create(threadData.createdBy),
-				creationTimestamp: threadData.creationDate.epochSecond,
+				// We need to support both epoch object and iso string dates
+				creationTimestamp: typeof threadData.creationDate === 'string' ?
```
mleczey: Could you extract method here?

front/main/app/models/discussion/domain/post.js
```diff
@@ -94,8 +94,12 @@ DiscussionPost.reopenClass({
 				upvoteCount: parseInt(threadData.upvoteCount, 10),
 				userBlockDetails: DiscussionUserBlockDetails.create(threadData.userBlockDetails)
 			}),
-			userData = Ember.get(threadData, '_embedded.userData.0'),
-			openGraphData = Ember.get(threadData, '_embedded.openGraph.0');
+			/*
+			 * if it's gotten from followed-by end-point we get userData and openGraph in plain object,
+			 * not in an array, we still need to support /threads end-point with an arrays result
+			 */
+			userData = Ember.get(threadData, '_embedded.userData.0') || Ember.get(threadData, '_embedded.userData'),
```
bkoval: We should have DiscussionUserData and DiscussionOpenGraphData models created for these fields, then this whole extraction logic could be moved to their constructors :)

front/main/app/models/discussion/domain/post.js
```diff
@@ -107,6 +111,15 @@ DiscussionPost.reopenClass({
 
 		return post;
 	},
+
+	/**
+	 * Gets timestamp from date that can be iso string or epoch object
+	 * @param date
```
bkoval: {*} - for mixed type values

front/main/app/models/discussion/domain/post.js
```diff
@@ -107,6 +111,15 @@ DiscussionPost.reopenClass({
...
+	 * @returns {number} - timestamp
+	 */
+	getThreadDataTimestamp(date) {
+		return date === 'string' ? (new Date(date)).getTime() / 1000 : date.epochSecond;
```
bkoval: no need for paratheses around new Date()

front/main/app/models/discussion/follow.js
```diff
@@ -17,16 +17,8 @@ const DiscussionFollowedPostsModel = DiscussionBaseModel.extend(
...
-			const requestUrl = M.getDiscussionServiceUrl(`/${this.wikiId}/threads`),
-				requestData = {
-					userId: user.get('userId'),
-					limit: this.get('loadMoreLimit'),
-					page: this.get('data.pageNum') + 1,
-					pivot: this.get('pivotId'),
-					viewableOnly: false
-				};
-			return this.loadThreadPage(requestUrl, requestData);
+			const requestUrl = M.getDiscussionServiceUrl(`/${this.wikiId}/threads/followed-by/${user.get('userId')}`);
```
bkoval: Is there a need for this const anymore? Maybe we could return it directly to preserve any unneeded memory allocation? :)

front/main/app/templates/components/discussion-forum-wrapper.hbs
```diff
@@ -52,13 +52,14 @@
...
-	{{discussion-wrapper-load-more
-		goToPage=(action this.attrs.goToPage)
-		hasMore=hasMore
-		isLoading=isLoading
-		loadNextPage=(action 'loadNextPage')
-		showButton=true
-	}}
-{{/if}}
+{{#unless isFollowedPostsView}}
+	{{#if showLoadMoreButton}}
```
bkoval: If in and unless is already a bit of logic - maybe moving it to the .js and combining into one var would help separating logic and view layers? :)

latata: Firstly I wanted to do this but finally I left it as it is because this is just temporary solution - disabling paging in followed by tab.

front/main/app/models/discussion/follow.js
latata: I think it looks more readable.

PR #2755, opened 2016-09-05
ADEN-3731 Add Ubisoft tracking

v3rth:
## Links

* https://wikia-inc.atlassian.net/browse/ADEN-3731

PR #2963, opened 2016-11-23
ADEN-3731 removed Ubisoft pixel

nandy-andy:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-3731
* Previous PR: https://github.com/Wikia/mercury/pull/2755

## Description

Removal of sunset functionality.


nandy-andy:
Good question :+1: Most likely we should :)

PR #2968, opened 2016-11-25
XW-2195 | Setup mercury for staging

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2195

## Description

Changes needed to support mercury in a staging environment.
I took a moment to clean up a bit settings for mercury and rename some files.

Changes introduced with this PR

### localSettings renamed to settings
It is not local anymore. Local is just confusing in this context.

So all localSettings.*.js got renamed to settings.*.js
localSettings.js -> settings.js
localSettings.base.js -> settings.base.js
localSettings.dev.js -> settings.dev.js
localSettings.example.js -> settings.example.js

I have also added 
settings.prod.js
and 
settings.staging.js

### Settings cascading
When settings.js is imported it loads
settings.base.js and all settings are based on this file
then it loads
settings.${environment}.js
and merges it with base

if such a file is not found (e.g. for sanbox)
it just loads base settings

in settings.js - you can override all settings for your local instance

BREAKING CHANGE - move your local config from
localSettings.js to settings.js

### Environments.Prod -> environments.prod
in utils.js
you could find Environments that is not a num and is not a class
I renamed it to environments
and all its keys to lower case (prod, dev etc.)

## Reviewers

@Wikia/x-wing


server/app/lib/utils.js
```diff
@@ -79,21 +61,21 @@ export function stripDevboxDomain(host) {
...
 
 /**
- * @param {LocalSettings} localSettings
+ * @param {Settings} settings
  * @returns {string}
  */
-export function getCDNBaseUrl(localSettings) {
-	return localSettings.environment !== Environment.Dev ? localSettings.cdnBaseUrl : '';
+export function getCDNBaseUrl(settings) {
+	return settings.environment === environments.prod ? settings.cdnBaseUrl : '';
```
rogatty: Let's check if this works properly on sandboxes.

server/app/lib/utils.js
```diff
@@ -280,37 +264,37 @@ export function createServerData(localSettings, wikiDomain = '') {
...
  * @param {Hapi.Request} request
  * @returns {string}
  */
-export function getStaticAssetPath(localSettings, request) {
-	const env = localSettings.environment || Environment.Dev;
+export function getStaticAssetPath(settings, request) {
+	const env = settings.environment || environments.dev;
 
-	return env !== Environment.Dev ?
+	return env === environments.prod ?
```
rogatty: Let's check if this works properly on sandboxes.

rogatty:
I think these could use some fixes too:
- https://github.com/Wikia/mercury/blob/dev/front/common/baseline/build-url.js#L48-L69
- https://github.com/Wikia/mercury/blob/dev/server/app/facets/auth/auth-view.js#L47
- https://github.com/Wikia/mercury/blob/dev/front/main/app/utils/domain.js - I'm not sure, we should check if it still works

hakubo:
@rogatty thanks for links - copied them to tickets

server/app/lib/utils.js
Warkot: It might be a good point. And what about `preview`, `verify` and other envs different then `dev`?

bkoval:
@Wikia/social-frontend

bkoval:
LGTM!

PR #2971, opened 2016-11-28
ADEN-4296 Add background color to BFAB

v3rth:
## Links

* https://wikia-inc.atlassian.net/browse/ADEN-4296

## Description

Add background color to BFAB slot


front/main/app/styles/module/_ads.scss
```diff
@@ -74,6 +74,7 @@ $player-collapse-button: 'data:image/svg+xml;base64,PHN2ZyBpZD0iTGF5ZXJfMSIgZGF0
 	.bfab-template {
 		padding: 0;
 		position: relative;
+		background-color: #000000;
```
drets: Please, change `#000000` to `#000` or change `#000` to `#000000` in different place in this file 
(I prefer `#000` since it's less bytes over the network).

PR #2981, opened 2016-12-02
XW-2290 | update Design System

hakubo:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2290

## Description

update Design System

## Reviewers

@Wikia/x-wing


bkoval:
https://github.com/Wikia/design-system/blob/master/package.json#L3 - is that outdated?

hakubo:
@bkoval you're right, I messed it up... fixing now

PR #2984, opened 2016-12-05
WW-621 | tracking impressions of curated main pages

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/WW-621

## Description

Adding tracking of impressions of curated main pages

## Reviewers
@Wikia/west-wing


front/main/app/routes/wiki-page.js
```diff
@@ -184,6 +184,13 @@ export default Ember.Route.extend(
 					controller: handler.controllerName,
 					model
 				});
+
+				if (model.isCuratedMainPage) {
+					track({
+						action: trackActions.impression,
+						category: 'main-page'
```
sebastianmarzjan: Maybe curated-main-page would be more distinguishable?

front/main/app/mixins/main-page-route.js
```diff
@@ -94,5 +94,10 @@ export default Ember.Mixin.create({
 				curatedContent: model
 			}
 		});
+
+		track({
+			action: trackActions.impression,
+			category: `main-page-${model.type}`
```
sebastianmarzjan: Same here, maybe curated- prefix would be useful to distinguish from 'normal' main pages

rybmat: I considered that, however decided to use only 'main-page' to be consistent with current tracking

PR #3001, opened 2016-12-21
SER-646 Added support for separate link for rejected images

jcellary:
## Links

* http://wikia-inc.atlassian.net/browse/SER-646

## Description

Changed link separate to reflect different schema when accessing rejected images


front/main/app/models/image-review/index.js
```diff
@@ -76,7 +76,7 @@ ImageReviewModel.reopenClass({
 				images.push(Ember.Object.create({
 					batchId,
 					imageId: image.imageId,
-					fullSizeImageUrl: status === 'REJECTED' ? `${image.imageUrl}?status=REJECTED` : image.imageUrl,
+					fullSizeImageUrl: image.imageUrl,
```
sqreek: what url is behind imageUrl?

jcellary: @tmszdmsk can tell you all about it :D

It's either http://vignette.wikia-dev.pl/14ce9de8-9537-40d0-b61e-9e19e918b620
or https://services.wikia-dev.pl/image-review/rejected-image-proxy/178e3695-3083-49c6-9a82-b4ef89bf7a1b for images that have the status "rejected"

PR #3019, opened 2017-01-10
XW-2577 | Retire Curated Main Pages Categories

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2577

## Description

We had 2 different pages that represent the same content: CMP category pages and mobile category pages. We decided to retire CMP category pages and use only mobile category page to not confuse search crawlers and users.

## Reviewers

Ping @Wikia/x-wing


front/main/app/models/curated-content.js
```diff
@@ -152,25 +150,11 @@ CuratedContentModel.reopenClass({
 			// MercuryApi (categories for section) returns article_local_url, ArticlesApi (subcategories) returns url
 			url = rawData.url ? rawData.url : rawData.article_local_url;
 
-			// TODO (CONCF-914): article_local_url is sometimes encoded and sometimes not, to investigate
```
rogatty: Will this work when the URL is encoded? What if it's not? Unfortunately I don't remember any examples, but it should be fairly easy to mock the data in the API and test it.

kvas-damian: The fix that was applied here was encoding `categoryName` which is not used now at all. We migrated to URL. 

The fix was applied in: https://wikia-inc.atlassian.net/browse/ZZZ-2096
I checked: http://de.clashofclans.wikia.com/main/section/Truppen - all categories still work after the fix. Looks that it's correctly handled by mobile category pages.

PR #3021, opened 2017-01-10
SOC-3558 decode only the URL

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-3558

### DIFF
--- a/front/main/app/mixins/discussion-parsed-content.js
+++ b/front/main/app/mixins/discussion-parsed-content.js
@@ -36,6 +36,8 @@ export default Ember.Mixin.create({
 
 		if (!this.get('shouldActivateLinks')) {
 			this.set('autolinkerConfig.replaceFn', this.wrapInSpan);
+		} else {
+			this.set('autolinkerConfig.replaceFn', this.decodeUri)
 		}
 		this._super();
 	},
@@ -48,6 +50,14 @@ export default Ember.Mixin.create({
 	wrapInSpan(match) {
 		if (match.getType() === 'url') {
 			return `<span class='url'>${decodeURIComponent(match.getUrl())}</span>`;
+		}
+	},
+
+	decodeUri(match) {
+		if (match.getType() === 'url') {
+			let tag = match.buildTag();
+			tag.setInnerHtml(decodeURIComponent(tag.getInnerHtml()));
+			return tag;
 		}
 	}
 });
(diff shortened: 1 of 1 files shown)


front/main/app/mixins/discussion-parsed-content.js
```diff
@@ -1,6 +1,14 @@
-import Ember from 'ember';
-import nl2br from 'common/utils/nl2br';
-import {truncate, shouldUseTruncationHack} from '../utils/truncate';
+import Ember from "ember";
```
mleczey: Linter is failing, please fix it.

front/main/app/mixins/discussion-parsed-content.js
```diff
@@ -1,6 +1,14 @@
...
+import nl2br from "common/utils/nl2br";
+import {truncate, shouldUseTruncationHack} from "../utils/truncate";
+
+function decodeUriSafely(uri) {
```
mleczey: Please move this method inside mixin.

front/main/app/mixins/discussion-parsed-content.js
```diff
@@ -36,6 +44,8 @@ export default Ember.Mixin.create({
 
 		if (!this.get('shouldActivateLinks')) {
```
mleczey: Maybe you could move here truthful block? (change condition and blocks of code)?

front/main/app/mixins/discussion-parsed-content.js
```diff
@@ -47,7 +57,17 @@ export default Ember.Mixin.create({
...
 		if (match.getType() === 'url') {
-			return `<span class='url'>${match.getUrl()}</span>`;
+			return `<span class='url'>${decodeUriSafely(match.getUrl())}</span>`;
 		}
-	}
+	},
+
+	decodeInnerHtml(match) {
+		if (match.getType() === 'url') {
+			let tag = match.buildTag();
```
mleczey: Please add new line after let.

front/main/app/mixins/discussion-parsed-content.js
```diff
@@ -47,7 +57,17 @@ export default Ember.Mixin.create({
...
+			tag.setInnerHtml(decodeUriSafely(tag.getInnerHtml()));
+			return tag;
```
mleczey: What should be returned when the condition in `if` is not met? Could you be explicit about it?

PR #3024, opened 2017-01-11
XW-2581 | displaying wiki description on Curated Main Page

weronika:
## Links

* https://github.com/Wikia/app/pull/12211
* https://wikia-inc.atlassian.net/browse/XW-2581

## Description

Added wiki description on Curated Main Page


front/main/app/styles/component/_wiki-description.scss
```diff
@@ -0,0 +1,5 @@
+.wiki-description {
+	line-height: 1.5;
+	margin-top: -$featured-content-margin-bottom;
```
kvas-damian: We probably should use use variable that is specific for other component. What would happen if we change order of imports or remove `featured-content`?
We could:
* have separate variables
* have shared variable that is stored in a file that can be included in both components.

weronika: Yes, you're right... These can't be separate variables, because that was the purpose of creating it - to have a common variable. So I guess I'll create separate file for shared variables, earlier I just thought it may be superfluous for only one variable.

front/main/app/styles/component/index.scss
```diff
@@ -1,4 +1,5 @@
 // file order is intended here - put new entries at the bottom
+@import 'variables';
```
kvas-damian: What do you think about using `_wikia-variables.scss` file?

weronika: I don't mind :)

front/main/app/styles/component/index.scss
```diff
@@ -1,4 +1,5 @@
 // file order is intended here - put new entries at the bottom
+@import 'wikia-variables';
```
kvas-damian: sorry for not being clear in my previous message. I was referring to reusing existing file: https://github.com/Wikia/mercury/blob/8f8cf192ab212f9d648a715b824b578bbd0666df/front/main/app/styles/_wikia-variables.scss#L119-L118

front/main/app/styles/_wikia-variables.scss
```diff
@@ -116,3 +116,8 @@ $discussion-header-height: 50px;
 $theme-bar-height: 10px;
 
 $discussion-post-type-smaller-than-min: .7rem;
+
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Main Page
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+$featured-content-margin-bottom: 20px;
```
kvas-damian: One last comment. Can we rename this variable as it is used not only for featured content? Maybe using `main-page` prefix?

weronika: Do you mean ```main-page``` instead of ```featured-content```? Or additionally? The second use is not placed in  "featured-content.scss" but it is certainly used in featured content's context, so I wouldn't say this name is misleading.

kvas-damian: makes sense! Let's leave it as it is.

PR #3026, opened 2017-01-13
XW-2580 | Prerender Curated Main Pages

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2580

## Description

Currently Curated Main Pages are rendered only on client side which makes it hard for search engine bots to crowl their content. To solve that problem we decided to prerender critical content of CMP on server side.

## Reviewers

Ping @Wikia/x-wing


server/app/views/_partials/curated-main-page/featured-content.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="featured-content mw-content">
+	<a href="{{featuredArticle.article_local_url}}" class="featured-content-item">
+		<figure>
+			<img src="{{{featuredArticle.thumb_url}}}"
```
latata: why triple brackets? why do we need to have it unescaped?

server/app/views/_helpers/truncate.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Truncate input string and add ellipsis if needed
+ *
+ * @see Copy of front/main/app/utils/truncate.js
+ *
+ * @param {String} text
+ * @param {Integer} maxLength
+ * @returns {string}
+ */
+module.exports = function truncate(text, maxLength = 48) {
```
latata: I think it would be great to have unit tests for this function

kvas-damian: As stated in jsDoc - it's a copy of frontend method. Do you think we should have a copy of unit tests as well?

server/app/views/_helpers/truncate.js
```diff
@@ -0,0 +1,37 @@
...
+
+	if (text.length <= maxLength) {
+		return text;
+	}
+
+	truncatedString = text.substr(0, maxLength);
+	lastWhiteSpacePos = truncatedString.search(/\s[^\s]*$/);
+
+	if (lastWhiteSpacePos === maxLength || lastWhiteSpacePos < 0) {
+		return truncatedString + ellipsisCharacter;
```
latata: it looks like it can produce somoething like this: "sample text ..." (whitespace before ellipsis character), doesn't it?

server/app/views/_partials/curated-main-page/featured-content.hbs
kvas-damian: Don't know why but escaping escapes `=` in query string.

server/app/views/_partials/curated-main-page/curated-content.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="curated-content mw-content">
+	<nav class="curated-content-items">
+		{{#each curatedContent as |curatedItem| }}
+			<a href="{{#if curatedItem.url}}{{curatedItem.url}}{{else}}{{curatedItem.article_local_url}}{{/if}}" class="curated-content-item section">
```
latata: I wonder if we could avoid these if/else statements in template and have prepared objects by js.

server/app/views/_helpers/truncate.js
kvas-damian: I don't think it's possible. We substring to `maxLength` then we check if space is located at `maxLength` position. Positions are indexed from 0 ;)

Having that said, I'm not sure why do we have `lastWhiteSpacePos === maxLength` condition. It'll never be true.

server/app/views/_helpers/truncate.js
latata: I'd add annotation that it's a copy of frontend method

kvas-damian: it's already there: https://github.com/Wikia/mercury/pull/3026/files#diff-4373b54ba72acae381d41344e1f26142R8

server/app/facets/mediawiki-page.js
```diff
@@ -137,6 +138,9 @@ function handleResponse(request, reply, data, allowCache = true, code = 200) {
 	// mainPageData is set only on curated main pages - only then we should do some special preparation for data
 	if (isMainPage && pageData.mainPageData) {
 		result = deepExtend(result, prepareMainPageData(data));
+		result = prepareCuratedMainPageData(result);
```
rogatty: Why new function / file? `prepareMainPageData` is for CMP already.

server/app/facets/operations/prepare-curated-main-page-data.js
```diff
@@ -0,0 +1,63 @@
...
+		options.mode = mode;
+	}
+
+	return Vignette.getThumbURL(imageUrl, options);
+}
+
+/**
+ * Get thumbs for curated main page
+ *
+ * @see front/main/app/mixins/curated-content-thumbnail.js
```
rogatty: Let's also add a backlink there.

server/app/views/_partials/curated-main-page/curated-content.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="curated-content mw-content">
```
rogatty: Can you add a comment to all these files stating that they are copies? Include the original filepath, too.

kvas-damian: Done

PR #3031, opened 2017-01-17
XW-2593 | Preprender categories module on articles and categories

kvas-damian:
## Links
* https://wikia-inc.atlassian.net/browse/XW-2593

## Description
Content rendered on server side is more attractive for Search Engine crawlers. This PR introduce prerendering categories module on article and category pages

## Reviewers
ping @Wikia/x-wing


server/app/views/_partials/article-footer/categories.hbs
```diff
@@ -0,0 +1,14 @@
+<nav class="collapsible-menu">
```
rogatty: Are these the same as on front-end? Please add a comment if yes (also backlinks).

kvas-damian: Done

server/app/views/_partials/article-footer/categories.hbs
```diff
@@ -0,0 +1,15 @@
+{{! Keep the same HTML structure as in front/main/app/templates/components/article-wrapper.hbs }}
+<nav class="collapsible-menu">
+	<div class="collapsed title">
```
hakubo: since they are collapsed - I imagine they are displayed: none - would anything index it?

kvas-damian: It's mobile view. it's acceptable that some content is hidden by default but accessible in UI.

PR #3032, opened 2017-01-17
XW-2590 | Remove /main/section route and render all sections on the main page

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2590
* https://github.com/Wikia/app/pull/12234

## Reviewers

@kvas-damian @rybmat @idradm @hakubo


front/main/app/components/curated-content-item.js
```diff
@@ -62,7 +62,7 @@ export default Ember.Component.extend(
 			});
 
 			if (itemType && itemType === 'section') {
-				this.sendAction('openCuratedContentItem', this.get('model'));
+				this.get('openSection')(this.get('model'));
```
idradm: Whats wrong with `sendAction`?

front/main/app/components/featured-content-item.js
```diff
@@ -10,17 +10,20 @@ export default Ember.Component.extend(
...
-			return this.generateThumbUrl(
-				this.get('model.image_url'),
-				this.get(`model.image_crop.${this.get('aspectRatioName')}`)
-			);
+			if (this.get('model.imageUrl')) {
+				return this.generateThumbUrl(
+					this.get('model.imageUrl'),
+					this.get(`model.imageCrop.${this.get('aspectRatioName')}`)
+				);
+			} else {
```
idradm: You could omit this `else`.

front/main/app/components/featured-content-item.js
```diff
@@ -10,17 +10,20 @@ export default Ember.Component.extend(
...
 		imageWidth: 400,
 		cropMode: Thumbnailer.mode.zoomCrop,
 		thumbUrl: Ember.computed('model', function () {
-			return this.generateThumbUrl(
-				this.get('model.image_url'),
-				this.get(`model.image_crop.${this.get('aspectRatioName')}`)
-			);
+			if (this.get('model.imageUrl')) {
+				return this.generateThumbUrl(
+					this.get('model.imageUrl'),
```
idradm: You can reuse `get` value.

front/main/app/styles/component/_curated-content.scss
```diff
@@ -98,33 +94,17 @@ $grid-icon-size: 22px;
...
-		left: -$back-button-padding;
-		margin-left: 10px;
-		padding: $back-button-padding;
-		position: absolute;
-		top: -$back-button-padding;
-		transform: rotate(90deg);
-		white-space: nowrap;
-		width: $back-button-with-padding-size;
+		svg {
+			margin-bottom: 3px;
```
idradm: Use vars for magic numbers.

front/main/app/templates/components/curated-content.hbs
```diff
@@ -1,16 +1,16 @@
...
-		</button>
+{{curated-content-section
+	model=model
+	isVisible=(equal activeLabel null)
+	openSection=(action 'openSection')
+}}
+{{#each model.items as |item|}}
+	{{#if (equal item.type 'section')}}
+		{{curated-content-section
+			isVisible=(equal activeLabel item.label)
```
idradm: Maybe this should be a computed field instead?

front/main/app/utils/wiki-handlers/curated-main-page.js
```diff
@@ -7,10 +7,10 @@ import CuratedContentModel from '../../models/curated-content';
  * @returns {Object}
  */
 function getCuratedContentModel(mainPageModel) {
-	if (mainPageModel.mainPageData.curatedContent) {
+	if (mainPageModel.curatedMainPageData.curatedContent) {
```
idradm: Is it possible that `curatedMainPageData` object is not set? What about using `get` method?

front/main/app/utils/wiki-handlers/curated-main-page.js
```diff
@@ -7,10 +7,10 @@ import CuratedContentModel from '../../models/curated-content';
...
 		return CuratedContentModel.create({
 			type: 'section',
-			items: CuratedContentModel.sanitizeItems(mainPageModel.mainPageData.curatedContent)
+			items: mainPageModel.curatedMainPageData.curatedContent.items
```
idradm: You could extract `curatedContent` value to var and reuse it here.

server/app/facets/operations/prepare-curated-main-page-data.js
```diff
@@ -20,44 +20,62 @@ function getImageThumb(imageUrl, width, height, mode, imageCrop) {
...
+				);
+			});
+		}
 
-	if (mainPageData && mainPageData.curatedContent) {
-		mainPageData.curatedContent.forEach((item) => {
-			item.url = item.url || item.article_local_url;
-			item.label = item.label || item.title;
+		if (curatedMainPageData.curatedContent && curatedMainPageData.curatedContent.items) {
+			curatedMainPageData.curatedContent.items.forEach(setThumbsForCuratedContent);
```
idradm: Can we use `map` function instead? It would be more clear that the data is changed here.

front/main/app/components/curated-content-item.js
rogatty: This is the syntax used in the official docs:
https://guides.emberjs.com/v2.5.0/components/triggering-changes-with-actions/#toc_passing-the-action-to-the-component

Also, IIRC `sendAction` works only if the passed action is an action name (a string), not a closure action (a function).

front/main/app/components/main-page.js
```diff
@@ -33,7 +33,7 @@ export default Ember.Component.extend(
 		/**
 		 * @returns {void}
 		 */
-		didRender() {
+		didInsertElement() {
```
rogatty: `didRender` was called twice for some reason and we were injecting the ads twice. I used the opportunity to fix it.

front/main/app/styles/component/_curated-content.scss
rogatty: This is debatable. This whole file has a lot of "magic numbers" and the reasoning was that when the number isn't reused in at least 2 places then putting it in a variable doesn't give you any value, just overhead.
Which one is easier to read:
```
$curated-content-section-back-svg-margin-bottom: 3px;
...
...
...
...
a lot of other stuff
...
...

.curated-content-section__back {
  svg {
    margin-bottom: $curated-content-section-back-svg-margin-bottom;
  }
}
```
vs. what we have currently?

front/main/app/templates/components/curated-content-section.hbs
```diff
@@ -0,0 +1,23 @@
+{{! Keep the same HTML structure as in server/app/views/_partials/curated-main-page/curated-content.hbs }}
+{{#if model.items}}
+	{{#if model.label}}
+		<div class="curated-content-section__header">
+			<a href="#" class="curated-content-section__back"{{action closeSection}}>
+				{{svg 'wds-icons-arrow' viewBox='0 0 20 16' role='img' class='icon'}}
+				{{i18n 'app.navigation-back-label'}}
```
rogatty: FYI this string isn't used anywhere else, it was forgotten and now _back_ in the game.

front/main/app/templates/components/curated-content.hbs
rogatty: In this template we invoke `curated-content-section` component multiple times and decide if each one should be visible. Computed property would return a single value for all of them.

Or do you propose to pass `activeLabel` to the component and compute `isVisible` in there? It would work but I think it adds complexity.

front/main/app/components/curated-content-item.js
hakubo: Please check what will happen if you don't pass it to the component

for now we can use Ember.K as a default handler.
openSection: Ember.K

front/main/app/components/curated-content.js
```diff
@@ -1,35 +1,24 @@
 import Ember from 'ember';
-import CuratedContentModel from '../models/curated-content';
```
hakubo: const {Component, $} = Ember;

front/main/app/components/main-page.js
hakubo: 👍

front/main/app/router.js
```diff
@@ -18,10 +18,6 @@ Router.map(function () {
 
 	this.route('infobox-builder');
 
-	this.route('mainPageSection', {
```
hakubo: you served us well!

server/app/facets/operations/prepare-curated-main-page-data.js
```diff
@@ -20,44 +20,60 @@ function getImageThumb(imageUrl, width, height, mode, imageCrop) {
...
-				400,
-				400 / 16 * 9,
-				Vignette.mode.zoomCrop,
-				item.image_crop && item.image_crop.landscape
-			);
-		});
-	}
+function prepareCuratedMainPageModules(curatedMainPageData) {
+	if (curatedMainPageData) {
+		const featuredContent = curatedMainPageData.featuredContent,
```
hakubo: const {a, b} = c;

front/main/app/styles/component/_curated-content.scss
idradm: Then why do we have so many unique vars? Maybe padding vars should be shared.

front/main/app/components/curated-content.js
```diff
@@ -1,35 +1,24 @@
...
 	actions: {
 		/**
+		 * @param {CuratedContentItem} item
 		 * @returns {void}
 		 */
-		loadMore() {
-			this.set('isLoading', true);
+		openSection(item) {
+			const navHeight = Ember.$('.site-head').outerHeight() + Ember.$('.site-head-fandom-bar').outerHeight(),
+				scrollTop = this.$().offset().top - navHeight;
```
kvas-damian: What about smart banner? Are we going to include it in these calculations?

front/main/app/templates/components/curated-content-section.hbs
```diff
@@ -0,0 +1,23 @@
+{{! Keep the same HTML structure as in server/app/views/_partials/curated-main-page/curated-content.hbs }}
+{{#if model.items}}
+	{{#if model.label}}
+		<div class="curated-content-section__header">
+			<a href="#" class="curated-content-section__back"{{action closeSection}}>
```
kvas-damian: space before `action` helper ;)

front/main/app/components/curated-content.js
rogatty: No, because it stays in its place at the top of the page and doesn't cover the section title.

PR #3034, opened 2017-01-18
XW-2585 refactor wiki-card-list, reuse search component

idradm:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2585

## Description

Appears on section styles.

## Reviewers

@kvas-damian


front/main/app/models/wiki/file.js
```diff
@@ -116,6 +116,14 @@ FileModel.reopenClass({
 
 		model.setProperties(pageProperties);
 	},
+
+	prepareFileUsageItem(item) {
+		return {
+			href: get(item, 'titleDBkey'),
```
kvas-damian: what do you think about calling it `prefixedTitle` or something that doesn't suggest valid url?

kvas-damian: Additionally please make sure we don't allow character that are not valid in URLs.

front/main/app/styles/component/wikia-ui-components/_wikia-card.scss
```diff
@@ -1,9 +1,24 @@
...
 	background: white;
 	box-shadow: 0 1px 1px 0 rgba(0, 0, 0, 0.1);
 	display: block;
 	margin-bottom: $card-margin;
 	width: 100%;
+
+	.wikia-card__item {
+		font-size: $type-smallest;
+		line-height: $line-height-min;
+		padding: $search-vertical-padding $search-horizontal-padding;
```
kvas-damian: Probably we should have variables specific for this component and stop reusing search variables.

front/main/app/templates/components/wikia-ui-components/wikia-card.hbs
```diff
@@ -1,3 +1,8 @@
-<div class="wikia-card">
-	{{yield}}
-</div>
+{{#link-to 'wiki-page' item.href trackingCategory=trackingCategory trackingLabel=trackingLabel}}
+	<div class="wikia-card__item">
```
kvas-damian: While we wrap content in a link it should be possible to get rid of this `div` and put styles directly on `<a>`

front/main/app/models/wiki/file.js
idradm: This is the format used in `wikia-card` component shared with `search` page results. Do we want to refactor it there as well?

kvas-damian: If it's used in the same context then it makes sense to refector it in search as well.

PR #3036, opened 2017-01-19
SER-731 add COPPA image review

pwojcik86:
## Links

* http://wikia-inc.atlassian.net/browse/SER-731

## Description

Added a new view to `image-review`, which shows all images uploaded by a specific user, and allows to remove them.

## Reviewers

@Wikia/services-team


front/main/app/components/image-review.js
```diff
@@ -7,6 +7,8 @@ export default Ember.Component.extend({
 	thumbnailModel: {},
 	newestImages: Ember.computed.equal('order', 'NEWEST'),
 	oldestImages: Ember.computed.equal('order', 'OLDEST'),
+	isCoppa: false,
```
tmszdmsk: It's probably "easier" to do it like that, but think if in the long term it's not better to split it into "features configuration" list.

PR #3039, opened 2017-01-20
SOC-3768 added the forum=html query parameter to all resources with content

slayful:
https://wikia-inc.atlassian.net/browse/SOC-3768


front/main/app/mixins/discussion-forum-model-static.js
```diff
@@ -6,7 +6,7 @@ export default Ember.Mixin.create(
 		findThreads(modelInstance, requestUrl, requestData) {
 			return new Ember.RSVP.Promise((resolve, reject) => {
 				request(requestUrl, {
-					data: requestData,
+					data: this.addFormat(requestData),
```
bkoval: This method returns, let's name it in a getter convention.

front/main/app/mixins/discussion-forum-model-static.js
```diff
@@ -22,6 +22,16 @@ export default Ember.Mixin.create(
 					reject(modelInstance);
 				});
 			});
+		},
+
+		/**
+		 * @private
+		 */
+		addFormat(requestData) {
```
bkoval: Please, use mixin to reuse this logic and compose it into all the models.

front/main/app/models/discussion/post.js
```diff
@@ -199,7 +199,8 @@ DiscussionPostModel.reopenClass({
 					responseGroup: 'full',
 					sortDirection: 'descending',
 					sortKey: 'creation_date',
-					viewableOnly: false
+					viewableOnly: false,
+					format: 'html',
```
bkoval: Let's use an abstracted decorator method here as well.

front/main/app/models/discussion/reported-posts.js
```diff
@@ -23,7 +23,8 @@ const DiscussionReportedPostsModel = DiscussionBaseModel.extend(
 					page: this.get('data.pageNum') + 1,
 					pivot: this.get('pivotId'),
 					viewableOnly: true,
-					reported: true
+					reported: true,
+					format: 'html',
```
bkoval: and here

front/main/app/models/discussion/reported-posts.js
```diff
@@ -79,6 +80,7 @@ DiscussionReportedPostsModel.reopenClass({
 					limit: reportedPostsInstance.get('postsLimit'),
 					reported: true,
 					viewableOnly: true,
+					format: 'html',
```
bkoval: and here

bkoval: If there are any data object fields that are repetitive, move them to abstracted methods.

front/main/app/models/discussion/post.js
```diff
@@ -194,13 +195,13 @@ DiscussionPostModel.reopenClass({
 				urlPath = replyId ? `/${wikiId}/permalinks/posts/${replyId}` : `/${wikiId}/threads/${threadId}`;
 
 			request(M.getDiscussionServiceUrl(urlPath), {
-				data: {
+				data: postInstance.getRequestDataWithFormat({
```
slayful: So I'm using the Mixin to extend the `DiscussionPostModel`.
Calling `this. getRequestDataWithFormat..` here won't work since it's used in a function introduced in ` DiscussionPostModel.reopenClass({`.

PR #3041, opened 2017-01-23
SER-755 passing query param with currently selected source to backend

tmszdmsk:
## Links

* http://wikia-inc.atlassian.net/browse/SER-755

## Description

Getting number of images by status & SOURCE.

## Reviewers
@Wikia/services-team


front/main/app/models/image-review/batch-id.js
```diff
@@ -77,13 +76,17 @@ ImageReviewModel.reopenClass({
...
 			getUserPermissions(),
 			getImageSources(),
-			getBatch(batchId)
-		]).then(([permissions, sources, batch]) =>
-			ImageReviewModel.create(Ember.assign(permissions, sources, batch, {status})));
+			getBatch(batchId),
+			getImageCount(status, source)
+		]).then(([permissions, sources, batch, imagesCount]) => {
+			return ImageReviewModel.create(
+				Ember.assign(permissions, sources, batch, {status}, {imagesToReviewCount: imagesCount.countByStatus})
```
pwojcik86: Perhaps we should change `countByStatus` to just `count`

front/main/app/routes/image-review/batch-id.js
```diff
@@ -6,11 +6,10 @@ const {Route, Logger} = Ember;
 export default Route.extend({
 
 	model(params) {
```
pwojcik86: I think we can use destructuring here - `{batchId, status, source}`

front/main/app/models/image-review/batch-id.js
tmszdmsk: ofc, I need to change backend first ;)

PR #3042, opened 2017-01-23
SOC-3590 | Catch the exception and serve appropriate error page

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-3590


server/app/facets/show-discussions.js
```diff
@@ -17,12 +17,14 @@ export default function showDiscussions(request, reply) {
...
 		if (!variables.enableDiscussions) {
-			return reply('Not Found').code(404);
+			return reply('Not Foundxx').code(404);
 		}
 
 		context.documentTitle = `Discussions | ${variables.siteName} | Fandom powered by Wikia`;
 		context.showSpinner = true;
 
 		showApplication(request, reply, wikiVariables, context, true);
+	}).catch(() => {
```
rogatty: Would one of the existing exceptions work in this case?
https://github.com/Wikia/mercury/blob/63184f1ab1688f7a529b27b933dc35ad39de35bd/server/app/facets/mediawiki-page.js#L60-L73

PR #3037, opened 2017-01-20
XW-2585 | File pages on mercury

kvas-damian:
## Links
* https://wikia-inc.atlassian.net/browse/XW-2585

## Description
* introducing file pages on mobile
* mediawiki models refactoring
* introducing `wikia-card-list` component

## Reviewers
Ping @Wikia/x-wing 

Don't merge as it's not full feature.

PR #3050, opened 2017-01-25
XW-1667 | Category page - show categories on empty article

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/XW-1667

## Description

Show categories on category pages without article.

## Reviewers

Ping @Wikia/x-wing


front/main/app/templates/category.hbs
```diff
@@ -1,33 +1,24 @@
...
+		showComments=false
+		siteName=siteName
+		toggleDrawer='toggleDrawer'
+		wikiPageHeaderSubtitle=(i18n 'app.category-page-subtitle')
+	}}
 		{{category-sections
 			adsContext=model.adsContext
 			sections=model.sections
-			setupAds=true
+			setupAds=false
```
rogatty: What does it mean? Do we still need it?

kvas-damian: I have no idea. It was like this before so I left it in place ;)

rogatty: Not really, it was `true` before so at first glance it looks like you disabled the ads ;)

hakubo: yeah, from what I see we should remove it as
this only tells category-sections to setupAdsContext

something that will already happen in articleWrapper

front/main/app/mixins/ads.js
```diff
@@ -90,7 +91,7 @@ export default Ember.Mixin.create({
 		}
 
 		if (showPreFooter) {
-			this.appendAd(adsData.mobilePreFooter, 'after', $articleBody);
+			this.appendAd(adsData.mobilePreFooter, 'before', $articleFooter);
```
kvas-damian: Ping @nandy-andy 
We moved this add to be displayed after yielded content in `article-wrapper` to avoid displaying ad under the ad on category page when there is no article.

front/main/app/templates/category.hbs
```diff
@@ -1,33 +1,23 @@
...
-				adsContext=model.adsContext
-				sections=model.sections
-				setupAds=false
-				loadBatch=(action 'loadBatch')
-			}}
-		{{/article-wrapper}}
-	{{else}}
-		{{wikia-ui-components/wiki-page-header title=model.displayTitle subtitle=(i18n 'app.category-page-subtitle')}}
+	{{#article-wrapper
+		addPhoto='addPhoto'
```
rogatty: I think `addPhoto` and `edit` are not necessary.

kvas-damian: Maybe next time...

front/main/app/mixins/ads.js
nandy-andy: It doesn't seem to change anything on regular articles since there is nothing between `$articleBody` and `$articleFooter`. It may change viewability of `MOBILE_PREFOOTER` on category pages but it will definitely fix UX on those pages. Thanks for letting me know, I'll share the knowledge :bowing_man:

rogatty:
Merge after https://github.com/Wikia/mercury/pull/3037 is merged.

PR #3051, opened 2017-01-25
XW-2583 | Premiumize the Category page

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2583
* https://github.com/Wikia/app/pull/12292


front/main/app/controllers/category.js
```diff
@@ -20,33 +20,44 @@ export default Controller.extend(
 
 		actions: {
 			/**
+			 * @returns {void}
+			 */
+			articleRendered() {
+				this.get('article').send('articleRendered', ...arguments);
```
hakubo: It looks more readable to me when you have ...params
```function (...params) {
   func(...params)
}```
This way you don't have to use 'magic' arguments

File: front/main/app/styles/component/_category-members-grouped.scss [hunk 2]
```diff
@@ -0,0 +1,34 @@
+.category-members-grouped {
+	padding: 1.5rem $article-horizontal-padding;
```
hakubo: this could be on line 6, yes?

front/main/app/controllers/category.js
rogatty: `...arguments` seems to be the current convention:
* it's in [the Ember guides](https://guides.emberjs.com/v2.11.0/object-model/classes-and-instances/#toc_overriding-parent-class-methods)
* we have 59 occurences of it in our code

hakubo: kk!

front/main/app/styles/component/_category-members-grouped.scss
rogatty: Moved, despite Github not acknowledging it.

front/main/app/components/category-members-grouped.js
```diff
@@ -0,0 +1,61 @@
...
+					.then(() => {
+						const navHeight = $('.site-head').outerHeight() + $('.site-head-fandom-bar').outerHeight(),
+							scrollTop = this.$().offset().top - navHeight;
+
+						run.scheduleOnce('afterRender', this, () => {
+							$('html, body').animate({scrollTop});
+						});
+					})
+					.catch(() => {
+						this.addAlert({
```
hakubo: addAlert, type: alert

ok.... :)

front/main/app/controllers/category.js
```diff
@@ -20,33 +20,44 @@ export default Controller.extend(
...
-			/**
-			 * @returns {void}
-			 */
-			articleRendered() {
-				this.get('article').send('articleRendered', ...arguments);
+							this.get('target').send('updateDynamicHeadTags');
+						});
+				}
+
+				return RSVP.Promise.reject();
```
hakubo: how about adding a reason of rejection?

front/main/app/models/wiki/category.js
```diff
@@ -31,6 +37,36 @@ const {get} = Ember,
...
+					controller: 'MercuryApi',
+					method: 'getCategoryMembers',
+					title: this.get('title'),
+					categoryMembersPage: page,
+					format: 'json'
+				}
+			});
+
+			return request(url)
+				.then((response) => {
```
hakubo: if you don't check if data is there you could also do
.then({data})

front/main/app/models/wiki/category.js
```diff
@@ -50,7 +86,14 @@ CategoryModel.reopenClass({
 			// Category Basic Data
 			// This data should always be set - no matter if category has an article or not
 			pageProperties = {
-				sections: get(data, 'nsSpecificContent.members.sections')
+				categoryMembersGrouped: get(data, 'nsSpecificContent.membersGrouped'),
```
hakubo: how about adding a var nsSpecificContent
and then just
nsSpecificContent.membersGrouped ...

front/main/app/routes/wiki-page.js
```diff
@@ -223,6 +237,14 @@ export default Ember.Route.extend(
 				}
 
 				return true;
+			},
+
+			/**
+			 * When we load another page for category members, we don't reload the route's model
+			 * Because of that, we need to trigger the head tags update manually
+			 */
+			updateDynamicHeadTags() {
```
hakubo: actually we don't need to do that right?
what is a benefit of updating head tags?

rogatty: We're not sure what the Google bot is doing so we keep it on the safe side.

front/main/app/styles/component/_category-members-grouped.scss
```diff
@@ -0,0 +1,59 @@
...
+	&__subcategory {
+		$icon-size: 18px;
+
+		&::before {
+			background-image: $wds-icons-pages-small-svg;
+			content: '';
+			left: -4px;
+			height: $icon-size;
+			position: absolute;
+			width: $icon-size;
```
kvas-damian: square mixin? ;)

PR #3052, opened 2017-01-26
SOC-3868 I've got over 999 posts but can't see how many exactly

slayful:
* http://wikia-inc.atlassian.net/browse/SOC-3868

### DIFF
--- a/front/main/app/components/discussion-moderator-tools.js
+++ b/front/main/app/components/discussion-moderator-tools.js
@@ -5,6 +5,8 @@ import ResponsiveMixin from '../mixins/responsive';
 export default Ember.Component.extend(
 	ResponsiveMixin,
 	{
+		currentUser: Ember.inject.service(),
+		canUseModeratorTools: Ember.computed.bool('currentUser.permissions.discussions.canUseModeratorTools'),
 		classNames: ['discussion-moderator-tools'],
 
 		actions: {
--- a/front/main/app/mixins/discussion-moderation-model.js
+++ b/front/main/app/mixins/discussion-moderation-model.js
@@ -32,12 +32,15 @@ export default Mixin.create({
 	deleteAllPosts(posts) {
 		return request(M.getDiscussionServiceUrl(`/${this.wikiId}/users/${posts.get('0.createdBy.id')}/posts/delete`), {
 			method: 'PUT'
-		}).then(() => {
+		}).then((response) => {
 			posts.forEach((post) => {
 				post.setProperties({
 					isDeleted: true,
-					isReported: false
+					isReported: false,
 				});
+				if (post.lastDeletedBy == null) {
+					post.set('lastDeletedBy', response.deletedBy);
+				}
 			});
 		}).catch(() => {
 			this.setFailedState('editor.post-error-general-error');
@@ -53,7 +56,10 @@ export default Mixin.create({
 		return request(M.getDiscussionServiceUrl(`/${this.wikiId}/threads/${post.get('threadId')}/undelete`), {
 			method: 'PUT'
 		}).then(() => {
-			post.set('isDeleted', false);
+			post.setProperties({
+				isDeleted: false,
+				lastDeletedBy: null
+			});
 		}).catch(() => {
 			this.setFailedState('editor.post-error-general-error');
 		});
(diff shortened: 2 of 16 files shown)


front/main/package.json
```diff
@@ -66,6 +66,7 @@
     "ember-sinon": "0.3.0",
     "ember-sortable": "1.8.0",
     "ember-wormhole": "0.4.0",
+    "ember-tether": "0.3.1",
```
rogatty: Doesn't ember-wormhole do the same thing?

slayful: So we already do have ember-tether as a dependency to something else, just requiring it explicitly.

AFAIK ember-wormhole does not have the capability to nicely calculate the element's position and attaching to another element, which is what I aim to do.

I'd rather use this than re-implement it myself, as many different things may affect such positioning (width and height of both the element and the element to which we're attaching, viewport resizing and so on).

rogatty: Makes sense :thumbsup:

front/main/app/components/discussion-user-activity-value.js
```diff
@@ -5,9 +5,7 @@ export default Ember.Component.extend({
 	classNames: ['user-activity-value'],
 
 	displayedValue: Ember.computed('value', function () {
-		const value = this.get('value');
-
-		return value <= 999 ? value.toString() : '999+';
+		return this.get('value');
```
bkoval: Seems like we don't need this computed at all, since it's the same as 'value', right?

front/main/app/components/discussion-user-activity-value.js
```diff
@@ -19,6 +17,27 @@ export default Ember.Component.extend({
...
 		return classNames;
-	})
+	}),
+
+	// showing and hiding the tooltip
+
+	/**
+	 * @private
+	 */
+	hoovering: false,
```
bkoval: *hovering

front/main/app/components/discussion-user-activity-value.js
```diff
@@ -19,6 +17,27 @@ export default Ember.Component.extend({
...
+
+	showTooltip: Ember.computed('value', 'hoovering', function () {
+		return this.get('hoovering') && this.get('value') >= 1000;
+	}),
+
+	mouseEnter() {
+		this.set('hoovering', true);
```
bkoval: Can't you just call showTooltip here (assuming it's abstracted into a method)? It would allow you to get rid of "hovering" whatsoever.

front/main/app/components/discussion-user-activity-value.js
```diff
@@ -19,6 +17,27 @@ export default Ember.Component.extend({
...
+		return true;
```
bkoval: Why does it need to return true?

front/main/app/helpers/shorten-large-number.js
```diff
@@ -0,0 +1,45 @@
...
+ * shortenLargeNumber(0.12345)
+
+ * @param {Array} [number, digits]
+ * @returns {string}
+ */
+export default Ember.Helper.helper((params) => {
+	const number = params[0];
+	const digits = params[1];
+
+	const units = ['k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'];
```
bkoval: Please, use one const here and define these in a comma-delimited way.

front/main/app/helpers/shorten-large-number.js
```diff
@@ -0,0 +1,45 @@
...
+
+	for (let i = units.length - 1; i >= 0; i--) {
+		const decimal = Math.pow(1000, i + 1);
+
+		if (number <= -decimal || number >= decimal) {
+			return +(number / decimal).toFixed(digits) + units[i];
```
bkoval: Let's use Number(n) instead of +n for readability and safety - http://stackoverflow.com/questions/1133770/how-do-i-convert-a-string-into-an-integer-in-javascript

front/main/app/helpers/shorten-large-number.js
slayful: I really don't like that we're doing that, but I guess it's the style here, so I'll conform. ;(

front/main/app/components/discussion-user-activity-value.js
slayful:
```
Browser events may bubble up the DOM which potentially target parent component(s) in succession. To enable bubbling return true; from the event handler method in your component.
```
https://guides.emberjs.com/v2.1.0/components/handling-events/

front/main/app/components/discussion-user-activity-value.js
slayful: 👍  I'll give it a try

front/main/app/helpers/shorten-large-number.js
slayful: Actually we want to return a string, so I'll remove it altogether.

PR #3055, opened 2017-01-27
SOC-3867 | cleanup unused stuff and fix badge/hero images in Insights

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/SOC-3867


front/main/app/routes/discussion/moderator/user-activity/moderations.js
```diff
@@ -13,12 +13,10 @@ export default DiscussionBaseRoute.extend(
 		 * @returns {*}
 		 */
 		model(params) {
-			const discussionModel = this.modelFor('discussion'),
-				parentParams = this.paramsFor('discussion.moderator.user-activity');
+			const parentParams = this.paramsFor('discussion.moderator.user-activity');
 
 			return Ember.RSVP.hash({
 				current: DiscussionUserActivityModerationsModel.find(Mercury.wiki.id, parentParams.days),
-				index: discussionModel
```
bkoval: Don't we need to pass site attributes to this route? Why remove it for just user activity?

slayful: AFAK this isn't even available in `user-activity.hbs` because moderations, posts and reports is rendered using the `{{outlet}}`.

I'm not sure how exactly, but both `attributes` and `categories` are available in the model in `user-activity.hbs`

slayful: I believe the error pages are working because of the same thing.

PR #3057, opened 2017-01-30
SOC-3892 | Bring back the old ways of loading scripts to not break re…

bkoval:
…vving

## Links

* http://wikia-inc.atlassian.net/browse/SOC-3892
* https://wikia-inc.atlassian.net/browse/IRIS-3926

## Description

This should bring back revving assets on auth pages. We'll address fixing it properly (using partials and revving them as well) as a technical debt story.

## Reviewers

@slayful @rogatty


bkoval:
Ping @slayful @rogatty

PR #3061, opened 2017-02-02
XW-2430 | Remove Recent Wiki Activity

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2430

## Description

Remove Recent Wiki Activity

## Reviewers

Ping @Wikia/x-wing


front/main/app/components/article-diff.js
```diff
@@ -1,165 +0,0 @@
...
-	{
-		classNames: ['diff-page'],
-		currentUser: Ember.inject.service(),
-		revisionUpvotes: Ember.inject.service(),
-		currentUserUpvoteId: Ember.computed('revisionUpvotes.upvotes.@each.count', function () {
-			const upvotes = this.get('revisionUpvotes.upvotes').findBy('revisionId', this.get('model.newId'));
-
-			return upvotes.userUpvoteId || 0;
-		}),
-		userNotBlocked: Ember.computed.not('currentUser.isBlocked'),
```
rogatty: Should we remove `isBlocked` from the user model? It's used only here:
https://github.com/Wikia/mercury/commit/ea4675b9008ca6904be24b9cbd2b0a8bc5f74635

On the other side, it's not a lot of data nor logic and maybe it can be useful in the future. What do you think?

kvas-damian: I don't have strong opinion about it. 
I thought it's used to prevent article edition by blocked users, but it's not...

rogatty: Maybe it should? :slightly_smiling_face:

kvas-damian: Done: https://github.com/Wikia/mercury/pull/3061/commits/a4bd07f5b40872aa1ead77ec0c605783865e4225

PR #3091, opened 2017-02-22
XW-2823 | use json to send curated content data

idradm:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2823
* https://github.com/Wikia/app/pull/12457

## Description

Json enables bigger form data. 

## Reviewers

@Wikia/x-wing


kvas-damian:
We still have a limit on post data length. What do you think about using `multipart` ?

front/main/app/models/curated-content-editor.js
```diff
@@ -93,10 +93,12 @@ CuratedContentEditorModel.reopenClass({
 	 * @returns {object} converted object
 	 */
 	prepareDataForSave(model) {
-		return {
-			data: [].concat(model.featured, model.curated.items, model.optional),
-			community_data: model.communityData
-		};
+		let fd = new FormData();
```
hakubo: 1. I think you can use const for that, as your reference will stay the same
2. what do you think about more descriptive name for that? communityData, formData etc.

PR #3094, opened 2017-02-24
XW-2808 | provide env at service discovery query

idradm:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2808

## Description

Send environment value to service discovery.

## Reviewers

@Wikia/x-wing


kvas-damian:
Can you give us some more details? Why is it required?

idradm:
Mercury is discovering service urls through querying consul, however, both prod and staging services are at the same datacenter. Services team will add all missing environment values as tags, that way we will get only valid hosts on query with `tag` param.

PR #3096, opened 2017-03-01
XW-2828 | Don't log connection and parse errors together

hakubo:
## Links

* http://wikia-inc.atlassian.net/browse/XW-2828

### DIFF
--- a/server/app/lib/wikia-session.js
+++ b/server/app/lib/wikia-session.js
@@ -15,7 +15,7 @@ export default function scheme() {
 		 * @param {*} reply
 		 * @returns {*}
 		 */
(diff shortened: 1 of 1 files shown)


server/app/lib/wikia-session.js
```diff
@@ -15,39 +15,40 @@ export default function scheme() {
 		 * @param {*} reply
 		 * @returns {*}
 		 */
-		authenticate: (request, reply) => {
+		authenticate(request, reply) {
 			const accessToken = request.state.access_token,
 				/**
 				 * @param {*} err
```
rogatty: Please update.

PR #3121, opened 2017-03-22
SER-795 Added product_person field to i18n configs

jcellary:
## Links

* http://wikia-inc.atlassian.net/browse/SER-795

## Description

Added new product person field to i18n crowdin config files

## Reviewers

@pwojcik86

### DIFF
--- a/crowdin/main.conf
+++ b/crowdin/main.conf
@@ -1,6 +1,6 @@
 [crowdin.project]
 projectid = "mercury"
-product_person = "cristina"
+product_person = "stanley"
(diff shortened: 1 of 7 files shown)


crowdin/main.conf
```diff
@@ -1,5 +1,6 @@
 [crowdin.project]
 projectid = "mercury"
+product_person = "cristina"
```
rogatty: This should be Stanley, right?

jcellary: True!

PR #3125, opened 2017-03-24
SER-852 Change download-translations task, add crowdin tool instalation step to Dockerfile

nikodamn:
## Links

* http://wikia-inc.atlassian.net/browse/SER-852
* https://github.com/Wikia/deploytools-plugins/pull/186 - deploytools-plugins part
* https://github.com/Wikia/mobile-wiki/pull/40 - mobile-wiki part 

## Description

* Update Dockerfile to install i18n-tools (crowdin) to download translations during building Mercury application
* Update shell script for downloading translations to not break if there are no translations available
* Update shell script for downloading translations to pass also a global config file for i18n-tools

## Reviewers

@Wikia/services-team


rogatty:
Did you check how much it will slow down the deploy? I worry that installing Python might take significant time, which will impact each deploy...

nikodamn:
@rogatty I spoke with @harnash about it and it deployment will be longer only in the first run (it already happened), because of Docker's caching of layers, where downloading python is one, separated layer

PR #3145, opened 2017-04-05
IRIS-4094 | When user navigates to an invalid Discussion URL he's red…

slayful:
…irected to /d

## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4094


kvas-damian:
How user gets into to the invalid discussions url?
If it's not an ember app maybe you should consider performing redirect on server side?

PR #3146, opened 2017-04-10
ADEN-5009 Center absolute IMA div

fraszczakszymon:
It turns out that we should not override IMA's inline styles because some vpaids are styled base on those defaults. Instead of overriding IMA properties everywhere we can adjust styles for UAP units only and keep defaults for all standalone players.

Connected with: https://github.com/Wikia/app/pull/12765

## Links

* http://wikia-inc.atlassian.net/browse/ADEN-5009

### DIFF
--- a/front/main/app/styles/module/_ads.scss
+++ b/front/main/app/styles/module/_ads.scss
@@ -16,10 +16,11 @@ $vpaid-z-index: 10;
 
 @mixin absolute-fill() {
 	height: 100%;
-	left: 0;
+	left: 50%;
 	position: absolute;
 	width: 100%;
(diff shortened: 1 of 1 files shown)


front/main/app/styles/module/_ads.scss
```diff
@@ -74,6 +73,7 @@ $vpaid-z-index: 10;
 		}
 	}
 
+	// TODO: remove below styles
```
dianafa: when those styles should be removed?

fraszczakszymon: Next time when we touch this file ;) (after release)

PR #3149, opened 2017-04-11
IRIS-3993 | Redirect back to piggyback after signing in

slayful:
* http://wikia-inc.atlassian.net/browse/IRIS-3993

### DIFF
--- a/server/app/lib/auth-utils.js
+++ b/server/app/lib/auth-utils.js
@@ -6,7 +6,7 @@ import url from 'url';
 import querystring from 'querystring';
 import settings from '../../config/settings';
 import authLocaleSettings from '../../config/authLocaleSettings.js';
-import {setUrlPathname} from './url-utils'
+import {setUrlPathname} from './url-utils';
 
 /**
  * @param {Hapi.Request} request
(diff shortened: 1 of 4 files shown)


server/app/lib/auth-utils.js
```diff
@@ -62,10 +63,7 @@ export function getCacheBusterUrl(redirect) {
  * @returns {string}
  */
 export function getRedirectUrlWithQueryString(route, request) {
-	const redirectUrl = request.url;
-
-	redirectUrl.pathname = route;
-	return redirectUrl.format();
```
slayful: This implementation was actually pretty error prone because it modified the request's URL object. If some other parts of the code used that, we'd have trouble.

server/app/lib/url-utils.js
```diff
@@ -1,6 +1,30 @@
...
+/**
+ * @param {string} url
+ * @param {object} queryParams
+ * @returns {string} url
+ */
+export function setUrlQuery(url, queryParams) {
+	const urlObject = Url.parse(url);
+	// if search is not undefined then query won't have an effect
+	urlObject.search = undefined;
+	urlObject.query = queryParams;
```
latata: can't we just replace the search property?

slayful: Discussed offline - query is an object not a string so it's more elegant this way.

PR #3153, opened 2017-04-12
IRIS-4268 | Increased the line-height and align text right

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4268


front/main/app/styles/component/_wikia-user-profile.scss
```diff
@@ -66,17 +66,18 @@
 		padding: 0 0 11px;
 
 		.wds-notifications__mark-all-as-read {
-			line-height: 12px;
+			line-height: 14px;
```
kvas-damian: Why don't we use variables from DS?

front/main/app/mixins/notifications-unread-count.js
```diff
@@ -3,6 +3,12 @@ import Ember from 'ember';
...
-	unreadCount: computed.oneWay('notifications.model.unreadCount'),
-	hasUnread: computed.gt('notifications.model.unreadCount', 0),
+	unreadCount: computed('notifications.model.unreadCount', function() {
+		const count = this.get('notifications.model.unreadCount');
+		if (count > 99) {
+			return '99+';
+		} else {
+			return count;
+		}
+	}),	hasUnread: computed.gt('notifications.model.unreadCount', 0),
```
kvas-damian: missing `nl`

front/main/app/styles/component/_wikia-user-profile.scss
slayful: In DS there's only `$wds-line-height: 1.5em;`, which's not that.
I could use `$wds-typescale-size-` but I'm not sure this is appropriate.

kvas-damian: In this case I think you should contact Designer and advice them to use DS defined sizes.

PR #3154, opened 2017-04-13
IRIS-4163 | Load notifications only when needed

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4163


front/main/app/models/notifications/notifications.js
```diff
@@ -18,32 +18,30 @@ const NotificationsModel = EmberObject.extend({
...
-		this.setProperties({
-			data: new A()
-		});
-
-		const notifications = apiData.notifications;
-
-		if (notifications && notifications.length) {
-			this.addNotifications(notifications);
+	loadFirstPage() {
+		if (stubbingOn) {
```
kvas-damian: Do we really need testing stubs in production code?

front/main/app/models/notifications/notifications.js
```diff
@@ -18,32 +18,30 @@ const NotificationsModel = EmberObject.extend({
...
+			return new RSVP.Promise((cb) => setTimeout(() => cb(_notifications(), 1000)));
 		}
+		return request(M.getOnSiteNotificationsServiceUrl('/notifications'))
+			.then((data) => {
+				this.addNotifications(data.notifications);
+				return this.getNext(data);
```
kvas-damian: It's not fully intuitive that `loadFirstPage()` returns link to next page.
Can we rethink API of this model?

PR #3163, opened 2017-05-19
Hover for dropdowns

hakubo:
## Links

* https://wikia-inc.atlassian.net/browse/XW-3400

## Description

Install new DS and don't render suggestions container when there are no suggestions

## Reviewers


rybmat:
will be done together with https://wikia-inc.atlassian.net/browse/XW-3296

PR #3165, opened 2017-05-24
IRIS-1821 - toggle deleted post and reply visibility

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-1821


front/main/app/services/discussion-toggle-deleted.js
```diff
@@ -0,0 +1,93 @@
...
+
+		request(M.getUserPreferencesServiceURL(`/${userId}/global/hideDeleted`), {
+			method: 'GET'
+		}).then((preference => {
+			this.setHideDeleted(preference.value === 'true');
+			this.applyHideDeleted(false);
+		}).bind(this), (reason => {
+			this.setHideDeleted(false);
+			this.applyHideDeleted(false);
+		}).bind(this));
```
katanka: Q: Whats the correct way to use bind and not upset the linter?

front/main/app/services/discussion-toggle-deleted.js
```diff
@@ -0,0 +1,93 @@
+import request from 'ember-ajax/request';
+
+export default Ember.Service.extend({
+	willHideDeleted: false,
+	hideDeleted: false,
```
slayful: What's the difference between `willHideDeleted` and `hideDeleted`?

Are you doing something like a two-stage update?

front/main/app/services/discussion-toggle-deleted.js
```diff
@@ -0,0 +1,93 @@
+import request from 'ember-ajax/request';
```
slayful: Perhaps `discussion-show-deleted` or `discussion-filter-deleted` would be a better name?

front/main/app/components/discussion-post-card-base.js
```diff
@@ -19,6 +21,10 @@ export default Component.extend(
 		shouldActivateLinks: Ember.computed.alias('isDetailsView'),
 		shouldTruncateContent: Ember.computed.not('isDetailsView'),
 
+		isHidden: computed('isDeleted', 'discussionToggleDeleted.hideDeleted', function () {
+			return this.get('isDeleted') && this.get('discussionToggleDeleted').get('hideDeleted');
+		}),
```
slayful: I thought it would be a query param to Discussions rather than filtering in FE? I'm not sure how will it affect FE logic (like paging) to be honest.

front/main/app/services/discussion-toggle-deleted.js
katanka: Yeah, because it has to behave differently in the mobile and desktop views. In the desktop view, the posts should be hidden/shown as soon as the checkbox is clicked. But in mobile view, the posts shouldn't be hidden/shown until the user clicks the "apply" button.

So `hideDeleted` is whether we are currently hiding deleted posts, and `willHideDeleted` is the future state of `hideDeleted` after `applyHideDeleted()` is called.

front/main/app/components/discussion-post-card-base.js
katanka: I talked with James last week and we decided it would be simpler as a FE implementation, rather than adding another layer of complexity to the discussions service. I've tested it and paging seems to work fine, there are just fewer posts sometimes (if a page contains deleted posts).

front/main/app/services/discussion-toggle-deleted.js
```diff
@@ -0,0 +1,93 @@
...
+
+export default Ember.Service.extend({
+	willHideDeleted: false,
+	hideDeleted: false,
+	currentUser: Ember.inject.service(),
+
+	setup: Ember.on('init', function () {
+		this.retrieveStoredPreference();
```
katanka: Yes, I confirmed with Cristina.

The request is asynchronous with the request to the discussions service and doesn't seem to affect page load times, since it's a much faster request.

front/main/app/components/discussion-filters.js
```diff
@@ -12,9 +12,11 @@ export default Ember.Component.extend(
 		classNames: ['discussion-filters'],
 		currentUser: Ember.inject.service(),
 		discussionSort: Ember.inject.service(),
+		discussionHideDeleted: Ember.inject.service(),
```
bkoval: Alphabetical order! 🚓  👮 👮‍♀️

PR #3172, opened 2017-06-08
IRIS-4331 - use rawContent for replies when in user post list

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4331

## Description

When viewing all posts by a user, display `rawContent` rather than `parsedContent`.

## Reviewers


kvas-damian:
I'm not sure if it's the right direction:
1. most likely it's not vulnerable to XSS
2. this change is affecting all content - original and migrated
3. it's dropping existing features as autolinking and newlines for existing content.

katanka:
If this is an XSS vulnerability then we're already vulnerable, `rawContent` is used on the main page list. I believe all HTML is escaped already.

https://github.com/Wikia/mercury/blob/dev/front/main/app/templates/components/discussion-post-card-detail.hbs#L82


@garthwebb @jsutterfield any thoughts?

garthwebb:
@kvas-damian , @katanka , a few things:
For content created in discussions:
* We store content as-is in rawContent
* Rendered content is always null

For content migrated from Forums
* We store the MW-Parser rendered HTML in renderedContent
* We strip tags from renderedContent and store it in rawContent

Given this, we should **always escape rawContent when we use it**.  Currently the post list page escapes rawContent, though I'm not sure where that happens.  Its not escaped in the DB and its not escaped by the backend in the API call, but it is definitely escaped in the browser.

jsutterfield:
I think `rawContent` is being escaped in the file @slayful linked to. Specifically [here](https://github.com/Wikia/mercury/blob/89b76455e0f9364e2c21b321e75620951e8398b2/front/main/app/mixins/discussion-parsed-content.js#L23-L25). As @garthwebb mentioned, `renderedContent` doesn't need to be escaped. I was trying to figure out how that escaping works, and it looks like there's a fair amount of Ember magic but my best guess is that it's looks this:

That file (discussion-parsed-content.js) is a mixin which is extended by discussion-post-card-base. discussion-post-card-base is then extended by 3 components: discussion-post-card-detail, discussion-post-card-reply, and discussion-editor-overlay-message. Those components are then used in the postList, postDetail, and allPostByUser page. So when handlebars is using that `{{{rawContent}}}` that @katanka linked to, it's getting dispatched to [here](https://github.com/Wikia/mercury/blob/89b76455e0f9364e2c21b321e75620951e8398b2/front/main/app/mixins/discussion-parsed-content.js#L11) which is then escaping then content.

In addition to escaping the content, it also handles the linking which happens [here](https://github.com/Wikia/mercury/blob/89b76455e0f9364e2c21b321e75620951e8398b2/front/main/app/mixins/discussion-parsed-content.js#L33).

So AFAICT, Andrew's change does end up using discussion-parsed-content, which is ensuring that the content is escaped and that it continues to render links properly. I pushed these changes to my devbox, and when I add breakpoints inside of discussion-parsed-content the escaping and linking methods are getting called.

In terms of:
> this change is affecting all content - original and migrated

Personally I think that's fine. All original content is `rawContent`, and all migrated content has a `rawContent`. When using `renderedContent`, if it doesn't exist, then it falls back to `rawContent`. So really this change only affects our migrated content.

kvas-damian:
@katanka @jsutterfield you're right. I missed the part where we escape `rawContent` (BTW. should the name still be `rawContent` after we made transformations like escaping?). In the past it wasn't there so I was worried about it.

PR #3173, opened 2017-06-09
IRIS-3887 - add undelete button to deleted post top note

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-3887

## Description

Adds an undelete button to the top note of deleted posts and replies.

## Reviewers

### DIFF
--- a/front/main/app/components/discussion-post-card-top-note.js
+++ b/front/main/app/components/discussion-post-card-top-note.js
@@ -11,7 +11,8 @@ export default Component.extend({
 	canModerate: computed.readOnly('post.userData.permissions.canModerate'),
 	isLocked: computed.readOnly('post.isLocked'),
 	showButtons: computed('isReported', 'post.isDeleted', 'canModerate', function () {
-		return (this.get('isReported') || this.get('post.isDeleted')) && this.get('canModerate');
+		return (this.get('isReported') || this.get('post.isDeleted'))
+				&& this.get('canModerate');
 	}),
 	modalDialog: inject.service(),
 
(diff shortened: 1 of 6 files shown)


front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -10,7 +10,9 @@ export default Component.extend({
 	canDelete: computed.readOnly('post.userData.permissions.canDelete'),
 	canModerate: computed.readOnly('post.userData.permissions.canModerate'),
 	isLocked: computed.readOnly('post.isLocked'),
-	showButtons: computed.and('isReported', 'canModerate'),
+	showButtons: computed('isReported', 'post.isDeleted', 'canModerate', function () {
+		return (this.get('isReported') || this.get('post.isDeleted')) && this.get('canModerate');
```
slayful: Could you break the line so that the condition is more clear?

front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -187,6 +189,15 @@ export default Component.extend({
 		},
 
 		/**
+		 * Unelete item
```
slayful: Typo

front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -187,6 +189,15 @@ export default Component.extend({
...
+		 * @param {Object} item - post or reply
+		 * @returns {void}
+		 */
+		undelete(item) {
+			this.get('undelete')(item);
```
slayful: There's no need to show confirmation dialog, correct?

katanka: I don't believe so, since there isn't if you undelete from the dropdown. This is just a shortcut. Also, the reporting process can only be done once per post, whereas deleting/undeleting can be done repeatedly.

PR #3177, opened 2017-06-16
IRIS-3817 - generate thread opengraph image from content

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-3817

## Description
Use the image (either in the OG card or uploaded to the post) in the first post
If there is no image in the first post, use the largest image in the discussion replies
If there are no images in the replies, use the community avatar
If there is no community avatar, fall back to the fandom logo

## Reviewers


server/app/lib/open-graph.js
```diff
@@ -60,11 +55,33 @@ export function getPromiseForDiscussionData(request, wikiVars) {
 							i18n.t('main.share-default-title', {siteName: wikiVars.siteName, ns: 'discussion'});
 						// Keep description to 175 characters or less
 						openGraphData.description = content.substr(0, 175);
-						if (wikiVars.image) {
+
+						const postsWithOpenGraph = response.payload._embedded['doc:posts'].filter(post => post._embedded.openGraph);
+						if (response.payload._embedded.openGraph) {
```
jsutterfield: This method is stating to get a little hairy. Is there a way we could pull all these ifs into a new method?

server/app/lib/open-graph.js
```diff
@@ -48,23 +80,29 @@ export function getPromiseForDiscussionData(request, wikiVars) {
 			return new Promise((resolve, reject) => {
 				// Fetch discussion post data from the API to complete the OG data
 				MW.fetch(apiUrl)
+					.then((response) => injectDesignSystemData({
```
rogatty: `injectDesignSystemData` is already used in `show-application.js` which calls `OpenGraph.getAttributes()`. We would make call to DS API twice if it stays here. Instead you can refactor https://github.com/Wikia/mercury/blob/42cf1e6b726ff8ed4e19c3fea4c8c17cb8cce922/server/app/facets/show-application.js#L78-L106 into something like this (change the order, basically):

```
wikiVariables
	.then((wikiVariables) => {
		Utils.redirectToCanonicalHostIfNeeded(settings, request, reply, wikiVariables);

		context.wikiVariables = wikiVariables;
		context.isRtl = isRtl(wikiVariables);
	})
	.then((templateData) => injectDesignSystemData({
		data: templateData,
		request,
		showFooter: showGlobalFooter
	}))
	.then((templateData) => {
		return OpenGraph.getAttributes(request, templateData.wikiVariables).then((openGraphData) => {
			// Add OpenGraph attributes to template data
			templateData.openGraph = openGraphData;
			return templateData;
		});
	})
	.then((templateData) => injectSassVariables(templateData))
```

server/app/lib/open-graph.js
```diff
@@ -15,6 +16,42 @@ import {getStaticAssetPath} from './utils';
...
+	} else if (data.communityHeader !== null) {
+		// Use the wordmark image
+		openGraphData.image = data.communityHeader.wordmark['image-data'].url;
+	} else {
+		// Use Fandom logo as default image
+		openGraphData.image = `http:${getStaticAssetPath(settings, request)}` +
+				'common/images/og-fandom-logo.jpg';
+		openGraphData.imageWidth = 1200;
+		openGraphData.imageHeight = 1200;
+	}
```
slayful: Can we extract these into methods?
Something in lines of:
```
const image = fetchImage()
openGraphData.image = image.url;
openGraphData.imageWidth = image.width;
openGraphData.imageHeight = image.height;
```
and
```
function fetchImage(data, request) {
    if (response.payload._embedded.openGraph) {
        return getMainPostOpenGraph(response.payload._embedded.openGraph[0]);
    }
   const postsWithOpenGraph = findPostsWithOpenGraph(response.payload._embedded['doc:posts']);
   if (postWithOpenGraph.length > 0) {
      return findLargestOpenGraphImage(postsWithOpenGraph);
  }

 [....]
}
```

katanka:
@slayful I changed the function to be a bit cleaner by returning an object of the OG info, but I don't think that all the contents of the `if/else`s need to be factored out, since those functions wouldn't be reused and I think it's readable enough in its current state.

server/app/lib/open-graph.js
```diff
@@ -2,6 +2,7 @@ import Promise from 'bluebird';
 import settings from '../../config/settings';
 import * as MW from './mediawiki';
 import {getStaticAssetPath} from './utils';
+import injectDesignSystemData from '../lib/inject-design-system-data';
```
slayful: This line seems to be no longer used.

server/app/lib/open-graph.js
```diff
@@ -15,11 +16,55 @@ import {getStaticAssetPath} from './utils';
...
+			: [];
+	if (response.payload._embedded.openGraph) {
+		// Use OpenGraph in main post
+		const openGraph = response.payload._embedded.openGraph[0];
+		return {
+			image: openGraph.imageUrl,
+			imageHeight: openGraph.imageHeight,
+			imageWidth: openGraph.imageWidth
+		};
+	} else if (postsWithOpenGraph.length > 0) {
```
slayful: Since you're returning in each if-else branch, you don't have to use else and you could move
```
const postsWithOpenGraph = response.payload._embedded['doc:posts']
 +			? response.payload._embedded['doc:posts'].filter(post => post._embedded.openGraph)
 +			: [];
```
just before this check `if (postsWithOpenGraph.length > 0)`.

slayful:
@katanka I've left two comments, besides that LGTM.
I agree it's pretty clear and could be left as is.

I do think that it would benefit from separate functions for two reasons:
- the function takes up the whole screen, and it would be easier to follow what's happening with a few if-condition then-return
```
if (condition) {
    return wellNamedGenerator();
}
```
- you're already using comments to clarify what's happening, so instead you could use named functions

Doing that would clearly separate two levels of abstraction
- choosing which open graph generator is the best one
- generating an open graph

katanka:
@slayful sounds good, made those changes

PR #3186, opened 2017-07-06
IRIS-4649 - use better foreach on queryselectorall

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4649

## Description

Call `foreach` in a backwards-compatible manner on `querySelectorAll` statements.

## Reviewers


front/auth/app/common/password-form.js
```diff
@@ -194,7 +194,7 @@ export default class PasswordForm {
 	 * @returns {void}
 	 */
 	clearErrors() {
-		this.form.querySelectorAll('small.error').forEach(element => {
+		Array.prototype.forEach.call(this.form.querySelectorAll('small.error'), element => {
```
slayful: This should be also possible I think:
https://github.com/Wikia/upstream/blob/master/www/wp-content/themes/upstream/modules/design-system/scripts/wds-global-navigation.js#L69

slayful: Perhaps it's a bit cleaner.

jsutterfield: +1 that does look a bit cleaner.

PR #3191, opened 2017-07-13
XW-3747 | Handle Post List type

JamesHawking:
## Links

* https://wikia-inc.atlassian.net/browse/XW-3747

## Reviewers

@Wikia/x-wing

### DIFF
--- a/front/common/public/locales/en/discussion.json
+++ b/front/common/public/locales/en/discussion.json
@@ -127,7 +127,7 @@
     "rename-category-auth-error": "You are logged out. Please sign in.",
     "rename-category-oermissions-error": "Category cannot be updated.",
     "rename-category-general-error": "Something went wrong, please try again.",
-    "guidelines-link-title": "Discussions Guidelines",
+    "guidelines-link-title": "Guidelines",
     "guidelines-header-title": "Discussions Guidelines",
     "welcome-message-header": "Awesome to have you here!",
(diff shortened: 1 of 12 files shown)


front/common/public/locales/en/discussion.json
```diff
@@ -127,7 +127,7 @@
     "rename-category-auth-error": "You are logged out. Please sign in.",
     "rename-category-oermissions-error": "Category cannot be updated.",
     "rename-category-general-error": "Something went wrong, please try again.",
-    "guidelines-link-title": "Discussions Guidelines",
+    "guidelines-link-title": "Guidelines",
     "guidelines-header-title": "Discussions Guidelines",
```
rogatty: Is this used anywhere?

front/main/app/components/left-rail-user-details.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	userProfile: Ember.computed('users', function () {
```
rogatty: I think this would work too:
```userProfile: computed.reads('users.0')```

File: front/main/app/styles/component/discussion/_left-rail-user-details.scss [hunk 3]
```diff
@@ -0,0 +1,26 @@
+.left-rail {
```
rogatty: Let's scope components and classes to `discussion-*`

PR #3210, opened 2017-07-28
IRIS-4650 - add custom error for phalanx blocks

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4650

## Description

This adds a custom error dialog to inform users when their post has been blocked by Phalanx.

## Reviewers
@jsutterfield


front/main/app/mixins/discussion-contribution-controller.js
```diff
@@ -197,13 +197,47 @@ export default Ember.Mixin.create({
...
+			part2 = i18n.t('main.phalanx-block-message-part2-singular', {ns: 'discussion'});
+		}
+
+		brokenRuleIds = brokenRuleIds.map(id => `#${id}`);
+
+		let formattedBrokenRuleIds = '';
+		if (brokenRuleIds.length === 1) {
+			formattedBrokenRuleIds = brokenRuleIds[0];
+		} else if (brokenRuleIds.length === 2) {
+			formattedBrokenRuleIds = brokenRuleIds.join(' and ');
```
jsutterfield: `and` won't be translated here. Maybe we could get by with just joining the `brokenRuleIds` with a comma followed by a space? I think if we took that approach we should be able to avoid this whole if/else if/else block, and it would save us from the i18n troubles.

PR #3212, opened 2017-08-04
XW-3813 | Localize number of posts in Community Header

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-3813

## Description

- Fix paths to numeral assets
- Use `{{localize-number}}` and remove `{{numeral}}`
- Remove unused numeral from the server
- Update numeral to 2.0.6
- Patch thousand delimiters for consistency with MediaWiki

## Reviewers

@Wikia/x-wing


front/main/app/services/numeral-locale.js
```diff
@@ -37,7 +37,23 @@ export default Ember.Service.extend({
 	 * @return {void}
 	 */
 	setLocale(locale = 'en') {
-		numeral.language(locale);
+		numeral.locale(locale);
+
+		// Patch locales for consistency with MediaWiki
+		switch (locale) {
+			case 'de':
+				numeral.locales.de.delimiters.thousands = '.';
```
latata: `numeral.locales[locale].delimiters.thousands` and combine `it` and `pt-br` as they are the same?

rogatty: You're right, fixed.

front/main/app/helpers/localize-number.js
```diff
@@ -2,32 +2,33 @@ import Ember from 'ember';
...
+ * If you pass format as the second parameter, it will override the default one
  * @param {Array} params
  * @returns {string}
  */
 export default Ember.Helper.extend({
 	numeralLocale: Ember.inject.service(),
 	onLocaleChange: Ember.observer('numeralLocale.isLoaded', function () {
-		this.recompute();
+		// It won't recompute if we don't wait
+		Ember.run.schedule('afterRender', () => {
```
hakubo: does that mean it'll re-render views after loading locale?
but we do load it synchronously right?

rogatty: We load them asynchronously and yes, this observer is to re-render the numbers after locale is loaded. This fix is only for the English language where `compute()` wasn't called.

PR #3216, opened 2017-08-07
User activity glossary more conspicuous

shaedrich:
## Links
_no links provided_

## Description
Foregrounds user activity glossary so the user can see what he has chosen (title (which is the tab) and description). This is because the user reads from top to bottom and the middle rail is focussed.

## Reviewers
_noone assigned yet_


rogatty:
Discussions Front End is no longer a part of Mercury.

PR #3218, opened 2017-08-07
Output category description if it exists

shaedrich:
## Links
_no links provided_

## Description
This allows to display the category description if only one category is choosen.

## Reviewers
_noone assigned yet_


rogatty:
Discussions Front End is no longer a part of Mercury.

PR #3219, opened 2017-08-07
IRIS-4786 - show moderation tools header on main page

katanka:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4786

## Description

This standardizes the right sidebar on the main page, reported page, and post page.


shaedrich:
When approved and merged close #3220 because it modified the same case the other way round.

shaedrich:
I wonder why this is still unchanged when I look at discussions.

PR #3221, opened 2017-08-09
IRIS-4788 | Make sure blockquotes and headings look nice on dark theme

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4788


front/main/app/styles/runtime.scss
```diff
@@ -20,3 +20,6 @@
 @import 'auth/app/styles/base/input';
 @import 'auth/app/styles/module/forms';
 @import 'auth/app/styles/module/form-icons';
+
+// discussions specifics
+@import '_discussion-quotes';
```
rogatty: We do `@import 'component/index';` above so this isn't needed.

front/main/app/styles/theme/_dark.scss
```diff
@@ -172,6 +171,10 @@ body.dark-theme {
 						color: $dark-discussion-post-title-color;
 					}
 				}
+
+				h1,h2,h3,h4,h5,h6 {
```
rogatty: Where do these tags come from? Can we use put classes in them and use here?

front/main/app/styles/_discussion-quotes.scss
```diff
@@ -12,3 +12,11 @@
 	-webkit-margin-start: 0;
 	-webkit-margin-end: 0;
 }
+
+.discussions blockquote {
+	border-left: 4px solid $color-links;
+	color: #999;
+	p {
+		color: #999;
```
rogatty: I think 3 times is enough for a variable.

front/main/app/styles/theme/_dark.scss
slayful: We cannot as these come from migrated forums.

front/main/app/styles/component/discussion/_discussion-quotes.scss
```diff
@@ -12,3 +12,11 @@
 	-webkit-margin-start: 0;
 	-webkit-margin-end: 0;
 }
+
+.discussions blockquote {
+	border-left: 4px solid $color-links;
+	color: #999;
+	p {
+		color: #999;
```
rogatty: I think 3 times is enough for a variable.

front/main/app/styles/theme/_dark.scss
rogatty: How about an explanatory comment, then?

PR #3222, opened 2017-08-10
IRIS-4771 - fix cancel button on report action dialog

katanka:
## Links

https://wikia-inc.atlassian.net/browse/IRIS-4771

## Description

This updates the confirmation callback to accept the boolean confirmed parameter it is sent.

## Reviewers


front/main/app/components/discussion-post-card-top-note.js
```diff
@@ -221,7 +221,7 @@ export default Component.extend({
 				header,
 				name: 'modal-dialog-approve',
 				confirmButtonText: i18n.t('main.modal-dialog-approve', {ns: 'discussion'}),
-				confirmCallback: (() => this.get('approve')(item)),
+				confirmCallback: ((confirmed) => confirmed ? this.get('approve')(item) : Ember.K()),
```
slayful: I recall Ember.K being deprecated, though I'm not sure what should be done here instead.
https://github.com/emberjs/ember.js/pull/14751

slayful: I believe you might get away with simply returning `this`, since it's an inline function.

rogatty: It can be `function () {}` as it's used in this way:
```
modalDialogService.confirmCallback(true);
```

PR #3234, opened 2017-09-08
IRIS-4830 | Add a variable which will override color links from theme designer in Discussions

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4830

### DIFF
--- a/front/main/app/mixins/theme.js
+++ b/front/main/app/mixins/theme.js
@@ -90,7 +90,13 @@ export default Ember.Mixin.create({
 			return;
 		}
 
-		this.set('themeColors', Ember.get(Mercury, 'wiki.theme'));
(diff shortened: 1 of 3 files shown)


front/main/app/mixins/theme.js
```diff
@@ -72,6 +72,20 @@ export default Ember.Mixin.create({
 
 		return (color.getLuminance() < luminanceThreshold) ? defaultColor : fallbackColor;
 	},
+
+	/**
+	 * @private
+	 */
+	setupTheme() {
+		let theme = Ember.get(Mercury, 'wiki.theme');
+		let override = Ember.get(Mercury, 'wiki.discussionColorOverride');
```
bkoval: We usually add one blank line after var definitions.

+ These seem to not mutate after declaration, so they can be declared as consts.

PR #3237, opened 2017-09-18
IRIS-4920 | Use width and height returned by BE

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4920


front/main/app/models/discussion/domain/content-images.js
```diff
@@ -17,17 +17,16 @@ const {Object, A} = Ember,
...
 			const images = this.get('images');
 			const position = images.filterBy('visible')
 				.reduce((previous, item) => Math.max(previous, item.position), 0);
-			// TODO BE should send the height and width
 			images.pushObject(Object.create({
-				height: 200,
+				height: image.height,
 				position: position + 1,
-				url,
+				url: image.url,
```
rogatty: Is it possible that the server doesn't throw an error but one of these properties is empty?

slayful: Currently it is not - either full response or failure.

PR #3242, opened 2017-09-21
IRIS-1382 | Show placeholder and spinner during image upload

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/IRIS-1382

## Description

Show placeholder and spinner, disable submit while uploading. I also did the first step toward moving away from app-dark-theme.css which required moving a lot of SCSS variables.

## Reviewers

@slayful


front/main/app/components/discussion-editor.js
```diff
@@ -56,8 +56,10 @@ export default Ember.Component.extend(
 			}
 		}),
 
-		submitDisabled: Ember.computed('content', 'title', 'currentUser.isAuthenticated', 'showOverlayMessage',
-			'isReply', 'category', 'isGuidelinesEditor', function () {
+		submitDisabled: Ember.computed(
+			'content', 'title', 'currentUser.isAuthenticated', 'showOverlayMessage',
+			'isReply', 'category', 'isGuidelinesEditor', 'contentImages.images.@each.isUploading',
```
slayful: I'm wondering if we could add a method to contentImages which would describe if they're loading rather than exposing their internals here.

front/main/app/components/discussion-image-upload.js
```diff
@@ -63,23 +62,23 @@ export default Component.extend(
...
-				})
+			this.get('contentImages')
+				.addContentImage(imageFile, this.get('staticAssets'))
 				.catch((err) => {
-					Logger.error('Error uploading image', err);
-					if (err.status === 400) {
-						this.handle400Response(err.response);
-					} else {
-						this.showErrorMessage('image-upload.upload-failed');
+					if (!this.get('isDestroyed')) {
```
slayful: Early cut-off?
```
if (this.get('isDestroyed') {
    return;
}
```

front/main/app/components/discussion-image-upload.js
```diff
@@ -63,23 +62,23 @@ export default Component.extend(
...
 			}
 
-			this.set('isLoadingMode', true);
-
-			this.uploadImage(imageFile)
-				.then((result) => {
-					this.get('contentImages').addContentImage(result);
-				})
+			this.get('contentImages')
+				.addContentImage(imageFile, this.get('staticAssets'))
```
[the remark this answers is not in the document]
rogatty: This line won't be called when the component is destroyed. Only after this method we're running async and have to check if a user didn't close the editor before the image was uploaded.

front/main/app/components/discussion-editor.js
rogatty: Good idea, I used a computed property.

PR #3249, opened 2017-10-02
IRIS-4934 | Fix and simplify post image thumbnailing process

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/IRIS-4934
* https://stackoverflow.com/questions/44991155/prevent-upscaling-with-scrset

## Description

Use `` instead of `<picture><source></picture>`. It's simpler as the browser handles pixel ratio by itself. 

## Reviewers

@slayful

### DIFF
--- a/front/main/app/components/discussion-post-image.js
+++ b/front/main/app/components/discussion-post-image.js
@@ -42,7 +42,7 @@ export default Component.extend(
 		srcset: computed('crop', 'image', function () {
 			const image = this.get('image');
 
-			if (!image.url) {
+			if (!image || !image.url) {
 				return '';
 			}
 
--- a/front/main/app/components/discussion-post-image.js
+++ b/front/main/app/components/discussion-post-image.js
@@ -86,6 +86,11 @@ export default Component.extend(
 			return sources;
 		},
 
+		/**
+		 * Crop to 16:9 aspect ratio
+		 * @param {Object} image
+		 * @returns {Array}
+		 */
 		getCroppedSources(image) {
 			const sources = [];
 
--- a/front/main/app/components/discussion-post-image.js
+++ b/front/main/app/components/discussion-post-image.js
@@ -27,7 +27,7 @@ export default Component.extend(
 		/**
 		 * @public
 		 */
-		crop: false,
+		shouldCrop: false,
 
 		/**
 		 * @public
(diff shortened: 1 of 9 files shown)


front/main/app/components/discussion-post-image.js
```diff
@@ -1,40 +1,35 @@
...
-			mobile: [420, 767, 1063]
-		},
+		availableThumbSizes: [
+			420,
+			// Desktop < 1575px
+			520,
+			// Desktop >= 1575px
+			640,
+			767,
+			1063
```
slayful: Perhaps it makes sense to extract those as constants elsewhere?

front/main/app/components/discussion-post-image.js
```diff
@@ -44,208 +39,73 @@ export default Component.extend(
...
+		srcset: computed('crop', 'image', function () {
+			const image = this.get('image');
 
-		/**
-		 * @public
-		 *
-		 * Accepts null or undefined, fallbacks to 1.
-		 */
-		widthMultiplier: 1,
+			if (!image.url) {
```
slayful: what if image is `null`?

front/main/app/components/discussion-post-image.js
```diff
@@ -44,208 +39,73 @@ export default Component.extend(
...
+				return '';
+			}
 
-		displayedSources: computed('crop', function () {
-			return this.get('crop') ? this.get('croppedSources') : this.get('sources');
-		}),
+			const sources = this.get('crop') ? this.getCroppedSources(image) : this.getUncroppedSources(image);
```
slayful: Perhaps the field's name `shouldCrop` rather than `crop` would be clearer?

front/main/tests/unit/components/discussion-post-image-test.js
```diff
@@ -0,0 +1,77 @@
+import {moduleForComponent, test} from 'ember-qunit';
+
+moduleForComponent('discussion-post-image', 'Unit | Component | discussion post image', {
+	unit: true
+});
+
+test('uncropped sources', function (assert) {
```
slayful: Maybe a more descriptive test names could be used.

front/main/app/components/discussion-post-image.js
rogatty: I'd wait until they're used somewhere else. They are based on CSS layout and no other JS uses them.

front/main/app/components/discussion-post-image.js
rogatty: Good catch, fixed.

PR #3250, opened 2017-10-03
IRIS-1635 | Show image upload errors inside editor

rogatty:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-1635

## Description

Use editor errors instead of global `alert-notifications`.
Remove input reset from `discussion-image-upload` as the component is destroyed when the upload starts.

## Reviewers

@slayful 
/cc @bkoval @kvas-damian

### DIFF
--- a/front/main/app/components/discussion-image-upload.js
+++ b/front/main/app/components/discussion-image-upload.js
@@ -10,7 +10,6 @@ export default Component.extend(
 		staticAssets: inject.service(),
 
 		isDragActive: false,
-		resetFileInput: false,
 
 		allowedFileTypes: [
 			'image/jpeg',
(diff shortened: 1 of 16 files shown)


front/main/app/components/discussion-image-upload.js
```diff
@@ -65,30 +62,18 @@ export default Component.extend(
...
 					if (err.status === 400) {
 						this.handle400Response(err.response);
 					} else {
 						this.showErrorMessage('image-upload.upload-failed');
 					}
-				})
-				.then(() => {
-					if (!this.get('isDestroyed')) {
-						this.set('resetFileInput', true);
-					}
```
slayful: Are we sure we don't need that?

rogatty: Yup, quoting the PR description:
>Remove input reset from discussion-image-upload as the component is destroyed when the upload starts.

`isDestroyed` was _always_ true after I moved the loading spinner.

rogatty:
I agree about the flow being hard to reason about. Not sure what would be the best way to improve it without major refactor.

I smoke tested all variants that came to my mind and noted them in the testing notes. Had to fix a few missed places.

PR #3251, opened 2017-10-04
IRIS-4921 | Handle content images in ogImage generation for Discussions

slayful:
## Links

* https://wikia-inc.atlassian.net/browse/IRIS-4921

### DIFF
--- a/server/app/lib/open-graph.js
+++ b/server/app/lib/open-graph.js
@@ -4,7 +4,6 @@ import * as MW from './mediawiki';
 import {getStaticAssetPath} from './utils';
 
 /**
- * @typedef {Object} OpenGraphAttributes
  * @property {string} [description]
  * @property {string} [image]
  * @property {number} [imageHeight]
@@ -13,12 +12,7 @@ import {getStaticAssetPath} from './utils';
  * @property {string} type
  * @property {string} url
  */
-
-/**
- * @param {*} response
- **/
-function getThreadOpenGraph(response) {
-	const openGraph = response.payload._embedded.openGraph[0];
+function fromOpenGraph(openGraph) {
 	return {
 		image: openGraph.imageUrl,
 		imageHeight: openGraph.imageHeight,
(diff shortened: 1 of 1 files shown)


server/app/lib/open-graph.js
```diff
@@ -13,40 +12,41 @@ import {getStaticAssetPath} from './utils';
...
  */
-
-/**
- * @param {*} response
- **/
-function getThreadOpenGraph(response) {
-	const openGraph = response.payload._embedded.openGraph[0];
+function fromOpenGraph(openGraph) {
 	return {
 		image: openGraph.imageUrl,
```
rogatty: There is no `imageUrl` in JSDoc. Please update it.

server/app/lib/open-graph.js
```diff
@@ -79,13 +79,26 @@ function generateFandomLogoOpenGraph(request) {
  * @param {*} context
  */
 function fetchOpenGraphImage(response, request, context) {
-	if (response.payload._embedded.openGraph) {
-		return getThreadOpenGraph(response);
+	const thread = response.payload._embedded;
+
+	if (thread.contentImages && thread.contentImages.length > 0) {
+		return fromContentImage(thread.contentImages[0]);
```
rogatty: I think we want the largest image, for both posts and replies.

rogatty: Oh, right, there is only one image in the post 😛

server/app/lib/open-graph.js
```diff
@@ -79,13 +79,26 @@ function generateFandomLogoOpenGraph(request) {
  * @param {*} context
  */
 function fetchOpenGraphImage(response, request, context) {
-	if (response.payload._embedded.openGraph) {
-		return getThreadOpenGraph(response);
+	const thread = response.payload._embedded;
```
rogatty: In Ember we use post and reply. Can we stick to it? It'll be less confusing when introducing FastBoot.

server/app/lib/open-graph.js
```diff
@@ -13,40 +12,41 @@ import {getStaticAssetPath} from './utils';
...
+function getLargestImage(images) {
+	images.sort((a, b) => b.imageWidth * b.imageHeight - a.imageWidth * a.imageHeight);
+	return images[0];
+}
+
+function getPostOpenGraphImages(posts) {
+	return posts
+		.map(post => post._embedded.openGraph)
+		.filter(openGraphs => openGraphs && openGraphs.length > 0)
+		.map(openGraphs => openGraphs[0])
```
rogatty: Method is named getPostOpenGraphImage**s** and we return an array. I don't think this line fits. The same goes for getPostContentImage**s**.

slayful: I believe there's a misunderstanding here - both open graph and content images are a list per post/reply in the response. At the moment BE returns up to one element in that list.

I was considering re-using find the biggest image code for selecting the open graph and content image from the list returned by BE, but that's neither necessary (up to one element for unforseeable future) nor the logic we'd end up choosing if/when we allow more than one content image or open graph per post/reply.

So this code filters posts without or an empty a list of opengraphs and then maps a list of one element into an element. Similar thing goes for content images.

rogatty: Yup, I thought we return an array with one item but there are actually as many items as posts with images.

PR #3252, opened 2017-10-04
IRIS-4933 | Tracking for image upload

slayful:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-4933
* https://wikia-inc.atlassian.net/wiki/spaces/DIS/pages/64946499/Discussions+Web+Tracking+Events


front/main/app/components/discussion-image-upload.js
```diff
@@ -50,11 +59,13 @@ export default Component.extend(
 		handleImageSelected(imageFile) {
 			if (this.get('allowedFileTypes').indexOf(imageFile.type) === -1) {
 				this.showErrorMessage('image-upload.invalid-file-type');
```
rogatty: Because of this call, we'll run both `trackInvalidFileType` and `trackImageUploadFailure`. Is that ok?

front/main/app/utils/image-upload-tracker.js
```diff
@@ -0,0 +1,47 @@
...
+		category: 'image-upload',
+	});
+}
+
+export function trackFileDropped() {
+	track(labels.fileDropped, 'gesture');
+}
+
+export function trackImageSelected() {
+	track(labels.imageSelected, 'click');
```
rogatty: How about using https://github.com/Wikia/mercury/blob/dev/front/common/utils/track.js#L34-L89 and adding there the missing actions? At least the not-dynamic ones.

front/main/app/components/discussion-image-upload.js
slayful: I'd think so, I'll ask Korni.

PR #3253, opened 2017-10-04
IRIS-5013 | Use thumbnails for avatars

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/IRIS-5013

## Description

Use thumbnails instead of full-size images. Doesn't work for default wiki avatars like http://images.wikia.com/messaging/images/e/e5/Avatar4.jpg.

I chose the sizes based on CSS.

## Reviewers

@slayful

### DIFF
--- a/front/main/app/components/discussion-reply-editor-entry-point.js
+++ b/front/main/app/components/discussion-reply-editor-entry-point.js
@@ -1,13 +1,15 @@
 import Ember from 'ember';
 import DiscussionStickyComponentMixin from '../mixins/discussion-sticky-component';
 
-const {Component} = Ember;
+const {Component, inject} = Ember;
 
 export default Component.extend(
 	DiscussionStickyComponentMixin,
 	{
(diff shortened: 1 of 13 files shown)


front/main/app/components/discussion-reply-editor-entry-point.js
```diff
@@ -1,14 +1,16 @@
...
-const {Component} = Ember;
+const {Component, inject} = Ember;
 
 export default Component.extend(
 	DiscussionStickyComponentMixin,
 	{
 		classNames: ['discussion-reply-editor-entry-point'],
 		containerSelector: '.discussion-editor-entry-point-container',
 
+		currentUser: inject.service(),
```
slayful: Why's that needed now?

rogatty: It was always needed but missing. This fixes the bug where we showed default avatar instead of the user's one in the mobile reply editor entry point.

PR #3255, opened 2017-10-06
IRIS-1620 | Hide open graph when there is content image

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/IRIS-1620

## Description

A content image has priority over an open graph. Save both to service, it's handled by mobile apps as expected.

## Reviewers

@slayful

### DIFF
--- a/front/main/app/mixins/discussion-editor-opengraph.js
+++ b/front/main/app/mixins/discussion-editor-opengraph.js
@@ -2,11 +2,40 @@ import Ember from 'ember';
 import {track, trackActions} from '../utils/discussion-tracker';
 import {getLastUrlFromText} from 'common/utils/string';
 
-export default Ember.Mixin.create({
+const {
+	Mixin,
+	computed,
+	get,
+	observer,
+	run
+} = Ember;
+
+export default Mixin.create({
 	isOpenGraphLoading: false,
 	openGraph: null,
-	showsOpenGraphCard: false,
 	contentLength: 0,
+
+	showOpenGraphCard: computed('openGraph', 'contentImages.images.@each', function () {
+		const contentImages = this.get('contentImages');
+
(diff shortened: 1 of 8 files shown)


front/main/app/models/discussion/domain/content-images.js
```diff
@@ -20,7 +20,7 @@ const DiscussionContentImages = Object.extend(
 		}),
 
 		hasImages() {
-			return this.get('images.length');
+			return this.get('images.length') > 0;
```
slayful: Will this be fine if images are null?

front/main/app/mixins/discussion-editor-opengraph.js
```diff
@@ -99,39 +116,31 @@ export default Ember.Mixin.create({
 	},
 
 	setOpenGraphProperties(url) {
-		if (this.get('showsOpenGraphCard')) {
+		if (this.get('openGraph') || this.get('contentImages').hasImages()) {
```
slayful: Perhaps extract to a separate method?
`hasOpenGraphOrContentImage()`?

front/main/app/models/discussion/domain/content-images.js
rogatty: They will never be null, see https://github.com/Wikia/mercury/blob/6c9272a3c736fcaae21cb81903d5263023119681/front/main/app/models/discussion/domain/content-images.js#L9-L16

front/main/app/mixins/discussion-editor-opengraph.js
slayful: In the method above you check if contentImages exist before calling has images, shouldn't this happen here as well?

rogatty: It's not used anywhere else and is pretty straightforward.

PR #3262, opened 2017-10-20
Parse payload JSON to make sure requested-closure-date gets user id

rogatty:
## Links

* https://kibana5.wikia-inc.com/goto/b0f9cbca59264ff26706acbc4eca3fbd

## Description

We were passing a stringified object from server to client and then treat it as an object, see https://github.com/Wikia/mercury/blob/09935dee9800bc3e87364a5e60c6e7929e97ce6a/front/auth/app/common/login.js#L128
Because of that, we were sending request to `/user-preference/undefined/global/requested-closure-date`

## Reviewers

@slayful

### DIFF
--- a/server/app/facets/auth/reset-password.js
+++ b/server/app/facets/auth/reset-password.js
@@ -105,7 +105,7 @@ export function post(request, reply) {
 		updatePasswordFor(username, password, token, request)
 			.then(data => {
 				reply({
-					payload: data.payload
+					payload: JSON.parse(data.payload)
 				}).code(200);
 			}).catch(data => {
 				const errors = translateError(data, (error) => {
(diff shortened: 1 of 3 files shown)


server/app/facets/auth/reset-password.js
```diff
@@ -104,9 +104,7 @@ export function post(request, reply) {
 	} else {
 		updatePasswordFor(username, password, token, request)
 			.then(data => {
-				reply({
-					payload: data.payload
-				}).code(200);
+				reply(JSON.parse(data.payload)).code(200);
```
slayful: `HttpStatus.OK` to make it more consistent?

rogatty: Good idea, but I don't want to spend any more time on this now 😉

PR #3263, opened 2017-10-20
Use HttpStatus.OK for consistency

shaedrich:
## Links
* Requested in https://github.com/Wikia/mercury/pull/3262/files/ba76db4ffacb53bcdf10da5a72207e8dc923e969#r145929935
* Depends on #3262

## Description
Use of `HttpStatus.OK` constant instead of hard coded `200` value for consistency

## Reviewers
@slayful

### DIFF
--- a/server/app/facets/auth/reset-password.js
+++ b/server/app/facets/auth/reset-password.js
@@ -104,7 +104,7 @@ export function post(request, reply) {
 	} else {
 		updatePasswordFor(username, password, token, request)
 			.then(data => {
-				reply(JSON.parse(data.payload)).code(200);
+				reply(JSON.parse(data.payload)).code(HttpStatus.OK);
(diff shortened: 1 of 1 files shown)


server/app/facets/auth/reset-password.js
```diff
@@ -104,7 +104,7 @@ export function post(request, reply) {
 	} else {
 		updatePasswordFor(username, password, token, request)
 			.then(data => {
-				reply(JSON.parse(data.payload)).code(200);
+				reply(JSON.parse(data.payload)).code(HttpStatus.OK);
```
rogatty: `HttpStatus` is `unefined` in this context.

shaedrich: I imported HttpStatus so it should be available now.

PR #3272, opened 2017-11-27
SER-1419 use k8s domain for user-registration, remove consul

pwojcik86:
## Links

* http://wikia-inc.atlassian.net/browse/SER-1419
## Description

Use the internal k8s domain when communicating with `user-registration`. Remove Consul-based service discovery.

## Reviewers

@Wikia/iris @Wikia/services-team

### DIFF
--- a/server/config/settings.dev.js
+++ b/server/config/settings.dev.js
@@ -4,7 +4,7 @@
 
 // These variables are used in multiple places in config
 const devDomain = (process.env.WIKIA_DATACENTER === 'poz') ? 'pl' : 'us',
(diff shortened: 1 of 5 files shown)


server/config/settings.dev.js
```diff
@@ -14,6 +14,9 @@ export default {
 	devDomain,
 	authCookieDomain: `.wikia-dev.${devDomain}`,
 	servicesDomain,
+	userRegistationService: {
```
rogatty: It should be `userRegistrationService`

PR #3275, opened 2017-11-29
use production Helios by default

pwojcik86:
## Links

* https://wikia-inc.atlassian.net/browse/MAIN-12070

## Description

Production Helios should be used by default. This should fix login problems on `verify` and `preview`, which was broken after removing Helios' url from the base config.

## Reviewers

@rogatty @tortila @nikodamn


rogatty:
I would add `userRegistationService.internalUrl` to `settings.base.js` too

PR #3286, opened 2018-02-27
SUS-4207 Respect scriptPath in build-url method

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/SUS-4207


fraszczakszymon:
@TK-999 it's used in unit tests only. Is english wiki going to have `scriptPath = '/en'`?

PR #3288, opened 2018-03-26
XW-4535 | Remove Optimizely

latata:
https://wikia-inc.atlassian.net/browse/XW-4535
@Wikia/x-wing


latata:
Tests were failing only on jenkins so we agreed to merge it.

PR #3297, opened 2018-05-21
IRIS-5918 | opt-in for auth-pages

xkxd:
## Links
https://wikia-inc.atlassian.net/browse/IRIS-5918

## cc
@nikodamn


front/common/utils/track.js
```diff
@@ -169,11 +173,14 @@ export function trackPageView(uaDimensions, overrideUrl) {
 	if (M.prop('initialPageView')) {
 		M.prop('initialPageView', false);
 	} else {
-		window.trackQuantservePageView();
-		window.trackComscorePageView();
-
-		M.tracker.Internal.trackPageView(context);
-		M.tracker.UniversalAnalytics.trackPageView(uaDimensions, overrideUrl);
+		M.trackingQueue.push(window.trackQuantservePageView);
```
rogatty: Files in `front/common` might be used in the Ember app (`front/main`). Let's make sure that it still works correctly as there is some tracking.

rogatty:
There are some linter issues.

PR #3299, opened 2018-05-22
IRIS-5918 | Install tracking-opt-in in ember app

rogatty:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-5918


front/main/lib/include-node-modules/index.js
```diff
@@ -0,0 +1,28 @@
...
+module.exports = {
+	name: 'include-node-modules',
+
+	included(app) {
+		this._super.included.apply(this, arguments);
+		this.app = app;
+	},
+
+	treeForVendor() {
+		let app = this.app;
```
xkxd: variable is not needed

PR #3301, opened 2018-05-25
IRIS | Bump opt-in modal to 1.0.31

xkxd:


package.json
```diff
@@ -103,5 +102,8 @@
     "sinon": "1.17.3",
     "svgo": "0.6.1",
     "through2": "0.6.5"
+  },
+  "dependencies": {
+    "@wikia/tracking-opt-in": "1.0.31"
```
rogatty: Isn't it a dev dependency if we include the files in the concatenated build? We don't need source files on prod.

hakubo: hmm I have just merged
https://github.com/Wikia/discussions-front-end/pull/206/files
where it's also dependency
ok?

xkxd: true, I've just npm installed this without flag

rogatty: @hakubo it doesn't make any practical difference, it's just for clarity.

